PR #26, opened 2015-01-20
Feature/quant

Alex-Jordan:
Chris and I are adapting `mathbook` to at least one project we have been working on. As we add/modify our fork of `mathbook`, we'll check if you are interested in accepting the modifications here too.

This pull request is our first test of our work flow. It adds the element `quant` to each of `mathbook-latex.xsl` and `mathbook-html.xsl`. `quant` has a `mag` attribute, which should be a number (pos or neg), and may be omitted if there are units. `quant` can have `unit` and `per` elements as children. A `unit` child is a "numerator" unit, and a `per` child is a "denominator" unit. `unit` and `per` elements must have a `base` attribute like "gram" or "second". They may additionally have a `prefix` attribute (like "kilo") and an `exp` attribute.

The latex side needs the package `siunitx` to be loaded. I'm not sure where you would add that in `mathbook-latex.xsl`. For Chris and my purposes, we have our own .tex hub file built outside of MBX. But if this feature is useful with MBX, `siunitx` needs to be added in preambles somehow. If you'd like me to add that, tell me where and I will update this branch.

Test both `.xsl` files on:
```
        <p>An empty <quant/> gives an error.</p>
        <p>The number <quant mag="42"/> is special.</p>
        <p>We had <quant mag="42"><unit base="newton" prefix="kilo"/><unit base="meter" /></quant> of torque.</p>
        <p>And we had some amount of<quant><unit base="gram" prefix="kilo"/><unit base="meter" /><per base="second" exp="2" /></quant> of force.</p>
        <p>Or maybe we had  <quant mag="42"><unit base="watt" prefix="micro"/><unit base="hour"/></quant> of energy.</p>
        <p>The quantity <quant mag="13"><unit base="gram" prefix="kilo" exp="-23"/><unit base="meter" /><per base="second" exp="2" /></quant> has a crazy exponent.</p>
        <p>test <quant><per base="gram" prefix="kilo"/><unit base="meter" /><per base="second" exp="2" /></quant> test</p>
        <p>test <quant mag="42"><per base="gram" prefix="kilo"/></quant> test</p>
        <p>test <quant mag="23"><unit prefix="kilo" base="gram" exp="2"/><unit prefix="micro" base="second"/><per prefix="centi" base="meter" exp="3"/><per prefix="milli" base="ohm"/></quant> test</p>
```

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -1478,8 +1478,8 @@ is just flat out on the page, as if printed there.
 
 
 <!-- Numbers, units, quantities                     -->
-<!-- quant                                          -->
-<xsl:template match="quant">
+<!-- quantity                                       -->
+<xsl:template match="quantity">
     <!-- warning if there is no content -->
     <xsl:if test="not(descendant::unit) and not(descendant::per) and not(descendant::mag)">
         <xsl:message terminate="no">
@@ -1517,7 +1517,7 @@ is just flat out on the page, as if printed there.
     <xsl:apply-templates />
 </xsl:template>
 
-<!-- unit and per children of a quant element       -->
+<!-- unit and per children of a quantity element    -->
 <!-- have a mandatory base attribute                -->
 <!-- may have prefix and exp attributes             -->
 <!-- base and prefix are not abbreviations          --> 
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -1477,6 +1477,154 @@ is just flat out on the page, as if printed there.
 </xsl:template>
 
 
+<!-- Numbers, units, quantities                     -->
+<!-- Numbers with no units:                         -->
+<!-- quant element with mag attribute (magnitude)   -->
+<xsl:template match="quant[not(descendant::unit) and not(descendant::per)]">
+    <xsl:choose>
+        <xsl:when test="@mag">
+            <xsl:value-of select="@mag"/>
+        </xsl:when>
+        <xsl:otherwise>
+            <xsl:text>??magnitude needed??</xsl:text>
+            <xsl:message terminate="no">
+                <xsl:text>magnitude needed</xsl:text>
+            </xsl:message>
+        </xsl:otherwise>
+    </xsl:choose>
+</xsl:template>
+
+<!-- Quantity with units, but no per parts,         -->
+<!-- with or without a magnitude                    -->
+<xsl:template match="quant[descendant::unit and not(descendant::per)]">
+    <!-- print the magnitude if there is one -->
+        <xsl:if test="@mag">
+              <xsl:value-of select="@mag"/>
+        </xsl:if>
+        <xsl:apply-templates select="unit" />
+</xsl:template>
+
+<!-- Quantity with units, that do have per parts,   --> 
+<!-- and at least one numerator unit part,          -->
+<!-- with or without a magnitude                    -->
+<xsl:template match="quant[descendant::unit and descendant::per]">
+    <!-- print the magnitude if there is one -->
+        <xsl:if test="@mag">
+            <xsl:value-of select="@mag"/>
+        </xsl:if>
+        <!-- numerator units, slash, denominator units -->
+        <sup> <xsl:apply-templates select="unit" /> </sup>
+        <xsl:text>&#8260;</xsl:text>
+        <sub> <xsl:apply-templates select="per" /> </sub>
+</xsl:template>
+
+<!-- Quantity with units, that do have per parts,   -->       
+<!-- but no numerator unit parts,                   -->
+<!-- with or without a magnitude                    -->
+<xsl:template match="quant[not(descendant::unit) and descendant::per]">
+    <!-- print the magnitude if there is one -->
+        <xsl:if test="@mag">
+            <xsl:value-of select="@mag"/>
+        </xsl:if>
+        <!-- numerator 1, slash, denominator units -->
+        <sup> <xsl:if test="@mag"> <xsl:text>&#8239;</xsl:text> </xsl:if> <xsl:text>1</xsl:text> </sup>
+        <xsl:text>&#8260;</xsl:text>
+        <sub> <xsl:apply-templates select="per" /> </sub>
+</xsl:template>
+
+<!-- unit and per children of a quant element       -->
+<!-- have a mandatory base attribute                -->
+<!-- may have prefix and exp attributes             -->
+<!-- base and prefix are not abbreviations          --> 
+<xsl:template match="quant/unit|quant/per">
+    <!-- if the unit is 1st and no mag, no need for thinspace. Otherwise, give thinspace -->
+    <xsl:if test="position() != 1 or (local-name(.)='unit' and ../@mag)">
+        <xsl:text>&#8239;</xsl:text>
+    </xsl:if>
+    <!-- prefix is optional -->
+    <xsl:if test="@prefix">
+        <xsl:if test="@prefix='yocto'">  <xsl:text>y</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='zepto'">  <xsl:text>z</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='atto'">   <xsl:text>a</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='femto'">  <xsl:text>f</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='pico'">   <xsl:text>p</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='nano'">   <xsl:text>n</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='micro'">  <xsl:text>&#xb5;</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='milli'">  <xsl:text>m</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='centi'">  <xsl:text>c</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='deci'">   <xsl:text>d</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='deca'">   <xsl:text>da</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='hecto'">  <xsl:text>h</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='kilo'">   <xsl:text>k</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='mega'">   <xsl:text>M</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='giga'">   <xsl:text>G</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='tera'">   <xsl:text>T</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='peta'">   <xsl:text>P</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='exa'">    <xsl:text>E</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='zetta'">  <xsl:text>Z</xsl:text> </xsl:if>
+        <xsl:if test="@prefix='yotta'">  <xsl:text>Y</xsl:text> </xsl:if>
+    </xsl:if>
+    <!-- base unit is *mandatory* so check to see if it has been provided -->
+    <xsl:choose>
+        <xsl:when test="@base">
+            <xsl:if test="@base='ampere'">           <xsl:text>A</xsl:text> </xsl:if>
+            <xsl:if test="@base='candela'">          <xsl:text>cd</xsl:text> </xsl:if>
+            <xsl:if test="@base='kelvin'">           <xsl:text>K</xsl:text> </xsl:if>
+            <xsl:if test="@base='gram'">             <xsl:text>g</xsl:text> </xsl:if>
+            <xsl:if test="@base='meter'">            <xsl:text>m</xsl:text> </xsl:if>
+            <xsl:if test="@base='metre'">            <xsl:text>m</xsl:text> </xsl:if>
+            <xsl:if test="@base='mole'">             <xsl:text>mol</xsl:text> </xsl:if>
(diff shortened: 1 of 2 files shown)


rbeezer:
Dear Alex and Chris,

Thanks very much for the contribution, and for packaging it as a pull request.

To load ```siunitx``` find the preamble part of  ```xsl/mathbook-latex.xsl```  and do a condition on the presence of a ```<quant>``` tag.  You'll find similar examples.  In other words, you can easily set up the latex side so that the package is loaded if and only if it is needed.  I have not done this, but it might be nice to have an XSL message (```<xsl:message>```) produced if a package is required and not around (```\IfFileExists```, iirc).  Let me know if you need more guidance.

At first blush, it strikes me that the numerical value of "mag" is content (not metadata) and would be best inside of a tag and not as an attribute.  It becomes a bit wordier:

```
<quant>
  <mag>42</mag>
  <unit base="newton" prefix="kilo"/>
  ...
</quant>
```

but my experiences tell me this will work better long-run.  I know it might be some work to change this up.  Besides that hassle, how do you feel about the actual structuring of the source?

I'm teaching four courses on two continents at the moment, so it'll be a couple weeks before that fixes itself and I can test this out thoroughly.

Thanks,
Rob

Alex-Jordan:
I'm converting `mag` to be a tag instead of an attribute. I'll take your word for it that it's the right choice. It does make authoring a bit more tedious, but you are probably right. Although, maybe mag really is meta info? If you have a quantity like 1 meter, it's the same as 39.3701 inches, so is the magnitude we write really 'content' or just a feature of how we are choosing to present the content? Any way, too late now, since I'm almost done converting :)

I'm having one last difficulty. If I'm in the template for a `unit` tag on the html side. Each `unit` needs a preceding thin space _except_ under these conditions:

* There is no `mag` and the `unit` is the first `unit` child in its `quant`. I can handle identifying if it is first, but how to identify whether or not there is a `mag` sibling?
* The `unit` is the first `unit` and it is part of a fractional unit. So basically if the `unit` is first and it has a `per` sibling.

In each case I need a conditional that evaluates whether or not a certain type of sibling (`mag` or `per`) exists. Do you have any suggestions?

----

`siunitx` is a pretty standard package. I wouldn't worry about someone not having it.  Is this what you meant about adding an `xml:message`?

rbeezer:
Dear Alex,

A quick hint on your sibling need - more later.  XSLT has "axes": ancestors, 
children, etc.  You want "preceding-sibling" and/or "following-sibling"  It is 
quite powerful for exactly the situation you have.  I really like the diagram at:

http://our.umbraco.org/wiki/reference/xslt/xpath-axes-and-their-shortcuts

Then it is a matter of seeing if your desired elements are in the node-set you 
get back (or something close to this?).  If you get the selection syntax right 
you can just test on the result, with empty being false and thus not present. 
It is all pretty clean when you wrap your head around it.

Search the MBX HTML XSL on "::" and/or "sibling" and you may find an example 
that is exactly what you need.  If this hint is not enough, report back and I 
can dig deeper.

Rob

Alex-Jordan:
OK, `mag` is now a tag instead of an attribute. And after learning about sibling axes, I tidied up the conditional code too. Test with:
```
<p>In the middle of this <quant></quant> there is an empty quant.</p>
<p>There is a number <quant><mag>42</mag></quant>.</p>
<p>The value of gravitational constant <m>g</m> is <quant><mag>9.8</mag><unit base="meter"/><per base="second" exp="2"/></quant>.</p>
<p>A quantity with lots of unit craziness is <quant><mag>23</mag><per base='degreeCelsius'/><per base='second' exp='2'/><unit prefix='micro' base='hectare' exp='23'/></quant></p>
<p>One Hertz is a <quant><per base='second'/></quant></p>
<p>Force is measured in <quant><unit prefix="kilo" base="gram"/><unit base="meter"/><per base="second" exp="2"/></quant>.</p>
<p>Here we are missing a base: <quant><unit prefix='kilo'/></quant></p>
<p>A temperature in <quant><unit base='degreeCelsius'/></quant> and another in <quant><unit base='degreeFahrenheit'/></quant>.</p>

```

Note that this is all meant to make unit constructs _outside_ of math mode. In LaTeX, `\num{}`, `\si{}`, and `\SI{}` handle it, and in HMTL, it's more direct.

However in LaTeX, `\num{}`, `\si{}`, and `\SI{}` also work _inside_ math mode. So I have a separate project to add these macros and the unit macros to MathJax. Once that is done, then in a math environment, we can feel free to use  `\num{}`, `\si{}`, and `\SI{}` inside math mode directly.

rbeezer:
Very good!  Thanks for the contribution.  I'm very anxious to pull it in and 
have a good look, but it will have to wait a few days.  I'll apologize now for 
the delay.

Rob

rbeezer:
Dear Alex,

Yes, you are right, maybe a decision on units makes mag a presentation and the 
object is really just its inherent physical quantity.  Very perceptive.  But as 
I have said, there is no bright line separating content from presentation and if 
there was, then this might land right on top of it.

Yes, my comment about an XSL "message" was because I didn't know how standard it 
was.  I definitely do not want to see preambles full of usepackage's that are 
not necessary - it would be gravy to warn somebody exactly how their TeX 
installation is insufficient if they use an MBX feature using a not-to-standard 
package.

Thanks for being agreeable to switching up "mag".

Rob

Alex-Jordan:
No rush. Please feel free to accept, reject, or request modification to any
of these items at your own pace. There may be a lot more coming your way.
We're making them for our own fork, and if they work with the master
branch, all the better.

On Wed, Jan 21, 2015 at 9:12 PM, Rob Beezer <notifications@github.com>
wrote:

> Very good! Thanks for the contribution. I'm very anxious to pull it in and
> have a good look, but it will have to wait a few days. I'll apologize now
> for
> the delay.
>
> Rob
>
> On 22/01/2015 01:22, Alex Jordan wrote:
> > OK, |mag| is now a tag instead of an attribute. And after learning about
> sibling
> > axes, I tidied up the conditional code too. Test with:
> >
> > |<p>In the middle of this <quant></quant> there is an empty quant.</p>
> > <p>There is a number <quant><mag>42</mag></quant>.</p>
> > <p>The value of gravitational constant <m>g</m> is
> <quant><mag>9.8</mag><unit base="meter"/><per base="second"
> exp="2"/></quant>.</p>
> > <p>A quantity with lots of unit craziness is <quant><mag>23</mag><per
> base='degreeCelsius'/><per base='second' exp='2'/><unit prefix='micro'
> base='hectare' exp='23'/></quant></p>
> > <p>One Hertz is a <quant><per base='second'/></quant></p>
> > <p>Force is measured in <quant><unit prefix="kilo" base="gram"/><unit
> base="meter"/><per base="second" exp="2"/></quant>.</p>
> > <p>Here we are missing a base: <quant><unit prefix='kilo'/></quant></p>
> > <p>A temperature in <quant><unit base='degreeCelsius'/></quant> and
> another in <quant><unit base='degreeFahrenheit'/></quant>.</p>
> >
> > |
> >
> > Note that this is all meant to make unit constructs /outside/ of math
> mode. In
> > LaTeX, |\num{}|, |\si{}|, and |\SI{}| handle it, and in HMTL, it's more
> direct.
> >
> > However in LaTeX, |\num{}|, |\si{}|, and |\SI{}| also work /inside/ math
> mode.
> > So I have a separate project to add these macros and the unit macros to
> MathJax.
> > Once that is done, then in a math environment, we can feel free to use
> |\num{}|,
> > |\si{}|, and |\SI{}| inside math mode directly.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/rbeezer/mathbook/pull/26#issuecomment-70942705>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rbeezer/mathbook/pull/26#issuecomment-70972488>.
>



-- 
Alex Jordan
Mathematics Instructor
Portland Community College

Alex-Jordan:
At https://groups.google.com/forum/#!topic/mathbook-xml-support/La6yGblwIt0 you had some comments about this feature. I can't tell if they all apply to this feature in its current form, or if they were comments for an earlier iteration. Certainly 6) applies, but I haven't had success acting on it.

rbeezer:
Dear Alex,

Only looking at the code finally now.  I was trying to get my (our) workflow 
right with something smaller, so I was stalling on this one.  ;-)  Mostly that 
post was advice to help you with code, in response to your questions.

"mag" was important, and you have done that.  I'd really love to see all the 
unit translations done with keys.  There was an example someplace: could you add 
it to the sample-article by making a whole section all of its own -  "Scientific 
Units" or something similar?

I'd like to accept this largely as-is, since it is very self-contained and will 
be valuable, so those high-level suggestions are the only changes I'd suggest.

Rob

Alex-Jordan:
I'm having trouble figuring out how to use keys in the case where I am looking up stuff in a different file. I can't seem to mimic the localization stuff, partly because that relies on global parameters I think, so it's not exactly the same.

I have made a `mathbook-units.xsl` that has, among other things:
```
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
...
<prefixes>
    <prefix full='yocto' short='y'/>
    <prefix full='zepto' short='z'/>
    ...
</prefixes>
...
</xslstylesheet>
```

Then in `mathbook-html.xsl`, there is:
```
<xsl:key name="prefix-key" match="prefix" use="@full"/>
<xsl:template match="unit|per">
    ...
    <xsl:if test="@prefix">
        <xsl:for-each select="document('mathbook-units.xsl')">
            <xsl:for-each select="key('prefix-key',@prefix)">
                <xsl:value-of select="@short"/>
            </xsl:for-each>
        </xsl:for-each>
    </xsl:if>
    ...moves on to base and exponent...
</xsl-template>
```
As a reminder, `xml` will have stuff like
```
<quantity>
    <mag>9.8</mag>
    <unit prefix='kilo' base='gram'/>
    <unit base='meter'/>
    <per base='second' exponent='2'/>
</quantity>
```

Any suggestions on what I can do to make it right in `mathbook-html.xsl`? Every online example/tutorial about using keys doesn't have the complication of using a different file. And I feel like extra complications with the localization as a model are leading me astray.

rbeezer:
By "different file" you mean your units show up in the source XML, not in an XSL 
stylesheet?

Too late this evening to think clearly - I'll try again later.

Alex-Jordan:
No, there are three files in play: xml source, xsl mathbook-html, and now
xsl mathbook-units. The last is small-ish, and its only purpose is to keep
that big list of prefixes and bases in a separate place from mathbook-html.
mathbook-html will process unit and per elements, by looking up their
prefix and base attributes in mathbook-units, finding the abbreviations,
and then writing them to the output.

Perhaps I misunderstood, but this is what I thought you were asking for
when you suggested I look at how the localization was handled.

I won't be able to come back to this until Wednesday or later. As always,
no rush.

On Sun, Mar 1, 2015 at 9:47 PM, Rob Beezer <notifications@github.com> wrote:

> By "different file" you mean your units show up in the source XML, not in
> an XSL
> stylesheet?
>
> Too late this evening to think clearly - I'll try again later.
>
> On 03/01/2015 09:29 PM, Alex Jordan wrote:
> > I'm having trouble figuring out how to use keys in the case where I am
> looking
> > up stuff in a different file. I can't seem to mimic the localization
> stuff,
> > partly because that relies on global parameters I think, so it's not
> exactly the
> > same.
> >
> > I have made a |mathbook-units.xsl| that has, among other things:
> >
> > |<xsl:stylesheet version="1.0" xmlns:xsl="
> http://www.w3.org/1999/XSL/Transform">
> > ...
> > <prefixes>
> > <prefix full='yocto' short='y'/>
> > <prefix full='zepto' short='z'/>
> > ...
> > </prefixes>
> > ...
> > </xslstylesheet>
> > |
> >
> > Then in |mathbook-html.xsl|, there is:
> >
> > |<xsl:key name="prefix-key" match="prefix" use="@full"/>
> > <xsl:template match="unit|per">
> > ...
> > <xsl:if test="@prefix">
> > <xsl:for-each select="document('mathbook-units.xsl')">
> > <xsl:for-each select="key('prefix-key',@prefix)">
> > <xsl:value-of select="@short"/>
> > </xsl:for-each>
> > </xsl:for-each>
> > </xsl:if>
> > ...moves on to base and exponent...
> > </xsl-template>
> > |
> >
> > As a reminder, |xml| will have stuff like
> >
> > |<quantity>
> > <mag>9.8</mag>
> > <unit prefix='kilo' base='gram'/>
> > <unit base='meter'/>
> > <per base='second' exponent='2'/>
> > </quantity>
> > |
> >
> > Any suggestions on what I can do to make it right in
> |mathbook-html.xsl|? Every
> > online example/tutorial about using keys doesn't have the complication
> of using
> > a different file. And I feel like extra complications with the
> localization as a
> > model are leading me astray.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/rbeezer/mathbook/pull/26#issuecomment-76660593>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rbeezer/mathbook/pull/26#issuecomment-76661700>.
>



-- 
Alex Jordan
Mathematics Instructor
Portland Community College

Alex-Jordan:
I've got this working with a key, all values of which are stored in a separate mathbook-units.xsl file. I want to consider what to do (or not do) about `<quantity>` within a math environment before I am ready for review. But just wanted to let you off the hook for teaching me more about keys.

Alex-Jordan:
Do math environments do any scrubbing of elements within them? For some reason I can't understand, when `<quantity>` as below:
```
<m>
  <quantity>
     <unit />
  </quantity>
</m>
```
is tested with `"descendant::unit and not(descendant::per)"` it comes up false. I can't understand this.

rbeezer:
<m> and friends apply no templates.  They are meant to contain straight 
unadulterated LaTeX.  Or more precisely, LaTeX that MathJax can cope with.

rbeezer:
Thanks, I seem to be on many hooks, so it was still going to be a few days 
before I got to building a keys example.

Rob

Alex-Jordan:
A cleaned up version of this feature is now in https://github.com/rbeezer/mathbook/pull/63.

PR #32, opened 2015-01-21
feature/sfrac

cmhughes:
Hi Rob,
As Alex mentioned in https://github.com/rbeezer/mathbook/pull/26, we're working on a project which uses the `xml` approach you demonstrated to us in June 2013. We've made quite a few new features, and are in the process of re-doing them in a way that allows us to submit them to your original project, and `pull` your updates; hopefully some of our features will be useful!

This pull request implements a 'beveled' fraction in both the `.tex` and the `.html` versions. It conditionally loads the `xfrac` package in the `.tex` version, and conditionally updates the `MathJax` in the `.html` version. You'll see that I've provided links within the code, but for reference, here they are:
- https://github.com/mathjax/MathJax-docs/wiki/Beveled-fraction-like-sfrac,-nicefrac-bfrac
- http://tex.stackexchange.com/questions/3372/how-do-i-typeset-arbitrary-fractions-like-the-standard-symbol-for-5-%C2%BD

To test it, add (for example) `\sfrac{2}{13}` to any of the math elements in `examples/sample-article.xml` and then run the translation scripts. You should see that the `\sfrac` produces a beveled fraction :)

Best
Chris

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -2605,6 +2605,24 @@ MathJax.Hub.Config({
         scale: 88,
     },
 });
+    <xsl:if test="//m[contains(text(),'sfrac')] or //md[contains(text(),'sfrac')] or //me[contains(text(),'sfrac')] or //mrow[contains(text(),'sfrac')]">
+    /* support for the sfrac command in MathJax (Beveled fraction)
+        see: https://github.com/mathjax/MathJax-docs/wiki/Beveled-fraction-like-sfrac,-nicefrac-bfrac */
+MathJax.Hub.Register.StartupHook("TeX Jax Ready",function () {
+  var MML = MathJax.ElementJax.mml,
+      TEX = MathJax.InputJax.TeX;
+
+  TEX.Definitions.macros.sfrac = "myBevelFraction";
+
+  TEX.Parse.Augment({
+    myBevelFraction: function (name) {
+      var num = this.ParseArg(name),
+          den = this.ParseArg(name);
+      this.Push(MML.mfrac(num,den).With({bevelled: true}));
+    }
+  });
+});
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks very much, Chris.  I'll put it through its paces in a few days.

rbeezer:
Thanks very much, Chris.  That is a fun feature to have.  OK, here's the workflow discussion.  Thought this would be a good one to start with.  I like a (mostly) linear history with very few merge objects, so here's what I did.

I added your `feature/sfrac` branch to my local repository, rebased it onto the tip of my current dev branch.  I edited your commit message slightly, and added a new commit with an example use (and credit!) in the sample article.  These were fast-forward merged into dev and now are public.

This changes the commit hashes, but I do not think this should be a problem.  When you next pull from "official" mathbook you'll get your work back on dev and can delete your `feature/sfrac` branch.

Let me know if this is a workable procedure for contributions like this.  I guess I'll need to learn a bit more about using a branch to go back and forth (like for tables).  I'll chip away at the other contributions in the next few days.  Keep 'em coming!

Rob

cmhughes:
Hi Rob,
Thanks very much for the update, and for accepting this feature. I've mainly worked with merging in the past; I can see that rebasing gives a more linear history, but need to do some more reading to understand more subtle differences. 

Alex, I think that Rob's non-merging workflow shouldn't make much (any?) difference to OMD development; here's what I did to update our OMD development branch:

```
git checkout dev         (which tracks origin/dev)
git fetch
git pull
git fetch omd dev
git pull omd dev
git push omd dev
```
I guess maybe I could have used `git fetch --all`?...

This means that our `dev` branch is now up to date, as can be seen in the Network screenshot below.

screenshot from 2015-01-28 21 23 03

PR #35, opened 2015-01-25
option not to have typeout for chapter, sections, etc in latex version

cmhughes:
This pull request introduces

     <xsl:param name="latex.typeout" select="'yes'"/>

to `xsl/mathbook-latex.xsl`, which can be changed to `yes` or `n`o to allow or dis-allow the verbose `\typeout` associate with each chapter, section, etc.

To test, try running `xsltproc xsl/mathbook-latex.xsl examples/sample-article.xml` and then change the value of `latex.typeout` to `no` and re-run. You should see a removal of the `typeout` when the option is set to `no`.

### DIFF
--- a/xsl/mathbook-latex.xsl
+++ b/xsl/mathbook-latex.xsl
@@ -78,11 +78,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- For a non-electronic copy, mostly links in black -->
 <xsl:param name="latex.print" select="'no'"/>
 <!--  -->
-<!-- Typeout Option                                     -->
-<!-- Some users may not want the \typeout{Section ***}  -->
-<!-- Simply change latex.typeout to 'no' and typeout    -->
-<!-- will be removed -->
-<xsl:param name="latex.typeout" select="'yes'"/>
+<!-- Typeout Option                                                       -->
+<!-- By default, each heading (chapter, section, subsection, etc)         -->
+<!-- adds verbose \typeout{Chapter ***} information to the .tex file.     -->
+<!-- This feature can be switched on/off as follows:                      -->
+<!--    latex.logging=1  means that \typeout will be used                 -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Dear Chris,

I'd like to include this.  One tangential question, one request, and one favor.

1.  Did you find this was too much information?  Or did it get in the way of something?  Simply curious why you thought to turn this off?

2.  Could we make this switch more general?  How about "latex.logging" with a value of 0 being no extra output.  So for now, you could test on the level being greater than zero, and if so use the  typeout stuff.  Maybe set the value to 1 as the default.  If I add more information, we can make a hierarchy of levels.  Mostly, I think "latex.typeout" may not make sense to somebody who does not know the `\typeout` command.

3.  Would you mind making these changes in yur repository as a way for me to test my workflow?  Can you update `feature/optionalTypeout`?  I'd just like to test updating my side and this seems like a simple enough change to experiment with.

Thanks,
Rob

cmhughes:
Hi Rob,
Thanks for the follow-up, and for the questions. I'll respond to each :)

1. I think that the `typeout` feature is very useful for someone new to `MBX` or new to `latex`, as they can be fairly sure that they're getting the output that they want. As I've mentioned before, Alex and I are using the `MBX` idea in a *big* document that has many chapters, sections, subsections, and lots more; the `log` file and the output is already very full with all kinds of other information. Additionally, each chapter `.tex` file is around 8000 lines long--adding 3 lines per heading makes it even bigger :) 

2. That sounds like a good idea; I've udpated the language as you described in the latest commit.

3. Yes, happily :) Alex and I have agreed that each new feature we make will always be branched from *your* development branch, `origin/dev`, so `git pull` brings in any updates. You'll see that I have updated the branch to include the most recent updates from `origin/dev`, which can be verified graphically by looking at the network graph: https://github.com/rbeezer/mathbook/network

I hope this helps; please do feel free to follow-up if you have other feedback :)
Best
Chris

PR #36, opened 2015-01-25
added option to only put LaTeX label on objects that have an ID

cmhughes:
This pull request adds a switch so that numbered objects such as chapters, sections, figures will only receive a `\label` in the `latex` version when they have an `xml:id` attribute.

The parameter that controls this is in `xsl/mathbook-common.xsl`:

    <xsl:param name="labelObjectsWithoutID" select="'yes'" />

If you run `xsltproc xsl/mathbook-latex.xsl examples/sample-article.xml` and you have `labelObjectsWithoutID` set to `yes`, then you'll see that everything that has a number will receive a generic label, such as `\label{section-1}` (i.e, the original behaviour). If you change `labelObjectsWithoutID` to `no` then you'll see that only the objects that have an `xml:id` will receive a label.

----
PS: Rob, we're making quite a few updates to `openMathDocs` and are submitting them to you with no pressure at all :) If you like them, we'd love them to be part of your project, but if you don't like them or you need parts of them tweaked, please do let us know--we certainly could benefit a lot from your feedback.

### DIFF
--- a/xsl/mathbook-common.xsl
+++ b/xsl/mathbook-common.xsl
@@ -74,6 +74,10 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- on an <xref> where it is unjustified or a problem    -->
 <!-- Default is to have this feature off                  -->
 <xsl:param name="autoname" select="'no'" />
+<!-- Chapters, sections, figures, tables, exercises, etc         -->
+<!-- can receive a generic \label{<object name>-<object number>} -->
+<!-- or not, according to the value of labelObjectsWithoutID     -->
+<xsl:param name="labelObjectsWithoutID" select="'yes'" />
 <!-- How many levels to table of contents  -->
 <!-- Not peculiar to HTML or LaTeX or etc. -->
 <!-- Sentinel indicates no choice made     -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Dear Chris,

Looking for ultra-clean LaTeX source?  Me too.  ;-)

This one makes me a bit nervous, I rely on the "internal-id" stuff quite heavily.  For example, it seems that the exception for notation elements is "hidden" in the labeling routine.  One of the things I like about XSL is the maintainability if you modularize stuff carefully (I won't claim to have *always* done so).  

Maybe the labeling template needs a parameter?  I can't pinpoint exactly how I'd make this more robust.  Am I making sense?  Any thoughts?

cmhughes:
Thanks for the follow-up, Rob; I appreciate the caution :)

In the latest commit, I've added
```
<xsl:param name="alwaysGiveLabel" select="'notation section'" />
```
The idea is that if you have `labelObjectsWithoutID` set to `no` but you would still like to (generically) label certain elements, you can specify them in `alwaysGiveLabel`.

The check in the `label` routine contains:
```
<xsl:if test="$labelObjectsWithoutID='yes' or @xml:id or contains($alwaysGiveLabel, local-name(.))">
```

Does that seem a bit better? I hope that this removes something of the 'hidden' nature that I had previously, and I'm very open to further feedback :)

PR #41, opened 2015-02-04
Slight update to doc in sample article

kcrisman:
The initialism appears to be used in more than one place - it is in the HTML also.


rbeezer:
I've merged this into mathbook/dev (with edited commit message).

Thanks for your first contribution.

PR #58, opened 2015-02-26
id attribure for figure in html

Alex-Jordan:
I noticed when I clicked on a link to a figure reference, there was no target id. Thought about submitting an issue, but then I thought this might be more helpful as long as I did it right.


rbeezer:
Looks correct, I'll double-check and merge tonight perhaps.  Thanks - oversight on my part, I think figures were implemented very early in development.

Rob

rbeezer:
Works great.  Added an explicit test case to the "Cross-Referencing" section of the sample article.  I'll push it up before the night is out.

cmhughes:
I've also found that `table` elements don't receive an ID. It is part of https://github.com/rbeezer/mathbook/pull/52

PR #63, opened 2015-03-06
quantity, quantity/mag, quantity/unit, quantity/per

Alex-Jordan:
This is a cleaned up version of https://github.com/rbeezer/mathbook/pull/26.

### DIFF
--- a/examples/sample-article.xml
+++ b/examples/sample-article.xml
@@ -166,7 +166,7 @@ the xsltproc executable.
         which have subsubsections.These are all "containers"
         and so line breaks and extraneous whitespace gets
         scrubbed.  Use indentation and line breaks liberally
-        within containers to make the structure of your 
+        within containers to make the structure of your
         source readable.
         -->
 
@@ -253,7 +253,7 @@ the xsltproc executable.
 
             <p>Because a definite integral can be computed using an antiderivative, we have the following definition.</p>
 
-            <!-- 
+            <!--
             Notation may be associated with a definition,
             or it can be placed standalone.  If associated
             with a definition, future work will reflect that
@@ -1555,7 +1555,7 @@ the xsltproc executable.
             </exercise>
 
          </exercises>
-        
+
         <!--
         Various author tools are planned.  Note the embedded "todo"
         elements above.  An XSL transform in-testing will extract them
--- a/examples/sample-article.xml
+++ b/examples/sample-article.xml
@@ -1748,6 +1748,15 @@ the xsltproc executable.
 
         </section>
 
+        <section>
+            <title>Units of Measure</title>
+
+            <p>Units of measure can be given xml treatment too with the <c>quantity</c> element. In <latex />, the <c>siunitx</c> package is loaded to achive unit handling. Since that package only offers SI units, some other common units will be added by MBX in the preamble. In HTML, the capabilities of <c>siunitx</c> are simulated, weakly. Note that at present, you should not attempt to use the <c>quantity</c> element within a math environment.</p>
+            <p>The value of gravitational constant <m>g</m> is <quantity><mag>9.8</mag><unit base="meter"/><per base="second" exp="2"/></quantity>. Force is measured in <quantity><unit prefix="kilo" base="gram"/><unit base="meter"/><per base="second" exp="2"/></quantity>, also known as one <quantity><unit base="newton"/></quantity>. A quantity with rather ridiculous units is <quantity><mag>23</mag><per base='degreeCelsius'/><per base='second' exp='2'/><unit prefix='micro' base='hectare' exp='23'/></quantity>. One <quantity><unit base="hertz"/></quantity> is the same as <quantity><per base='second'/></quantity>. You can have a unitless quantity, like <quantity><mag>42</mag></quantity>, which may help with consistency between such numbers and units in the <latex /> output. Some non-SI units are available, such as the absurd <quantity><unit base="degreeFahrenheit"/><unit base="foot"/><unit base="pound"/><per base="gallon"/></quantity>.</p>
+            <p>For a full list of the allowed units and prefixes, see <c>mathbook-units.xsl</c>. If you have a need for more units, they need to be added to <c>mathbook-units.xsl</c> in the section that deals with units which are not part of <c>siunitx</c> by default.</p>
+
+        </section>
+
         <!-- Bibliography/References can appear anywhere -->
 
         <references>
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -20,7 +20,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 *********************************************************************-->
 
 <!-- Identify as a stylesheet -->
-<xsl:stylesheet 
+<xsl:stylesheet
     xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0"
     xmlns:xml="http://www.w3.org/XML/1998/namespace"
     xmlns:exsl="http://exslt.org/common"
@@ -1890,7 +1890,7 @@ is just flat out on the page, as if printed there.
         <xsl:when test="not(*) and not(normalize-space())">
             <xsl:value-of select="@href" />
         </xsl:when>
-        <xsl:otherwise>         
+        <xsl:otherwise>
             <xsl:apply-templates />
         </xsl:otherwise>
     </xsl:choose>
@@ -2988,8 +2988,8 @@ $(function () {
 
 <!-- Knowl header -->
 <xsl:template name="knowl">
-<script type="text/javascript" src="http://code.jquery.com/jquery-latest.min.js"></script> 
-<link href="http://aimath.org/knowlstyle.css" rel="stylesheet" type="text/css" /> 
+<script type="text/javascript" src="http://code.jquery.com/jquery-latest.min.js"></script>
+<link href="http://aimath.org/knowlstyle.css" rel="stylesheet" type="text/css" />
 <script type="text/javascript" src="http://aimath.org/knowl.js"></script>
 
 </xsl:template>
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -1695,6 +1695,102 @@ is just flat out on the page, as if printed there.
 </xsl:template>
 
 
+<!-- Numbers, units, quantities                     -->
+<!-- quantity                                       -->
+<xsl:template match="quantity">
+    <!-- warning if there is no content -->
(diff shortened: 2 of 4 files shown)


Alex-Jordan:
This is ready for review. Note there is a section in the sample article. In the future, when MathJax has an siunitx extension, then siunitx unit commands can be used in math mode. But for now, these units are for outside of math mode only (even though they would technically work in math mode in the LaTeX side.)

There are several warnings for if you don't use these elements properly. 

If we add more units not already listed in `mathbook-units.xsl`, then simultaneous new siunitx unit macros need to be added to `mathbook-latex.xsl`. Perhaps all of those macros should also be stored in `mathbook-units.xsl`, and simply be called somehow from `mathbook-latex.xsl` when needed.

Alex-Jordan:
I'm going to try bringing the extra siunitx macros out of mathbook-latex and into mathbook-units. My goal is to leave mathbook-latex with no explicit mention of any units, and make it so that all additions/revision of units would take place in mathbook-units. I'll ping again when I think it's *relaly* ready for review.

pkra:
forgive me for just lurking but @Alex-Jordan would you be interested in creating an extension for MathJax? We've had requests before but we don't know siunitx enough. There's also currently a third party extension PR implementing the physics package so perhaps that developer might be intereted in joining forces.

Alex-Jordan:
@pkra I'd love to have such an extension too, and I even looked into writing MathJax extensions with this one in mind last fall. The problem for me is that `siunitx` does so much. Despite its name, it is much more than unit handling. It also: 
* handles number formatting, e.g do you want `\num{300}` to output "300", "3E2", "3\times10^2", etc.; do you want numbers output like "10{,}000", "10\,000", or "10.000", etc.
* can deal with different _input_ formats. `\num{300}` is the same as `\num{3e2}`
* provides the `S` column type, great for tables of data aligned by decimal point, and other things. (`S` can take optional arguments.)

Many more things could be listed. And most of what it does works inside and outside of math mode, but I suppose for MathJax, we would only try to replicate the math mode parts.

Much like what I'm trying to do here, it shouldn't be too hard to have a MathJax extension that _weakly_ imitates `siunitx`, but for users who are expecting such an extension to behave just like `siunitx`, they will be disappointed. Here, I'm only trying to get weak copies of the `\num{#1}`, `\si{#2}`, and `\SI{#1}{#2}` commands, where `#1` is a number in plain old "300" input form, and `#2` is a restricted, but reasonable subset of the `siunitx` unit macros. That's like only 10% of what `siunitx` does. I have to wonder if it would be better to get a true TeXpert involved like Joseph Wright (the author of `siunitx`, currently writing `siunitx 3.0`) who could do it right.

Alex-Jordan:
@rbeezer, I need some xsl help. I'm trying to have lines like this in `mathbook-units.xsl`:

```
<base full='degreeFahrenheit'  short='&#176;F' siunitx-macro='\DeclareSIUnit\degreeFahrenheit{\SIUnitSymbolDegree{F}}&#xa;' ></base>
```

The `full` and `short` attributes are used by a key in `mathbook-html.xsl`, and I have that part working fine. The `siunitx-macro` attribute is code that needs to go into a `.tex` preamble if the source uses the `quantity` element. So in `mathbook-latex.xsl`, I have 

```
    <xsl:if test="//quantity">
        <xsl:text>%% Used for units and number formatting&#xa;</xsl:text>
        <xsl:text>\usepackage[per-mode=fraction]{siunitx}&#xa;</xsl:text>
        <xsl:text>%% Common non-SI units&#xa;</xsl:text>
        <xsl:for-each select="document(mathbook-units.xsl)/base">
            <xsl:value-of select="@siunitx-macro" />
        </xsl:for-each>
    </xsl:if>
```
This is outputting the two comments and the `\usepackage`, but none of the `siunitx-macro` attributes. Any thoughts on what I'm doing wrong?

rbeezer:
Just a thought - deep into recursive routines for tables at the moment.

Try a // just before your "base" in the "for-each".  I do not fully grok 
"document()" yet, but any XML file has exactly one root element.  Maybe for an 
XSL file it is "stylesheet"?  So maybe

document(...)/stylesheet/base

is correct and more exact?  Let me know either way.

I know you are trying to put lots of stuff into one file, which is plowing new 
ground for me.  I only understand keys well enough to mimic and adapt working 
examples.

Rob

Alex-Jordan:
Figured it out. It needs to be `document('mathbook-units.xsl')`, not `document(mathbook-units.xsl)`.

rbeezer:
Ooh.  ;-)

Alex-Jordan:
OK, this is ready for review. Check the sample-article for examples. And see if you think the way mathbook-units.xsl is built looks OK.

rbeezer:
Dear Jordan,

Looks good.  A few nits.

1.  The warnings are fine.  But lots of this will get caught when I get a DTD together and then  xmllint  will see that elements are out of order or that required bits are not present.  So you only need to "worry" about other kinds of mistakes.  There is now a "examples/sample-errors-and-warnings.xml" as the Bad Bank where you can test catching warnings.  Don't remove any warnings (unless you really want to).  You can add to the  sample-errors  if you like, but do not feel it is required.

2.  I want  xelatex  to be available for Unicode input (which I do not totally understand yet).  I just barely noticed that the micro-hectare lost the micro under xelatex.  Maybe a fix could go into the preamble?  I wonder if there are other such situations?
http://tex.stackexchange.com/questions/118894/unicode-math-interferes-with-siunitx

3.  There are some changes that look like 100% whitespace.  Are they tabs of mine?  Superfluous?  I can't tell.  In any event, I'd collect them into a "whitespace-only" commit, which I can do once this is merged.

4.  mathbook-units.xsl looks fine.  But I don't know keys so well yet.  Looks fine and works - that should be good for now.

5.  How about something with English units for the sample article to test that functionality?

6.  I had to install the Ubuntu  texlive-science  package.  :-(   No helping that.  I should make a totally minimal tex install and see how much else the sample article requires.  Or make a totally simple sample article that will build always.

7.  Can you replace `test="local-name(.)='per'"` with `test="self::per"`  Will that work?  I'm liking the use of axes much better these days.

Alex-Jordan:
1.  Good to know for the future. These warning messages are to let the user know that their elements are going to output blanks because they aren't structured right. That doesn't cause a problem with compilation though.

2.  I don't have `xelatex` working locally, so I can't test this, but it looks like I can use `ifxetex`to redefine how micro is made, and use a literal unicode micro character. Sound OK?

3.  I took to hunting and eliminating whitespace by using `/\s$` in `vim`.  I only wanted to cut out my own whitespace, but as long as I found some of yours I thought I would cut it. Should I put the whitespace back in to match `dev` right now? Then a follow up commit would remove it?

4.  Good, I was wondering if you think the attribute names sound OK.

5.  Done.

6.  Ah I see. As a big TeX user, I tend to just have everything installed. I have the impression that `siunitx` is well-used, but I don't know how to confirm that kind of thing. [Here](https://trac.macports.org/wiki/TeXLivePackage) you can see what LaTeX packages are in which Ubuntu packages. You have packages in `mathbook-latex.xsl` from `texlive-latex`, `texlive-basic`, `texlive-xetex`, `texlive-math-extra`, (now) `texlive-science`, `texlive-latex-recommended`, and `texlive-latex-extra`. Additionally, these packages are not listed as part of any of the Ubuntu packages (some of which I find suspicious, and wonder if there is an accidental omission): `fixltx2e`, `amssymb`, `amsthm`, `graphicx`, `beramono`, `tikz`, `multicol`, `xfrac`, `makeidx`, `showidx`, `longtable`, `showkeys`.

7.  Done.

I'll see if you have anything to say about any of these before I repush.

pkra:
@Alex-Jordan a partial implementation of the easier parts would be a great start, I think. It could go on https://github.com/mathjax/MathJax-third-party-extensions (which is mirrored on our CDN, see http://docs.mathjax.org/en/latest/options/ThirdParty.html#mathjax-third-party-extension-repository). It could also attract other contributors.

rbeezer:
On 03/08/2015 11:31 PM, Alex Jordan wrote:
>  2.
>
>     I don't have |xelatex| working locally, so I can't test this, but it looks
>     like I can use |ifxetex|to redefine how micro is made, and use a literal
>     unicode micro character. Sound OK?

Yes, and I can test.  I already use  ifxetex  to load a monospace font two 
different ways.  So look for that in the preamble.  I did not study the 
stackexchange discussion to closely, so if a literal Unicode will solve the 
problem, then give it a go.

>  3.
>
>     I took to hunting and eliminating whitespace by using |/\s$| in |vim|. I
>     only wanted to cut out my own whitespace, but as long as I found some of
>     yours I thought I would cut it. Should I put the whitespace back in to match
>     |dev| right now? Then a follow up commit would remove it?

You can edit your commits to collect all these into one of its own, or I can do 
it.  Just let me know.  I'd prefer to have "whitespace-only commits that folks 
can ignore, rather than having them as distractions in more substantial commits.

>  6.
>
>     Ah I see. As a big TeX user, I tend to just have everything installed. I
>     have the impression that |siunitx| is well-used, but I don't know how to
>     confirm that kind of thing. Here
>     <https://trac.macports.org/wiki/TeXLivePackage> you can see what LaTeX
>     packages are in which Ubuntu packages. You have packages in
>     |mathbook-latex.xsl| from |texlive-latex|, |texlive-basic|, |texlive-xetex|,
>     |texlive-math-extra|, (now) |texlive-science|, |texlive-latex-recommended|,
>     and |texlive-latex-extra|. Additionally, these packages are not listed as
>     part of any of the Ubuntu packages (some of which I find suspicious, and
>     wonder if there is an accidental omission): |fixltx2e|, |amssymb|, |amsthm|,
>     |graphicx|, |beramono|, |tikz|, |multicol|, |xfrac|, |makeidx|, |showidx|,
>     |lo ngtable|, |showkeys|.

Big TeX user as well, but that tool could be helpful.  I should probably make 
something interesting (but simple) for folks first test use.

> I'll see if you have anything to say about any of these before I repush.

Three exams today, sorry I couldn't reply sooner.  Almost there.

Rob

Alex-Jordan:
Grading exams is the only part of an academic position that I just outright dislike. You may feel similarly :)

OK, all re-arranged into two commits; one for content, one for white space. I had to rebase, so that likely means that if you straight pull, you will have conflicts. I guess if you force pull, or just delete your local branch before pulling again, you'll be OK.

I added some new things since you last checked. You may want to put it through the ropes again just to make sure. No rush of course.

rbeezer:
All done.  Looks great, just merged.  I'll advertise on the list and you can pile on if you wish.  Thanks for sticking with this one.  (As usual, I edited the commit messages some.)

PR #76, opened 2015-03-25
cols and by attributes for exercisegroup in latex

Alex-Jordan:
This implements a `cols` attribute for `exercisegroup` on the latex side of things. Additionally, the `by` attribute can be set to "row" for the exercises to progress by row. See the `sample-article.xml` for more. (But you'll need to see the pdf to see the columns. It's not implemented for HTML.)

I was wondering if Rob or David would like to handle the HTML side. I guess that whether expansion is by row or by column, CSS will control most of what is going on. So simultaneous writing of some CSS is needed alongside additions to `mathbook-html.css`. I'm not much of a CSS expert, which is why I'm wondering if someone else would care to tackle that. Whoever ends up doing this, it would be nice to indent the exercises in an exercise group as a way to make the introduction and conclusion stand out.

### DIFF
--- a/examples/sample-article.xml
+++ b/examples/sample-article.xml
@@ -1999,7 +1999,217 @@ the xsltproc executable.
                 </hint>
             </exercise>
 
-         </exercises>
+            <exercisegroup cols="4">
+                <introduction><p>An <c>exercisegroup</c> can have a <c>cols</c> attribute (which defaults to <q>1</q>). Exercises will progress downward in columns unless the <c>by</c> attribute is set to <q>row</q> (see a later example). The stringparam <c>exercisegroup.pregression</c> can be set to <q>row</q> to globally override all <c>exercisegroup</c>s that do not have a <c>by</c> attribute specified. So you can also just leave all <c>by</c> attributes unspecified and use <c>exercisegroup.pregression</c> to globally control this behavior.</p></introduction>
+
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,x</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,x^2</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,x^{23}</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,e^x</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,2^x</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,\ln(x)</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,\sin(x)</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,\arctan(x)</m></p>
+                    </statement>
+                </exercise>
+
+                <conclusion>This has been a collection of derivative calculations.</conclusion>
+
+            </exercisegroup>
+
+            <exercisegroup cols="3">
+                <introduction><p>Here are the same exercises in 3 columns.</p></introduction>
+
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,x</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,x^2</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,x^{23}</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
+                        <p><m>\frac{d}{dx}\,e^x</m></p>
+                    </statement>
+                </exercise>
+                <exercise>
+                    <statement>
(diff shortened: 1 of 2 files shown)


rbeezer:
Dear Alex,

Feedback/Suggestions/Questions.

I spent a fair amount of time yesterday doing David's trick of looking at existing books.  Seems that when a group of exercises is arranged in columns the individual exercises are extremely short, and usually just an expression.  I think this is true of the existing examples in Judson's AATA, such as within

http://abstract.ups.edu/beta/html/sets-exercises.html

So I'd like to think about just allowing one-line per item in a multicolumn exercise group.  Then much of the alignment and wrapping into new columns becomes a non-issue, I believe.

I like how you got the minipage code into the preamble, but I'm very hesitant to give up source that has lists which only need adjustment to their enclosing environments to be re-purposed.  Even given your recent post about wrapfig and lists.  In other words, I'd rather have exercises prefaced by  \item[2.]  than enclosed in \begin\end{exercisegroupitem}{2. }

Do you think this is possible?

XML: I think maybe going across should be the default.  Though "by" makes very good sense here, I'd rather use more expressive attribute names and usually nouns.  For example, here 

direction="across"  and  direction="down"

would preferable, along with "cols".

XSLT: If no "cols" present, then I'd rather the generated LaTeX was as simple as possible.  In other words, the \RenewDocumentEnvironment would not be in the preamble to affect the multicols package, there would just not be a multicols in play at all for a one-column list.

I like what is going on here:

http://tex.stackexchange.com/questions/98969/enumeration-with-several-items-per-line-several-lines-and-nicely-formatted

http://tex.stackexchange.com/questions/149101/inline-arrangement-using-enumitem/149268

*except* that it seems the "shortlst" package is key and this is not a popular package (meaning you have to install it "by hand.")

There are several overlapping, and often contradictory, goals/demands/requirements for the LaTeX source in my mind, and I'm not even sure I have them all totally reconciled with each other.  So I appreciate your patience as we try to find our way.

HTML: I may not have the skill for the HTML side.  I sort of just hacked up the multicolumn lists as best I could.  You could look at what I did there, it shouldn't be too hard to move over in this case.

Lists: Eventually, I'd like to harmonize this with "regular" multicolumn lists.  Maybe  direction="down" goes multicolumn style and wraps into the next column, while direction="across" goes shortlst style (and is limited to short items)?

Rob

Alex-Jordan:
Thanks Rob, I'll factor all this in next time I have a chance.

> I like how you got the minipage code into the preamble, but I'm very hesitant to give up source that has lists which only need adjustment to their enclosing environments to be re-purposed. Even given your recent post about wrapfig and lists. In other words, I'd rather have exercises prefaced by \item[2.] than enclosed in \begin\end{exercisegroupitem}{2. }

>Do you think this is possible?

I can investigate and get back to you. One of the links has a solutions that looks promising using the `environ`, `multido`, and `calc` packages. Are those packages OK to use?

>XSLT: If no "cols" present, then I'd rather the generated LaTeX was as simple as possible. In other words, the \RenewDocumentEnvironment would not be in the preamble to affect the multicols package, there would just not be a multicols in play at all for a one-column list.

What you see in this branch right now is necessary to ensure that there is no formatting difference between say one column and two (other than the columns of course). In particular, the same vertical spacing after the introduction and before the conclusion was causing trouble if some separate mechanism was used for one column. (And as you may have picked up on, you can't just do \begin{multicols}{1} without that block of code.) But I may have learned enough in dealing with other issues since this started to get uniform spacing even through different constructions. I'll give it a try along with the other suggestions.

rbeezer:
Thanks for the reply.  The  calc  package is bad actor (redefines \setlength), 
but I think we have adequate defenses in place, look at the preamble when you 
have a tabular in the document and see notes about the extpfeil package and the 
conflict.  The other two packages look pretty standard as well.

Alex-Jordan:
I vote to close this. I didn't even use this with CLM. Instead I used what is at https://groups.google.com/forum/#!searchin/mathbook-xml-support/columns/mathbook-xml-support/gJcWvpN4r4s/ZUOOguhcvNgJ. I could submit this for approval instead, but it does use that `tasks` package.

Alex-Jordan:
Here is an MWE for LaTeX for exercises in columns, either reading across rows or down columns. The body is pretty clean. The preamble isn't too bad either. Compile and play. 


```
\documentclass{article}

\usepackage{tasks}
\usepackage{multicol}
\usepackage{enumitem}
\usepackage{showframe}

\NewTasks[
style=enumerate, 
column-sep=1cm, 
label-align=left, 
item-indent=2.5em,
label-offset=0.5em,
label-width=2em,
after-item-skip=12pt,
]{rowexercises}[\item]

\newenvironment{colexercises}[1]
{\setlength{\columnsep}{1cm}
\begin{multicols}{#1}
\begin{enumerate}[
align=left,
leftmargin=2.5em, 
labelsep=0.5em, 
labelwidth=2em, 
itemsep=12pt]}
{\end{enumerate} 
\end{multicols}
}

\begin{document}

\begin{rowexercises}(3)
\item[1.1] one\dotfill one
\item[1.2] two\dotfill two
\item[1.3] three\dotfill three
\item[1.4] four\dotfill four
\item[1.50] five\dotfill five
\end{rowexercises}

\begin{colexercises}{3}
\item[2.1] one\dotfill one
\item[2.2] two\dotfill two
\item[2.3] three\dotfill three
\item[2.4] four\dotfill four
\item[2.50] five\dotfill five
\end{colexercises}

\begin{rowexercises}(3)
\item[3.1] A lot of short words to fit in more than one line.
\item[3.2] two\dotfill two
\item[3.3] three\dotfill three
\item[3.4] A lot of short words to fit in more than one line.
\item[3.50] five\dotfill five
\end{rowexercises}

\begin{colexercises}{3}
\item[4.1] A lot of short words to fit in more than one line.
\item[4.2] two\dotfill two
\item[4.3] three\dotfill three
\item[4.4] A lot of short words to fit in more than one line.
\item[4.50] five\dotfill five
\end{colexercises}

\end{document}
```

Alex-Jordan:
Closing this as #370 is better and up to date with current dev.

PR #77, opened 2015-03-26
preface/acknowledgment was duplicating title in body in latex

Alex-Jordan:


Alex-Jordan:
This is probably moot after all the work on related topics since March.

rbeezer:
I think I once meant to merge this, sorry it slipped through the cracks.  I think this is all handled pretty much the same universally, so they all work properly, or not.  So I'll close it while I am here.

Issue #33, opened 2015-01-22
Document various things that aren't documented.


kcrisman:
As requested in #31, here is a ticket with a list of things that need to be documented.  I'll keep changing this list as things are done.
- The hierarchy of containers, e.g. that appendices only are at the level of chapters for LaTeX (and probably shouldn't work for HTML, though it does)
- The strict hierarchy in such environments - no `<p>` or other stuff outside the few legal containers in them.
- When you must use `<p>`, for instance apparently inside remarks.
- Math mode, ala percent discussion at #43
- What the heck the `<acro>` tag is for (?)
- _Exactly_ how to make an index - namely, `makeindex filename` for `filename.tex`
- The difference between `<md>`, `<me>`, `<mdn>`, `<men>`.  (And using `number="no"`.)  Eventually I figured it out, but because the `d` doesn't convey "rows" to me, I wasn't sure at first exactly what was going on.
- More to come.

rbeezer:
Better than OK - it'll work great.

rbeezer:
There is now an extensive discussion in the sample article about all the special, reserved and escape characters.

kcrisman:
Awesome.

rbeezer:
Definitions:  Now a complete example in sample article.

Theorem-like: Subsection 4.3 has been expanded.  It seems unnecessary to give an example of each, as they only vary by the actual element name.  I think there is now enough examples and enough guidance for folks to get started.

davidfarmer:
If any of these items remain, a new issue should be created.

PR #110, opened 2015-06-18
slightly updated index verbiage

kcrisman:
It should be clear right away that the default isn't creating an index in latex.


kcrisman:
Actually, I can't get the index to appear even after using `makeindex`.  Is this user error?

rbeezer:
I think this is sorted now on the support list.

https://groups.google.com/forum/#!topic/mathbook-xml-support/KBdDdFcvFow

I did make corrections to the source output in upcoming push at ac60ccef and will adress more thoroughly at #33.

PR #111, opened 2015-06-19
Finish algorithm

kcrisman:
The algorithm should make it 100% clear what the output is; after all, an algorithm could create a list and then not do anything with it.  More importantly, Python requires a `return` statement, unlike (say) R!


rbeezer:
It's in and will be part of upcoming push.  Thanks.

PR #137, opened 2015-07-13
Latex image to png

Alex-Jordan:
Adds png, eps, and all (tex, pdf, svg, png, eps) options for the output from latex-image components when the mbx script is used. 

Note that previously, mbx had no newline character at the end of the file. According to http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline, files should end with a newline character.

### DIFF
--- a/examples/sample-article.xml
+++ b/examples/sample-article.xml
@@ -855,7 +855,7 @@ the xsltproc executable.
             <subsection>
                 <title><latex /> images</title>
 
-                <p>There are several graphics engine packages that a <latex /> document can employ. Code from these packages renders diagrams automatically as part of normal processing of <latex /> files.  For HTML output the <c>mbx</c> script produces SVG versions of the pictures.  The packages should be loaded in docinfo/latex-image-preamble, which is also where global package settings should be made. These first examples are from the <url href="http://www.texample.net/tikz/examples/">TeXample.net</url> site.</p>
+                <p>There are several graphics engine packages that a <latex /> document can employ. Code from these packages renders diagrams automatically as part of normal processing of <latex /> files.  For HTML output the <c>mbx</c> script produces SVG versions of the pictures.  The script can also produce standalone TEX source files, PDFs, PNGs, and EPSs. The packages should be loaded in docinfo/latex-image-preamble, which is also where global package settings should be made. These first examples are from the <url href="http://www.texample.net/tikz/examples/">TeXample.net</url> site.</p>
 
                 <!-- http://www.texample.net/media/tikz/examples/TEX/noise-shaper.tex -->
                 <figure xml:id="figure-tikz-electronics">
--- a/script/mbx
+++ b/script/mbx
@@ -190,23 +190,25 @@ def latex_image_conversion(xml_source, dest_dir, outformat):
             latex_cmd = [tex_executable, latex_image]
             _verbose("converting {} to {}".format(latex_image, latex_image_pdf))
             subprocess.call(latex_cmd, stdout=devnull, stderr=subprocess.STDOUT)
-            if outformat == 'svg':
+            if outformat == 'all':
+                shutil.copy2(latex_image, dest_dir)
+            if (outformat == 'pdf' or outformat == 'all'):
+                shutil.copy2(latex_image_pdf, dest_dir)
+            if (outformat == 'svg' or outformat == 'all'):
                 svg_cmd = [pdfsvg_executable, latex_image_pdf, latex_image_svg]
                 _verbose("converting {} to {}".format(latex_image_pdf, latex_image_svg))
                 subprocess.call(svg_cmd)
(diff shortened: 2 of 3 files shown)


rbeezer:
Applied these commits just fine, though the third one seems to be empty (I can deal with it on a rebase once I merge this).

Testing with the tikz language images in Judson's AATA and the processing all goes smoothly.

PDF created by `mbx` script seems to have a resolution of `72dpi`, as reported by ImageMagick's command
```bash
identify --verbose <filename>
```
The conversion to PNG preserves this resolution and then the result is of very poor quality, lots of jaggies and fine detail lost in math glyphs.

Tried adding `-density 300` to the `convert` command in `mbx` with some success, though it seems the physical size information may not be preserved.  A `-resize 24%` before output file sent the image back to poor quality.

Maybe it would be best if original PDF had 300dpi?  This answer seems to suggest how to make this happen from within the `*.tex` file?   Though it is just using `shell-escape` to call `convert` anyway.  Unsure what impact this would have on other outputs (file sizes?).

http://meta.tex.stackexchange.com/questions/2435/how-can-i-provide-a-better-resolution-of-uploaded-pictures-created-with-the-docu

It seems the PDF is just a container with some sort of `PBM` (Portable BitMap) in it.  In other words, not a vector graphics image.  Which makes me wonder why it looks so good, and how the SVG is being built.

I'm in no hurry to solve all these conversion and scaling issues across all of MBX, but we might as well get started with some small steps.

Alex-Jordan:
You tested on `tikz` components, as in `<tikz>` elements? I only added options for `<latex-image>` components. Or have his `<tikz>` elements been converted to `<latex-image>`?

Alex-Jordan:
I remember now about the empty commit. I thought I had restored the lack of a newline character at the end of the file. But it turns out I hadn't. Then I learned that you are "supposed" to have such a character there anyway.

Anyway, should have undone the commit, but too late now if you have made a branch. Otherwise I could still take care of it.

rbeezer:
No, everything like this in AATA is inside of "latex-image-code".  They just 
happen to all be written in tikz.

I'm not intending to maintain (or test) "tikz" in any way.

And the processing all "works" just fine, it is just that the default properties 
of the resulting PNGs seem insufficient to me.

We should produce high-quality versions, and let people downsize if they feel it 
is necessary.  Impossible to do a good job in the other direction.  ;-)

rbeezer:
I have a branch already, but will clean up eventually anyway.

Not a problem, just wanted to make sure there was no missing content.

Alex-Jordan:
I lost track if I should still be working on this. Should I keep this branch open and (later) try different resolutions? Or did you plan to make those kinds of decisions? I'm not sure I know how to decide.

rbeezer:
I thought you might have some ideas about getting better resolution.  Really, I 
want better visual quality.

Can you try inserting

`"-density", "300",`

into the convert command before the "infile" and "outfile" (needs two items in 
the list as written) and see what you think?  Visual quality, file size, 
physical size?

I think that might be enough of a solution, but I had to move on when 
experimenting before.

Alex-Jordan:
So if I were running the `convert` command from the command line, it would go:
``convert foo.pdf -density 300 foo.png``
?

rbeezer:
I was getting good results placing it before *both* of the file names.

Alex-Jordan:
OK, so it needed to be 
`convert -density 300 foo.pdf  foo.png`
This makes the png pretty nice looking, just a bit noticeably less sharp than the pdf. The downside is the file size went from 4kb to 34kb. Is that OK?

Alex-Jordan:
Visually, I can't tell the difference between 
`convert -density 300 foo.pdf foo.png`
and 
`convert -density 300 foo.pdf -quality 100 foo.png`

but the latter is only 25kb on my test image compared to 34kb for the former. I'm not sure what "quality 100" really means, but it was someone's favorite on a thread I found. Should I go with this latter option for now?

rbeezer:
4x on resolution would naively suggest 16x for filesize, so I think that is good.

I'm not worried about bandwidth and storage at this stage.  If it becomes an 
issue we can chip away at it.

rbeezer:
Sounds good, we can always adjust later.  Done for tonight...

Rob

Alex-Jordan:
I messed things up somehow. I think I rebased my local branch for this onto dev after new changes to dev, and now GitHub won't accept my push. Rather than try to git my way out of this, would you like to make the change before you merge? All that has changed is line 203 of `mbx` has changed to 
``
png_cmd = [pdfpng_executable, '-density', '300', latex_image_pdf, '-quality', '100', latex_image_png]
``
(In case the line number is off, its a line that begins `png_cmd = [pdfpng_executable,...`)

rbeezer:
Added the `convert` commands and rebased to two commits (both with your name on them).  At 7ad01143

Thanks very much for working this one up.

PR #178, opened 2015-10-27
extract tex from server for MBX source problems that use PGgraphmacro…

Alex-Jordan:
…s.pl

Here it is. I'm iffy on one last thing, that perhaps you can declare nothing to worry about. When I copied your old `extract-webwork-tex.xsl` into  `extract-ww-tex-server.xsl` and `extract-ww-tex-mbx.xsl`, in the latter, I had to comment out some lines that were at the top. You can see some import/includes that are commented out. If I left them in there, then the whole base64 conversion process was only creating parts of the problem. It was like they were applying templates that I didn't want, killing templates I did want. 

Since my example is working, maybe we don't care? But since you understand why those lines were there in the first place, it's worth a thought. Will email you the pdf just cause.


rbeezer:
This code has been merged at 6c84f859

PR #204, opened 2016-02-19
Support for \pi in mag in HTML

Alex-Jordan:
Added the `str` namespace to `html.xsl` in the same way `common.xsl` does this, so that I could use `str:replace` to replace `\pi` with `\(\pi\)`. Now an author is closer to being able to treat the contents of mag in the same way they would treat the argument to `\num{}` or first argument to `\SI{}{}`.

Note: `siunitx` input for these arguments is much more sophisticated. This is still far from what an `siunitx` MathJax extensions would give us.


rbeezer:
Added an illustrative example to the sample article.  Works well.  Thanks.

Merge at aeb3da37

PR #209, opened 2016-02-27
Add mpg and kph to units

Alex-Jordan:
Adds "kilometerperhour", "kilometreperhour", and "milepergallon" to the options for a base attribute for a unit or per element. Result is "kph", "kph", and "mpg" respectively.

### DIFF
--- a/schema/dtd/mathbook.dtd
+++ b/schema/dtd/mathbook.dtd
@@ -519,7 +519,7 @@
 
 <!ENTITY % prefix "yocto|zepto|atto|femto|pico|nano|micro|milli|centi|deci|deca|deka|hecto|kilo|mega|giga|tera|peta|exa|zetta|yotta">
 
-<!ENTITY % base "ampere|candela|kelvin|gram|meter|metre|mole|second|becquerel|gray|sievert|degreeCelsius|celsius|coulomb|henry|ohm|siemens|tesla|volt|weber|hertz|katal|joule|newton|pascal|watt|lumen|lux|radian|steradian|degree|arcminute|arcsecond|day|hour|minute|hectare|liter|litre|percent|degreeFahrenheit|fahrenheit|pound|foot|inch|yard|mile|mileperhour|gallon">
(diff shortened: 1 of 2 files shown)


davidfarmer:
I think "milespergallon", with the s, and similarly for the metric
quantities, is closer to standard usage.  Also, milesperhour.

Alex-Jordan:
We use "meter" for this, not "meters", because "1 meter" is correct and "1 meters" is not. Similarly, "1 mile per gallon" is correct, while "1 miles per gallon" is not. Sticking to the pattern "1 _____" will make it easier to consistently guess the syntax without looking it up. At least that has been the thinking thus far. That, and making things match siunitx (when siunitx has a corresponding unit in the first place).

rbeezer:
Matching  siunitx  is important for the day when MathJax supports it and we rip 
out a lot of this stuff.  We want minimal damage to author's source.  Though I 
am getting pretty good at my XSL for fixing deprecations automatically.

davidfarmer:
I am arguing about author usage, not absolute facts.  Ask someone
what "mpg" stands for, and almost everyone will respond "miles per
gallon".

If you are going to type the same thing no matter what the actual
number is, then either "1 milepergallon" is correct and almost every
other case is wrong, or vice-versa.  But I guess your point is that
the users will know to model the syntax after the "1" case.  If
people know to think that way, then I can't argue against it.

Alex-Jordan:
Mentioning `siunitx` confused my point---sorry. The exact `siunitx` syntax is separated from the author input syntax and controlled within `mathbook-units.xsl` for this kind of reason. But using the singular for units would keep that particular consistency of `siunitx` in place.

Alternatively, we can just open it up to all variants. Allow "foot" and "feet", etc. Before I became acclimated to `siunitx`, I would get caught with stuff like `8\meters\per\second` and it would throw an error at me. So I really do understand your point about what is natural for an author.

rbeezer:
Has this discussion run its course (singular/plural on units)?  I'm happy with this as-is and prepared to merge.  Alex?  David?

davidfarmer:
Previous discussions persuaded me that singular is a good
way to do it.  We just need to be consistent and then the user
can always know what to do.

rbeezer:
Thanks for the reply, David.  I really like teh consistency on this one - singular/plural on my directory names always messes me up (and I never learn).  

Merged locally and to be pushed shortly.

PR #220, opened 2016-04-04
Fix bug with \amp inside a tikz \matrix

Alex-Jordan:
This addresses https://github.com/rbeezer/mathbook/issues/156, and the forum post: https://groups.google.com/forum/#!topic/mathbook-xml-support/Bqd87biU83Y.

Sets `\amp` as the ampersand replacement for within a tikz `\matrix`. Both in `mathbook-latex.xsl` and in `extract-latex-image.xsl`.

Added remarks about `\amp` to the sample-article, and put it to work by making the pgfplots-demo consist of a matrix of two pgfplots `axis` environments. The new one demos a 3D plot.

Tested by making tex and pdf, and using mbx script on the target image, and seeing html results. Updated the pdf and svg for this image that the repo stores.

If this is pulled, https://github.com/rbeezer/mathbook/issues/156 can be closed.


Alex-Jordan:
This has sat here too long; now is in conflict. Looks like a trivial thing to address, but since I'm not a git wizard, I'll just close this and resubmit.

PR #242, opened 2016-04-20
Jupyter Reference Spacing

Jahrme:
Fix reference spacing in Jupyter output by adding space between reference type and reference number.


Jahrme:
Going to work on the other issue with Jupyter generation I noticed  then package up as one large pull request.

PR #246, opened 2016-04-25
possible grammatical improvement?

kcrisman:
I never know what to do in such situations, but I like this better.  Feel free to ignore.


rbeezer:
Merged locally and pushed.  Thanks very much!

PR #250, opened 2016-04-27
Ww example doc changes

Alex-Jordan:
Removed the other MBX goodies section from ww sample chapter. Moved ww templates into a section of ww sample chapter. Made it so that if you use `<webwork />` in a project without any actual WeBWorK problems, that failing to declare a WeBWorK server does not produce a warning.


rbeezer:
Two edits:

* Edited remark about templates being somewhere else previously.  Reader is unlikely to be interested in the history and they've not been publicly available as completed anyway.
* The test on the `webwork` element needs `node()` for the matching.  Fixed and tested in the `sample-errors-and-warnings.xml` example.

Merged locally and pushed.

PR #252, opened 2016-04-28
`mbx` works in python 2 and 3 now

leingang:


rbeezer:
Fantastic!  I'll review carefully tonight.

leingang:
Yes, it probably needs more testing.  I was able to run the sample batch you were doing in the session.  The changes are minor syntax changes, except for one: in Python 2 the config parser module is called `ConfigParser`, but in 3 it's called `configparser`.  I think they are functionally identical, though.

rbeezer:
Dear Matthew,

This was a **huge** help.  Thanks.  First of many, I hope. ;-)

I had to fix up a little `exec/eval` confusion distinct from image generation, which was an improvement under 2 or 3.  Added a Python version report to the full `-vv` debugging output.

Merged locally, commit message edited, pushed.

Rob

PR #254, opened 2016-04-29
Fix WeBWorK table columns bug

Alex-Jordan:
For a table with cells that have multicol, we need to correctly calculate the underlying left column and right column for a particular cell correctly. The change here does that.


rbeezer:
Excellent.  Very tricky, no?  ;-)

Merged locally and pushed publically.  Thanks for keeping this maintained.

Issue #156, opened 2015-07-24
tikz matrix entry separator conflicts with side-by-side in LaTeX


rbeezer:
Reporter suggests that the ampersand used in a tikz matrix conflicts with the `\pushValignCaptionBottom` macro.  Example code: plain is problem-free, side-by-side fails.

https://groups.google.com/forum/#!topic/mathbook-xml-support/Bqd87biU83Y

```xml
<figure>
<image xml:id="image1">
<latex-image-code>
<![CDATA[

\begin{tikzpicture}
\matrix
{
  a & b\\
};
\end{tikzpicture}
]]>
</latex-image-code>
</image>
</figure>

<sidebyside>
<figure>
<image xml:id="image2">
<latex-image-code>
<![CDATA[

\begin{tikzpicture}
\matrix
{
  a & b\\
};
\end{tikzpicture}
]]>
</latex-image-code>
</image>
</figure>
```

cmhughes:
I'll experiment and report back.

Initial thoughts: it's weird that this causes issues, but that tabulars do
not.

rbeezer:
Thanks, Chris.  Yes, quite a corner case.  As I said on the support list, it'd be best if an author did not have to markup a matrix differently just for tikz inside a side-by-side.

So hopefully we can *configure* a fix.

Rob

cmhughes:
Before I look at this, did it turn out to be an xelatex issue?

Alex-Jordan:
This issue is really more of a `tikz` issue with `&` than a `sidebyside` issue, but I won't go down the rabbit hole of explaining. It means Chris is off the hook, and it's on me somewhat to come up with ideas since I'm responsible for the whole `latex-image-preamble` scheme.

The current standard is to use `\amp` inside tex code in MBX source, so rather than `\&` that A Pereira offered in the Google group post that started this, the solution should assume the author uses `\amp`.

Despite `\amp` having been defined in the preamble, it's not recognized within the `\matrix` command. I don't know why not, but that's beside the point. What is needed is the line:

``\tikzset{ampersand replacement = \amp}``

in the `latex-image-preamble` (which will be used in both tex output and for making image files). This should be used whenever the `tikz` package is loaded in someone's `latex-image-preamble`. Unfortunately, that's not something you can 100% automatically test for. Because `tikz` might be in play as a dependency of some other package the author put in their `latex-image-preamble`.

[...]

PR #257, opened 2016-05-02
Fix bug with \amp inside a tikz \matrix

Alex-Jordan:
This addresses #156, and the forum post: https://groups.google.com/forum/#!topic/mathbook-xml-support/Bqd87biU83Y.

Sets \amp as the ampersand replacement for within a tikz \matrix. Both in mathbook-latex.xsl and in extract-latex-image.xsl.

Added remarks about \amp to the sample-article, and put it to work by making the pgfplots-demo consist of a matrix of two pgfplots axis environments. The new one demos a 3D plot.

Tested by making tex and pdf, and using mbx script on the target image, and seeing html results. Updated the pdf and svg for this image that the repo stores.

If this is pulled, #156 can be closed.


rbeezer:
Nicely done - thanks.  Merged and pushed.  Sorry it took me so long to get to this one.

Shading works well for me with both `pdflatex` ans `xelatex`.  And that is a very nice looking graphic you added.

PR #265, opened 2016-05-13
Fix small misspellings

Alex-Jordan:
I ran spellcheck on the tex output from the webwork chapter I just added, and it caught a few more misspellings in the rest of the tex output. They traced back to either the latex.xsl or other xml parts from the author guide. Then I grepped for the same misspellings in all of mathbook and found them a few more places. (Often, just in comments.) So just because why not, I fixed them.

### DIFF
--- a/doc/author-guide/mbx-script.xml
+++ b/doc/author-guide/mbx-script.xml
@@ -26,7 +26,7 @@
             <li><p>Provide complete debugging output with bug reports</p></li>
         </ul>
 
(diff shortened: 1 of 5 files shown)


rbeezer:
Merged as-is, with just a note of the originating pull request.  Thanks for the corrections.

I have real trouble typing `environment`.  ;-)

Changes to DTD should be live at the website, with my new website-update script (partially implemented).

Alex-Jordan:
My biggest issues are `function` gets typed as `funciton`, and `ratio` gets typed as `ration`. That last one drives me crazy, because it's clearly bad wiring in my brain making my hands complete the `ion` suffix even when the `n` keystroke is well out of the way of the `io`.

PR #264, opened 2016-05-13
Author's Guide: WeBWorK chapter

Alex-Jordan:
WeBWorK chapter for author guide. Two subsections that are "planned". Also, my next project will be to remove the answer tag from webwork tags, putting var to that purpose by way of a form attribute. Documentation will change at that time.

PR #270, opened 2016-05-15
Webworkanswertovar

Alex-Jordan:
Various fixes discussed at #264, withholding a warning in one case where it is not necessary, and the answer -> var change discussed at https://groups.google.com/d/msg/mathbook-xml-support/gDKVt72u_Xc/mpC0Ibg1BgAJ


Alex-Jordan:
Added one more thing I had mentioned somewhere recently but forgotten about. An earlier iteration of what "category" eventually became had left category="popup" etc. as empty options in the category to format help converter. There is no need for this; in the current scheme, no author should be writing category="popup". If they want that, they should be using form="popup" in a var down in the body.

rbeezer:
Fixed up a conflict in the DTD due to a change that I made yesterday, light editing on one commit message.  Merged locally and pushed.

Thanks for all the clean-up and rearrangement.

PR #274, opened 2016-05-21
jade meets mathbook xml

haraldschilly:


rbeezer:
Thanks very much, Harald!  More praise through other channels.  Great way to start my weekend.

Light-editing of commit message, new commit so your first name survived, added to `examples/README.md` and pushed to `dev` branch (which is really the mainline these days).

haraldschilly:
nice!

and a small github pro tip: goto this url 
https://github.com/rbeezer/mathbook/settings/branches
to set the default branch of a repository to dev.

rbeezer:
Dear Harald - a never-ending source of good ideas.

`dev` branch is now primary, that's bugged me for a long time.  Reset - thanks!

davidfarmer:
What is the expected frequency of updating master?

If people are working on their own textbooks, but not thinking
in terms of developing MathBook XML, am I correct that they
should pull from master, not dev?

rbeezer:
Do you mean to ask me that before or after I merge master into dev (or something!), make a huge unrepairable mess, need to clone my own repo, and pull the many temporary branches I have going?  That was this morning's "valuable experience" - nothing lost.  But I **was** thinking last night about a geographically-distinct off-site backup location for my "private" work.  (My web host should suffice and I'll add that to website-update script.)   We should worry more about the bus factor.  ;-)

To answer your question.  Everybody is a contributor now and they should all pull `dev`, all the time.  I very rarely break things and the fixes are usually fast and trivial.  Also, I want these people to be in that habit long-term.  And things are changing fast, at leat fopr the next 3 months.  I am going to advance master now and then, but not on any kind of schedule.  I think this pro tip will help a lot.

PR #276, opened 2016-05-21
jade: bit cleanup

haraldschilly:
just some general cleanup, adding backmatter, etc.


rbeezer:
Thanks, Harald.  I panicked a bit when I saw the verbatim text!  (That's a great addition.)

PR #279, opened 2016-05-26
Windows documentation

daverosoff:
I have documented all the steps (I think) except the changes or additions we should make in the `mbx` script, which I think should be discussed elsewhere. Let me know about any changes you'd like me to make.

### DIFF
--- a/doc/author-guide/author-guide.xml
+++ b/doc/author-guide/author-guide.xml
@@ -20,6 +20,7 @@
         <xi:include href="topics.xml" />
         <xi:include href="mbx-script.xml" />
         <xi:include href="webwork.xml" />
+        <xi:include href="windows-installation.xml" />
         <xi:include href="backmatter.xml" />
 
     </book>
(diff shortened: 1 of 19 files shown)


daverosoff:
Note that one commit `Typos` ([259c816](https://github.com/rbeezer/mathbook/pull/279/commits/259c81635f079f1528070b1ba659de667be261d4)) was just me spotting what I thought were mistyped xml:id elsewhere in the author guide.

rbeezer:
Dear Dave,

Thanks very much for this.  Very helpful!  A few comments, a couple might be helpful for future contributions.

* I really like the pug/jade source, but I edited the xml version to be more in the form/style I like to use.  The source is available in the `examples/pug` directory for viewing.
* There are a few gotchas with the pug source, like if you split a raw MBX tag across lines, two attributes can end up adjacent (no space) it seems, which will be an error.
* The `<less />` and `<greater />` tags are just for "regular" text.  Inside verbatim (e.g. `<c>` you need to use the characters, but escaped for XML (e.g. `&lt;`).  I have more/better documentation on a branch.
* I had to "fix" `TikZ`.  We don't guarantee that `<em>` will be italic.  I'm not sure how to really accomplish the correct way to write that.
* I did use `<em>` on "evil" - I'm not sure you are defining evil, but maybe you are. ;-)
* Reorganized to put this in an appendix, where I want to place specfics of *other* programs, not MBX itself.
* Reorganized commits - I hope you don't mind.  Your name is on everything that is your work and I split out my edits in case any of them are bad.

Thanks again.  First of many, I hope.  ;-)

Rob

daverosoff:
Thanks Rob! I certainly approve of all the changes. For me using pug was an opportunity to hard-wrap lines at 80 characters, a habit I'm reluctant to break. For some reason I thought the XML output from pug used long lines, but now I see otherwise. Thanks for taking the time to get my contributions into house style.

Dave

rbeezer:
Eventually we should have a filter to remove the extra newlines and this will be 
a great way to go (with no extra editing).

PR #285, opened 2016-06-01
Add chapter on reading the DTD to Author Guide

mitchkeller:
I've taken a first stab at something for the author guide on reading the DTD. It covers all the things I've found useful in perusing the DTD, but I'm happy to take suggestions for additional things to cover.

There's also a placeholder section for for the Relax NG schema, but I've just noted that will be written once we have a Relax NG schema that's not just coming from `trang`.


rbeezer:
Dear Mitch,

This looks great!  Is it ready to go, or do you plan more edits?  Two small things I noticed, which I can fix in the course of merging this, so just let me know (I know you have a conference coming up, and I'm happy to proofread for these little things).

* `specifcation` very early
* `subsection inside a subsection` looks like a typo

Thanks for doing this, it'll be a big help.  I'm working on #279 right now and will get to this next.

Rob

mitchkeller:
This is ready to go as far as I'm concerned. You're correct on the two things you noted. Please fix in the course of merging.

rbeezer:
Dear Mitch,

Thanks again for this!  I did a bit of editing to get things into house style.  You may not have realized there is "add on" XSL for a "tag" and "attribute" element.

* Lots of "c" now use the extra two elements
* Wrapped "DTD" in "init"
* Removed contractions

If you do a `git diff` on the edit commit and its predecessor and use the `--word-diff` flag you can readily see exactly what I edited.

Squashed commits, merged locally, and pushed.

Rob

mitchkeller:
- Ah, "tag" and "attribute" are very nice!
- For my own knowledge, what's the advantage of always wrapping DTD in "init"?
- Your (sensible) cleanup of "label{blah}" and "ref{blah}" to include backslashes reveals why I didn't have them there: look at the HTML version and see that for some reason, the \ref{blah} appears to be passed off to MathJax and rendered as ???, so that should be fixed.

rbeezer:
`init` is short for "initialism"  Right now LaTeX renders it in small caps, so looks a bit less jarring.  `abbr` tag in HTML, which could get some styling eventually.  Companion to MBX `abbr` and `acro`.

Yep, I suspected the backslashes had given you some trouble.  I think MathJax is trying to interpret the `\ref` even though not wrapped as math.  Took me a couple hours late last night to see the light on that one.  I'll need to experiment.  See #288

PR #286, opened 2016-06-02
Fixwebworkknowls

Alex-Jordan:
This is all to address the issues from https://groups.google.com/d/msg/mathbook-xml-support/FLYOt5XaUpg/21k_T1HaBwAJ

Ultimately this is just about reordering js loads, so that iframeResizer works. I tested for unintended consequences by making a simple article with every js utility that MBX uses all on one page, and there were no errors on Firefox, Chrome, or Safari (aside from the errors that espy is failing to load properly, a separate issue that has been around for a while).

### DIFF
--- a/examples/webwork/Makefile
+++ b/examples/webwork/Makefile
@@ -127,6 +127,16 @@ sample-chapter-latex:
 	xelatex sample-ww-chapter.tex; \
 	xelatex sample-ww-chapter.tex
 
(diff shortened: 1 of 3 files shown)


rbeezer:
Excellent!  Thanks very much.  Your JS-fu just went up a notch.  ;-)

Hoped-for coincidental fix of #190 didn't happen.  :-(

PR #287, opened 2016-06-02
Mac pathnames not quite case-sensitive

kcrisman:
See http://apple.stackexchange.com/questions/22297/is-bash-in-osx-case-insensitive


daverosoff:
Thanks for the clarification.

rbeezer:
Thanks.  I edited the commit message slightly (and put the stackexchange URL into the source nearby).

Merged locally and pushed.

PR #311, opened 2016-06-16
Optional List Conclusion

Jahrme:
According to the DTD the `conclusion` of a `list` is not optional:

```
<!ELEMENT list (title?, index*, introduction?, (ol|ul|dl), conclusion)>
```

The DTD should have the `conclusion` of a `list` be optional just like the `introduction`:

```
<!ELEMENT list (title?, index*, introduction?, (ol|ul|dl), conclusion?)>
```


rbeezer:
Thanks!  Good catch.  Merged locally, pushed, etc.

PR #322, opened 2016-06-21
Untangle my previous PR

daverosoff:
Dear Rob,

I hope you'll find this much simpler. I still did a sequence of commits to keep changes logically separated. I believe all your style edits (contractions, etc.), use of `<cd`/`<tag`/`<ellipsis` are preserved, and I have tried to be careful with whitespace. Please feel free to close my other open PR.

Thanks,
Dave

### DIFF
--- a/doc/author-guide/editors.xml
+++ b/doc/author-guide/editors.xml
@@ -93,7 +93,7 @@
                     <p>Use Quick Add Next again to add the next instance to the selection, which will then typically be disconnected.</p>
                 </li>
                 <li>
-                    <p>Continue to Quick Add Next as many times as you like. Use Quick Skip Next (<c>Ctrl+K, Ctrl+D</c>/<c>Cmd+K, Cmd+D</c>) to jump over instances you'd like to leave alone. If you go too far and select in error, hit <c>Ctrl+U/Cmd+U</c> to undo.</p>
+                    <p>Continue to Quick Add Next as many times as you like. Use Quick Skip Next (<c>Ctrl+K, Ctrl+D</c>/<c>Cmd+K, Cmd+D</c>) to jump over instances you would like to leave alone. If you go too far and select in error, hit <c>Ctrl+U/Cmd+U</c> to undo.</p>
                 </li>
                 <li>
                     <p>Make your modification, only one time.</p>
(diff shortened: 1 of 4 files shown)


rbeezer:
Nicely done.  Doesn't that feel better?  ;-)

Dropped a single period haging off a `cd` and causing a goofy extra nearly-blank line.  And modified commit messages some (in particular to a single line on one).

Rebased, mild edits, merge all done locally.  So trash your branch and pull your work from mainline.

Thanks for getting this all shipshape - looking forward to more.

Content question:  I thought Python's `os.path` module, etc meant you didn't have to make accomodations for different platforms.  Is it *how* you suggest installing programs that make this necessary?  I wonder what platform WSL will report?

daverosoff:
Yes, it does feel better.

As for your question, I reflexively want everything to work in the MSYS2 shell that ships with Git on Windows. It already emulates a lot of GNU/Linux—useful stuff like `find`, `xargs`, `grep`, `sed`, and so on. It expects path names differently from how `os.path` presents them. But since you ask, I will try the old `mbx` in the Windows command shell and see how it works. I never thought to do this, since I would never use it if a Linux-like were available but a lot of Windows folks don't have any Linux experience. It might be possible to detect the shell and switch among several different options.

I think WSL will behave as though it were an honest-to-god Linux. I'll be able to test it when those features are released into a mainstream build, probably in about a month. If I bug my sysadmin he'll let me use a dev build, which I'm willing to do if people think it's urgently necessary.

My prediction would be that the usual (Linux) instructions will work with no modification on WSL, including `apt-get` and so on.

rbeezer:
> Yes, it does feel better.

I thought so. ;-)

> As for your question, I reflexively want everything to work in the MSYS2 shell

That's what I thought.

> It already emulates a lot of GNU/Linux—useful
> stuff like |find|, |xargs|, |grep|, |sed|, and so on. It expects path names
> differently from how |os.path| presents them. But since you ask, I will try the
> old |mbx| in the Windows command shell and see how it works.

It would be good to detect the shell, or maybe we just have a switch for MSYS32, 
-w (for "windows").  Truth in advertising.  ;-)

> I think WSL will behave as though it were an honest-to-god Linux. I'll be able
> to test it when those features are released into a mainstream build, probably in
> about a month. If I bug my sysadmin he'll let me use a dev build, which I'm
> willing to do if people think it's urgently necessary.

That would be my guess, too.  And no rush.  Thanks again.

daverosoff:
I wasn't able to find a Python trick that could tell the difference between the Git Bash/MSYS shell and the CMD shell. But I think the point is moot since I was not able to get the CMD shell to work with `pdfcrop`; see my most recent comment at #328.

IMO, the incorporation of `pdfcrop` removes the CMD shell as a viable interface to MBX.

PR #334, opened 2016-07-13
Update mbx script for windows and document in AG 

daverosoff:
As you can see I outsmarted myself with the commits and `git rebase -i`. I'm getting better, though.


script/mbx
```diff
@@ -224,7 +234,10 @@ def latex_image_conversion(xml_source, xmlid_root, data_dir, dest_dir, outformat
             if (outformat == 'svg' or outformat == 'all'):
                 pdfsvg_executable = get_executable(config, 'pdfsvg')
                 _debug("pdfsvg executable: {}".format(pdfsvg_executable))
-                svg_cmd = [pdfsvg_executable, latex_image_pdf, latex_image_svg]
+                if plat == "Windows":
+                    svg_cmd = [pdfsvg_executable, "-density", "300", latex_image_pdf, latex_image_svg]
```
daverosoff: The images were very blurry since the default density for ImageMagick's `convert` is something crazy low like 72dpi.

rbeezer:
As always, thanks!

Two edits that I put onto a commit with my name on it (to spare you responsibility) and one comment follow.  Also `author/personname` is still my fault.

* `latex_image_pdf = "{}.pdf".format(filebase)` was duplicated, removed it
* I kinda liked "get_platform()" rather than a global variable, but its OK
* `_debug()` writes its own prefix, and I don't care how the lines break, so no need to use two calls

Rob

PR #341, opened 2016-07-18
Resolve DTD Violations in Sample-Article

Jahrme:
Adjust the Sample-Article and DTD resulting in fewer DTD violations in the Sample-Article.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -241,9 +241,9 @@ the xsltproc executable.
             -->
             <theorem xml:id="theorem-FTC">
                 <title>The Fundamental Theorem of Calculus</title>
-                <index>Fundamental Theorem of Calculus</index>
+                <index><main>Fundamental Theorem of Calculus</main></index>
                 <statement>
-                    <p>If <m>f(x)</m> is continuous, and the derivative of <m>F(x)</m> is <m>f(x)</m>, then <me>\definiteintegral{a}{b}{f(x)}{x}=F(b)-F(a)</me><index>Buried in p/statement/theorem</index></p>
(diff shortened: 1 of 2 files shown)


davidfarmer:
I think it is more typical to put the statement of a definition before the notation.
So I suggest reverting the changes around line 290.

rbeezer:
Thanks, Jahrme.  `%markup` should be contained within `%linear` so one edit in the DTD looks duplicative.  I'll experiment/fix when I merge, so don't worry about it.

Thanks for doing this, KDC will be pleased. ;-)  The remainder "need work" so are not worth fixing now.

rbeezer:
@davidfarmer - the `notation` will not be displayed at all here - placement is for "List of Notation" to function.  It is "metadata" of sorts, very similar to `index`.

I want all this stuff up top, so I can relax the necessity of `statement` predictably.

davidfarmer:
Since many people use the sample article to figure out what to do,
I didn't want it to give the impression that notation should come first.

Eventually we want to develop an MBX style guide, and that will
likely have notation after statement.

rbeezer:
"notation" will not be visible to the reader there.

It is information about the "definition", like its "title".  It is (a) a target 
for the cross-reference from the "List of Notation", (b) content for the "List 
of Notation".

It needs to handled specially, and typically drops a target - we want that 
target at the top of the "definition" in practice (eg PDF).

To put it at the end (a) is contrary to practice with "title", "caption" and 
"index", (b) unnecessarily complicated for implementation.

IF we make a new "block" titled "Notation", then I agree that the right 
presentation is after the definition (and so should be authored there), and we 
can talk about other places where it might appear.

Rob

rbeezer:
Thanks, Jahrme.  Lists are going to change (must lice in `list` or `p`) so I reversed a few things.  DTD **will** catch up soon.  At  770dd766  but forgot to put PR number in commit message, so it won't magically appear here.  Has your name on it anyway.

PR #343, opened 2016-07-22
Modified assemblage to make pretty blue boxes

oscarlevin:
The actual changes are around lines 741 and 2857 (other changes are line endings that got updated to LF automatically by my local machine).  Tested and this works with my book, the sample article is giving me trouble for (unrelated, I believe) font reasons.  Instead of having an mdframed environment, I changed to creating an assemblage environment, which gets styled using mdframed.

I did originally get an error because xcolor was added after mdframed in the latex preamble.  I put an (extra) xcolor invocation earlier, and this removed the pdflatex error.  Depending on how everything gets loaded eventually, this could be fixed (perhaps the TODO comment near assemblage relates to this?)

### DIFF
--- a/xsl/mathbook-latex.xsl
+++ b/xsl/mathbook-latex.xsl
@@ -738,9 +738,9 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:text>%% assemblage: minimally structured content, high visibility presentation&#xa;</xsl:text>
         <xsl:text>%% Package for breakable highlight boxes&#xa;</xsl:text>
         <!-- TODO: load just once, see webwork -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Dear Oscar,

Thanks very much for this.

* I got all the line-ending stuff sorted, **and** was able to put your name back onto the resulting commit.
* A bit wonky under `xelatex` with sample article example - I'll post on support list to give it broader exposure.
* Changes pushed, website examples updated later tonight.

When `git diff` shows lots of line-ending nonsense, go

```
git add -p xsl/mathbook-latex.xsl  (y/n on a per-hunk basis)
git commit -m "message for commit"  (note, no -a)
```

Should then have un-commited changes that are all undesirable

```
git stash save
git stash drop
```

And you have the desired changes in the commit and nothing unwanted anywhere.

PR #344, opened 2016-07-23
Add Music Capabilities

Jahrme:
Added examples and documentation to *Humanities in Action* (available in `/examples/humanities/`).

Added empty tags for standard accidentals in western music.
`<doublesharp/>`, `<sharp/>`, `<natural/>`, `<flat/>`, and `<doubleflat/>`

Added markup for scale-degrees.
`<scaledeg>5</scaledeg>`

Added note construction which properly formats both alphabetic and numeric pitch classes with accidentals and octave designations used in scientific pitch notation.
`<n pc="C" acc="sharp" octave="4"/>`

Added chord construction for creating chord labels with optional alterations.
Simple Chord:
```
<chord root="A" mode="major"/>
```
Complex Chord:
```
<chord root="C sharp" mode="minor" bps="7" suspended="yes" bass="E" parentheses="no">
    <alteration>omit 5</alteration>
</chord>
```

### DIFF
--- a/examples/humanities/hia.xml
+++ b/examples/humanities/hia.xml
@@ -1,5 +1,6 @@
-<?xml version="1.0" encoding="UTF-8" ?>
+<?xml version="1.0" encoding="UTF-8"?>
 <mathbook xmlns:xi="http://www.w3.org/2001/XInclude" xml:lang="en-US">
+    <docinfo/>
     <book xml:id="hia">
         <title>Humanities in Action</title>
         <frontmatter xml:id="index">
@@ -15,5 +16,6 @@
             </preface>
         </frontmatter>
         <xi:include href="poetry.xml"/>
+        <xi:include href="music.xml"/>
     </book>
 </mathbook>
--- a/examples/humanities/music.xml
+++ b/examples/humanities/music.xml
@@ -1,0 +1,172 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<chapter xml:id="music">
+    <title>Music</title>
+    <section xml:id="symbols">
+        <title>Musical Symbols</title>
+        <example>
+            <title>Accidentals</title>
+            <ul>
+                <li><p><c>&lt;doublesharp/&gt;</c>: <doublesharp/></p></li>
+                <li><p><c>&lt;sharp/&gt;</c>: <sharp/></p></li>
+                <li><p><c>&lt;natural/&gt;</c>: <natural/></p></li>
+                <li><p><c>&lt;flat/&gt;</c>: <flat/></p></li>
+                <li><p><c>&lt;doubleflat/&gt;</c>: <doubleflat/></p></li>
+            </ul>
+        </example>
+    </section>
+    <section xml:id="scaledegs">
+        <title>Scale Degrees</title>
+        <p>Scale degrees are created using the <c>&lt;scaledeg&gt;</c> tag which places a <q>diacritic circumflex</q> over the degree. The circumflex looks best when attached to single-digit numbers.</p>
+        <example>
+            <title>Scale Degrees</title>
+            <ul cols="2">
+                <li><p><c>&lt;scaledeg&gt;0&lt;/scaledeg&gt;</c>: <scaledeg>0</scaledeg></p></li>
+                <li><p><c>&lt;scaledeg&gt;1&lt;/scaledeg&gt;</c>: <scaledeg>1</scaledeg></p></li>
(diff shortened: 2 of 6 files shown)


Jahrme:
Conflict created by commit 3718488 regarding `xcolor` in LaTeX.
I will look at the conflict more closely tomorrow.

rbeezer:
I was afraid of that.  :-(  (I recall your earlier warning.)  Welcome to 
software development.

I can restict xcolor use/loading, but generally, we need to get this sorted. 
Let me know what more you discover and how I can help.

Jahrme:
I have resolved the issue locally, but as it required a rebase I was not sure how to handle updating the pull request. Suggestions?

rbeezer:
Reviewing right now.

1.  I'll have at least one change request.  Add a new commit here for that, I'll 
incorporate it into correct current commit.

2.  Just post instructions for xcolor fix - I think I know what it is and it is 
not too involved.  Otherwise, tell me it is complicated and we can try plan B.

Review in a minute.

rbeezer:
Looked through this carefully, despite needing work.

Does the "pass option" drill just need to move to the new load location? Is the conflict just a merge conflict, or is it more functional? I can make that fix.

One question: `root="C sharp"` (for example) can be many different combinations? It would not make sense to use "C-sharp" say? In other words, there is a 2-item space-separated list, not a single attribute value? I didn't examine the code very closely.

`fontspec` is loaded iff xelatex is used. Where you condition on "n", "chord", etc. please check if pdflatex is being used and fail gracefully (use \typeout{} for informative message?) Otherwise xelatex is in use and you don't need to load fontspec.

There is a pull request for lualatex support. Maybe peek at that and plan for that being a possibility relative to previous paragraph.

Does HIA ever say you need to load a "lilypond" package? I had to.

Looks great!

rbeezer:
If it is easier - you can open a new pull request and reference it as a continuation of this one.  No problem for me.

davidfarmer:
I am always leery of one-letter tags, such as "n".  Those are
valuable, and we should have a high bar for assigning them.

Am I correct that all of the existing "n" tags must occur inside
some musical environment, and so we have not stopped
"n" from being used in another context?

If not, then it would be good to hear the argument for why
"n" means "note" and cannot ever mean anything else.

rbeezer:
On 07/23/2016 12:14 PM, davidfarmer wrote:
> I am always leery of one-letter tags, such as "n". Those are
> valuable, and we should have a high bar for assigning them.

Agreed.

> Am I correct that all of the existing "n" tags must occur inside
> some musical environment, and so we have not stopped
> "n" from being used in another context?

No, they will show up in the course of almost any sentence in any paragraph 
discussing some aspect of music.

> If not, then it would be good to hear the argument for why
> "n" means "note" and cannot ever mean anything else.

In a music theory textbook they will be as ubiquitous as "m" in a math textbook, 
or a "c" in a computer science textbook.  We (Prof Hutchinson, Jahrme, R, me) 
have had many discussions to get to this one.  Rob Hutchinson has a good head 
for what we are doing.  This is definitely a place to make authoring easier with 
a sensible abbreviation.

And I could not think of a natural use for "n" in a future scenario.

Rob

Jahrme:
I can change the structure of `root` and `bass` attributes to process notes like `C-sharp` and `G-doubleflat` as they function as a single entity. My only concern is creating confusion by using a `-` in `root` and `bass` attributes, while using a space in the `mode` and `bps` attributes (where one might have `mode="min major"` or `bps="6 5"`). With `mode` and `bps`, we are processing two related but independent pieces of information (chord quality vs quality of the seventh/ninth/etc. for `mode`; intervals from the bass for `bps`).

As for loading a "lilypond" package, I have never had that happen with *HIA*. The only package that should be necessary is the "lilyglyphs" package (which is included in the preamble). However, it is possible that I have simply never encountered this requirement as I have LilyPond installed on my computer. If you could elaborate slightly on what you encountered I can investigate and address this in *HIA*.

Currently:
I have made the necessary changes regarding lilyglyphs & fontspec with xelatex vs pdflatex support. If you think `root` and `bass` should use `-` as their connective character I will make that change, close this PR, and make a new PR including all of these adjustments (including changing HIA to reflect these changes).

rbeezer:
We can leave root and bass as-is, I was mostly thinking of attributes for the 
DTD (remember "not tested"?).  I think Relax NG will handle this situation better.

Its only  lilyglyphs.sty  that is a problem.  Sorry to be vague.  Under  xelatex 
  we have fontspec already, under  pdflatex  including  fontspec  raises an 
error.   So we should just assume fontspec and xelatex, and die with an error if 
someone tries using pdflatex on musical content.

Go ahead and make a new PR - reference the old one, and also fix  xcolor?

PR #346, opened 2016-07-24
Add Music Capabilities

Jahrme:
Continuation of PR #344

New PR contains updates in response to feedback as well as resolves the conflict that arose.

Additionally, there is now support for accidentals in pdfLaTeX (without `fontspec`).

### DIFF
--- a/examples/humanities/hia.xml
+++ b/examples/humanities/hia.xml
@@ -1,5 +1,6 @@
-<?xml version="1.0" encoding="UTF-8" ?>
+<?xml version="1.0" encoding="UTF-8"?>
 <mathbook xmlns:xi="http://www.w3.org/2001/XInclude" xml:lang="en-US">
+    <docinfo/>
     <book xml:id="hia">
         <title>Humanities in Action</title>
         <frontmatter xml:id="index">
@@ -15,5 +16,6 @@
             </preface>
         </frontmatter>
         <xi:include href="poetry.xml"/>
+        <xi:include href="music.xml"/>
     </book>
 </mathbook>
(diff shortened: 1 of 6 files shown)


rbeezer:
Close!

* The `musix` font commands definitely need some additional line breaks in the XML to migrate to the LaTeX.
* My `pdflatex` run chokes at the first use of `\sharp`  (wants to use math mode?).  Maybe related to above?  (I added some line breaks to get this far, but am unsure about the format of these commands.)
*  Will you want to make the announcement on teh list, once done and merged?
*  Would Prof Hutchinson let you put his 4-page handout in sample article?   I can't recall.  I guess it'll have scores so needs to wait.

You can just add a new commit onto this branch and I'll get it OK and bury it into the right place.  Don't work too hard on the commit message.

Jahrme:
I added the line breaks (an oversight in my haste to add `pdflatex` support) and fixed the `\sharp` issue by grabbing all five accidentals from the musix font so that they are available in both math and text mode.

I can make the announcement on the list, and I will ask Prof. Hutchinson if he would let us use his handout in the sample article. If so, I can submit a PR adding the example.

rbeezer:
Nicely done!  That's great that you got `pdflatex` support working.

* Squashed to two commits, with your name and PR # on them.
* When you advertise, mention that HTML lists are wonky, and that most of the work is in `mathbook-common.xsl` since you are leveraging LaTeX and MathJax for music notation (people will find that interesting).  (I should put this out on the MathJax list!)
* If you think Hutchinson's handout will happen fast, you may want to wait until we get that in before announcing.  Make it a new section of the "Music" chapter.  Give him credit for the material.
* Live on website, but I think I will turn off knowling examples right now.

PR #360, opened 2016-08-02
Change mbx and Author's Guide for alternative pdf2svg on Windows

daverosoff:
These changes fix the issues described in these posts:

https://groups.google.com/d/msg/mathbook-xml-support/RlMq2zTd-AY/wRnjFlO1AAAJ
https://groups.google.com/d/msg/mathbook-xml-support/f48l5LBQCjA/pCl-KzChAAAJ

The installation procedure is slightly different since another utility has to be installed.


rbeezer:
Nice to see `get_platform()` is back.  ;-)

I'll review more carefully in the  morning.  Thanks.

rbeezer:
Excellent!  Thanks for the quick work.  Merged with just an addition to the commit message with PR number.

*  I like the 64-bit check.  I was going to suggest that.
*  If you are in SageMathCloud, I would say you are no longer in Windows, even if your browser is being hosted there.  No need to adjust that warning now, but maybe you can pick it up next time.

Thanks again.  -Rob

PR #366, opened 2016-08-08
fix trivial typo

kcrisman:
found while looking at how lists now work


rbeezer:
Thanks!  Merged at 497345d7

PR #369, opened 2016-08-12
Fix typo in case entry in DTD.

mitchkeller:
Went to try to validate my book this morning and the validator threw an error from the DTD. As best as I can tell, this is what was intended.


rbeezer:
Merged, and seems to take care of it, sample article does not fail on any `case` elements.  Thanks for the catch and the fix!  -Rob

PR #370, opened 2016-08-12
Let exercisegroups have cols

Alex-Jordan:
A PR to let exercisegroups have cols. This is version 2.0 of #76 which I'll close and xref to here.

Output: compare 
http://spot.pcc.edu/~ajordan/temp/derivatives.pdf page 28 
to 
http://mathbook.pugetsound.edu/examples/sample-article/derivatives.pdf page 28

And compare
http://spot.pcc.edu/~ajordan/temp/section-11.html
to
http://mathbook.pugetsound.edu/examples/sample-article/html/section-11.html

HTML uses a div around the exercises within the exercisegroup that uses `display:flex` and has a class for the number of cols. But on a narrow screen the items will reconfigure into fewer columns. The extra CSS for this is at http://spot.pcc.edu/~ajordan/temp/sample-article.css. I suspect that the same classes `cols2`, etc. along with flex box styling could be used to improve how lists get their multiple columns. [Something is not right with that at present. When I look at http://spot.pcc.edu/~ajordan/temp/section-5.html in Firefox the ul items are not aligned with each other..]

LaTeX uses [tasks](https://www.ctan.org/tex-archive/macros/latex/contrib/tasks?lang=en) package. It has LPPL status "maintained", and is maintained by the author [Clemens Niederberger](http://www.mychemistry.eu/)  whose name is on [a lot of LaTeX contributions](https://www.ctan.org/author/niederberger). tasks was born out of a much larger package exsheets that is used for exercise sheets. tasks is just the item enumeration method, but by popular demand he split it off into its own package.

It is possible to use `\item` for the item demarker in a tasks environment. However if you do that, you cannot have regular lists inside such an item. So I have it set to `\exercise` for now. 

#76 has been there for a long time, and no rush to give feedback on this one. But I will use this for the next CLM release, and so if it sits here as a PR I will try to rebase it regularly.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -1836,7 +1836,7 @@ the xsltproc executable.
 
                     <exercise>
                         <statement>
-                            <p><m>f(x)=x^3</m>, <m>\frac{df}{dx}</m>.  This sentence is just a bunch of gibberish to be certain that the problem's left margin aligns with the problem number's indentation.</p>
+                            <p><m>f(x)=x^3</m>, <m>\frac{df}{dx}</m>.  This sentence is just a bunch of gibberish to check where the second line of the problem begins relative to the first line.</p>
                         </statement>
                     </exercise>
 
@@ -1907,6 +1907,26 @@ the xsltproc executable.
                         <statement><p>Second exercise of second group.</p></statement>
                     </exercise>
                     <conclusion><p>Conclusion to second exercise group.</p></conclusion>
+                </exercisegroup>
+
+                <exercisegroup cols="4">
+                    <introduction><p>An <c>&lt;exercisegroup&gt;</c> can have a <c>cols</c> attribute taking a value from 1<ndash />6. Exercises will progress by row, in so many columns. On a small screen, the HTML exercises may reorganize into fewer columns.</p></introduction>
+                    <exercise>
+                        <statement><p><m>1+2</m></p></statement>
+                    </exercise>
+                    <exercise>
+                        <statement><p><m>3+4</m></p></statement>
+                    </exercise>
+                    <exercise>
+                        <statement><p><m>5+6</m></p></statement>
+                    </exercise>
+                    <exercise>
+                        <statement><p>Add seven to eight.</p></statement>
+                    </exercise>
+                    <exercise>
+                        <statement><p><m>9+10</m></p></statement>
+                    </exercise>
+                    <conclusion><p>This feature was designed with short <q>drill</q> exercises in mind.</p></conclusion>
                 </exercisegroup>
 
             </exercises>
--- a/schema/dtd/mathbook.dtd
+++ b/schema/dtd/mathbook.dtd
@@ -580,6 +580,7 @@
 
 <!ELEMENT exercisegroup (introduction?, exercise+, conclusion?)>
     <!ATTLIST exercisegroup xml:id ID    #IMPLIED>
+    <!ATTLIST exercisegroup cols   CDATA #IMPLIED>
 
 <!ELEMENT exercise (title?, index*, ((statement, hint*, answer*, solution*)|(introduction?, webwork, conclusion?)))>
     <!ATTLIST exercise xml:id ID    #IMPLIED>
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -2761,10 +2761,30 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- Never hidden so calling hidden-knowl-text raises error -->
 <!-- There is no head ever -->
 <xsl:template match="exercisegroup" mode="head" />
-<!-- Body is just all content             -->
-<!-- introducttion, exercises, conclusion -->
+<!-- Body is introduction,       -->
+<!-- exercises wrapped in a div, -->
+<!-- conclusion                  -->
 <xsl:template match="exercisegroup" mode="body">
-    <xsl:apply-templates />
+    <xsl:apply-templates select="introduction"/>
+    <xsl:element name="div">
+        <xsl:attribute name="class">
+            <xsl:text>exercisegroup-exercises</xsl:text>
+            <xsl:text> cols</xsl:text>
+            <xsl:choose>
+                <xsl:when test="not(@cols)">
+                    <xsl:text>1</xsl:text>
+                </xsl:when>
+                <xsl:when test="@cols = 1 or @cols = 2 or @cols = 3 or @cols = 4 or @cols = 5 or @cols = 6">
+                    <xsl:value-of select="@cols" />
+                </xsl:when>
+                <xsl:otherwise>
+                    <xsl:message terminate="yes">MBX:ERROR: invalid value <xsl:value-of select="@cols" /> for cols attribute of exercisegroup</xsl:message>
+                </xsl:otherwise>
+            </xsl:choose>
+        </xsl:attribute>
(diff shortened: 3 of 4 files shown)


rbeezer:
Dear Alex,

Looking good.  Sorry to be slow working up to the big contributions.  Paragraph columns soon.

Questions and comments from a careful, but not final, review.

1. Max of six due to the CSS?  Reasonable, me thinks, but maybe @davidfarmer wants to peek at the proposed CSS as part of this?
2. I was never wild about "by" as an attribute name, so I see that is moot for the moment.
3. LaTeX is looking much better (indentation of conclusion).  I'll have to check my HTML.
4. Removed lines can go away (not just commented out) since the git history has them. (\newlist, \setlist here).
5. I have "console-typeout" for introductions that are peers of subdivisions, but I don't see a need for them at every last instance of an introduction (now that they are appearing in more places).  Ditto for conclusion.
6. Can you do all the `tasks` lengths in `ex` or `em` units?  We have a wide range of possible fonts now.
7.  Does the width of a column seem skinny compared to the width of the space between?  Or is it just me?

Updating this branch as-is should be fine, it'll rebase onto `dev` on my end with no problem since development has slowed down some.

Thanks,
Rob

Alex-Jordan:
> Max of six due to the CSS? Reasonable, me thinks, but maybe @davidfarmer wants to peek at the proposed CSS as part of this?

I wanted to avoid inline styling, and instead use classes like "cols3" for three-column exercisegroups. These classes would ultimately be defined in mathbook.css, so there can only be finitely many. All they specify is the basis width for flex items, and I imagine they could be reused in a rewiring of lists that used flex display. So, six seemed a reasonable cap.

> I was never wild about "by" as an attribute name, so I see that is moot for the moment.

Yes, I opted to *only* (perhaps, just for now) allow progression by row. Simplified a lot, and to me at least it seems like the more reasonable option for exercises in an exercisegroup. [Any refactor of general lists though should perhaps permit either direction.] Side note: over the years, I've passed out in-class worksheets with column numbering that sometimes went vertically (via multicols) and sometimes went horizotally (via tasks). Students mess it up both ways.  Many will inevitably ignore the actual enumeration, and either progress horizontally or progress vertically based on some sort of bias they bring with them. I have not observed one direction being the bias more commonly.

I'll get to the changes you mentioned and update the PR, plus maybe one more change that I'll only bother explaining if my experiments tell me it is worth including.

Alex-Jordan:
Addressed 4, 5, and 6. 

Not sure what you mean by 7, but perhaps you are just perceiving something that is a consequence of the specific content I put into the example. If you add `debug=true` to the `\NewTasks` options, you will get boxes outlining the "cells" for the content. In HTML, I use Firebug to highlight the exercise's `<article>` so I can see the full width being reserved. In both output forms, it looks wide enough to me (although maybe not if you go up to 6.)

----

Up until about three weeks ago, subsequent lines of a tasks item would not run flush with the number. Rather, they would run flush with where the first line of content started. See what I mean in page 28 of the pdf linked above, where you see "Add seven to eight" and the "eight" goes on to the next line.

Roughly three weeks ago, inspired by [my SE post](http://tex.stackexchange.com/questions/324624/can-tasks-be-used-to-make-a-list-with-itemindent-labelwidth-labelsep), the author enhanced `tasks` so that you can have subsequent lines run flush with the number. This requires updating to version 0.12. 

Whether you are in 0.12 or something older, the execution is the same if you don't need subsequent lines to align with the number. Good. If you are in version 0.12, to achieve the alignment with the number, you have to adjust certain options for the `\NewTasks`. If you change these options, and someone has an older version of `tasks`, it will compile but look bad.

So question: leave it alone, with subsequent lines not aligning with the number? (like the "eight" in the linked pdf?) Or code it to version 0.12 of `tasks` and encourage people to use up-to-date tex distributions? I suspect leave it alone, but the indentation behavior will not match what is currently used for exercisegroup exercises. [Which is not a problem to me, but maybe to others.]

davidfarmer:
The CSS seems okay.  I added it to mathbook-add-on.css, with somewhat more specific selectors.
Please test (by omitting sample-article.css) before merging.

rbeezer:
Thanks - this was going to be disruptive if we didn't coordinate.  Having the 
CSS in place will help.  I'll try to close this one over the weekend.

rbeezer:
Re: Version 0.12.

Would you like to experiment with conditional setup in the preamble based on package date?  I think it would be straightforward enough with `\@ifpackagelater` as described in this post:

http://tex.stackexchange.com/questions/13304/which-package-version-am-i-using

It'd be a good example to get this right the first time, and it can be copied later in similar situations.

I don't care how messy the preamble is, that is part of the beauty of what we are doing.  But do include some sort of TeX comment that always gets written in front of the dual-setup so people (publishers) know what is going on and can edit knowledgeably.

I want a new MBX user to have a very good experience out-of-the-box, so "too bad it looks crummy, please get a modern TeX distribution" should be avoided whenever possible.

I think current PR does not require 0.12, and if so I plan to merge it.  0.12 setup/features can be a new PR.

Rob

Alex-Jordan:
> I think current PR does not require 0.12,

That's right. This PR sets it up regardless of the tasks version, to look like what is on page 28 of http://spot.pcc.edu/%7Eajordan/temp/derivatives.pdf. Leveraging new formatting tools in 0.12, we could make the word "eight" in "Add seven to eight." fall in line under the exercise number (which happens to be 8). That would be desirable (by a small epsilon) only because horizontal space is so precious in print.

I'll act on 0.12 later once this PR is processed.

rbeezer:
Thanks!  This is an excellent contribution.  I'll let you announce (before/after 0.12), unless you are too busy (and I can do it).

I got a skinny width to scrunch down to three columns.  Very nice.

Spacing was an optical illusion.  ;-)  Interesting:  I could get `3+4+5+6+7` to fit for LaTeX, but only three terms in MathJax (which must have non-elastic spacing, or something).

I added a comment about `debug=True` for LaTeX output and made Exercise 6 be `3+4+5`, and smashed all the commits together.

I think you only have one PR outstanding. ;-)

@davidfarmer - thanks again for getting the CSS in place in advance.

Rob

Alex-Jordan:
There is an issue with pursuing the 0.12 enhancement. Since you would want the subsequent lines to be left flush with the number, you would want the number to be left-aligned in its box. 

OK, but the box for the number is of a fixed width, even though the number itself is not. The content of the item only starts after the number's box, possibly after a small skip forward of some size. This presents a problem if you do not know how many digits will be in the number. If you set it up to handle possible three-digit numbers with no bad overlap, then when a one-digit number is there it's a very wide gap owing to the extra space you reserved for three-digit numbers. 

This is not an issue with the current configuration, because I set the number to be right-aligned in its box (and you recall, the entire box floats off to the left of the item.) This kind of thing would only become an issue in the current configuration if like, five-digit numbers are used, and I don't think anyone will have that many exercises :)

I could do the 0.12 thing planning for two-digit numbers, and it looks OK for one-digit as well. But then three-digits looks bad. 

Or I could leave this alone until a better solution presents itself.

rbeezer:
Dear Alex,

Thanks for all the explanation.  That all makes sense, having played with the 
debug boxes when I did the final review.

Let's wait until we get a better idea - there's more pressing things to do. 
Let's leave this PR closed, and make a "wish list" item to remind us to keep 
0.12 in mind.

Can you make a new issue, and maybe just cut/paste this post and anything else 
into it, or maybe just link back to parts of this?  Maybe I'll make a new label 
- "future possibilites" or something?

Rob

PR #375, opened 2016-08-20
WeBWorK version

Alex-Jordan:
This is to address the issue discovered in https://groups.google.com/d/msg/mathbook-xml-support/4ThAGRu4xFI/bHY9LGNXBQAJ. 

For xsltproc, you can declare `webwork.version` to be 2.11 and you will get "password=". Otherwise, you get "course_password=". This will handle all the webwork knowls.

For the tex, that is extracted using the mbx script. My feeling was that it would be better to pass both "password=" and "course_password=" when the script makes its server call. But if you would prefer, some sort of switch could be added to the script to single out 2.11. It just seems like a really narrow thing to make a switch for.

### DIFF
--- a/doc/author-guide/webwork.xml
+++ b/doc/author-guide/webwork.xml
@@ -205,6 +205,7 @@
                 <prompt>$ </prompt>
                 <input>xsltproc --stringparam webwork.server https://webwork.myschool.edu &lt;xsl&gt; &lt;xml&gt;</input>
             </console>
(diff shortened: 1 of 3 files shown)


rbeezer:
Thanks for hopping on this one, Alex.  (Small typo in closing `p` tag of Author's Guide, which I fixed.)  I figured this was more urgent than the other two outstanding PR's.

PR #376, opened 2016-08-26
Oneof

Alex-Jordan:
I converted the first subsection of APEX exercises to WeBWorK, and it led to these additions/edits to WeBWorK XSL. Also, a method to control randomization for the official HTML/print version, now with example in the webwork sample chapter.


Alex-Jordan:
BTW, the name of this branch comes from how it started: I only intended to add detection for loading the `parserOneOf.pl` macro library. But the branch grew.

rbeezer:
Thanks!  Merged as-is.  -Rob

Issue #363, opened 2016-08-05
Brackets in LaTeX Titles


Jahrme:
LaTeX titles (specifically short titles) need to be handled more carefully. Currently, if there are brackets in the `title` of a subdivision, the brackets wreak havoc on the title. The simplest solution is to wrap the short title in braces to hide the brackets.

Currently:
```
\subsection[A [short] title]{A [short] title}
```

Better:
```
\subsection[{A [short] title}]{A [short] title}
```

Minimal example (saved as `txt` due to GitHub rejecting `xml` and `mbx`):
[brackets-in-titles.txt](https://github.com/rbeezer/mathbook/files/404517/brackets-in-titles.txt)

(HTML does not appear to have any issues with brackets in titles.)

Jahrme:
Fix at proposed PR #377

Jahrme:
Solution (PR #377) accepted, closing issue.

PR #377, opened 2016-08-31
Protect Titles in LaTeX

Jahrme:
LaTeX short-titles cannot contain brackets (see issue #363). This request would protect brackets within short-titles by wrapping the short-title in braces.

The solution was found on StackExchange [here](http://tex.stackexchange.com/questions/99495/inside-an-optional-argument).


rbeezer:
Dear Jahrme,

Thanks very much for the fix.  Prof. Jackson will appreciate it.

* I thought to perhaps insert the group conditionally (on a right bracket), but the SE post convinced me otherwise - thanks for including that.

* I added a commit with a test case in the sample article and snuck the SE link into the comments in one of the places where it happens.

Rob

PR #383, opened 2016-09-09
Add titles to induction examples.

davidfarmer:
The examples are immediately after Principle 2.1.1.

http://mathbook.pugetsound.edu/examples/sample-book/html/section-math-induction.html

new version looks like this:

http://sl2x.aimath.org/development/MBX/html/section-math-induction.html


rbeezer:
Thanks, David! I knew those colons were sitting there waiting for something! ;-)  Merged and live at website examples.

Rob

PR #390, opened 2016-09-14
HTML Style Options for `ups-writers` customizations

Jahrme:
This PR changes the CSS color scheme to a UPS specific color scheme and prevents examples being knowlized.

### DIFF
--- a/contrib/ups-writers/ups-writers-html.xsl
+++ b/contrib/ups-writers/ups-writers-html.xsl
@@ -11,6 +11,8 @@
 <xsl:import href="../xsl/mathbook-html.xsl" />
 
 <xsl:output method="html" />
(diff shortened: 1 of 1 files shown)


davidfarmer:
On a recent thread it was discussed that non-knowlized examples
could be marked with an "L" along the left margin.  This pull
request will not enable that, because there needs to be a new
css class attached.  I am happy to work on the css once we decide
what class to add (assuming you want that feature).

> This PR changes the CSS color scheme to a UPS specific color scheme and prevents examples being
> knowlized.

rbeezer:
Merged.  Looks great!  Thanks for taking care of this.

rbeezer:
And put the hash in the wrong place.  Again.  So just did a forced push, which I hope does not cause problems if I did it quick enough.

PR #406, opened 2016-09-29
Added some useful index entries

davidfarmer:
Added some index entries for terms I repeatedly need to look up,
and also a few others that I saw.

After committing I saw that there were already some index entries.
It would be good to compile the index as part of the sample article.

I didn't know the standard for adding index entries: end of a paragraph,
immediately after the term first appears, or between the section title
and the first paragraph.  So I did examples of all three possibilities.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -377,7 +377,7 @@ the xsltproc executable.
                 <output>12.005</output>
             </sage>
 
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, David.  And for restoring the PR queue!  I'll get to this over the weekend.

rbeezer:
Merged at  0dbed72a

Thanks for your first contribution!  ;-)

One bad element, which I corrected and combined into your commit.  Contribution exposed two shortcomings:

* #409 
* Some stray debugging stuff on level-3 index entries, ecd4ad9bf

Not live yet at website, but perhaps very soon.

PR #420, opened 2016-10-25
Detect additional WeBWorK macros from pg-code

Alex-Jordan:
ORCCA, APEX, and the MT Hood project are all using these macros (and I'm tired of manually adding them when I need them :) )

### DIFF
--- a/xsl/mathbook-webwork-pg.xsl
+++ b/xsl/mathbook-webwork-pg.xsl
@@ -282,27 +282,19 @@
         <xsl:text>  "MathObjects.pl",&#xa;</xsl:text>
         <xsl:text>  "PGML.pl",&#xa;</xsl:text>
     </xsl:variable>
-    <!-- accumulate macros evidenced by some aspect of problem design -->
+    <!-- accumulate macros evidenced by some aspect of problem design      -->
+    <!-- for details on what each macro file provides, see their source at -->
+    <!-- https://github.com/openwebwork/pg/tree/master/macros              -->
+    <!-- or                                                                -->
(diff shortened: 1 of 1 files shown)


rbeezer:
"bizarro"?

Alex-Jordan:
You pushed a button that I am always excited to talk about :) It was one time I put abstract algebra to use in coding a WeBWorK contribution.

At the level of math I teach, we often need to seriously distinguish between x^2+2x+1 and (x+1)^2; between sqrt(12) and 2sqrt(3); etc.

WeBWorK is not a CAS, and so it doesn't naturally distinguish these kinds of things. Normally, it would declare sqrt(12) and 2sqrt(3) equal by computing their decimals. And it would declare x^2+2x+1 and (x+1)^2 equal by evaluating both at random inputs.

bizarroArithmetic redefines the arithmetic operators to put a "bizarre" field structure on R. If you use this tool in the right way, you can expect the answer (x+1)^2, and when a student enters x^2+2x+1, (x+1)^3/(x+1), or whatever, get a message like "Your answer is algebraically equivalent to the correct answer, but not in the right form. Perhaps it should be simplified, factored, expanded..." It checks with both regular arithmetic and then bizarroArithmetic.

rbeezer:
> and I'm tired of manually adding them when I need them

That's a good reason to act fast. ;-)  Merged as-is.

Building sample chapter makes three warnings about macros being "implied" by the problem's construction.  I'm not sure if they are new or not.  Maybe you'd want to remove the redundancies?

Thanks for the explanation of bizarro arithmetic - I was curious ...

PR #423, opened 2016-10-27
Additional units of time for units handling

Alex-Jordan:
Added "year (yr)", needed for ORCCA.

Also added:
week (wk)
month (mo)
decade (decade)
century (century)
millennium (millennium)

[Note: don't misspell "millennium" as "millenium", since "annus" and "anus" mean very different things.]


rbeezer:
Merged as-is, but for edit of commit message.  Thanks for keeping this up-to-date.

Issue #399, opened 2016-09-22
LaTeX for WebWork problems from a WW server have changed


rbeezer:
So we don't parse it properly and there are error messages about poorly formed XML and nothing gets rendered.

See https://groups.google.com/d/topic/mathbook-xml-support/QQBsGWQFNPs/discussion

rbeezer:
@Alex-Jordan When this is resolved are you able to make a PR to the WW code with comments in the vicinity of these markers, so they don't change without warning?

Alex-Jordan:
I'll do that.

rbeezer:
Fix is now at #475

rbeezer:
Looks like this is done, except for some mop-up.  See #475 for details.

PR #440, opened 2016-11-29
Protect path to .pg file in LaTeX

Alex-Jordan:


rbeezer:
First commit looks good.  I presume you will be piling on some more (to extend second commit?).  Post here when you need feedback or are ready to merge.

Alex-Jordan:
More coming. The second commit here is an isolated upgrade, although it feels like it intersects with the next thing because both involve server tex. Second commit here is like problem 6 from page 12 of the chapter 1 APEX draft I posted recently in the APEX forum. Everything about that problem is MBX-generated, not PG-generated. _Except_ the image.

Next big commit will be the one to fix #399. But it's no good to merge that until the WeBWorK project merges my PR over there. I can't say how long it might be until they get around to looking at it. If we want to, we could tap Mike's shoulder about it.

I might keep piling on some small commits related to WeBWorK & PG. But maybe don't wait for the commit to fix #399, because we don't know the timeline on the WeBWorK merge.

Alex-Jordan:
1465ba7 does several related things. Throughout, I examined all the combinations of:

- a problem is free-range versus within an `exercises`
- a problem has a `title` versus does not
- a problem has a `introduction` versus does not
- a problem has a `conclusion` versus does not
- a problem has `stage`s versus does not

and settled on this layout here. Counting `introduction`, `statement`s, `hint`s, `solution`s, and `conclusion`, they all are preceded by a `\medskip` unless they are the very first thing, in which case they are inline with the exercise number (which may be followed immediately with a `title`). 

Except `stage`s ("Part N") will always get a carriage return. They also get the `\medskip` if they are not the first stage. [So, the first stage has no extra whitespace between it and the (exerciseNumber|exerciseNumber-title|exerciseNumber-introduction|exerciseNumber-title-introduction).]

(Note: In the sample article, the one place where there is an edrcise with stages, it so happens that vertical spacing is tight on that page and a lot (all?) of the rubber space from the `\medskip`s is being compressed. )

Alex-Jordan:
I've rebased from dev and I see now that messes with my ability to keep pushing this branch. Going to close this, delete this branch on GitHub, then make a new PR.

PR #453, opened 2016-12-09
Webwork issues 2

Alex-Jordan:
New attempt at #440.

Latest commit addresses two non-WeBWorK issues, but they were too trivial and I was too lazy to make a new branch. One issue is that the unit list in the DTD wasn't in the same order as mathbook-units.xsl. My fault from when I added those extra time units about a month ago. The other issue was that kilometerperhour, kilometreperhour, and milepergallon had been removed from the DTD accidentally (I think) on the same day they had been added. See 9c1148c189bfed9de8541b5573622130a199c8ef and d0517a3f5a1b577a83ac1ac405287b66b886fc44.


Alex-Jordan:
Oops, screwed something up with the last commit. CLosing, and I'll be submitting these as individual PRs.

PR #455, opened 2016-12-11
Protect path to .pg file in LaTeX

Alex-Jordan:
Sometimes the .pg file ends up with an underscore in its name. This is a cheap way to protect against that.


rbeezer:
Merged, will push shortly.  Thanks for the maintenance.

PR #457, opened 2016-12-11
Adjust layout of WeBWorK exercise components in print

Alex-Jordan:
Array of adjustments to WeBWorK print output. All informed by now having the opportunity to examine output from ORCCA, APEX, and the MHCC project.

### DIFF
--- a/xsl/mathbook-latex.xsl
+++ b/xsl/mathbook-latex.xsl
@@ -2575,6 +2575,15 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:apply-templates select="*" />
 </xsl:template>
 
+<!-- Most introductions are followed by other sectioning blocks (e.g. subsection) -->
+<!-- And then there is a resetting of the carriage. An introduction preceding a   -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged and pushed recent changes.

Comments:
* `count(...)=0` is a red flag.  An empty node-set tests as false, so usually that is sufficient.
* Above attracted me to code for `stage/statement`.  It seemed cleaner to break this out, then I consolidated code duplication.  Tested with diffs on sample-chapter and saw no changes.  Then a
```xsl
<xsl:text>\noindent%&#xa;</xsl:text>
```
seemed apparent and its removal created diffs that seemd to reinforce that view.  Holler if I've exceeded my authority.  I put this on a separate commit to accept full blame.
* Added a modal `serial-number` routine for stages/parts to enforce consistency (but there is nowhere else yet to use it!).  Might enable good xref's someday to stages.
* I've only meant the `console-typeout` for subdivisions (of which a "sectional" introduction is one).  Emitting them for each WW problem will be too much noise and swamp the signal, so I took those out.

Thanks for this one - visually-appealing high-quality LaTeX is an important feature.

Issue #374, opened 2016-08-19
implement quantity inside math mode


Alex-Jordan:
The good folks at MathJax built this:

https://github.com/mathjax/MathJax-third-party-extensions/tree/master/siunitx

I'm not sure it is entirely working though, based on loading their provided sample page. Monitor this, experiment, and when time is available, sit down and enable `<quantity>` inside math elements.

Alex-Jordan:
siunitx due for MathJax 2.7, currently in beta: https://www.mathjax.org/mathjax-v2-7-beta-now-available/

rbeezer:
Excellent!

Alex-Jordan:
Hasn't been updated since 2017, but there is [github.com/burnpanck/MathJax-siunitx](https://github.com/burnpanck/MathJax-siunitx).

Alex-Jordan:
Lee McPherson (PCC chemistry) has developed this with funding from an internal PCC award:

https://github.com/limefrogyank/siunitx-pcc

Demo:
https://limefrogyank.github.io/siunitx-pcc/

Activate MathJax accessibility tools (and possibly reload the page) to see some speech strings. Awaiting an internal PCC meeting in a few weeks and some feedback from the MJ crew. With luck it will eventually become part of MJ and not a third party thing.

PR #459, opened 2016-12-11
Add quantity element to WeBWorK

Alex-Jordan:
Up to now, a quantity in a webwork would do nothing at all for the archival version of the problem. And the HTML version would be broken due to either a thinspace character, special fraction slash, or incorrect formatting of \pi. This version of the quantity templates just uses plain space, plain slash, and puts \pi in PGML delimiters.

Additions to the stylesheet tag needed for the \pi -> [`\pi`] conversion.

### DIFF
--- a/xsl/mathbook-webwork-pg.xsl
+++ b/xsl/mathbook-webwork-pg.xsl
@@ -22,6 +22,8 @@
 
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0"
     xmlns:xml="http://www.w3.org/XML/1998/namespace"
+    xmlns:str="http://exslt.org/strings"
+    extension-element-prefixes="str"
 >
 
 <!-- This file is a library of routines to convert parts of      -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Two comments:

* Does this get stress-tested somewhere?  I mean like in the sample chapter?  I suspect it may be being used in production, but if it breaks we want to hear about it early.   I don't see `unit` or `quantity` in a grep of the sample chapter.  Is it somewhere else?
* `position()=1` is dangerous, as it does not do what you think, I think.  Insert an XML comment in front of whatever is being probed and the comment is the first node inside that element.  See if you can break your code. ;-)  Try something like  `not(preceding-sibling::*)` to just cruise for elements - you can add a test for `text()` if that is a possibility here.

Alex-Jordan:
Harder than I thought this would be. Special characters that were causing trouble were just thinspace and special-fraction-slash. But upon building a stress test, more things from mathbook-units.xsl are causing trouble too:
micro
ohm
degree
arcminute
arcsecond

Alex-Jordan:
Just figured out how to add characters to base64 vocabulary. So I think I can make this very clean. Except exponents may end up with `^` since there is no sup element in PGML to use. Not sure how to handle sup and sub for fractional units. Need to see if PGML can be made to do those things.

Alex-Jordan:
FYI, going to put this on hold. Working on #374 might make this much cleaner and save duplicate work. If #374 were implemented then all WeBWorK units could go inside math mode and just use siunitx syntax. I think that would require adding siunitx to the native WeBWorK print preamble, but that would be no big deal.

PR #469, opened 2016-12-17
Add PGML line breaks before list and after image

Alex-Jordan:
These are needed to fix a pressing issue with the MT Hood project. 

The main one is putting a line break at the start of a list. If there is a p with text content and then a list, it fails to build the list correctly without this line break.

The other change is putting line breaks following the image command. Without this there are instances of a sentence that should be starting on a new line starting to the right of an image. The same pair of line breaks is already there for tabular, so this also makes image parallel to tabular in this way.


rbeezer:
Merged and pushed.  Will likely update website examples later today.  THANKS!

PR #472, opened 2016-12-20
windows 10 instructions for seeing file extensions

kcrisman:
And of course thanks to @daverosoff for actually doing this so I could try it out!


rbeezer:
Thanks, KDC, and thanks Dave for having a look.  Merged, etc.

kcrisman:
Great!

(By the way, interesting note is that dll files have their extensions whether you turn this feature on or off.)

PR #475, opened 2016-12-21
Webwork212 2

Alex-Jordan:
Now with `xsl:number`.

One thing that might raise an eyebrow is removing lines 357--372 of the mbx script. These two things are not necessary with 2.12 and above, and we now indicate in the author guide that with 2.11, server tex problems were never truly an option (since I never made the appropriate PR to WeBWorK.)

One of these things comes from the PGML preamble, and is commented out in the new block pf PGML preamble code. The other thing comes from a WeBWorK subroutine which now (appropriately) does nothing when the problem is called via the anonymous route.

Some things were formerly not accounting for stages, hence certain changes in `mathbook-webwork-pg.xsl`.

Thanks for your patience and careful reviewing.

### DIFF
--- a/doc/author-guide/webwork.xml
+++ b/doc/author-guide/webwork.xml
@@ -12,12 +12,12 @@
     <author>Alex Jordan</author>
 
     <introduction>
-        <p>With a <webwork /> server (version 2.11 or higher) and a little setup work, you can embed <webwork /> exercises in your MBX project. HTML output will have interactive problem cells. PDF output will contain static versions of exercises. And all such exercises can be archived by the <c>mbx</c> script into a file tree to be uploaded onto the <webwork /> server for use in the <q>traditional</q> way.</p>
+        <p>With a <webwork /> server (version 2.12 or higher, limited support with 2.11) and a little setup work, you can embed <webwork /> exercises in your MBX project. HTML output will have interactive problem cells. PDF output will contain static versions of exercises. And all such exercises can be archived by the <c>mbx</c> script into a file tree to be uploaded onto the <webwork /> server for use in the <q>traditional</q> way.</p>
     </introduction>
 
     <section>
         <title>Configuring a <webwork /> Course for MBX</title>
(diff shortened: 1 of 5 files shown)


Alex-Jordan:
Forgot to note: WeBWorK 2.12 is not displaying hints when `displayMode=tex` when the problem is accessed through the anonymous route. There is a hint in the new sample-chapter stress test section, but I commented it out for this reason (because it causes a compilation error when there is an image in the hint). 

If you uncomment it and `make sample-chapter-server-tex` and examine the resulting files in `scratch/pdf/webwork-tex`, you'll see no hints section. This is not a result of bad `mbx` script code, because even if you look directly at a url with `displayMode=tex`, there is no hint section present. The mechanism that gives you permission to view hints is somehow broken in this anonymous, `displayMode=tex` situation. But it's a WeBWorK issue, not a MBX issue. 

I need to move on to other pressing matters and won't investigate on the WeBWorK end for a while. What this means is that MBX users wouldn't have access to WeBWorK hints in print output. They still get them in HMTL.

rbeezer:
Looks good.

*  I am 99% sure the `webwork//` prefix in the `<xsl:number />` is redundant.  I removed it and tested.
*  Serial numbers of stages are broken.  Not sure why.  Will investigate.  #479
*  Hints are at  #480
*  Also closing  #399

As usual, thanks for plowing through this one.

Alex-Jordan:
Is this ready for an alert to JT and others in the forum? Looks like you are having a busy day; ping me when I should say something. (Or maybe we wait until the stage serial numbers are settled?)

rbeezer:
Dear Alex,

Thanks for checking.  Just wrapping up a week of fixing outstanding bugs, 
starting new projects (musical scores are going to be very cool!), etc.  One son 
is still sleeping....

Yes, blast away on the list.  Mention serial numbers at your discretion - it 
should be a really simple fix.  Prod WW'ers on hints at your discretion.

Mountain trip with the "boys" Mon - Wed, so not so active for a bit.  Then must 
get serious about Joint Math Meetings.

Thanks for all your contributions.

Rob

rbeezer:
Dear Alex,

Thanks for checking.  Yes, blast away on the list.  Mention serial numbers at your discretion - it should be a really simple fix.  Prod WW'ers on hints at your discretion.

Thanks for all your contributions.

Rob

PR #499, opened 2017-01-12
Document perl-free WeBWorK problems

Alex-Jordan:
It's always been possible to take a shortcut and not use any perl in a pg-code section with a WW problem. This makes it official by describing how to do that, and giving the right warnings about doing things that way.

### DIFF
--- a/doc/author-guide/webwork.xml
+++ b/doc/author-guide/webwork.xml
@@ -95,10 +95,51 @@
 
         <subsection>
             <title>Perl-free Problems</title>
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks!  Merged at 58757dea.

*  Capitalized one instance of `perl` in the body (yes, I know it is not an acronym).
*  No new code, but a testing example in the sample WW chapter would be welcome.
*  You can advertise on the Google group as you see fit.

Thanks again for the contribution.

PR #504, opened 2017-01-13
Add WW component configuration survey to sample chapter

Alex-Jordan:
A useful addition for establishing the CSS that handles WW exercise layout.


rbeezer:
Got it.  Thanks!

PR #509, opened 2017-01-15
Pgml indentation

Alex-Jordan:
Two commits here. First one only corrects DTD violations I found in the WW sample chapter.

Next one is all for the sake of list items. Prior to this, there are issues with list enumeration when the list items have too much going on inside. It all comes down to how you basically have to maintain the same level of indentation to persist with a list at a certain depth.

ol/li and ul/li got merged into one template, so that might make for an ugly section of the diff. 

Added lots of lists in the stress test section. Some of them are still broken! But I am convinced it is not our fault, after examining the PG output. As noted in the section, there may be a PGML bug (or we are trying to use PGML for something it was not intended to do.)

One other thing that this commit does is cut down on over liberal line breaks in the PG output. Too many line breaks made the PG code hard to read with human eyes, with long voids of vertical white space. Much better now, and everything in the sample chapter still works as before.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -248,11 +248,11 @@
 
                 <introduction>
                     <p>One of the strengths of <webwork /> is its ability to give intelligent feedback for incorrect answers.</p>
-                    <ul>
+                    <p><ul>
                         <li><p>There is general feedback for when the student's answer is in an entirely different ballpark from the correct answer. Try entering something like <q>y</q>.</p></li>
                         <li><p>There is general feedback for when the student's answer is not in the right form. Try entering <q>x<circumflex />2*x<circumflex />3</q>, which, right or wrong, is unsimplified.</p></li>
                         <li><p>And problems can be written to detect and respond to common mistakes. Try entering an answer where you multiply the two exponents (instead of adding them, which would be correct.)</p></li>
-                    </ul>
+                    </ul></p>
                 </introduction>
 
                 <webwork>
@@ -1011,24 +1011,24 @@
                     </setup>
 
                     <statement>
-                        <ol label="a.">
+                        <p><ol label="a.">
                             <li>
                                 <p>Suppose the correct answer is <m><var name="$answer1"/></m>.</p>
-                                <ul label="square">
+                                <p><ul label="square">
                                     <li>
                                         <p><var name="$answer1" width="15"/></p>
                                     </li>
-                                </ul>
+                                </ul></p>
                             </li>
                             <li>
                                 <p>Suppose the correct answer is <m>\displaystyle <var name="$answer2"/></m>.</p>
-                                <ul label="square">
+                                <p><ul label="square">
                                     <li>
                                         <p><var name="$answer2" width="15"/></p>
                                     </li>
-                                </ul>
+                                </ul></p>
                             </li>
-                        </ol>
+                        </ol></p>
 
                     </statement>
 
@@ -1062,24 +1062,24 @@
                     </setup>
 
                     <statement>
-                        <ol label="a.">
+                        <p><ol label="a.">
                             <li>
                                 <p>Suppose the correct answer is <m>\displaystyle <var name="$answer1"/></m>.</p>
-                                <ul label="square">
+                                <p><ul label="square">
                                     <li>
                                         <p><var name="$answer1" width="15"/></p>
                                     </li>
-                                </ul>
+                                </ul></p>
                             </li>
                             <li>
                                 <p>Suppose the correct answer is <m>\displaystyle <var name="$answer2"/></m>.</p>
-                                <ul label="square">
+                                <p><ul label="square">
                                     <li>
                                         <p><var name="$answer2" width="15" evaluator="$answer2evaluator"/></p>
                                     </li>
-                                </ul>
+                                </ul></p>
                             </li>
-                        </ol>
+                        </ol></p>
 
                     </statement>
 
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Added one more commit. Prior to this, all ordered list items were arabic regardless of depth (unless specified otherwise in source.) Similarly with unordered list items and bullets. This commit makes both conform to HTML and LaTeX defaults. Except that PGML is failing in both cases. I've examined the PG output from MBX and it is correct, but PGML is not handling deep-nested lists correctly (if it ever intended to). This is explained in the new problem in the sample-article.

rbeezer:
Hi Alex,

Sorry to be so slow on this one.  I guess my first read made me think some of the "is this PGML's fault?" ambiguity would be resolved.

Not sure how I feel about stressing PGML to the point of breaking?  Is it predictable enough (a certain depth?) that we should instead be warning?  Do you think the WW crew would fix any obvious bugs in a timely way (like before we merged this)?

Rob

Alex-Jordan:
@dpvc Davide, will tagging you here reach you? (If not, I'll catch you Friday.)

If so, would you take a look at the forum post below and give your thoughts on whether those issues are fixable?
http://webwork.maa.org/moodle/mod/forum/discuss.php?d=4045

rbeezer:
Thanks, Alex.  Holler if there is activity there and I can get caught up.  I found a bug in the Jupyter notebook Markdown rendering when I went about 6 levels deep with the MBX sample article.  This stuff will drive you crazy!

dpvc:
Looks like a bug.  I'll have to take a look and see what I can find out.

dpvc:
OK, I think I tracked it down (this is code that I wrote *long* ago, and it is complicated).  It looks like of you change [line 757](https://github.com/openwebwork/pg/blob/master/macros/PGML.pl#L757) to be

``` perl
    } elsif ($id eq 'indent' && $top->{type} eq 'indent' && $prev->{type} eq 'list') {
      $prev = $prev->topItem;
      if ($top->{indent} > $value && $value > 0) {
        splice(@{$self->{stack}},$i,1);
        if ($par) {splice(@{$self->{stack}},$i,1); $prev->pushItem($par)}
        $top->{indent} -= $value;
        $prev->pushItem($top);
        $prev->combineTopItems;
      }
    }
```

that should take care of it.  I'm pretty sure this won't mess up anything else, but as I said, this is pretty complicated, and I don't remember the details very well, so let me know if this causes problems for you.  If not, I'll submit it as a patch.

Alex-Jordan:
Thanks @dpvc, I'll test it out when I have some time this week and report here.

Alex-Jordan:
OK, that cleans up all the numbering issues I saw in the big collection of list layout testing I made for this MBX pull request. And you would know best, but it does seem unlikely that tacking on that last `elsif` would have a bad side effect somewhere. Thanks for this, and please make the PR to pg. I'll post my .pg test file there for someone else to try and confirm.

Rob, in the mean time, I will make Davide's edit on the PCC WeBWorK server. Then I'll edit the sample-chapter here to remove the language about PGML maybe not working with deeply nested lists. Davide's PR to WeBWorK may take some time to make it out into the real world in WeBWorK 2.13, but we can merge this into mathbook anyway I hope?

----

Davide, there is another issue with unordered lists. If I understand what I see, PGML treats `-` and `*` as the same when it comes to making list item bullets. No bullet style is specified, and either the browser default or CSS takes over. This means that you might use a `*`, intentionally wanting a solid disc, but you could get say a square if that's what the CSS says to do at that depth. Could this be changed so that `-` retains the current behavior, but `*` forces a disc in the way that `+` and `o` force squares and circles?

Using line numbers subsequent to inserting your above fix, I added line 1086:
```
  disc    => 'ul type="disc"',
```
and changed line 360 to use `'*'=>'disc'`. My experiments suggest this works for HTML.

I added line 1393:
```
\def\pgmldiscItem{\par\indent\llap{$\bullet$ }\ignorespaces}%
```
which I believe will then make `*` behave for hardcopy output. But then there is going to be a discrepancy between how `-` behaves with HTML (depends on the list dept plus the CSS) versus how it behaves with hardcopy (is always a `\bullet`). I'm not sure how to address that or if to address it at all. So I was wondering what you think.

dpvc:
@Alex-Jordan, I think the better solution is to simply change

```
  bullet  => 'ul',
```
to
```
  bullet  => 'ul type="disc"',
```
at line 1080 (in the numbing with the added `elsif`), without your additions above.  This will make `*` and `-` both produce solid bullets in both HTML and TeX output.  Nothing is depth-specific.

If you really want `-` to stay the same as it is, then your solution is the right one.  But I'm not sure there is a real need for that.

Alex-Jordan:
My regex is not fancy, but I searched the OPL for `(?s)BEGIN_PGML.*?^\s*- .*?END_PGML` trying to find all places where a `-` is used in PGML for a bullet. I was wondering if anyone built a nested list with `-` at all depths, relying on the CSS to style differently at different depths.

I only got one hit, and the `-` was inside math, so not used as a PGML bullet. 

Then I did the same search, replacing `-` with `\*` had a ton of positive matches, as I was expecting. So for whatever reason, PGML authors have adopted `*` and don't use `-` for bullets.

Furthermore, I only found one file (a PCC file!) where `*` was used as a list bullet at the second level of depth. And no instances at depth 3 or more. Of course, this is really not surprising, since how often do you really want nested lists in a WeBWorK problem?

Anyway, I just wanted to run this check before changing the behavior of `-` and `*` to _always_ be bullets in HTML.

If you open a PR for the indentation issue, would you mind tossing in a commit that also applies your suggested change to line 1080?

dpvc:
Thanks for checking the library for it.  I will make the pull request include both fixes.

dpvc:
I finally made the pull request for it.

PR #534, opened 2017-02-26
Pgml indentation2

Alex-Jordan:
New version of #509. Two commits are the same, with different hashes. One commit altered to remove the language about PGML list shortcomings. A fourth commit added that makes `<c>` content have monospace font within a WeBWorK.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -248,11 +248,11 @@
 
                 <introduction>
                     <p>One of the strengths of <webwork /> is its ability to give intelligent feedback for incorrect answers.</p>
-                    <ul>
+                    <p><ul>
                         <li><p>There is general feedback for when the student's answer is in an entirely different ballpark from the correct answer. Try entering something like <q>y</q>.</p></li>
                         <li><p>There is general feedback for when the student's answer is not in the right form. Try entering <q>x<circumflex />2*x<circumflex />3</q>, which, right or wrong, is unsimplified.</p></li>
                         <li><p>And problems can be written to detect and respond to common mistakes. Try entering an answer where you multiply the two exponents (instead of adding them, which would be correct.)</p></li>
-                    </ul>
+                    </ul></p>
                 </introduction>
 
                 <webwork>
@@ -1011,24 +1011,24 @@
                     </setup>
 
                     <statement>
-                        <ol label="a.">
+                        <p><ol label="a.">
                             <li>
                                 <p>Suppose the correct answer is <m><var name="$answer1"/></m>.</p>
-                                <ul label="square">
+                                <p><ul label="square">
                                     <li>
                                         <p><var name="$answer1" width="15"/></p>
                                     </li>
-                                </ul>
+                                </ul></p>
                             </li>
                             <li>
                                 <p>Suppose the correct answer is <m>\displaystyle <var name="$answer2"/></m>.</p>
-                                <ul label="square">
+                                <p><ul label="square">
                                     <li>
                                         <p><var name="$answer2" width="15"/></p>
                                     </li>
-                                </ul>
+                                </ul></p>
                             </li>
-                        </ol>
+                        </ol></p>
 
                     </statement>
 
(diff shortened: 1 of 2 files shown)


rbeezer:
Lots of delicate XSL probing here!  In order of the combined diff.  Code review, no actual testing.

```xsl
test="preceding-sibling::p|preceding-sibling::image|preceding-sibling::tabular"
```
is true if *any* predecessor of the image, not matter how far away,  is a `p` or `image` or `tabular`.  Is it possible you want to check the immediately preceding interesting node?
```xsl
test="preceding-sibling::node()[self::* or self::text()][1][self::p or self::image or self::tabular]
```

You use `parent::` and `..\` in close proximity, and the latter is a synonym for something similar to the former.  I'm liking the axes syntax better lately.

```xsl
ancestor::ul|ancestor::ol
```
is functionally identical to 

```xsl
ancestor::*[self::ul or self::ol]
```
I don't know which is preferable, your call.

What is `PF()`?

I didn't know that `[normalize-space()]` was possible!  I sort of prefer `normalize-space(<something goes here>)` but not critical.

The `list-indent` template had me confused until I finally got to it.  It produces zero space outside of a list context.  At least say so close by its definition.  Is there a better name that will convey its null-effect if images, etc are *outside* of a list context?  Maybe "optional-list-indent"?  You could wrap it in an "if" everytime, but I suspect you would find that tedious.

Alex-Jordan:
>> test="preceding-sibling::p|preceding-sibling::image|preceding-sibling::tabular"

>is true if any predecessor of the image, not matter how far away, is a p or image or tabular. Is it possible you want to check the immediately preceding interesting node?

I think your proposed alternative will test false on, say, the second `p` inside an `li` with children `p`, `todo`, `p`. Maybe `todo` is the only such thing to worry about, but I was worrying about things like this.

----

>You use parent:: and ..\ in close proximity, and the latter is a synonym for something similar to the former. I'm liking the axes syntax better lately.

Are you suggesting I change things like:
```
<xsl:if test="parent::li and not(following-sibling::*) and not(../following::*[1][self::li])">
```
to
```
<xsl:if test="parent::li and not(following-sibling::*) and not(parent::li/following::*[1][self::li])">
```
?

----

I guess I prefer `ancestor::ul|ancestor::ol` because it seems easier for an XSL newbie to understand.

----

`PGML::Format()` is a subroutine we needed to use a lot, that you don't normally need when you author PGML by hand.  We created `PF()` as a shorter alias, to cut down on url string length.

----
From the documentation:

``normalize-space( [string] )``

**_string_ (optional)**
    The string to be normalized. If omitted, string used will be the same as the context node converted to a string. 

So I'm using it here many times applied to a node set, to sieve away all the nodes in that set that that are entirely harmless white space that the author may be using for source readability. Such nodes are converted to empty, and then that is "false" as a selector. Used this way, in most cases, I'm not sure that there is a "something" that you can put in there, since I want it applied to all the variety of things in whatever node set has been gathered.

----

I'll change `list-indent` to `potential-list-indent`. I like that better than `optional`, which sounds like a user has a choice of some sort.

rbeezer:
On 02/26/2017 03:02 PM, Alex Jordan wrote:
> I think your proposed alternative will test false on, say, the second |p| inside
> an |li| with children |p|, |todo|, |p|. Maybe |todo| is the only such thing to
> worry about, but I was worrying about things like this.

You are right, but it *felt* like you wanted to examine the node right next 
door.  So I thought you would get too many false positives.

You are absolutely right about `todo`.  I need to make some canned filters so 
these get ignored, along with a similar situation inside paragraphs (like an 
`index`).  Eventually the DTD will clearly mark where a "todo" can go - likely 
not in an "li" and not in a "p".

> Are you suggesting I change things like:
>
> |<xsl:if test="parent::li and not(following-sibling::*) and
> not(../following::*[1][self::li])"> |
>
> to
>
> |<xsl:if test="parent::li and not(following-sibling::*) and
> not(parent::li/following::*[1][self::li])"> |

Yes.  `..` is `parent::node()`.  I use `parent::*` often when I already "know 
where I am."

> I guess I prefer |ancestor::ul|ancestor::ol| because it seems easier for an XSL
> newbie to understand.

Sure.

> |PGML::Format()| is a subroutine we needed to use a lot, that you don't normally
> need when you author PGML by hand. We created |PF()| as a shorter alias, to cut
> down on url string length.

Can you add the long form to the comment?

> So I'm using it here many times applied to a node set, to sieve away all the
> nodes in that set that that are entirely harmless white space that the author
> may be using for source readability. Such nodes are converted to empty, and then
> that is "false" as a selector.

I didn't know you could do that. ;-)

> I'll change |list-indent| to |potential-list-indent|. I like that better than
> |optional|, which sounds like a user has a choice of some sort.

Sounds good.

Alex-Jordan:
OK, just pushed 3 more commits. As you merge, you may want to squash 0dca25a, 35f2296, 7c8b28c, and 5d7f982 together, and use the commit message from 0dca25a.

rbeezer:
Perfect - thanks.  I'd have likely done the identical reorganization, but the 
guidance is greatly appreciated since it'll save me some fiddling.

I'll run another review with actual testing, soon.

rbeezer:
Thanks - all merged as-is.

PR #554, opened 2017-04-07
Remove science/math focus

davidfarmer:
Update README to remove specific mention of science and mathematics.


rbeezer:
Thanks, David.  This one has been on my hit-list.  Going public now at website, I should look for other references on the web pages right now.

PR #571, opened 2017-05-23
Add electronvolt to units.

Jahrme:
This pull request adds the `electronvolt` unit which is a common unit of energy within physics (see [here](http://units.wikia.com/wiki/Electronvolt)).

### DIFF
--- a/schema/dtd/mathbook.dtd
+++ b/schema/dtd/mathbook.dtd
@@ -690,7 +690,7 @@
 
 <!ENTITY % prefix "yocto|zepto|atto|femto|pico|nano|micro|milli|centi|deci|deca|deka|hecto|kilo|mega|giga|tera|peta|exa|zetta|yotta">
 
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Since `quantity` was first added, it appears `siunitx` now supports several units that it did not used to support (unless you manually defined them). electronvolt is one of these. 

How about one bigger commit that updates `mathbook-units.xsl` and the DTD to include all the new units. Go to the documentation for `siunitx` at http://ctan.mackichan.com/macros/latex/contrib/siunitx/siunitx.pdf, and see that everything from tables 1--5 is present in PreTeXt. 

Also, I initially organized the units with a scheme I'm not sure I could rigorously define. But under that scheme, I would either put electronvolt into the "Energy" category or make a new category for atomic-level measurements.

rbeezer:
Thanks, Jahrme.  I've merged this, in an effort to keep the PR queue under control.

It can move if/when a more complete list gets added as @Alex-Jordan suggests.

PR #585, opened 2017-05-25
Update the `video` Tag

Jahrme:
This will make the `video` tag (when not using YouTube) use the HTML5 `video` tag instead of `flowplayer`.

This will allow videos other than `webm` files and should be more cross-platform compatible.


rbeezer:
Thanks, Jahrme!  For a long time, I've wanted to see this fixed.

I added the `controls` attribute so those show by default.  There is also a modal `image-width` template that does error-checking and respects a document-wide default width, so I added that in.

And f312e4ee makes the UPS video work in the sample article.

PR #586, opened 2017-05-26
Collect thumbnails for GeoGebra alongside YouTube

Alex-Jordan:
Pay no attention to the branch name. Other things are happening on this branch that I haven't committed and it was simplest just to do this commit here.


rbeezer:
I'm going to wait on #587 so there is something to test with.

PR #598, opened 2017-06-03
GeoGebra

Alex-Jordan:
These commits are all GeoGebra Materials-related. The first commit additionally combines YouTube thumbnail scraping with GGM thumbnail scraping.

The rest give an example of a `<geogebra materials='...' />` in the WeBWorK sample chapter. You can test how it behaves in the HTML output. Or you can archive to PG, upload to WeBWorK and see behavior there (including through WeBWorK's hardcopy output mode).

The last commit lets `<geogebra materials='...' />` behave in LaTeX output. You can test what the WeBWorK sample chapter looks like in LaTeX, but I did not provide a WeBWorK-independent test example because that is a job for #587. Also there is the enhancement of `sidebyside` so that it can support a `geogebra` child, because at Tacoma, Rob outlined a future where a one-panel sidebyside would be the equivalent of a captionless figure, which is how WeBWorK needs to handle `geogebra`.

This replaces #586, which I will now close.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -915,7 +915,7 @@
         </section>
 
         <section>
-            <title>Graphics in Exercises</title>
+            <title>Graphics and Interactives in Exercises</title>
             <p>It is natural to want to write exercises that rely on graphics. For example, an exericse might produce a chart of some kind, and ask the reader to extract some information from that chart.</p>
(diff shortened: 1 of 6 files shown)


rbeezer:
Alex - do we still want to include `interactives` within a WW problem?  I'm close to finishing work on `interactive` and will have pageres doing screen shots for the static versions.

I've not experimented with `interactive` inside a plain `exercise` yet, so I think that probably needs to be done first.

rbeezer:
How much of this is still relevant?  Should we preserve the multiple panel warning (easy on my end)?  Any of the remainder?

I'd like to close this one after preserving anything worth keeping.

Alex-Jordan:
About the `sidebyside` thing. Do you still plan to get rid of `sidebyside`s with only one child, and introduce `display`? If so, then I wouldn't bother keeping this. `sidebyside` would no longer be OK at all within a WW, and `display` would take its place.

Alex-Jordan:
I sign off on closing this. I'll go ahead, but you see my comment about `display`.

PR #606, opened 2017-07-03
New scheme for WeBWorK in print

Alex-Jordan:
I'll be pushing commits here for the new scheme for handling WeBWorK in print over the next few weeks. There are some moving parts, because for this to work it depends on the version of the WeBWorK server. WW 2.14 will come late summer, and that is where I will have everything ready to go. 

But there are a few things that are independent of such core changes to WW, and I'd like to start pushing those. If you like, don't merge and just let it all accumulate here. I would still appreciate any observations you might make about these commits as they come in though.

This first commit is about a book's custom LaTeX macros. Right now, they are inserted into every little bit of math in a problem. Not an issue for MathJax, which can handle repeat \newcommand's. But this will be an issue for the new print scheme. So this commit allows `mathbook-webwork-pg.xsl` to see "know" it is being called from (the new, uncommitted) extraction template. If so, it omits the macro definitions.

### DIFF
--- a/xsl/mathbook-webwork-pg.xsl
+++ b/xsl/mathbook-webwork-pg.xsl
@@ -56,6 +56,11 @@
 <!-- to provide AnswerFormatHelp link.                            -->
 <xsl:param name="pg.answer.form.help" select="'yes'" />
 
+
+<!-- If the extraction style sheet is calling this style sheet, $static -->
+<!-- will be 'yes', and that influences certain templates here          -->
+<xsl:param name="static" />
+
 <!-- ################# -->
 <!-- File Organization -->
 <!-- ################# -->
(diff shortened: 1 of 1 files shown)


rbeezer:
The stylesheet you are editing should stand on its own, without passing in this parameter.  And I think it would be cleaner to have "normal" behavior in the `when` part, not the `otherwise`.  So three suggestions:

*  Reverse the name (`dynamic`) or reverse the tests (` = 'no' `).
*  Supply a default with `select` that preserves current behavior so existing uses do not now need a parameter passed.
* Consider computing a boolean from the passed-in string (`yes|no`) and then the tests of the boolean are easier, especially with many of them.  I've been using variable names such as `b-static` to recognize booleans.

PR #608, opened 2017-07-03
New scheme for WeBWorK in print

Alex-Jordan:
Revision from #606.


Alex-Jordan:
OK, I acted on your suggestions from #606.

rbeezer:
Looks good. Will merge in AM.

rbeezer:
Even better: no more `choose` at all.

Right now, this creates no change in behavior until extraction worksheet is used., so no need to think too hard about it.

PR #621, opened 2017-07-08
Add instruction element for WeBWorK

Alex-Jordan:


rbeezer:
You need to actively kill this on the LaTeX side.  I guess before it gets sent to the WW server for rendering?  A template that is empty will do the job.  In the right place.  Or maybe it should be killed as part of some common routines, then selectively enabled in an overriding module (as here in HTML).

In the template, do you think we can go
```xslt
match="webwork//statement//instruction"
```
I forget where `stage` comes in, maybe one of these double-slash can be a single?  Let me know if we can use something more specific.

Alex-Jordan:
We've got at least: `webwork/statement/instruction`, `webwork/stage/statement/instruction`, and `webwork/statement/p/ol/li/instruction`. So that's why the double slash.

The PG problem has turned `instruction` into `KeyboardInstructions(...)` which is a macro in `PCCmacros.pl`. This macro actively kills it _unless_ the display mode is some flavor of HTML. In particular with the new PTX display mode, nothing comes back.

I can kill it in common. Seems smarter to me. Activating it back up in `mathbook-webwork-pg.xsl` will work. In the new scheme, I don't think this matters for print unless you respond a certain way in a forum post I am about to make :) But probably still smart to kill it.

rbeezer:
In the discussion on the Google Group, we had "I think we've established this is only for within a webwork/statement or webwork/stage/statement, as a peer of p."

I don't want this to be abused as a per-output hack in author's source, so I'd like the match condition to be very rigourous and strict. Like

```xsl
match="webwork/statement/instruction | webwork/stage/statement/instruction
```

so as to be a top-level element of the problem statement (allowing stages).  In particular, do you want to try to convince me what this needs to be down in a list item?

Alex-Jordan:
By "within", I meant possibly deeper than a child.

Principle: an "instruction" distracts from the mathematics, and should not be used "far away" from the answer blank it refers to.

Small example:

```
Let's examine f(x) = ln(x).

a. Evaluate f(1).
    ____
b. Can f have both positive and negative outputs?
    <popup with yes or no>
c. What is the domain of f, using set-builder notation?
    ________
```

The instruction here would be about how to type set-builder notation. Students don't know about the `|` character on a keyboard, so you say something like "The | character is above the enter key and is typed using shift-`\`." Maybe there should also be something there about using `>=`, `<=`, and `!=` as needed.

Putting a message like this before the list, instead of within part c (after the question but before the answer blank), can distract students as they set out to answer part a. And possibly throws a red herring at them if they think it's some kind of hint for answering part a.

Here, the instruction is really intimately linked to the answer blank. So I considered having it be a child of the `var`. But there are also plenty of examples where an instruction applies to multiple/all of the answer blanks in a problem, so I thought not to put it in any one `var`.

How about 
```
webwork/statement/instruction
webwork/stage/statement/instruction
webwork/statement//li/instruction
webwork/stage/statement//li/instruction
webwork/statement//tabular/row/cell/instruction
webwork/stage/statement//tabular/row/cell/instruction
```
?  I think that actually covers "within a webwork and a peer of p".

Alex-Jordan:
This just came up in my schema validation for ORCCA, so I thought I'd check if it's on your radar. No rush, because I'm compiling ORCCA on a branch that has this implemented (although, not in the schema).

PR #622, opened 2017-07-08
Allow activity, exploration, and exercises inclusions in DTD

Alex-Jordan:
This PR is because Active Calculus has separated each `exercises`, `activity`, and `exploration` into its own file, which is then included via `xi:include`. These edits to the DTD make that DTD-compliant.

I could see not approving this for a number of reasons. But maybe it's OK because the DTD is not being maintained anyway. I just want to cut down DTD violations as I try to get Active ready to print, and at the same time I want to cut down on the current level of branching that I have going on.


rbeezer:
No harm, and might be a useful reminder when wrapping up the new RELAX-NG schema.  Thanks.

PR #623, opened 2017-07-09
Musical Time Signatures

Jahrme:
Currently time signature support is very good for LaTeX. XeLaTeX provides the best output but pdfLaTeX is also acceptable (if visually less pleasing).

On the other hand, HTLM output is typographically lacking, however it is of comparable quality with what I could find online already. The issues arise from two technical difficulties:
- First, the lack of support for unicode music symbols which is why symbolic time signatures are inserted as inline SVGs which reference the unicode glyph that *should* be there. In addition, they are supplied with a title and description for accessibility.
- Second, typographically correct time signatures are normally set in a music specific font and stacked (as can be seen in the XeLaTeX output), however as Google Fonts does not provide a music enabled font (and as most browsers enforce a same-origin policy) it would be difficult to achieve the characteristic look of typographically correct time signatures. However, it has become standard practice when writing about time signatures online to use a simple inline layout (e.g., 3/8, 2/4, etc.) and this presentation is sufficient (for now) for Dr. Robert Hutchinson who is (to my knowledge) the only author currently using the music capabilities of PTX.

It is possible that we can improve the HTML output in the future, but the current implementation should be sufficient for now.

### DIFF
--- a/examples/humanities/music.xml
+++ b/examples/humanities/music.xml
@@ -168,36 +168,54 @@
             </ul></p>
         </example>
     </section>
-
+    <section xml:id="time-signatures">
+        <title>Time Signatures</title>
+        <p>Time signatures are created using <c>&lt;timesignature&gt;</c>, <c>&lt;meter&gt;</c>, <c>&lt;top&gt;</c>, and <c>&lt;bottom&gt;</c> tags, though not all are used in every case.</p>
+        <p>The <q>symbolic</q> time signatures<mdash/><timesignature symbol="common"/> and <timesignature symbol="cut"/><mdash/>are produced by using <c>&lt;timesignature symbol="common"/&gt;</c> and <c>&lt;timesignature symbol="cut"/&gt;</c> respectively.</p>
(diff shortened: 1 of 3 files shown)


rbeezer:
Good work.  Not easy, I can see.

*  Include the note about (2 1/2)/4 in the visible text.  No problem letting people know the limits (and potential reasons).
*  Named templates (eg "cut-time") are great for content that is unchanging (or maybe parameterized).  However, I would like to avoid named templates that implicitly presume the context (eg "single-meter"). Can you convert these to be defined like
```xslt
<xsl:template match="timesignature" mode="single-meter">
```
and call them like
```xslt
<xsl:apply-templates select="." mode="single-meter" />
```
It feels more object-oriented and the definition will not accidentally match just anything.  So the rule is to explicitly pass context node through match/select.
* With modal templates for the `otherwise` case, could the non-modal `match="timesignature"` template go into `mathbook-common.xsl`?  Would that be good/workable long-term?  If so, leave a comment in the HTML and LaTeX files about where to go to find it.

I can combine/reorder commits if you just want to pile on (maintaining current file divisions).

rbeezer:
Dear Jahrme,

Do we still need/want this?  It might all be OK, except that it modifies the old DTD (which is gone now).

PR #635, opened 2017-07-19
Restore Normal Emphasis Behavior

Jahrme:
To add the custom underline abilities of *Sound Writing*, we loaded `ulem`, but `ulem` makes `\emph` underline text instead of italicizing it. Adding `\normalem` returns `\emph` to italics, and this helps keep the LaTeX and HTML output visually similar.


rbeezer:
Good idea.  I'll get it in shortly.

rbeezer:
Merged, and available.  Thanks for keeping this up-to-date.

PR #683, opened 2017-08-15
Suggest Windows users ignore "./ant test" failure

sklarjk:


rbeezer:
Thanks.

There was a stray space deleted from an earlier paragraph.  Be careful about extraneous changes like that, they can cause problems on the receiving end.  Do a `git diff` before committing and look for things you didn't mean to do.  No problem this time.

sklarjk:
OK, sorry. I'll be more careful in the future.

PR #692, opened 2017-08-18
Update google groups

davidfarmer:
The README mentioned the mathbook-support Google group, so I changed
it to the new groups.

For some reason there are a bunch of extra old commits, even though the only
change is from the most recent commit.  Not sure what to do about that, or if it matters.

Also, for some reason I only put the support and announcement groups.


Alex-Jordan:
One commit is when you removed the math focus. But then Rob applied that change with his own commit (with you explicitly named as author). So this commit itself never went through, even though the changes it makes are already in dev. 

The other commits are all merge commits. Whenever there is a merge that is not simple fast-forward, then a record of that merge is made with a merge commit. Somehow a bunch of merge commits have stacked up for you. I think Rob avoids merge commits in the commit history by doing the kind of thing he did with your removal of the math focus.

After this PR is resolved, consider deleting your dev branch (both locally and on your GitHub account).  Then pull master dev to rebuild it. Then for future PR's, if you make each on its own branch while only ever pulling dev from master dev, this kind of thing won't happen.

rbeezer:
Fetched the branch, and cherry-picked the newest commit.  Added PR number to commit message so it shows here.

Thanks for the fix!

PR #694, opened 2017-08-20
Define sidebyside template for within a webwork

Alex-Jordan:
Recently, the WW sample chapter has not been grabbing images off the server to insert into the PDF. This was because these images were put inside a sidebyside with the recent sbs and schema work, and sidebyside was not defined within mathbook-webwork-pg.xsl. The PG output had nothing where the image should go, except its description. So when the base64 was sent to the server, there was no image to grab. The mbx script would have no way to know this. So the error only became visible when the pdf compilation was attempted.

I'm unsure if this is how you'd like to handle this. Perhaps you have something more specific in mind for sidebyside template here. I do too, even allowing more than one panel. But for now this makes the sample chapter compile again.

I'm also unsure if the schema is ready to distinguish between "sbs lite" with only one, captionless child. Can I leave it to you to do the right thing for the schema?

### DIFF
--- a/xsl/mathbook-webwork-pg.xsl
+++ b/xsl/mathbook-webwork-pg.xsl
@@ -689,6 +689,16 @@
     </xsl:choose>
 </xsl:template>
 
+<!-- ############################################## -->
+<!-- Sidebyside in a WeBWorK expects only one child -->
+<!-- It should be in the captionless family         -->
+<!-- Just applies its tempaltes                     -->
+<!-- ############################################## -->
+
+<xsl:template match="webwork//sidebyside">
+    <xsl:apply-templates />
+</xsl:template>
+
 <!-- ####################### -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Done, at 2ec32e0b.  I changed the surrounding comments to reflect the fact that this could be a stopgap of sorts.

I could do better output for single-child `sidebyside` anytime, but it is not a priority since most authors will never know.  Does WW have any facility for pushing a single image (say) left or right?  We could interpret the layout parameters to that extent, perhaps?

rbeezer:
Forgot to say: thanks for getting this one sorted out.  I'll sleep better knowing the sample output is all in good shape.

Alex-Jordan:
Yes, PGML can push a single image (or paragraph, or...) left or right. Precede with `>>` to push right. End the line with `<<` to push left. Do both to center.

PR #735, opened 2017-09-26
Edits to WW sample chapter

Alex-Jordan:
These edits are in preparation for bigger WW changes coming. By getting these in first, it should help with diff comparisons before and after. This leaves the sample-chapter with a few schema violations having to do with a sidebyside inside an li. I'm hoping that will someday be legal (with restrictions on what can be in the sbs).

### DIFF
--- a/examples/webwork/Makefile
+++ b/examples/webwork/Makefile
@@ -1,21 +1,21 @@
 ## ********************************************************************* ##
-## Copyright 2015                                                        ##
+## Copyright 2015-7                                                      ##
 ## Robert A. Beezer, Michael Gage, Geoff Goehle, Alex Jordan             ##
 ##                                                                       ##
-## This file is part of MathBook XML.                                    ##
+## This file is part of PreTeXt.                                         ##
 ##                                                                       ##
-## MathBook XML is free software: you can redistribute it and/or modify  ##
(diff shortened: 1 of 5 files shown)


rbeezer:
Merged as-is, just edit on commit message.  Thanks for continuing to chip away on this one.

Alex-Jordan:
The proper thing to do next is to change the existing extraction style sheets to use percentages for width. (So that I can edit the sample-chapter accordingly, so that I can then commit the new extraction style sheets.) 

Would you be OK with me changing the image template to use _either_ percentage or integer pixel count depending on presence of the `%` character? (And not getting the deprecation alert involved, and not telling people to change their source? So that it can all be one big announcement later?) 

Or would you rather just get the conversion to percentage widths fully addressed now? I'd be fine with that, but I'd like to ask if I can do the PG extraction work, and then you would handle the deprecation warning and fix.

Keep in mind that either way, I'll be modifying `mathbook-webwork-pg.xsl`, which will be killed from the repo a few commits further in this process.

rbeezer:
Better to deprecate sooner rather than later.  Postponing just creates more pain 
later. ;-)  I'm happy to do deprecation messages, they are easy.

I had trouble finding this again and almost totally forgot about it.  ;-)  Let's 
do new questions on pretext-dev, not closed pull requests!

Rob

PR #744, opened 2017-09-30
Typo

jdemeyer:


rbeezer:
Thanks, Jeroen!  I work a bit differently than your typical Sage release manager, so you'll want to trash your own version of the branch.

PR #755, opened 2017-10-21
Enable full screen on videos

Alex-Jordan:
Prior to this commit, if I play the YouTube videos at 
http://mathbook.pugetsound.edu/examples/sample-article/html/section-video.html

there is a YouTube-provided icon at the lower right, but clicking it says full screen is not enabled. With this small change, I can make the videos full screen.

All examples that I find leave out the `=""`. But `xsltproc` is not OK with that when processing this, since the `iframe` is built directly. The `=""` seems to work for me at least.

This W3C report claims it is implemented across browsers:
http://w3c.github.io/test-results/html51/implementation-report.html

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -5415,6 +5415,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             width="{$width}"
             height="{$height}"
             frameborder="0"
+            allowfullscreen=""
             src="{$source-url}" />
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, that is a big improvement.  I added a few comments, so I can remember what is going on whenever I happen to be back to these.

Yes, just a null value is good enough to make the attribute present and have the HTML detect the **presence** of trhe attribute.  At least, that is how I understand it.

Thanks,
Rob

PR #766, opened 2017-11-02
Update conversions.xml

davidfarmer:
Corrected typo:  < element > should be < example >.


rbeezer:
Thank-you!  In repo, but website version will update later.

PR #767, opened 2017-11-09
WW extraction script

Alex-Jordan:
An important change I am making here is the default WW server for the sample-chapter (and minimal) example is just http://localhost. (I leave the PCC server there, but commented out.) I have several reasons for needing to do this. So how will you test?

Well, changing it to webwork.pcc.edu won't help, because I have not yet merged the necessary changes on the WW end into my production WW server. And I don't want to do that until this term ends in early December.

So I'm wondering if you have a webwork installation on the machine you use, and can set up localhost to be its web server? Can you type `http://localhost/webwork2` into a web browser and see anything? If so, I could tell you exactly what branches to be on for WW's three relevant repos (webwork2, pg, and webwork-open-problem-library). In the one-in-a-million chance you already have that set up, I'll hold off on trying to help you set it up. Let me know.

Alternatively, I could explore how to let you access my localhost. A brief glance at the internet says this may be possible with port forwarding.

And then once we get this localhost issue settled, I'll give guidance on how to test what is in this PR.

### DIFF
--- a/examples/webwork/Makefile
+++ b/examples/webwork/Makefile
@@ -70,7 +70,7 @@ DTD = $(MB)/schema/dtd
 
 # XML to apply templates to
 SMPCHP = $(WWEXAMPLE)/sample-chapter/sample-chapter.xml
-MIN = $(WWEXAMPLE)/minimal/mini.xml
+MINI = $(WWEXAMPLE)/minimal/mini.xml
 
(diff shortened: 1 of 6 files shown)


Alex-Jordan:
Oops. Just realized I committed too large of a cut. Stand by.

PR #772, opened 2017-11-16
Parameter for a style file added

arechnitzer:
I've added a parameter which will let the script know which style file to grab latex style from.


davidfarmer:
Note that we want to make it very difficult for people
to choose a "boxedtheorems" option.  PTX Principle 7 says:

7. MBX output is styled by selecting from a list of available templates,
relieving the author of the burden involved in micromanaging the output
format.

What we want is a "UBCstyle" option that boxes theorems and also makes
many other changes.  It is okay if, for now, this option only exists
for LaTeX.

I understand that if you know enough, then you can figure out how to
only enable the boxed theorems option.  But we absolutely do not want
to document that in a way that makes it easy for anyone to do it.
This is a lesson learned from the mess of LaTeX.

On the other hand, we want to advertise and document the "UBCstyle"
LaTeX option, once it exists.

rbeezer:
When Andrew is done (post here, Andrew), I will likely put this on a public 
branch for folks to chew on.

David - look at "clp.xml" - that is the UBCStyle.  It is one package (which 
Andrew is building up slowly).

Andrew - no rush, it'll be a week before I can look.

arechnitzer:
Hi David + Rob

Agreed - low level mix and match hackery is not a good idea. 

I am trying to reproduce the CLP pdf look but from my pretext source. Two reasons for this. 

First = my coauthors and I want to keep approximately that "look", and at the same time don't want to maintain (separately) the latex and the xml. I would like us to migrate to xml-only, which we could then make public. But I need to be able to add some more colour to the latex.

Second = longer term? I am trying to add latex-styles in a way that might eventually be a bit like beamer-themes. Ie - author chooses a "CLP" theme and compiles to get a pdf that looks like the current CLP (boxes around this and that, and arrows around examples etc). Similarly an "AATA" theme will give the AATA nice boxed chapter numbers, pretty headings and clean pages. etc. etc.  

At present all very experimental...

Andrew

davidfarmer:
Sounds good.

Will the CLP happen just by changing the style file, or does the
conversion to LaTeX have to be different?

arechnitzer:
Hi David
Unfortunately I think that it requires more than a change of style file. At present I'm working with the idea that the author / publisher changes an option at xsltproc time... producing different latex source depending on the style chosen. I'm not sure how to encapsulate style changes inside a \usepackage{pretext-theme-clp} or \usepackage{pretext-theme-aata} or similar without getting into some very messy hackings of things like the figure environment. My latex is not good enough for such things.

If you look at the clp.xml and pretext-clp.sty - that gives you an idea of what I'm working at. Unfortunately I don't know how to keep all my required code outside the .sty file, but I suspect it can be done with some more work.

Andrew

rbeezer:
Dear Andrew,

You are on the right track.  An author can pick a new style and each will be a coherent collection.  And the work may help us understand other technical aspects.

Rob

davidfarmer:
I understand that, for now at least, different output PDF styles
will require generating different LaTeX first.

arechnitzer:
Is coming along - take a look using "clp2.xml" - I've migrated much of the code to use tcolorbox which is a bit cleaner than my original clp.sty hackings. Still some "fun" dues to differences between how amsthm and tcolorbox declare theorems. Still lingering issues with numberwithin. Getting there though.

Andrew

rbeezer:
Dear Andrew,

I  had a look at the accumulated diff of all of these.  An impressive amount of work.

Numbering should be tough.  Try varying the parameters (level) and see if the LaTeX matches the HTML.

Does
```
<xsl:if test="$latex.draft='yes'" >0
```

have a stray zero at the end of the linne?

arechnitzer:
Yes - that was a stray trailing 0. Fixed now (upload soon).
I think I've sorted numbering - some "fun" required. You cannot set a default numberwithin for tcolorbox - you need to do it as you define each environment. However, a new tcb theorem can inherit its numbering properties from another. So - I now define a dummy theorem env with the correct numberwithin parameter. Then all other tcb theorems inherit the numbering from that dummy theorem. A bit of a cludge but it seems to work.

I think the code now produces something like the CLP-style (there are still a few minor issues around frontmatter titles and the titlepage - nothing critical) and also the default style. If you have suggestions for another book's style to copy then I can hack that.

Andrew

rbeezer:
Dear Andrew,

I'm going to close this for now, it can always be re-opened.  This was a good experiment.  I've got some new ideas (along these lines) , which I hope to get to soon.

Rob

PR #773, opened 2017-11-16
Author's Guide Index

nwintersgill:
This creates a basic index for the author's guide that includes some key terms. Also updates the section on indexing with updated, proper syntax.

While the index is fairly standard, it's worth noting that I had difficulties building the author's guide on a windows machine. Whenever I attempted to build the guide using xsltproc, I would receive an error based on line 5886 of mathbook-html.xsl. I commented out the problematic code and things started running smoothly again. This problem _only_ occurred with the author's guide. I did not include the change to the xsl file in the pull request.


rbeezer:
Nathan's work at  f11160cc84fd395c6ce6d8700cfbe5e84ba9e8c0. Once again, put hash in the wrong place.  Sorry.

rbeezer:
Dear Nathan,

Very nice work.  Now we have to keep it complete. ;-)  I put some edits into a new commit.

*  Please never introduce tabs into source.  (Indentation on `index` in `backmatter`.)  They cause lots of problems for lots of reasons.  Figure out how to configure your editor, so the tab key "works", but spaces result.
*  Edited the overview addition.  There was a typo "wish/with" and an em-dash should be authored with `<mdash />` to always produce the right thing.  While there I tightened it up, since I want the overview to be a quick-ish introduction.  Detail can go in the "Topics" chapter.  "sets of tags" cumulatively are an "element", delimited by "tags" and containing "content."

Go  `git show --word-diff <RAB-commit>`  to see exactly what I did.  This is why I put it into a new commit.

Thanks for helping to make PreTeXt easier to use.

Rob

PR #784, opened 2017-12-13
Typo (twice) in mathbook-html.xsl

davidfarmer:
xsl/mathbook-common.html should be xsl/mathbook-common.xsl in two comments


rbeezer:
Merged.  Thanks!

davidfarmer:
It says merged, so should I close?  Not sure why that was not automatic.

rbeezer:
Yes, go ahead and close.  I forgot to.  Or I can catch it later (on fone now).  I prefer to merge locally and edit commit message for a better record, independent of GitHub, but in a way GitHub still understands.  In this case, I also made two very trivial edits.

So simply trash your original branch, that commit now has a different hash.

Issue #786, opened 2017-12-15
Omit header around section titles


davidfarmer:
Let's omit the "header" wrapping the hN.heading for the titles of sections and subsections,
and also the chapter title on the chapter pages of a book.

I am not sure if I made all the necessary CSS changes, but if not, I am close.
So, a temporary version of the sample article and sample book would be
helpful but not necessary if inconvenient.

The 
    header id="masthead"
at the top can stay.

Is there some type of header for lists (which I seemed to see in the XSLT,
starting line 993)?
If so, please let me know an example so I can see if it should be eliminated.

I will submit a pull request to make this change.

This is part of issue #741

rbeezer:
Fixed at #787, but see also #788

PR #787, opened 2017-12-15
Eliminate header wrapping section titles

davidfarmer:
Fixes issue  #786

The diff in the pull request looks confusing.  What I did was eliminate lines 521-524
and 545 in mathbook-html.xsl, and then shift left the in-between lines by 4 spaces.
So, there is no < header > around the hN containing title.

Built the sample-article and it looked as expected.


rbeezer:
Looks like the "tooltip text" moved inward, since its home got deleted, which led to the complicated diff.  Thanks for the change--I'll merge once back to big screen.

Thanks for the PR.

rbeezer:
Thanks again.  Merged, and rebuilt the whole set of examples on website, so look there if you need to do any double-checking.

PR #792, opened 2017-12-22
Remove remaining header tags from body

davidfarmer:
Eliminated header tag around hN.heading in list-of.

The List of Results in the sample-article (Appendix C, appendix-results.html)
should look the same, but the "header"s around the "h1.heading"s are gone.


rbeezer:
Looks good.  Merged without any testing.  Thanks for the clean-up.  Live in a minute or two.

PR #801, opened 2017-12-30
Streamline knowls

davidfarmer:
Eliminate knowlstyle.css, because that code is elsewhere.

Add an example of math in knowls.

Shorten the alt text for the logo image (because the existing alt text disrupts the
layout too much when the logo image is not available).

Tested on sample article and looks okay.


rbeezer:
Thanks - good-looking changes.  Being pushed up to website right now, give it two minutes.

Issue #804, opened 2018-01-01
Links on the chapter page of a book


davidfarmer:
On the chapter page of a book, for example,

http://mathbook.pugetsound.edu/examples/sample-book/html/integers.html

the links to the sections are in a structure like:

nav.summary-links
  a
    span.codenumber span.title
  a
    span.codenumber span.title
  etc

How about making a list inside the nav.summary-links?  That would be

nav.summary-links
  ol
    li
      a
        span.codenumber span.title
    etc

Obviously this adds two levels of markup, and the results would look the same,
but it would more accurately reflect the structure.  Maybe it would be more
usable without the CSS?  Maybe it would allow more options later.

We could collapse the nav.summary-links and ol into ol.summary-links .
That hides any clue provided by the nav tag, so maybe is a bad idea.

davidfarmer:
I can make this change if we decide to do it.

rbeezer:
Sounds good to me.

* The `nav` might be worth keeping, maybe a small device will do a good job with it, absent CSS.  Also screen readers may use the hint.  It will get 100% replaced in EPUB by a different ToC mechanism, so that is not a no-CSS situation to consider.
* Normally `ol` would seem right.  But a summary page of appendices will go A, B, C, and a summary page of frontmatter will mostly be un-numbered.  Since we have the `codenumber`, perhaps a `ul` would give a better no-CSS presentation?  Your call.

A comment on my HTML coding style: a "literal" expression of the HTML in the XSL is best, like is done here for `nav.summary-links`, or with "attribute value templates" (AVT), such as is done here with `href="{$url}"`.  Constructing HTML with `xsl:element` and `xsl:attribute` should be reserved for when it is absolutely necessary - I've tended to overdo this construction in places.

I'll watch for a PR, but will be in/out much of today.  Thanks!

rbeezer:
Accomplished on #806

PR #806, opened 2018-01-02
Summary markup

davidfarmer:
Put summary links in a ul, as suggested in issue #804.

Added a wide equation to sample article, for testing different display widths.


rbeezer:
Thanks!  Going live right now.

PR #808, opened 2018-01-03
Update schema-install.xml

davidfarmer:
Delete un-necessary numbers.

### DIFF
--- a/doc/author-guide/schema-install.xml
+++ b/doc/author-guide/schema-install.xml
@@ -34,12 +34,12 @@
                 </li>
 
                 <li>
-                    <title>1. JDK</title>
+                    <title>JDK</title>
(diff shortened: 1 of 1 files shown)


kcrisman:
I suggest that we make it have a second level of nesting instead of just deleting, since these are "observing the following notes keyed to the four steps" from the `readme.md` file.  I can confirm the steps work, though they require a little facility with paths and the command line.

rbeezer:
The four numbered steps reference each other by number.  We could change those to `xref` by title, which might work well.

Or a nested `ol` though I forget what restrictions have been palced on `dl` - proably just that they need to be top-level.

Advise, and I'd address #812 at the same time.

kcrisman:
I think nested list makes more sense, since it really is a sublist as per the `readme.md` file (I just used this and can confirm).

rbeezer:
I put a numbered list into one of the description list items.  I hope that satisfies everybody on this one - I think it will.  I did a whole lot of other work as well, so did not use this pull request itself.  Sorry about not including it.

See a607798c3023b8f7dca711902db42af81af49fd1

Issue #705, opened 2017-08-30
alert should have better HTML markup


davidfarmer:
< alert > in PTX becomes < b >< i > in HTML.

It would be better to have it as < em class="alert" > .
(The CSS for that should already be in place.)

This would be a good issue for a someone who wants to start contributing to the xslt.
Maybe there could be an appropriate tag for such issues.

rbeezer:
Done.  "beginner" label, in addition to existing "contributor project".

rbeezer:
Advice for a beginner:  write "bare" HTML for the output, don't use the `<xsl:element name=...` approach.  The former is encouraged, and the latter is only necessary when parts of the output change programatically.

davidfarmer:
Done

PR #817, opened 2018-01-16
Small fixes

davidfarmer:
Catch-up with various small changes I have been making:

- Mark answers as answers instead of hints
- Add index entries
- Fix HTML markup for alert (issue #705)
- Add instructions for code blocks to CONTRIBUTING

Note that the last item was difficult for me to check before doing the pull request.


rbeezer:
*  Thanks very much, merged as-is with just edits to commit messages.
*  Triple back-ticks seem to have survived!  I use CoCalc to preview Markdown - it has live preview.  I've not found another way to accomplish that locally.
*  Some of Tom's "hints" were identical to the answers, iirc.
*  CONTRIBUTING  also needs a wholesale rewrite due to name changes.  ;-(  https://github.com/rbeezer/mathbook/issues/818

Issue #783, opened 2017-12-12
Option for spacing around an em dash


rbeezer:
Make a publisher option for a thin space, or a full space, surrounding an em dash.  This should be a publisher option, so a command-line string parameter.  Documentation should advise no extra space in source, maybe XSL could remove any that exists in adjacent text nodes.

Default is no space, options could go
```
emdash.space  none|thin|full
```

Option should be an easy XSL exercise, stripping adjacent space is hard and could wait.

kcrisman:
Make a publisher option for a thin space, or a full space, surrounding an
> em dash. This should be a publisher option, so a command-line string
> parameter. Documentation should advise no extra space in source, maybe XSL
> could remove any that exists in adjacent text nodes.
>
> Default is no space, options could go
>
> emdash.space  none|thin|full
>
>
> Should html have a space?  Fernando G. recommends it for MAA Reviews
because it looks different.

daverosoff:
Are you sure about this, Rob? I learned that if spaces surround the dash, it should be an *en* dash rather than an em dash. Em dashes are flush with surrounding text.

davidfarmer:
There is a difference between a full space -- as in this sentence,
and a little tiny bit of thin space, which some typographic styles use.
(And which I can't replicate in ascii.)

rbeezer:
Implemented, with comments, on PR #821.  A full-space option could be added later if there is a real hue and cry.

PR #821, opened 2018-01-24
Em dash spaces #783

daverosoff:
Here's my attempt, my first XSL contribution. I did not try to figure out how to strip space adjacent to an `<mdash/>`. See #783.

### DIFF
--- a/xsl/mathbook-common.xsl
+++ b/xsl/mathbook-common.xsl
@@ -178,7 +178,8 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- Pointers to realizations of the actual document -->
 <xsl:param name="address.html" select="''" />
 <xsl:param name="address.pdf" select="''" />
-
+<!-- Publisher option to surround emdash with thin or full space -->
+<xsl:param name="emdash.space" select="'none'" />
 <!-- Whitespace discussion: http://www.xmlplease.com/whitespace               -->
 <!-- Describes source expectations, DO NOT override in subsequent conversions -->
 <!-- Strip whitespace text nodes from container elements                      -->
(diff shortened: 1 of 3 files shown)


rbeezer:
Dear Dave,

XSL looks very good.  I might twiddle a few things to my liking, but that will be just for the eventual maintainer's convenience.

Nowhere have I seen a style guide to suggest a full space is recommended or allowed.  Mostly `none` or else `thin` (Associated Press Style).  Do you know otherwise?

The obstinate among us can elect to put in a full space.  But they will lose an easy route to `none` or `thin`.  I'd rather not have a switch to suggest something that is usually banned (if I'm right above).  Documentation would explain this.

Needs at least a test example in the sample article (Author's Guide for this sort of thing is on a branch of mine right now).  I'm perfectly fine writing a commit on top of this one, if you'd like.

Stripping existing spaces is tricky, and can probably wait.

Sorry for the delay, I've been traveling long-distance and on arrival helping organize a conference.

Rob

daverosoff:
Thanks, Rob, and no worries about the delay! 2 days barely registers. I added the `full` option because it was part of the original issue at #783. David's comments there don't suggest it is inappropriate exactly, but not as sanctioned as flush or thin spacing. I think it is reasonable (and conformant with all the professional styles I have learned) to insist on `none` or `thin`.

It might be best if you added a commit for the sample article; I wouldn't know where the best place for the example is. Thanks for your review.

Dave

rbeezer:
Yes, I am always surprised at how often I change my mind, relative to a first posting on an issue.  ;-)  Its partly why I don't try to rush anything.  I'll likely layer on top of this commit, so it is clear what is yours and what I do to it.  Thanks for getting this one rolling.

rbeezer:
I added an extra layer of massaging variables.  I've found deprecations easier if the default value is a null string, then I can tell if the author actually used the option at all.  And some error-checking helps.

Removed the `full` option---we could easily add it later if there is need.  See Section 8 of the sample article, part way down for discussion and an example I used for testing.

Welcome to the club of contributors to the XSL!  I like the way you made the space character a variable, I wouldn't have thought of that.  I look forward to more.

PR #822, opened 2018-01-24
Author's Guide: the PreTeXt Vagrant box

daverosoff:


rbeezer:
Thanks very much, Dave, for the Vagrant setup, and this write-up.

Commits all smashed together and merged.  I need to clean-up something else before updating web version, but will post on Vagrant thread when I do.

Rob

PR #837, opened 2018-02-13
Add option to asymptote command for no-GUI support

daverosoff:
The `mbx` script's `asymptote_conversion` function uses a command that tries to open an X11 display when creating SVG graphics. This didn't work on the Vagrant box, which is CLI-only, but adding one switch at the command line seems to resolve the problem. Tested on a GUI Linux installation to make sure that `mbx -c asy -f svg ...` still works there. 😄


rbeezer:
Thanks for the fix, Dave.  Appreciate you looking after this one.  -Rob

PR #840, opened 2018-02-18
Add to faq

davidfarmer:
I moved part of the Author FAQ into the (newly created) Publisher FAQ,
and copied over the items from the online FAQ.  Still needs work, but it is better.

At the end I pulled upstream dev and then merged dev into this branch.
That cause a merge commit, so I don't see what was gained by doing it that way.
Maybe I was supposed to rebase?


rbeezer:
Sounds good!  Thanks.  I'll get to it first chance I have.

rbeezer:
Thanks, David, for doing this.  First of two.

The merge commit is/was not necessary, and mostly is a bit in the way for how I am working as the sole committer.  A rebase of your branch onto dev would be better, but again not necessary, since it was the first thing I did (after removing the merge commit).

I had already deleted `source-formatting.xml` in a commit **after** where you built your branch.  Interesting to note that
1.  Your merge recognized the duplication and did not do anything extra about it.
2.  My rebase automatically delteted your commit removing the file.

rbeezer:
The Publisher's FAQ has several `xref` to the principles in the Author's Guide.  These raise warnings when you build.  We do not yet have a good scheme for cross-book references (and I'm not ready to solve that now).  So I'd suggest duplicating the Principles in the Publisher's Guide.  As an `xincluded` file we can at least easily keep/check the two copies in sync.  I can do this - please do not add new commits since I do not want to do the rebase, etc., over again.  Let me know if there is a simpler solution.

davidfarmer:
Fine for you to do that.

I am ready to move on to thinking about the FAQ on the
public website.

rbeezer:
OK, done and public.  Lots of subsequent edits, and worked on duplicating the list of principles (and syncing to main README).  Put the principles into an appendix of the Publisher's Guide - there did not seem a good place for it in front.

Further FAQ discusion back on some earlier (non-PR) thread, in a minute.

PR #842, opened 2018-03-01
modify mbx to add eps output for asymptote

mdoob:
It is useful to allow asymptote to output its (default) eps files. Some 3d files will fail using pdf output but will succeed with eps. This helps in debugging, and, in the worst case, allows a really bad file to be converted to svg by exceptional special processing.


rbeezer:
Thanks very much, Michael!  I didn't think it would be that easy. ;-)

Minor edits to commit message, all now public, and will announce on `pretext-dev`.

Since you built your commit on the `dev` branch, you may have trouble pulling since it will be in the way.  See https://github.com/rbeezer/mathbook/blob/dev/CONTRIBUTING.md for some advice about avoiding this problem next time.

Thanks for being our Asymptote expert!

PR #847, opened 2018-03-19
Add variety to the entries in a cols2 list

davidfarmer:
CSS problems were revealed when a cols2 list contains long entries:
the words in one item would overlap the label of the next item.

The CSS has been fixed.  Now the sample article contains an example.


rbeezer:
Merged as 5295bf12e828463c3a995490e7362660e77bf051 (forgot to tag it with the issue number).

Not yet live, likely in about  36 hours from now.  Thanks for the upgrade to the testing examples.

PR #849, opened 2018-03-26
Allow math mode in titles of tcolorbox elements

oscarlevin:
Previously, titles of assemblages (and I assume asides) that contained math mode content would throw an error when compiling with pdflatex.  To fix this, I wrapped the title content of each of these with braces, as suggested here: https://tex.stackexchange.com/questions/204577/math-inside-title-in-tcolorbox.

Added example to sample article.  Compiles correctly (latex and html work).


rbeezer:
Very good!  Thanks for catching this - once I get serious about lots of `tcolorbox` stuff we'll want to do this routinely.

One spelling fix, and minor edits to the commit messages.  Thanks for a neat and tidy contribution.

PR #855, opened 2018-04-10
Update LaTeX error entry in the FAQ

davidfarmer:
The LaTeX entry now emphasizes the likelihood that problems come from the installation.


rbeezer:
Excellent!  (I forgot we already has an entry!)

I snuck in an `xref` to schema validation where you mentioned it.  Thanks for putting this together.

PR #860, opened 2018-05-01
Kernel selection for Jupyter

vincentcote87:
I have added the ability to choose a python 3 kernel instead of the default sagemath kernel when runing mathbook-jupyter.xsl. To use: `xsltproc --stringparam kernel python3 /xsl/mathbook-jupyter.xsl /myArticle.xml`.
If the user does not add --stringparam or simply tries to add a non-existent kernel it will default to using sagemath. It currently only works with python 3 but more kernels could be added inside the `<xsl:choose>` statement ass needed/requested.

### DIFF
--- a/xsl/mathbook-jupyter.xsl
+++ b/xsl/mathbook-jupyter.xsl
@@ -252,14 +252,14 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:text>"language_info": {</xsl:text>
         <xsl:text>"codemirror_mode": {</xsl:text>
         <xsl:text>"name": "ipython", </xsl:text>
-        <xsl:text>"version": 2</xsl:text>
+        <xsl:text>"version": 3</xsl:text>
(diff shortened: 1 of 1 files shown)


rbeezer:
Dear Vincent,

Looks great!  Thanks.  I'll get to it today or tomorrow, and will likely change the name to `jupyter.kernel` to make it clear that it is Jupyter-specific (without changing giving credit to you).

Rob

kcrisman:
Nice.  Perhaps eventually other kernels could be added as well ... great for flexibility!

rbeezer:
Made the parameter more specific, as suggested, and set it to the empty string initially (this makes eventual deprecation warnings easier to handle).  Light editing of the commit messages.

You'll want to trash your local branch with this work, since a pull will get this version.  I didn't do any testing, so holler if anything seems amiss.

Thanks for the contributions, I'll look forward to more.  If you decide to do something bigger, discussion on `pretext-dev` in advance is always welcome.

Thanks again.

Rob

PR #862, opened 2018-05-02
ePub progress

mitchkeller:
* Change build.sh to use mjpage from mathjax-node-page, since page2svg has been removed from mathjaxj-node
* That change required some changes to the sed scripts, since mjpage doesn't behave identically to page2svg.
* Add CSS. Involves, as a temporary measure, bundling the CSS file alongside build.sh. XSLT updated to add CSS to manifest and put link in the content XHTML files. Since mjpage makes the link tag not be self-closing, more sed magic required.
* To produce smaller .epub files, I've changed build.sh to remove all PDF files from the images directory.

### DIFF
--- a/examples/epub/build.sh
+++ b/examples/epub/build.sh
@@ -14,10 +14,10 @@ declare MB=${HOME}/mathbook/mathbook
 declare MBXSL=${MB}/xsl
 declare EPUBSCRIPT=${MB}/examples/epub
 
-# mathjax-node paths
+# mathjax-node-page paths
 # requires installation, see
-# https://github.com/mathjax/MathJax-node
-declare MJNODE=/opt/node_modules/mathjax-node
+# https://github.com/pkra/mathjax-node-page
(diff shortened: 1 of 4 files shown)


rbeezer:
OK, stream-of-consciousness.  May go for a few hours, wait before digesting it all?

* Wrestled package managers and seem to have a current enough `mathjax-node-page`
* The `""` in the `sed` commands cause failures for me.
* Switched to `-if` (combine flags).  Will that work for you?
* Then sample built just fine, by all appearances.

Obvious improvements - sample looks real good.  I'll see what I can chip away at on my end.

rbeezer:
Merged, with some editorial work on commit messages.  Rightly or wrongly,

* I don't use periods.
* I try to mention what conversion is involved (or in -common).
* I use EPUB.  I've seen both semi-officially.
* I rebase on dev, and add PR numbers, for a bit of independence from GH.

Thanks very much - great to see real progress on this one.

PR #861, opened 2018-05-01
Move WW sample chapter from PCC to AIM

Alex-Jordan:
Don't merge this (yet). But some testing and feedback would be appropriate and appreciated.

This moves the sample chapter to rely on AIM's -dev server instead of PCC's server. It required generalizing the mbx script some since it handles placement of temporary image files a little differently. Should be more robust now for a wider variety of server configurations. 

Test by making the sample-chapter-server-tex and the sample-chapter-latex PDF. I did this using both webwork.pcc.edu and webwork-dev.aimath.org, and the diff on the tex files was empty (except the time stamp).

Wait for merging. Two things still could be better. Once the -ptx server is caught up to -dev, this should use -ptx instead of -dev. And also even though -dev is on the develop branch of webwork, something is not right yet because the webwork extraction tool from the mbx script is not working. with the webwork-dev.aimath.org server. (It appears to work fine still with webwork.pcc.edu.)

PR #867, opened 2018-05-07
Move WW server from PCC to AIM

Alex-Jordan:
Some kind of gitastrophe happened when I rebased the branch from which comes PR #861. This should be OK.

Note this one uses the AIM -ptx server. Everything is working on my end. Also I set all the `verify` to `False` for the requests sessions. I'm wondering if that is the issue with your earlier attempts. (It would help confirm that if you have trouble with `make sample-chapter-extraction` even before this PR.)

So if this works for you, it's ready to merge.

### DIFF
--- a/examples/webwork/Makefile
+++ b/examples/webwork/Makefile
@@ -81,13 +81,10 @@ EPUBOUT    = $(SCRATCH)/epub
 WWOUT      = $(SCRATCH)/webwork-extraction
 
 # Some aspects of producing these examples require a WeBWorK server.
-# For all but trivial testing or examples, please look into setting
-# up your own WeBWorK server, or consult Alex Jordan about the use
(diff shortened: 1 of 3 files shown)


rbeezer:
Yes, I would guess the false could be the solution.  On a different machine now, I'll see if I can make progress.

rbeezer:
Seems `verify=False` is necessary for me.  Hopefully we can figure out why I can't get certificates to work properly.

I updated `pip` and `requests`, which updated `certifi`.  I forced `mbx` to use `python3`.  Still similar errors, and paths all suggest this is using the updated routines.

Merged as-is.

The `graphics_pattern` regex has changed, and I've included a generic-ish example just before it, to help a reader have some idea of what is being matched.  Could you update that comment when you get the chance, as part of anything else when convenient?

PR #869, opened 2018-05-11
convert argument to required binary type

daverosoff:
I couldn't process the sample article with `mbx -c mom` without this fix. But maybe I am missing something?


rbeezer:
Definitely a Python2 vs. Python3 deal.  My Python is still back on "2".  Lesson learned - I should explicitly test on both 2 and 3.

Sorry for the delay, I was in Idaho last week. ;-)  Merged, etc., so you can trash your branch.  Thanks for taking the time to put this together.

PR #870, opened 2018-05-12
Refactor WW LaTeX production

Alex-Jordan:
OK, here is the new scheme for making a PDF with WW. For the sample chapter:

```
make sample-chapter-extraction
make sample-chapter merge
make sample-chapter-pdf
```

and see the output. Viewing the diff with the old .tex will not be as helpful as usual since large chunks of tex for WW problems changes. But somewhat helpful.

As discussed (perhaps long ago) this is not backwards compatible. Once this is merged, people need to use either a version 2.14 WW server (which doesn't exist yet) or a development branch server (like the PCC server or the AIM server).

So it's your call how to move forward. One option is that you test this, and when you are ready, you commit it to some other branch (possibly tweaking the commit message and minor details as usual). Then make that branch public so I can work from it. Then I can do the HTML, further author guide edits, and a few other things on that branch, with more pull requests to your copy of that branch. Eventually, you could merge the entire branch into dev, and we could announce.

Or after reviewing this, you could merge into dev right now and we announce that "the old way for print no longer works, sorry. But there is the AIM server you can use instead..."

I know there is a lot here to look through. I think it might be best if I let you ask questions rather than preemptively answer everything I can think of.


rbeezer:
First pass/review looks real good.

I think I would want to just merge this and announce the change.  It is about the start of summer, and we do have the AIM servers now.  People can experiment without panic, we can fix stuff, and you can move forward.  When do you guess 2.14 will be universally available?  By September?

Minor comments.  I can make changes if you are in agreement.  Two minor typos (whitespace) that I will fix and just merge in.

*  Author Guide - `href="webwork-ptx.aimath.org"`, I use a http here and use monospace content for a more human-readable version.
*  I've gravitated to `<instructions>` as part of the new `interactive`.  Visible in HTML, maybe not in PDF.  Just multiple "p", I think.  Do you want to go larger?  (Yes, I was the one keeping this back.)  If so, lets do it on a new PR/commit, not now.
*  `extract-pg-ptx.xsl` gets `str` extension in early going.  Is this necessary since `-common` is imported?  (I forget how this goes.)  Yes/no?
* Any way to get a decent radio button without the `waysysym` package?  We have packages already for `\textbigcircle`, `\textopenbullet` and `\circledcirc`.
* The big six-part matches are good - I never answered your question on that one.  We should save the `\\` for when it really is undetermined how many levels, or how much stuff, comes in-between.
* The numbering for `stage` should go in `-common`.  I'll make this change, since numbering is a bit delicate.

Did you enjoy obliterating huge chunks of complicated code?  Or was it bittersweet?  ;-)

I'll do another pass soon, actually test the code, and might have more comments.

Rob

Alex-Jordan:
Just realized that the `mbx` script should lose the `-c webwork-tex` option and all the associated code. I'll see if I can cut that in a follow-up commit here.

rbeezer:
On 05/12/2018 03:43 PM, Alex Jordan wrote:
> Thanks Rob, this all sounds good. More specifically:

Thanks for the replies and the additional commit.  I'll get serious about this 
one first thing Monday morning.  This week is a good one for me to prioritize 
this, though I know you have to prepare for the workshop.

Rob

rbeezer:
* I used `\odot` for the radio button.  Yes, it might look "pressed", but they all are, so it looks to me more like a round 3-D pushbutton.  ;-)  I preserved options in a comment.  I'll take the blame on a separate commit.
* I think `instuctions/p` should be the way forward.  Would it be reasonable to say, "these are comments on the HTML interface" and therefore a `var` element would be inappropriate here?  It just clutters up the schema. ;-)
* I put the `webwork-tex` component back in, and have it throw a `NotImplementedError` - we should make sure people get some hint when something has changed.
* I see the problem with the number for `stage`.  We may not see it until the server sends back a static version of an OPL problem.  I put both versions into `mathbook-common.xsl` conditioning on `webwork` versus `static`.  Eventually we may not need the `webwork` version?
* I think the "rsq" element for contractions and the like is probably overkill, and could possibly lead to some variance with the output for the remainder where an author would not have done this (in more obscure conversions, like Jupyter, EPUB).  More and more, I'm inclined to to take keyboard-text and clean it up automatically.  For example, we should convert `#` to `\#` in LaTeX output rather than expect the use of `<hash />`.

Nice job.  Thanks for sticking with this one - I'm looking forward to having it all together.

Alex-Jordan:
Thanks for the review. Can I post a thing or two following up on -announce? I will keep it short and to the point.

> I used \odot for the radio button. Yes, it might look "pressed", but they all are, so it looks to me more like a round 3-D pushbutton. ;-) I preserved options in a comment. I'll take the blame on a separate commit.

It's fine for now. I think a slightly bigger concern is the use case of a student who chooses to actually fill in bubbles. But I'd like to some day have a small tikzpicture make a reasonable button image.

> I think instuctions/p should be the way forward. Would it be reasonable to say, "these are comments on the HTML interface" and therefore a var element would be inappropriate here? It just clutters up the schema. ;-)

An actual example from ORCCA: there is a problem that randomly is presented in inches, centimeters, meters, feet, etc. The answer includes units, entered as abbreviations. So we have: 
```
<instruction>Use <var name="$unitabbrev" /> for <var name="$unit" />.</instruction>
```
which might come out "Use cm for centimeters." I wonder if "cm" should be in a "c" tag, but then I'm asking for "var" inside "c" too :) Anyway this is one example where the instructions could vary with the random version, hence the need for "var" inside.

> I see the problem with the number for stage. We may not see it until the server sends back a static version of an OPL problem. I put both versions into mathbook-common.xsl conditioning on webwork versus static. Eventually we may not need the webwork version?

I think we won't need it. Part of what I want to add to announce touches on this. Until HTML is finished, I'm a tad worried that maybe some numbering will be off between HTML and PDF. But once both are generated from the merged output, then it will be OK. And once both come from merged, then the "webwork" stage number maybe can go away.

> I think the "rsq" element for contractions and the like is probably overkill, and could possibly lead to some variance with the output for the remainder where an author would not have done this (in more obscure conversions, like Jupyter, EPUB). More and more, I'm inclined to to take keyboard-text and clean it up automatically. For example, we should convert # to \# in LaTeX output rather than expect the use of <hash />.

Is this about the "rsq" that comes back from the server? PGML is intercepting quotes and balancing them, and I had to be careful. But maybe I can just have PGML output simple quotes to PTX output.

rbeezer:
On 05/14/2018 02:40 PM, Alex Jordan wrote:
> Thanks for the review. Can I post a thing or two following up on -announce? I 
> will keep it short and to the point.

Certainly. You have privileges, right?

> It's fine for now. I think a slightly bigger concern is the use case of a 
> student who chooses to actually fill in bubbles. But I'd like to some day have a 
> small tikzpicture make a reasonable button image.

Thought of that.  ;-)

> An actual example from ORCCA: there is a problem that randomly is presented in 
> inches, centimeters, meters, feet, etc. The answer includes units, entered as 
> abbreviations. So we have:
> 
> |<instruction>Use <var name="$unitabbrev" /> for <var name="$unit" 
> />.</instruction> |
> 
> which might come out "Use cm for centimeters." I wonder if "cm" should be in a 
> "c" tag, but then I'm asking for "var" inside "c" too :) Anyway this is one 
> example where the instructions could vary with the random version, hence the 
> need for "var" inside.

OK, I knew you would have an example.  When we clean-up, remind about this for 
the schema.

> I think we won't need it. Part of what I want to add to announce touches on 
> this. Until HTML is finished, I'm a tad worried that maybe some numbering will 
> be off between HTML and PDF. But once both are generated from the merged output, 
> then it will be OK. And once both come from merged, then the "webwork" stage 
> number maybe can go away.

If the numbering ever goes off, I'd like to see it!

> Is this about the "rsq" that comes back from the server? PGML is intercepting 
> quotes and balancing them, and I had to be careful. But maybe I can just have 
> PGML output simple quotes to PTX output.

Right.  Its the apostrophe in, for example, "don't".

Alex-Jordan:
Found an issue, and wondering if you can just apply a quick commit (or see a better solution). 

Symptom: If there is a sidebyside with an image (all within a webwork) and there is no widths attribute on the sidebyside, you get "NaN" for the image width in the PG output, leading to all kinds of follow-up issues.

This comes from `mathbook-common` where I put an exception for getting width percentage on such an image (currently at line 2911). A simple fix is to apply this patch:
```
@@ -2910,7 +2910,7 @@ Neither: A structural node that is simply a (visual) subdivision of a chunk
 
 <!-- The exception is an image inside a sidebyside in a webwork  -->
 <!-- where the parent sidebyside should only have a single percentage in its @widths -->
-<xsl:template match="image[ancestor::sidebyside][ancestor::webwork]" mode="get-width-percentage">
+<xsl:template match="image[ancestor::sidebyside/@widths][ancestor::webwork]" mode="get-width-percentage">
     <xsl:value-of select="parent::sidebyside/@widths" />
 </xsl:template>
```

Then this template only applies when there is a `@widths` and otherwise the preceding template defaults it to `100%`. Since this is so small, could you just apply this?

I wonder if -common is ultimately not the right place for this, but for now I think both extract-pg-common and mathbook-html rely on this.

rbeezer:
OK, done.  Once there is a one-shot sidebyside, we'll need to get this sorted 
out properly.

I adjusted the match some, I think it is equivalent in functionality.

PR #871, opened 2018-05-17
fixed server name in live html output instructions

oscarlevin:
The current instructions for creating live HTML webwork output suggest using the webwork.server string param as `webwork-ptx.aimath.org`, but this does not work: you need the `https://` in front of it.

### DIFF
--- a/doc/author-guide/webwork.xml
+++ b/doc/author-guide/webwork.xml
@@ -33,9 +33,9 @@
             my $out = MODES(
                 TeX => '',
                 Latex2HTML => '',
-                HTML => qq!<P>If you were logged into a WeBWorK course 
-                and this problem were assigned to you, 
-                you would be able to submit an essay answer 
+                HTML => qq!<P>If you were logged into a WeBWorK course
+                and this problem were assigned to you,
+                you would be able to submit an essay answer
                 that would be graded later by a human being.</P>!,
                 PTX => '',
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Sorry, rushing through the documentation, knowing a fully thorough review will come at the end of all this. I just opened #874, with the HTML move and more documentation edits, which I think makes this one moot.

rbeezer:
First, my fault.

Second, thanks Oscar, for the fix.

Third, I'd like to have Oscar's fix go into the record.

Fourth, Alex had about six stray blanks at the end of lines.  Oscar's commit removed them, making the commit way bigger than necessary.

(Review your commits, patrol whitespace, set your editor to show tabs and blanks visually.)

So, I put back Alex's blanks, so Oscar's commit looks right, and merged that.

@Alex-Jordan You can back out your edit on current branch before re-basing, **IF** it comes to that.  Don't do anything about this now, I'll deal with it tomorrow.

rbeezer:
On 05/18/2018 07:24 AM, Oscar Levin wrote:
> Sorry, it turns out that my editor was set up to remove trailing whitespace and 
> add a blank newline at the end of the document whenever it saves. I'll be more 
> careful of this in the future.

No problem - normally that would be a very good thing, just an accumulation of 
stuff here.

PR #872, opened 2018-05-17
The enclosing tag for an aside is 'aside'

davidfarmer:
Asides were an "article".  Now they are an "aside".

The change may seem trivial, but it is a step toward refactoring asides.
With this change, when the screen is narrow an aside will look different:
it will be smaller and shifted right, and the other text will flow around it.


rbeezer:
Excellent - will likely merge in the AM.

rbeezer:
Thanks.  Code is pushed - I'll update website examples before calling it a day.

PR #874, opened 2018-05-18
Refactor WeBWorK HTML production

Alex-Jordan:
Here it is. I deleted the base64 xsl files. Now I'm wondering if I should have waited until the problem set creation is ready too...


rbeezer:
Looking good at first blush.  I'm glad to see that final `xsl:include` going away!  That caused more trouble...  But I'll miss the base64 library...

I'll take a hard look in the morning.

rbeezer:
Dear Alex,

Thanks for all this!  Good to see so much old code going away.

*  Knowl production (non-WW) was injured.  I rewrote some of the base code, and a bit of your new code, and put as much of the base work as I could on a commit with my name on it.
*  The modal `structure-number` template matching `webwork-reps//*` seems overly broad, and my attempts to narrow it down all failed.  That'll need to be understood and fixed.
*  I removed the "inline math" delimiters from the end of the problem set generation.  Diff suggested no problem, and I think there is no way this would have any effect here anyway.

Those were the show-stoppers, it seemed to me, so I am merging this much.  There is some more work I want to do (typos, names, some tighter code) and I'll note that here (so am leaving this open for now).

Rob

Alex-Jordan:
Thanks for following up with all the cleanup. 

We had `@category` in the `setup/var` for two reasons. One was pretty vague about just capturing some metadata for future yet-to-be-invented purposes. But the other was to go with the AnswerFormatHelp macro from WeBWorK (specifically from Paul Pearson at Ft Lewis). That macro cuurently accepts the types listed here at line 12: https://github.com/openwebwork/webwork-open-problem-library/blob/master/OpenProblemLibrary/macros/FortLewis/AnswerFormatHelp.pl

Now, there you have words that are for such and such type of answer, like "decimals". But as an attribute for a single var in PTX, we felt other words were more appropriate, like "decimal". We also felt that marking a var as an "integer" could be meaningful, even though AnswerFormatHelp doesn't have "integers". (It does have "numbers".)

To answer your question, we don't have a perfectly defined list of options. We could make one. But it should be allowed to grow. For example I would like to enhance AnswerFormatHelp to cover formulas-with-domain-restrictions like here: https://webwork.pcc.edu/webwork2/html2xml?&amp;answersSubmitted=0&amp;sourceFilePath=local/RestrictedDomainsDemo.pg&amp;problemSeed=1&amp;displayMode=MathJax&amp;courseID=anonymous&amp;userID=anonymous&amp;course_password=anonymous&amp;outputformat=simple

So that would be a new string for AnswerFormatHelp, and a possibly different new string for `@category` to take on.

With `static/var` gone, the place to keep `@category` is in a `var` that serves as an answer blank. So a `var[@width or @form]`. I have not updated any documentation for this, knowing we have some sorting out to do first.

rbeezer:
Thanks for the reminders. ;-)  I found the list of @category we are supporting. 
I'm going to hard-code those in the schema.  It can grow, and will be an easy 
edit when you want to add something.  That'll all get us easy error-checking.

Are  @width  and  @form  mutually-exclusive?

rbeezer:
I'm going to call an `exercise` in an `exercises` "divisional" now (rather than "sectional").  So `webwork.sectional.static` is now `webwork.divisional.static`.

At 0c76d27fd8041ed1137f21623883228123c74c8f

Alex-Jordan:
> Are `@width` and `@form` mutually-exclusive?

No. Perhaps only because of something like `<var form="array" width="10" />` for a matrix or vector answer with an answer blank for each entry, with each answer 10 wide.

For how to chunk out the homework sets, we should discuss it this week. I'm not sure what we went with before is best. Every time there is a run of exercises in the same series, that should end up as an exercise set with the serial number being the problem number in the WW set. (Maybe that's impossible to make happen.) Here's a complicated stress case:

```
chapter
  exercise
  exercises
    exercise
  exercise
  exercises
    exercise
chapter
  section
    exercise
    exercises
      exercise
      exercise
    exercise
    exercises
      exercise
      exercise
```

Do I understand right that this would make Checkpoint 1.1--1.2, Exercise 1.1--1.2, Checkpoint 2.1.1--2.1.2, Exercise 2.1.1--2.1.4? If so, I would ideally want to see this lead to four problem sets. No declaring of chunk level.

rbeezer:
Thanks for the answer on schema question.

Short answer:  "exercises" must come at the end of their containing division now, so your stress cases are prohibited.

rbeezer:
I think the stress-test is easy and can happen just as you say.

It may mean a new "chunking" routine, but I think that will be good to do as 
well.  (Necessary for "solution-list" also.)

rbeezer:
@Alex-Jordan Any reason to not close this?

Alex-Jordan:
Closing this. Unsure what has already been done separately now. But anyway with the move from html&iframe to json&div coming up, there is surely nothing to save here.

PR #875, opened 2018-05-18
Fix spelling of 'occurence'

Alex-Jordan:


rbeezer:
Thanks!  That's pretty bad.  I hope it didn't trip you up someplace.

PR #876, opened 2018-05-18
New temporary CSS file

davidfarmer:
As part of the transition to the new CSS, there will be some backward-incompatible changes.
This pull request will make that less painful.
The only real change is to have the HTML use mathbook-add-on2.css instead of the old mathbook-add-on.css .

There was some clean-up of the CSS, but the only obvious difference is that the "pink-flash"
works better when you link to a complicated structure like a subsection (previously it looked "jagged").

I also fixed a couple typos and deleted some trailing white space.  I note that in the view on
GitHub it is not easy to see exactly what typos were fixed.  Score one for structured source.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -516,7 +516,7 @@ the xsltproc executable.
                 </output>
             </sage>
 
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks for thinking ahead to make the transition less painful.  And the drive-by clean-up.

Merged, and I'm rebuilding website examples now in case you are doing some testing.

PR #879, opened 2018-05-24
Add aria-label to TOC toggle button.

davidfarmer:
The "hamburger" icon in the TOC toggle button is distracting for users of screen
readers.  This  aria-label may correct that problem, unless there is an issue with the
character being inserted as a ":before" element from CSS.


rbeezer:
Thanks.  We'll get some user-testing with assistive tech.  I'll rebuild website once Alex is done.

PR #887, opened 2018-05-25
Make aside#sidebar-left into a div.

davidfarmer:
After thinking about it, I decided that wrapping the TOC in an aside can only cause confusion.
So I made it a div.

Visually it is at the side of the page, but it is not aside-like in terms of content.  The navigation
role on the div is appropriate.


rbeezer:
Very good!  Rebuilding now.

PR #901, opened 2018-06-15
Text after a multicolumn list

davidfarmer:
There was a long-standing but unnoticed CSS problem when a multicolumn list
has a paragraph after it.  This adds an example.

The CSS problem has not been fixed yet.


davidfarmer:
The CSS now should handle this case (please check).

The way it is handled is not optimal, so there could be some cases
of multicolumn that now have too much vertical spacing after.

rbeezer:
Thanks.  I'll get the test case merged in the next few days.

rbeezer:
Going live on the website now.  Looks good to my eye.

PR #902, opened 2018-06-16
Support multicolumn lists in static WW output

Alex-Jordan:
This makes it so that when you author a WW problem in your PTX source, and it has a list with `@cols`, then in static output (print and a static HTML version) your `@cols` will be respected. To do this it has to encounter the list in `static` and look back for the corresponding list in `authored`.

Test that this doesn't affect any output in an adverse way. Test that the print sample-chapter behaves as indicated in checkpoint 5.2. Check the same with the HTML output, but you need to set the stringparam `webwork.inline.static` to `yes`.

You may see good ways to compactify the XSLT.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -599,6 +599,79 @@
                 </webwork>
             </exercise>
 
+            <exercise>
+                <introduction>
+                    <p>This exercise illustrates what happens when source has a multicolumn list. In <webwork /> itself, there is no provisioning for multicolumn lists, so live output here should not respect the indicated <c>cols</c> attribute. However, when <pretext /> produces static output, we do recover the value of <c>cols</c>. So in PDF and static HTML these lists should appear with multiple columns.</p>
+                </introduction>
+                <webwork>
+                    <statement>
+                        <p>This list has three columns in static output.<ol cols="3">
+                            <li>
+                                <p>Argentina</p>
+                            </li>
+                            <li>
+                                <p>Belize</p>
+                            </li>
+                            <li>
+                                <p>Canada</p>
+                            </li>
+                            <li>
+                                <p>Denmark</p>
+                            </li>
+                            <li>
+                                <p>Ethiopia</p>
+                            </li>
+                        </ol></p>
+                        <p>This list has one column in static output.<ol>
+                            <li>
+                                <p>Finland</p>
+                            </li>
+                            <li>
+                                <p>Ghana</p>
+                            </li>
+                        </ol></p>
+                    </statement>
+                    <hint>
+                        <p>This (unordered) list has two columns in static output.<ul cols="2">
+                            <li>
+                                <p>Hungary</p>
+                            </li>
+                            <li>
+                                <p>Iceland</p>
+                            </li>
+                            <li>
+                                <p>Jamaica</p>
+                            </li>
+                            <li>
+                                <p>Korea</p>
+                            </li>
+                            <li>
+                                <p>Lesotho</p>
+                            </li>
+                        </ul></p>
+                    </hint>
+                    <solution>
+                        <p>This list has five columns in static output.<ol cols="5">
+                            <li>
+                                <p>Mali</p>
+                            </li>
+                            <li>
+                                <p>Nepal</p>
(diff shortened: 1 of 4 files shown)


rbeezer:
"get-cols" is clearly tricky business in the static case.  Some XSL suggestions:

* Return value could be an integer (or string that converts to an integer)?  An "otherwise" clause could provide an empty string to test against?  This does not appear to be a place where RTFs should be involved.  They have their own problems, and may make for trouble if we ever move to 2.0.
* name() is best replaced by use of self::
* If you counted a node set of the combination of all ul and ol together then the logic could be much simpler?  And previous comment becomes moot?

Alex-Jordan:
Thanks, and no rush.

Yes, this is the first place that `authored` would really be used. I can imagine some table features coming through that way too someday, but nothing else has yet occurred to me. The multiple columns in print is desirable because of wasted white space.

If `authored` stays, it could be introduced with the merge instead of the webwork extraction mbx script run. Now that I'm changing a `@cols` attribute to adjust PDF layout, it is irritating to have re-run the extraction.

It's possible to hack around this if you really want to jettison `authored`. In PG, the list could be preceded by some PGML comment. And if the comment says the right thing in the right way, we could take that to the list after the round trip. Or, there could be an entirely new PGML delimiter solely for the purpose of preserving PTX attributes. 

The cost (either way) is that PreTeXt would perform more as expected based on narrow PreTeXt-specific enhancements/modifications to another piece of software.

While thinking it over more, I'll look at XSLT improvements.

rbeezer:
I'm imagining a future where after the merge, there isn't even a "static".  Just an "exercise" with maybe an attribute saying it was born as a webwork problem.  Then it has the base64 versions as additional content, and anything else necessary for specialized conversions (eg, PG version).  But if you hit it with any other conversion, it just automatically does the right thing.  So for example, the LaTeX conversion might never even need to mention "webwork-reps".

Assuming this is possible (send counter-examples!), I'd like to see some mechanism where the number of columns makes the trip through the WW server and is placed in the post-merge XML just as if it was authored by a human.

Related, but more philosophical.  A WW problem is really an HTML thing, and the LaTeX/PDF version is just an imitation.  The new "interactive" has the same flavor.  Real in HTML, substandard in print.  Here the imitation/substandard version is getting a feature the real version does not have.  That's a (significant) deviation from how I think about incorprating items that **cannot** be rendered faithfully in all output formats.

Alex-Jordan:
> Assuming this is possible (send counter-examples!), I'd like to see some mechanism where the number of columns makes the trip through the WW server and is placed in the post-merge XML just as if it was authored by a human.

That sounds like I need to either do the comment hack on the WW end, or make the new PGML delimiter for preserving PTX attributes. I'll raise the latter with Mike and Davide and see if they have objections.

> Related, but more philosophical. A WW problem is really an HTML thing, and the LaTeX/PDF version is just an imitation. 

I could quibble with this. WeBWorK has always been intended for multiple output modes. It's just that HTML is the most popular. See WW hardcopies. In that sense it may be like PreTeXt.

> The new "interactive" has the same flavor. Real in HTML, substandard in print. Here the imitation/substandard version is getting a feature the real version does not have. That's a (significant) deviation from how I think about incorprating items that cannot be rendered faithfully in all output formats.

There is a difference worth noting. With WW we (can) author in PTX, using PTX elements. Not exclusively (because of the pg-code) but the statement block is PTX. The other interactives use their native language only, either included as a source attribute or something like a block of code maybe wrapped in a CDATA. Once we went that way (instead of just using pure PG all the time) we are in a place where authors will (reasonably, I think) expect basic markup to behave like it does elsewhere in PTX. So for example, `@cols` on a list. Then again, we do have to disappoint in some places. Like no `figure` within a webwork statement even though those are fine in a regular exercise statement.

Alex-Jordan:
I'm going to close this. I have a more streamlined approach now that I use in ORCCA PDF production that uses better XSL-fu. Is it worth pursuing for merging into mathbook? It's still based on the idea that:

1. When it is a WW that was authored in PTX source...
2. You look at the PTX that comes back from the WW server for ol/ul, which can't possibly have a @cols because of the limitations of PGML...
3. And you find the corresponding ol/ul in the authored part and see if it has @cols.

PR #905, opened 2018-06-22
Support true display math in webwork

Alex-Jordan:
New PGML delimiters ````[```...```]```` have been merged into the webwork develop branch. I pulled into both AIM WeBWorK severs, as well as the PCC webwork server. These new delimiters are for true display math, as opposed to ```[``...``]``` which just put `\displaystyle{}` in front of inline math. 

This commit changes how PTX converts to PG, now making use of the new display math delimiter instead of faking it with centering and ```[``...``]```.

### DIFF
--- a/xsl/extract-pg-common.xsl
+++ b/xsl/extract-pg-common.xsl
@@ -1204,7 +1204,7 @@
 <!-- ######### -->
 
 <!-- extract-pg.xsl documentation -->
-<!-- PGML inline math uses its own delimiters: [`...`] and [``...``]       -->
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I forgot about testing. In the sample-chapter, the solutions to checkpoints 1.4 and 1.5 use an md and an me respectively. If you build the pdf before this commit, then you will see the content left-aligned. After the commit, these things come out like real display math. You have to go through the extraction, merge, pdf process.

rbeezer:
Excellent!  Thanks.  Going live now...

PR #906, opened 2018-06-23
updated the pug instructions

alexjbest:
I also added some pitfalls and made it pass the schema.

I think using the tag interpolation feature of pug i.e.  `#[m e^x]` rather than `<m>e^x</m>` is better in general so I have changed it to use that.

I really think more people should be using pug, its great! Maybe I'll try and add some more reasons if there is interest (based on some of the tricks I use in <https://github.com/alexjbest/aut-forms-arthur-selberg> and other projects).

### DIFF
--- a/examples/pug/pug-include.pug
+++ b/examples/pug/pug-include.pug
@@ -1,13 +1,14 @@
 - var name = "content of variable <q>name</q>"
 
-p.
-    Besides these basics, Pug also offers <q>true</q> includes.
-    This means, just write the text and don't think about additional overheads,
-    like that it also has to be an XML document.
+introduction
+    p.
+        Besides these basics, Pug also offers #[q true] includes.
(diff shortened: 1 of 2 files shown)


rbeezer:
Dear Alex,

Thank-you very much for the contribution!  I don't use Pug (maybe I should!), so this has been a bit neglected.  At some point (soon), we should add your name as an author on the example.

I edited the commit message and made a new commit to get rid of the MBX references.  So you can trash your branch, your commit will have a new hash.

Your notes look great!  Maybe I should point to them from the PTX Gallery?

I'd be game to do anything you like to make Pug more visible to the PTX crowd, as an alternative "front end" of sorts:

* An informative post on pretext-announce.
* Something in the Author's Guide about alternative authoring formats.
* Rebuild and post Pug example routinely as part of the website examples page.

Thanks again,
Rob

PR #911, opened 2018-07-03
Update logic to include alphanumeric, to match comment

gondree:
The existing `mbx` sanitization logic for user strings is intended to allow alphanumeric and underscores; however, it was only allowing alpha and underscores. 

This is a tiny little edit that re-writes the check using more appropriate python logic, to allow WebWork UIDs that contain numbers, among other sanitized logic. This also causes the logic to match its documentation. The `ascii_letters` and `digits` macros work in Python 2 and Python 3.

There does not appear to be an existing ticket for this issue, but its pretty trivial fix.

### DIFF
--- a/script/mbx
+++ b/script/mbx
@@ -1086,7 +1086,7 @@ def sanitize_url(url):
 def sanitize_alpha_num_underscore(param):
     """Verify parameter is a string containing only alphanumeric and undescores"""
     import string
-    allowed = set(string.ascii_lowercase + string.ascii_uppercase + '_')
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
This sounds good to me. Thanks for finding this.

rbeezer:
Thanks very much, Mark!  I'll get it merged later today.

rbeezer:
I edited the commit message, so you'll want to trash your own branch.  Thanks again, Mark. -Rob

PR #928, opened 2018-07-27
Webkit bug

davidfarmer:
This upgrades the version of MathJax being used, and removes the Javascript that
was the work-around for the layout bug in Safari.  Now the layout should also look
good in Safari.

I am not sure that I have tested it adequately, but it works on my Macbook.

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -8928,7 +8928,7 @@ var </xsl:text><xsl:value-of select="$applet-parameters" /><xsl:text> = {
     <!-- mathjax javascript -->
     <xsl:element name="script">
         <xsl:attribute name="src">
-            <xsl:text>https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-AMS_CHTML-full</xsl:text>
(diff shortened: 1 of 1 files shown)


rbeezer:
It became necessary to deprecate the `html.js.server` parameter.  At  4375ac4e59db6560fbfc1632555c7cc360da83e4

Several things are still being served from `aimath.org` - I'm assuming you are aware of that.

davidfarmer:
Currently, $html.js.server is still used in at least two lines,
so I am not sure why it is being deprecated now.

davidfarmer:
Never mind: I see what you did.

rbeezer:
I hard-coded those two lines to the defaults.

Since the path for the principal JS changed, the behavior of the switch changed 
substantially, so something had to give.  Code warned it could go away, so it did.

Do you need something for testing?

davidfarmer:
I don't need any testing.  Things worked in the cases
I could test them.

I will be migrating things to the new server.  Not that I also
have started using versions.  It was getting too hard to always
be backward compatible.

rbeezer:
OK, glad we are pretty much on the same page.

> I will be migrating things to the new server. Not that I also
> have started using versions. It was getting too hard to always
> be backward compatible.

So we'll need to be even more insistent about live examples.  Note that each 
source file has a time stamp for build time.  (Maybe I should put PTX git repo 
commits in there as well?)

PR #932, opened 2018-08-01
A better layout for divisional exercises

oscarlevin:
Divisional exercises are presented as list items, with hanging indents matching the first line.  

I could not immediately see how to apply this fix to exercise group exercises, and the inconsistency looks a little strange.

Another small possible improvement I don't know how to make: currently the period in an exercise's title is bold, while the period in the "Solution." is not.

### DIFF
--- a/xsl/mathbook-latex.xsl
+++ b/xsl/mathbook-latex.xsl
@@ -1080,8 +1080,8 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:text>%% Divisional exercises are rendered as faux list items&#xa;</xsl:text>
         <xsl:text>%% with hard-coded numbers as arguments, not as LaTeX environments&#xa;</xsl:text>
         <xsl:text>\NewDocumentEnvironment{divisionexercise}{mo}&#xa;</xsl:text>
-        <xsl:text>  {\textbf{#1}.\IfValueT{#2}{\ \textbf{#2}}\quad}&#xa;</xsl:text>
-        <xsl:text>  {\par\smallskip\noindent}&#xa;</xsl:text>
+        <xsl:text>  {\begin{itemize}\item[\textbf{#1}.]\IfValueT{#2}{\textbf{#2}}}&#xa;</xsl:text>
+        <xsl:text>  {\end{itemize}}&#xa;</xsl:text>
     </xsl:if>
     <xsl:if test="$document-root//list">
         <xsl:text>%% named list environment and style&#xa;</xsl:text>


rbeezer:
Dear Oscar,

A one-item list gives a great look.  But now we need to deal with the `itemize` environment to get vertical spacing between `exercise` (and `exercisegroup`, and the top and bottom of `exercises`).  And `item` itself to control the horizontal indents and spacings.

I've done some experiments and I think I can get this look from `tcolorbox` with fine-grained adjustments to the `title` to get the hanging-indent.  We can then control the spacing in obvious ways and we can migrate this technique to other instances and get consistency and control there.

Thanks for the spacing experiments, we'll be sure to get those in place, too.

Rob

rbeezer:
> Another small possible improvement I don't know how to make: currently the period in an exercise's title is bold, while the period in the "Solution." is not.

And "bolding" the period in a title is correct?

Does the

rbeezer:
Good progress with `tcolorbox`.

You have left margin, extrapolated from "Exercises" heading.  Then indented, you have body of the exercise, which sets another margin.  What is the distance between them?

And in that gap, where do you align the number?  Left?  Right (on period)?

* I'd like to avoid LaTeX's "tell me how wide you widest label is, like in the bibiography"
* All worse/harder in `solutions` - we can worry about that later.
* Maybe we should count characters in all the numbers to set gap described above?  Or the last number?

oscarlevin:
A tcolorbox solution would definitely be preferred.  As for spacing, I would like to see a list of exercises look exactly like a enumerated list in LaTeX, except for the numbers being bold.  This would mean the "standard" extra margin for the indented body of the exercise, and numbers right justified.  I think that two digit numbers would still fall inside the main outer margin.  For 3.5.14, that would stick out a bit, but you wouldn't hardly notice in a list of solutions (there is no regular margin text around, except maybe the title).  Numbers in a `solutions` could be styled completely differently anyway.

rbeezer:
@oscarlevin - thanks again for taking a stab at this.  Hanging indent specifically is isolated at 5937d3afc295ea21588591ec7e04285f043d9920.  Work just prior is more substantial, getting all exercises into tcolorboxes, which then works very well when an `exercisegroup` has columns.

I built the indent to accomodate up to 3-digit numbers in the body of the text.  For solution lists, I decided to be more economical, maybe less pretty, but with lots of navigational clues (running heads to come).  So full numbers (like 4.2.67) but then punted on indent size and just made them "run-in" titles.

Periods are being added to titles **before** new styling work sees them, so they are bold if the title is bold.  Could still be some inconsistencies since not all of the styling infrastructure is in place.

I'm going to close this now, lets continue any discussion using the groups.  Thanks again.

Rob

PR #933, opened 2018-08-07
Load Inconsolata and define \mono when tag is used

mitchkeller:
Discovered while working on writing a worksheet that if you use `<tag>`, LaTeX code is produced that uses the \mono macro but it's not defined.


rbeezer:
Good catch, Mitch.  `tage` and `attr` need the same treatment, so I added those into your commit.  Thanks.

PR #935, opened 2018-08-09
Add missing Czech translations

jirilebl:


rbeezer:
Thanks very much for keeping this current!  Squashed into one, and commit message lightly edited.

PR #940, opened 2018-08-17
Initial CLP style for use with tcbox

arechnitzer:

PR #941, opened 2018-08-17
Latexstyle

arechnitzer:
Trying again... now the file should be in the right place.


Alex-Jordan:
Do I read this right?

First there was #940, where Rob asked for the file to be moved. Then at roughly the same time, Rob made the file move and merged #940, while Andrew opened #941 that also moved the file.

If I read this right, can we close this?

rbeezer:
Right.  I was wondering why it looked like so much of this was already into the repo.  ;-)  I had a mental note to go do a comparison sometime.  #940 is a much better explanation.

PR #943, opened 2018-08-20
Clarify why some knowls appear when viewing locally.

davidfarmer:
Expand an FAQ answer to address confusion in this thread:
https://groups.google.com/forum/?hl=en#!topic/pretext-support/YFxYnpygFBo


kcrisman:
Got it, that makes much more sense (and definitely is a good addition to the FAQ).  I thought _all_ knowls came from files!  Thank you.

rbeezer:
Thanks, David.  Sorry for the delay - I think I missed that this was a PR (and not just an issue).

Forgot to edit commit message, so no PR number - work at 78568702b6986e2cb984df8dd91bc7150ca439d1

PR #950, opened 2018-09-05
Inline CSS to stop 'flash' of MathJax macros.

davidfarmer:
This addresses the 'flash' discussed on pretext-dev:
https://groups.google.com/forum/?hl=en#!topic/pretext-dev/R-xlCNjyxkc


rbeezer:
A good solution, it seems.  Merged -- more on `-dev` thread in a minute.

Thanks.

Issue #952, opened 2018-09-07
Exercisegroup exercises do not resepect exercise.divisional.solution


Alex-Jordan:
Experiencing this with ORCCA. But you can see in the sample article that if you:

1. find an exercisegroup/exercise
2. add to it a hint, answer, and solution
3. run `xsltproc --stringparam exercise.divisional.hint no --stringparam exercise.divisional.answer no --stringparam exercise.divisional.solution no ../../xsl/mathbook-html.xsl sample-article.xml`

The output will be showing the hint, answer, and solution despite all the "no"s. This is probably HTML only, since I recently built ORCCA's PDF and the solutions were not appearing in print. Solutions are appearing in HTML. Curiously, it seems they only appear in HTML for non-WW exercises. Since most ORCCA exercises are WW, it took a while for us to notice this.

rbeezer:
Thanks for the confirmation of the fix over on #953.

PR #953, opened 2018-09-07
Apply global exercise string params to exercisegroup exercises too

Alex-Jordan:
This addresses #952. Haven't thought through everything or scanned for other places this may be needed, but it makes the test case I cited in #952 clear up. And it makes the ORCCA HTML clear up too.

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -2937,7 +2937,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <!-- switch on inline vs. divisional (could do this in match?) -->
             <xsl:choose>
                 <!-- divisional -->
-                <xsl:when test="parent::exercises">
+                <xsl:when test="parent::exercises or parent::exercisegroup">
                     <xsl:apply-templates select="."  mode="exercise-components">
                         <xsl:with-param name="b-original" select="$b-original" />
                         <xsl:with-param name="b-has-statement" select="true()" />


Alex-Jordan:
Take a look at line 2998, which has another `test="parent::exercises"`. I don't grok that section of code, but wonder if it too should be `test="parent::exercises or parent::exercisegroup"`.

In `mathbook-latex`, I see several tests on `"parent::exercises"` without an accompanying `"parent::exercisegroup"`. You may want to look there as well.

I considered `ancestor::exercises`. Thought you might prefer the specificity of `parent::exercises or parent::exercisegroup` though.

Alex-Jordan:
Oh, and feel free to tackle this all in one commit and don't worry about my name on anything.

rbeezer:
`ancestor::exercises` (and `ancestor::worksheet`) is best, since we are just trying to determine if we have an inline exercise or a divisional exercise.  And an `exercises` is likely to soon have a `subexercises` to pass through.

Good catch!  There are about six fixes, and maybe about 15 places where it is done right.

A/B testing indicates just one (desirable) change.  Can you test the effect of the switches?  I've not done that.  Feel free to close if all is well.

PS: Your name ended up on **all** of this.  ;-)

Alex-Jordan:
Thanks, I can confirm that ORCCA isn't showing solutions after pulling this.

Alex-Jordan:
Well, after this, it is true that non-WW exercises in exercisegroups are not showing their solutions.

However, now WW exercises *are* showing their solutions.

rbeezer:
I think that *needs work*.  Can you make a short issue?

PR #964, opened 2018-09-21
Use permid for HTML id, when it exists

davidfarmer:
The template for internal-id is now supplemented by a file-id template.  Use internal-id
for HTML id~s and LaTeX \label~s, and file-id for filenames.

The internal-id uses the permid attribute if it exists.  Otherwise it uses the file-id
(which is identical to what formerly was called internal-id).

Tested on LaTeX and HTML versions of sample-article.
This worked in all cases I checked, but I could not figure out how to check every possibility.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -391,7 +391,7 @@ the xsltproc executable.
         the first non-trivial character.
         -->
 
-        <section xml:id="section-sage-cells">
+        <section permid="Abc" xml:id="section-sage-cells">
             <title>Computing Integrals with Sage (<m>\int</m>)</title>
             <!-- Index testing, semi-realistic -->
             <idx><h>Sage</h><h>integration</h></idx>
@@ -573,7 +573,7 @@ the xsltproc executable.
         <section xml:id="interesting-corollary">
             <title>An Interesting Corollary</title>
 
-            <objectives xml:id="objectives-structures">
+            <objectives permid="dEf" xml:id="objectives-structures">
                 <title>Fundamental Structures</title>
 
                 <introduction>
@@ -602,13 +602,13 @@ the xsltproc executable.
 
                 <todo>Should equation references come with parentheses?</todo>
 
-                <corollary xml:id="corollary-FTC-derivative">
+                <corollary permid="ghI" xml:id="corollary-FTC-derivative">
                     <creator>Leibniz, Newton</creator>
                     <idx><h>Fundamental Theorem of Calculus</h><h>Corollary</h></idx>
                     <statement>
                         <p>Suppose <m>f(x)</m> is a continuous function.  Then <men xml:id="equation-alternate-FTC">\frac{d}{dx}\definiteintegral{a}{x}{f(t)}{t}=f(x)</men>.</p>
                     </statement>
-                    <proof xml:id="proof-FTC-corollary">
+                    <proof permid="Jkl" xml:id="proof-FTC-corollary">
                         <p>We simply take the indicated derivative, applying Theorem<nbsp /><xref ref="theorem-FTC" /> at <xref ref="equation-use-FTC" /><mdn>
                             <mrow xml:id="equation-use-FTC">\frac{d}{dx}\definiteintegral{a}{x}{f(t)}{t}&amp;=\frac{d}{dx}\left(F(x)-F(a)\right)</mrow>
                             <mrow number="no">&amp;=\frac{d}{dx}F(x)-\frac{d}{dx}F(a)</mrow>
@@ -632,7 +632,7 @@ the xsltproc executable.
                     <input>2+2</input>
                 </sage>
 
-                <example xml:id="example-mysterious">
+                <example permid="mNo" xml:id="example-mysterious">
                     <title>A Mysterious Derivative!</title>
                     <p>So if we define a function with its variable employed as a limit of integration, like so <me>K(z)=\definiteintegral{345}{z}{x^4\sin(x^2)}{x}</me> then we get the derivative of that function so easily it seems like a mystery, <me>\frac{d}{dz}K(z)=z^4\sin(z^2)</me>. That's it.</p>
 
@@ -667,7 +667,7 @@ the xsltproc executable.
                             <p>A case may also have a <c>title</c>, whose formatting and structure is entirely up to the author.  This then becomes the text of a cross-reference, as well.</p>
                         </case>
 
-                        <case direction="forward" xml:id="forward">
+                        <case permid="pqR" direction="forward" xml:id="forward">
                             <title>Necessity</title>
 
                             <p>If you like, you can have both indications.</p>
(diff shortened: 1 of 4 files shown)


rbeezer:
When the `mbx` script is used to create new files (such as those given by `latex-image`) the internal-id is used to coordinate filenames.  So there are changes that need to be made in the "extraction" stylesheets, so that a "file-id" is used there.  I've got some sense of the magnitude of the necessary changes via `grep`.

A totally different approach is to 
* don't touch `internal-id,` so no existing behavior breaks, especially non-obvious coordination across stylesheets.
* make a new "permid" template which prefers `@permid`
* change those places in the HTML conversion (only) where an HTML `id` is generated (easy to locate unambiguously)
* selectively introduce "permid" other useful places (eg LaTeX conversion)

What concerns me about missing a change from `internal-id` to `file-id` is we can't find it until the buggy situation gets a `@permid`.  The bugs (if any!) with the approach above will surface in new features during testing.

You've put a lot of work into this PR and I don't want it to go for naught.  But I'm convincing myself (with a lot of careful thought) that I'd rather not risk the existing code without better testing.  Let me do some testing with your sample article changes and a bit of coding in the HTML conversion - I think the approach above will be very easy to implement.

davidfarmer:
No problem if the effect of using permid appropriately does not
use any of my code.  But I am not clear how the main work I did
can be avoided in your scenario.

The main work I did: go through the files, locating every instance
of internal-id, and decide if it needs to change to file-id.

I see now that I missed the extract- files and a couple others.
I certainly can do those and add to this pull request.

I won't do that unless you suggest that I proceed.

But if we don't continue on the path I started, I don't see how
another approach can avoid also examining those exact same cases.

rbeezer:
If i search the -html sylesheet for `id=` and `name="id"`(latter as part of an `attribute` command) then I think it is very hard to miss the construction of an HTML `id`.  Those are (all the) places to change `internal-id` to `perm-id`.  So maybe it is the contrapositive of what you did?  I should have time this evening (off and on).  I'll report back.

davidfarmer:
Not the contrapositive.  You definitely have identified a smaller
subset of things that need to change.

rbeezer:
Thanks for splitting the sample article changes into a separate commit.  That made part of this very easy.

I built the HTML with your PR.  Then I followed the approach above, and compared the HTML output.  One or two things I would have missed without your work.  A few places to be careful, where a different attribute pairs up with an `id`.  For example, the `@data-scroll` in the ToC code has an `id` as a value.  It was not too hard to duplicate all the uses of the `@permid` that you accomplished.  

Excepting one question.  Material born as a knowl, here `example`, have an anchor with `@refid` to connect with the hidden content.  Your output, then mine:

```html
<a knowl="" class="id-ref" refid="hk-Stu" id="exercise-essay">
```
```html
<a knowl="" class="id-ref" refid="hk-Stu" id="Stu">
```

Which should it be?  Notes in code suggest the `id` is used as the target of an in-context link as part of an `xref`.  But if a reader opens the `example` given as a born-hidden knowl, and annotates the content within the knowl, where should the `id` be/go to preserve/associate those annotations?  Certainly, any "in-context" link pointing here is using the `@permid` as the fragment.

Work tonight is a bit cursory.  I'll review and clean-up sometime tomorrow.

Rob

davidfarmer:
Maybe both outputs are wrong?

The "Stu" HTML id comes from the permid on an "exercise" in the source.

It is unexpected (to me) that the permid goes to the "a" tag instead of
the 'article class="exercise-like"'.  That article does not have an id at
all, but probably it should?.  The "a" tag needs an id because it is the
target of an in-context link.  If we expect outside links to persist,
then that id should be persistent (at the moment I do not know of an
annotation-like purpose for that id.)

I think this reveals that we need a methodical way to derive additional
HTML id~s from a permid.  For example, an object with permid is
knowlified.  Both the "a" tag and the HTML version of the object need
an id, but we can only use the permid for one of them.  For the other one,
it is better to use a modification of the permid than to use the
current global-counter id.

Before making any more xsl changes, I would add a bunch more permid~s
near and inside the object with the "Stu" permid.  I'd also change the
"Stu" permid because it is a common string and hard to search for in the
output.

ps. Don't expect annotation of content in a knowl to work properly
at first!

davidfarmer:
Another place where you need to derive an HTML id:  when a PTX "p" gets split into
two HTML "p"s because it contains display math.

rbeezer:
Right.  Both wrong.  I think the `id` belongs on the `article` (and I think that should be the target of an in-context link).  But I don't even know if I've ever really tried an in-context link to a born-hidden knowl.  I'll experiment (status quo, and with `id` on `article`).  I'm less convinced that the `a` for the knowl itself needs an `id`.

rbeezer:
Paragraphs with display math and lists.  I saw you put some `@permid` onto intra-paragraph math (and list items).  My mental conception is that as author's source, in PTX land, a `p` is a sort of atomic, indivisable, item.  Of course, we are dividing them in HTML!

Every new HTML `p`, except the first, follows exactly one `md`, `ol`, etc.  So **if** these intra-paragraph items has `@permid`, a following `p` could have a standard `@permid` with just one standard addition.

However, it seems problemsome/burdensome for an author to decide which `@permid` to change when they make one or more changes to a PTX `p` split up by display items?

Yes, annotations on born-hidden knowls will be interesting!  Already went there.  ;-)

rbeezer:
Results of experiments.

* Status quo: in-context link to born-hidden knowl works well (includes pink flash).
* Similarly with my current changes.

A footnote (`fn`) is a born-hidden knowl, so just a bare `a` inline with other text.  That explains why the `id` is on the `a`.  So code needs to differentiate an "inline" born-hidden knowl from a "block" one.  That information is already being generated and almost makes it way to the construction of the `a`.  So not much to change.

Proposal: push changes to utiliize `@permid` for HTML `id` and incrementally adjust other code on other (subsequent) commits as we expose situations like this and the split `p`.  That'll let you get started experimenting more with real source.

davidfarmer:
> However, it seems problemsome/burdensome for an author to decide which @permid to change when they make one
> or more changes to a PTX p split up by display items?

In the case of a PTX p containing text - math - text,
as long as the HTML id~s of both HTML "p"s are derived from
the permid, everything should be fine.

The fact that the author changing the permid will change the
id of two HTML p~s, is a weird edge case that should not cause
concern.

rbeezer:
New template added to preferentially use `@permid`, and in use throughout HTML output where an HTML `id` is created or referenced.

Preserved commit with sample article additions.  If a source document has no `@permid` at all, then the output should not changed (testing indicates this is so).  So we can slowly introduce necessary changes as `@permid` comes into play.

Such as putting the `id` for a born-hidden-knowl item in the best location.  Done at 7329f702e18d16ec2f89494b236e455453281a02.  Marginally improved the pink flash.

No changes to sample article, no work on split paragraphs.  Start new threads/issues for those - closing this now.

Good to have this in place, you can announce plans to the group, and I'll be ready to implement more changes in the HTML as needed.

Rob

PR #977, opened 2018-11-26
Logging in

davidfarmer:
Enable logging in to the HTML version, and allow a logged-in user to answer reading
questions directly in the text.  See Section 4.2.5 of the sample article.

To log in, hover your mouse in the upper-right corner of the page (may be outside the
top banner, if the browser is wide).  By default, logging-in is enabled, and anyone can
log in for 3 hours as 'guest' with password 'guest'.

Added AsciiMath support in Mathjax, so that reading question answers can use AsciiMath.
This required changing the main MathJax file.  Not sure how to check if there were any
unintended consequences.

Logging in, and answering questions, probably has some accessibility issues.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -980,19 +980,36 @@ the xsltproc executable.
                 <reading-questions>
                     <introduction>
                         <p>A set of reading questions may have an <tag>introduction</tag>, perhaps for preparatory explanation.</p>
+                        <p>If a student has logged in to the HTML version,
+                           then they can answer the reading questions directly in the book.
+                           Inline math LaTeX can be entered using
+                           <c>$</c>...<c>$</c> or <c>\(</c>...<c>\)</c> delimiters,
+                           and inline AsciiMath using backticks <c>`</c>...<c>`</c> as delimiters.
+                        </p>
                     </introduction>
 
                     <exercise xml:id="first-reading">
-                        <statement>
-                            <p>This is a <term>reading question</term> that you might have a student read prior to a class session, based on reading part of the book.</p>
-                        </statement>
-                        <answer>
-                            <p>Which might help make the classroom session improve.</p>
-                        </answer>
+                      <p>
+                        This is a <term>reading question</term> that you might have a student answer prior to a class session,
+                        based on reading part of the book.
+                        A quick glance before class can help you tailor class time to the specific needs of your students.
+                        The perfect reading question will reveal whether the student has read and understood the material,
+                        and will be difficult to answer if they have not.
+                        What do you think of that?
+                      </p>
                     </exercise>
 
                     <exercise>
-                        <p>And a second one, with a cross-reference to the first, as a check on numbering: <xref ref="first-reading" text="type-global"/>.</p>
+                      <p>
+                        And a second one, with a cross-reference to the first,
+                        as a check on numbering:
+                        <xref ref="first-reading" text="type-global"/>.
+                        Reading questions are allowed to have answers,
+                        but providing answers misses the point of a reading question,
+                        and the answer knowl interacts poorly with the mechanism used to
+                        allow students to answer directly in the book.
+                        Do you think the schema should ban answers to reading questions?
+                      </p>
                     </exercise>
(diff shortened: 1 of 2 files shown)


examples/sample-article/sample-article.xml
```diff
@@ -980,19 +980,36 @@ the xsltproc executable.
...
-                        <p>And a second one, with a cross-reference to the first, as a check on numbering: <xref ref="first-reading" text="type-global"/>.</p>
+                      <p>
+                        And a second one, with a cross-reference to the first,
+                        as a check on numbering:
+                        <xref ref="first-reading" text="type-global"/>.
+                        Reading questions are allowed to have answers,
+                        but providing answers misses the point of a reading question,
+                        and the answer knowl interacts poorly with the mechanism used to
+                        allow students to answer directly in the book.
+                        Do you think the schema should ban answers to reading questions?
```
kcrisman: No, I do not think it should be banned; some reading questions I assign often need a sample answer.  Example which folks may use:

> Give an example of a real-life situation where you experienced a nonzero second derivative, and explain.

Actual type of answer students give if I provide a sample:
> I noticed that as I was doing my weight training, the increase in the number of pounds I could lift got smaller each week, though I kept on improving.

Potential answer without a sample:
> I noticed that something looked like a parabola, and I know that $x^2$ has a nonzero second derivative.

However, suggesting that most reading questions not have a provided answer seems quite sound to me.

rbeezer:
Visual review looks good.  I'll merge as soon as possible, but may be a bit.  My Mondays are super-busy.

examples/sample-article/sample-article.xml
rbeezer: We'll want answers in the source, so they can migrate to an Instructor's Manual. So the schema shouldn't ban them.  

But it should be an easy matter to ban them from appearing in the HTML conversion (or any other conversion) at the place where the question is born (and any `solutions` division as well.)  Once I get switches in place we can discuss which options to always turn off.

rbeezer:
Put two backticks into a "regular" paragraph, far apart, and watch the craziness that ensues when it is interpreted as ASCIImath.  ;-)  I think we need to "turn off" processing on the `body` element, and then turn it back on for the reader-entry field of the questions?  See two configuration options at the bottom of:

http://docs.mathjax.org/en/latest/options/preprocessors/asciimath2jax.html#configure-asciimath2jax

My code comments suggest we need to change the overarching MJ configuration file when we switch to 3.0 (now in beta?).  Did you try just adding the right extension for AM support?  That might be the less radical change, and we can revisit a different configuration file when we need to change it anyway.  Hints on differences at:

http://docs.mathjax.org/en/latest/config-files.html

rbeezer:
Maybe the issue below is more relevant and/or appropriate now, if we want to avoid stray characters setting off MathJax pre-processing?  Note that my two-backtick experiment runs similarly if you place the pair `\(, \)` into any non-verbatim text.

https://github.com/rbeezer/mathbook/issues/611

davidfarmer:
I am willing to say that anyone who puts a pair of backslash parens
in the text, deserves what they get.

What is the use case of normal text for having backticks?
If those occur in some other environment (tikz, for example)
then we can tell AsciiMath to ignore that.

There are different options for how we can avoid having
unwanted side-effects to allowing AsciiMath, but I'd like to
know that we are solving an actual problem.

davidfarmer:
I will revert to the original MathJax main file and just add what
(I think is) the minimal necessary extensions.

Also will turn off AsciiMath processing for everything except
the user-entered answers.

rbeezer:
Looking good.  And working well.  Thanks for the changes.

It is telling we've never had anybody report the backslash-parens.  Backticks just seem more dangerous since they are a (unlikely) single character.

Notice that we are plowing new ground now that readers (not authors) are adding (math) content.

Sample article is live on the website.  Would you like a beta of AATA with over 100 reading questions?  If not now, feel free to ask later - I can build and host easily.

davidfarmer:
I'll announce on pretext-dev.  A beta AATA can wait until I
make some more revisions on tracking.

PR #980, opened 2018-12-09
Validate WeBWorK XML output

Alex-Jordan:
This does some simple testing on the returned content from each webwork problem. It's highly likely my Python is crude and you will see improvements.

Take the minimal webwork example and replace the webwork with `<webwork source="Library/this/problem/does/not/exist.pg"/>`. Then run `make mini-extraction` and you should get an appropriate message terminating the extraction .

Take the minimal webwork example and delete a semicolon in the pg-code. Then run `make mini-extraction` and you should get an appropriate message terminating the extraction.

Take the minimal webwork example and replace the webwork with `<webwork source="Library/Rochester/setDerivatives1/ns2_8_31.pg"/>` which is a problem with elements that have not been given PTX output definitions. Then run `make mini-extraction` and you should get an appropriate message terminating the extraction.

### DIFF
--- a/script/mbx
+++ b/script/mbx
@@ -379,6 +379,10 @@ def webwork_to_xml(xslt_exec, ptx_xsl_dir, xml_source, server_params, dest_dir):
     # Choose one of the dictionaries to take its keys as what to loop through
     for problem in sorted(origin):
 
+        # It is more convenient to identify server problems by file path,
+        # and PTX problems by internal ID
+        problem_identifier = problem if (origin[problem]=='ptx') else source[problem]
+
         #remove outer webwork tag (and attributes) from authored source
         if origin[problem] == 'ptx':
             source[problem] = re.sub(r"<webwork.*?>",'',source[problem]).replace('</webwork>','')
@@ -411,32 +415,17 @@ def webwork_to_xml(xslt_exec, ptx_xsl_dir, xml_source, server_params, dest_dir):
                 else:
                     pgbase64[hint_sol][problem] = pgptx[hint_sol][problem].encode('base64').replace('\n', '')
 
-        # First write authored and pg
+        # First write authored
         if origin[problem] == 'ptx':
             try:
                 include_file = open(include_file_name, 'a')
                 authored_tag = '    <authored>\n{}\n    </authored>\n\n'
                 include_file.write(authored_tag.format(re.sub(re.compile('^', re.MULTILINE),'      ',source[problem])))
-                pg_tag = '    <pg>\n{}\n    </pg>\n\n'
-                formatted_pg = re.sub(re.compile('^', re.MULTILINE),'      ',pg[problem])
-                # opportunity to cut out extra blank lines
-                formatted_pg = re.sub(re.compile(r"(\n *\n)( *\n)*", re.MULTILINE),r"\1",formatted_pg)
-                include_file.write(pg_tag.format(formatted_pg))
                 include_file.close()
             except Exception as e:
                 root_cause = str(e)
-                msg = "There was a problem writing a problem to the file: {}\n"
-                raise ValueError(msg.format(include_file_name) + root_cause)
-        elif origin[problem] == 'server':
-            try:
-                include_file = open(include_file_name, 'a')
-                pg_tag = '    <pg source="{}" />\n\n'
-                include_file.write(pg_tag.format(source[problem]))
-                include_file.close()
-            except Exception as e:
-                root_cause = str(e)
-                msg = "There was a problem writing a problem to the file: {}\n"
-                raise ValueError(msg.format(include_file_name) + root_cause)
+                msg = "There was a problem writing the authored source of {} to the file: {}\n"
+                raise ValueError(msg.format(problem_identifier, include_file_name) + root_cause)
 
         # Now begin getting static version from server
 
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
What do you think about prompting with an option to bail or proceed? With the thousands of problems in ORCCA, I would prefer to bail, fix it, and start over. Rather than let the 30 minutes of subsequent ww extraction continue only to learn I have to repeat that.

rbeezer:
The future is that the scripts will run un-attended on a server, so I'd like there to be no need for human interaction.  Reports, yes; monitoring, no.

How about another (elective) step up in verbose-ness that will stop at severe errors?  Eventually the script would benefit from a better hierarchy of warnings, but I don't think either of us wants to deal with that technical debt right now.

Alex-Jordan:
I'm wrestling between what I know will work well for me with ORCCA (stopping cold on these WW issues) and your general desire to not bail and proceed. What about a switch? It could be general purpose. Something that says if you hit some of these issues that (will eventually) need attention anyway, stop? Maybe a switch like that could be useful outside of webwork processing too?

rbeezer:
> What about a switch? It could be general purpose.

Yes., that'd work, too.  But default is off/procede.  ;-)

Alex-Jordan:
Any preference on name/abbreviation?

Like `-t` for terminate upon error. Or something like that?

rbeezer:
Yep.  Or -p for "pause"?  I can't recall what's taken already.

Choose the most unlikely letter to be used for other purposes, but which still 
makes sense here.  ;-)

Alex-Jordan:
There are warnings and errors. To me, the things we are looking for here are errors, not warnings. Do you agree?

If you do, we are now about to classify these things as "recoverable errors". Meaning that the `-a` flag (for abort upon recoverable error) will apply to these. That could be different than an "error" that demands terminating the script.

Should the alert messages be like `PTX:RECOVERABLE ERROR`? Or something that distinguishes from `PTX:ERROR`?

rbeezer:
On 12/10/18 2:44 PM, Alex Jordan wrote:
> There are warnings and errors. To me, the things we are looking for here are 
> errors, not warnings. Do you agree?

Yes.

> If you do, we are now about to classify these things as "recoverable errors". 
> Meaning that the |-a| flag (for abort upon recoverable error) will apply to 
> these. That could be different than an "error" that demands terminating the script.

Sounds good.

> Should the alert messages be like |PTX:RECOVERABLE ERROR|? Or something that 
> distinguishes from |PTX:ERROR|?

Let's keep it as a "PTX:ERROR" for now.  we can split hairs once somebody cares 
enough to put a proper warning system in place (there is a Python package called 
"warnings", iirc).

Alex-Jordan:
OK, I just force pushed a new version. For testing:

1. Should be no difference when making the sample chapter.
2. In sample chapter, change some `webwork` to have an `@source` that is nonsense (not going to be an actual pg file on the server). Then `make sample-chapter-extraction`. You should see an error message scroll by, and then in the `webwork-extraction.xml` search for "failure".
3. In sample chapter, change some `webwork` with `pg-code` to to break the Perl (eg, remove a needed semicolon). Then `make sample-chapter-extraction`. You should see an error message scroll by, and then in the `webwork-extraction.xml` search for "failure".
4. In sample chapter, change some `webwork` to have an `@source="Library/Rochester/setDerivatives1/ns2_8_31.pg"` (which returns content that is not valid XML). Then `make sample-chapter-extraction`. You should see an error message scroll by, and then in the `webwork-extraction.xml` search for "failure".
5. In `Makefile`, add the `-a` option to `sample-chapter-extraction` definition, and repeat items 2--4. The script should abort upon encountering any of these errors.

Alex-Jordan:
When this boils down, possibly you will want to commit the `-a` option separately.

rbeezer:
*  I had to restore a closing quote on an attribute twice.  One fixed line looks like
```python
static[problem] = static[problem].replace('<static', '<static seed="' + seed[problem] + '" ')
```
Otherwise the creation of the merge file failed.  I've got the fixes and can mangage them, even under a force push.

* I see `static[problem]` being assigned a shell problem but it does not become the `server-url` version, so the live HTML just fails with the usual WW debugging.  Maybe that is what you want?  Can we insert the "shell" warning into the first part of the `statement` to indicate the problem is known by PTX?

* I'm not wild about the whole source getting spit out on the console.  Could we reserve that for the `-a` behavior?  (Have not tested that mode yet.)

Alex-Jordan:
> I had to restore a closing quote on an attribute twice.

Oops, wasn't testing on merge. That area was fiddled with because I could no longer assume the absence of an attribute on the `static` at that point.

>I've got the fixes and can mangage them, even under a force push.

OK, but if I do more here (as it looks like I will) I will fix that and make another force push. You don't mind trashing your branch and reconstituting? Would you prefer I just tack on commits here? (Sorry if you've answered this in the past.)

>I see static[problem] being assigned a shell problem but it does not become the server-url version, so the live HTML just fails with the usual WW debugging. Maybe that is what you want? Can we insert the "shell" warning into the first part of the statement to indicate the problem is known by PTX?

I can do better things here. STay tuned for next iteration.

> I'm not wild about the whole source getting spit out on the console.

I struggled with what to report when it is a PTX-authored problem. With a server problem, reporting the file name is obvious. With a PTX-authored problem, something like `webwork-14` is not super helpful to locate it in your source. What do you think helps? Maybe just the converted `pg`?

rbeezer:
Force push on your end is fine on my end.

Let's report the `@xml:id` of the containing `exercise` (which will encourage use of `@xml:id`) and then say something short like
```text
Use -a to halt with full version of the problem.
```
and than make good on that advice?  The goal is to reduce support questions by giving enough information for folks to find and correct their own errors, not to be super-convenient.  ;-)

rbeezer:
Reading the warnings is most helpful.  If folks won't survey the putput, then 
they get what they get.  ;-)  Which is why we should be very economical with 
what we say.  Warnings should stand out.

That said, don't put in a broken problem.  Put in a shell version where it is 
obvious to any reader that there is a coding-problem to be reported.

I've got nothing bad to say about any of your suggestions or approaches, but I 
have ideas (a philosophy?) about how to communicate problems, and am trying to 
keep that consistent so that authors know what to expect.

Alex-Jordan:
We don't have the ID of the containing exercise at present. We could get that, but would need to add it to one of the extraction style sheets. What we have is the internal ID of the webwork itself. Is that sufficient?

I hesitate to change things to get the exercise's ID, because
1) I am so far confined to `mbx` script with this, and it's nice to stay so contained
2) Are we _sure_ webwork will only ever exist as a child of an exercise? What if it's a child of interactive, project, or something new some day? The less the code assumes an exercise is the parent, the fewer revisions might be needed some day.

rbeezer:
On 12/10/18 9:47 PM, Alex Jordan wrote:
> We don't have the ID of the containing exercise at present.

Right.  Use the ID of the "webwork".  Maybe suggest in the documentation that 
meaningful  xml:id  on the "webwork" will help with debugging trips to the server?

Alex-Jordan:
OK, force pushed again. Same testing routine as before.

rbeezer:
Looks great!  This will be a big help for folks producing WW problems.  I only tested the "no compile" scenario, but the code all looks good.  Going up now.

(I couldn't split out the abort part without requiring a tedious fiddle to put your name back on the commits.)

PR #982, opened 2018-12-13
Put answers in WW extraction

Alex-Jordan:
In the sample chapter's Makefile, change the server to -dev. Then `make sample-chapter-extraction` should build a `webwork-extraction.xml` file with answer elements for each problem. `make sample-chapter-merge`, `make sample-chapter-pdf`, and `make sample-chapter-html` should all make good things.

Leaving the server as -ptx should still behave too, just with no answers. 

We have the perennial chick and egg problem with WW development. Part of making this work was adding code to WW on -dev. I'll need to open a PR there to WW's develop branch, get it approved, then wait until it merges into WW's master. Or I could bring these changes into -ptx right now while the WW repo chugs along with its development more slowly. Not sure what to do, except that I will wait to open the PR to WW until you've had a chance to comment. (In case I need to do a little more there.)

### DIFF
--- a/doc/author-guide/webwork.xml
+++ b/doc/author-guide/webwork.xml
@@ -12,55 +12,16 @@
     <author>Alex Jordan</author>
 
     <introduction>
-        <p>With a <webwork /> server (version 2.14 or higher, or <url href="https://webwork-ptx.aimath.org"><c>webwork-ptx.aimath.org</c></url>) and a little setup work, you can embed <webwork /> exercises<idx><h><webwork /></h><h><webwork /> exercise</h></idx> in your <pretext /> project. HTML output will have interactive problem cells. PDF output will contain static versions of exercises. And all such exercises can be archived by the <c>mbx</c> script into a file tree to be uploaded onto the <webwork /> server for use in the <q>traditional</q> way.</p>
+        <p>With a <webwork/> server (version 2.14 or higher, or <url href="https://webwork-ptx.aimath.org"><c>webwork-ptx.aimath.org</c></url>) and a little setup work, you can embed <webwork/> exercises<idx><h><webwork/></h><h><webwork/> exercise</h></idx> in your <pretext/> project. HTML output can have interactive problem cells or print problems in <q>static</q> form. PDF output will print static versions. And all such exercises can be archived into a file tree to be uploaded onto a <webwork/> server for use in the <q>traditional</q> way.</p>
+
+        <p>See the <pretext/> Publishing Guide for how to configure a <webwork/> course and server to process your <webwork/> problems, and then how to process them.</p>
     </introduction>
 
-    <section xml:id="webwork-configuration">
-        <title>Configuring a <webwork /> Course for <pretext /></title>
-        <p>We assume a mild familiarity with administrating a <webwork /> server. The version of <webwork /> needs to be 2.14 or later for use with <pretext />. Using the <c>admin</c> course, create a course named <c>anonymous</c>. In the course's Course Configuration menu, set all permissions to <c>admin</c> (or perhaps set some to the even more restrictive <c>nobody</c>). Except set <q>Allowed to login to the course</q> to <c>login_proctor</c>.</p>
-        <p>In the Classlist Editor, add a user named <c>anonymous</c>, and set that user's permission level to <c>login_proctor</c>, the permission level one higher than <c>student</c>. Set that user's password to <c>anonymous</c>. Note that because this is public information, anyone will be able to log into this course as user <c>anonymous</c>. This is why setting the permissions earlier is very important. (Especially preventing this user from changing its own password.)</p>
-        <p>Add the following lines to the <c>course.conf</c> file (which lives in the parent folder of the <c>templates/</c> folder.)</p>
-        <pre>
-        # Hide message about previewing hints and solutions for instructors
-        $pg{specialPGEnvironmentVars}{ALWAYS_SHOW_HINT_PERMISSION_LEVEL} = 100;
-        $pg{specialPGEnvironmentVars}{ALWAYS_SHOW_SOLUTION_PERMISSION_LEVEL} = 100;
-        </pre>
-        <p>In the <c>templates/macros/</c> folder, edit <c>PGcourse.pl</c> (or create it if need be) and add the lines:</p>
-        <pre>
-        <![CDATA[
-        #### Replace essay boxes with a message
-        sub essay_box {
-            my $out = MODES(
-                TeX => '',
-                Latex2HTML => '',
-                HTML => qq!<P>If you were logged into a WeBWorK course
-                and this problem were assigned to you,
-                you would be able to submit an essay answer
-                that would be graded later by a human being.</P>!,
-                PTX => '',
-            );
-            $out;
-        };
-
-        #### Suppress essay help link
-        sub essay_help {};
-
-        #### How many attempts until hint is available
-        $showHint = -1;
-        # May be a bug that WeBWorK requires -1 instead of 0
-        # for immediate access to hints
-
-        1;
-        ]]>
-        </pre>
-        <p>Now <pretext /> will be able to communicate with this course to retrieve what is needed.</p>
-    </section>
-
     <section xml:id="webwork-source">
-        <title><webwork /> Problems in Source</title>
+        <title><webwork/> Problems</title>
 
         <introduction>
-            <p>A <tag>webwork</tag><idx><webwork /></idx> tag must be inside an <tag>exercise</tag>, optionally preceded by an <tag>introduction</tag>, and optionally followed by a <tag>conclusion</tag>.</p>
+            <p>A <tag>webwork</tag><idx><webwork /></idx> tag must be inside an <tag>exercise</tag>, <em>optionally</em> preceded by an <tag>introduction</tag>, and <em>optionally</em> followed by a <tag>conclusion</tag>.</p>
             <pre>
             <![CDATA[
             <exercise>
@@ -75,23 +36,20 @@
             </exercise>
             ]]>
             </pre>
-            <p>There are several methods for putting content into the <tag>webwork</tag>. (Note that an empty <tag>webwork</tag> with no attributes will simply produce the camelcase <webwork /> logo.)</p>
+            <p>There are several methods for putting content into the <tag>webwork</tag>. (Note that an empty <tag>webwork</tag> with no attributes will simply produce the camelcase <webwork/> logo.)</p>
         </introduction>
 
         <subsection>
-            <title>Using an Existing <webwork /> Problem</title>
-            <p>If a problem already exists and is accessible from the <c>anonymous</c> course's <c>templates/</c> folder, then you can simply include it as a <attr>source</attr> attribute. For example, if it is a problem in the Open Problem Library (OPL) then relative to the <c>templates/</c> folder, its path is <c>Library/...</c> and you may use:</p>
-            <pre>
-            <![CDATA[
-            <webwork source="Library/PCC/BasicAlgebra/Exponents/exponentsMultiplication0.pg" />
-            ]]>
-            </pre>
-            <p>Or if you have a problem's PG file, you can upload it into the <c>anonymous</c> course's <c>templates/local/</c> folder and use it with:</p>
-            <pre>
-            <![CDATA[
-            <webwork source="local/my_problem.pg" />
-            ]]>
-            </pre>
+            <title>Using an Existing <webwork/> Problem</title>
+            <p>
+                If a problem already exists and is accessible from the hosting course's <c>templates/</c> folder, then you can simply include it as a <attr>source</attr> attribute. For example if it is a problem in the Open Problem Library (OPL), then relative to the <c>templates/</c> folder, its path is <c>Library/...</c> and you may use:
+
+                <cd>&lt;webwork source="Library/PCC/BasicAlgebra/Exponents/exponentsMultiplication0.pg" /></cd>
+
+                Or if you have a problem's PG file, you can upload it into the hosting course's <c>templates/local/</c> folder and use it with:
+
+                <cd>&lt;webwork source="local/my_problem.pg" /></cd>
+            </p>
         </subsection>
 
         <subsection>
@@ -142,13 +100,13 @@
             ]]>
             </pre>
 
-            <p>Many special contexts are automatically detected by <pretext />, and it loads the appropriate macro file into the PG problem. However you may need to explicitly load a macro file as described in <xref ref="webwork-pg-code">Subsection</xref>.</p>
+            <p>Many special contexts are automatically detected by <pretext/>, and it loads the appropriate macro file into the PG problem. However you may need to explicitly load a macro file as described in <xref ref="webwork-pg-code">Subsection</xref>.</p>
 
         </subsection>
 
         <subsection xml:id="webwork-pg-code">
             <title>PG code in Problems</title>
-            <p>To have randomization in problems or otherwise take advantage of the algorithmic programming capabilities of Perl and <webwork />'s PG language requires using a <tag>setup</tag> tag. Having at least a little familiarity with coding problems in <webwork /> is necessary, although for simpler problems you could get away with mimicking the sample article in <c>mathbook/examples/webwork/</c>. A <tag>statement</tag>, (optional) <tag>hint</tag>, and (optional) <tag>solution</tag> follow.</p>
+            <p>To have randomization in problems or otherwise take advantage of the algorithmic programming capabilities of Perl and <webwork/>'s PG language requires using a <tag>setup</tag> tag. Having at least a little familiarity with coding problems in <webwork/> is necessary, although for simpler problems you could get away with mimicking the sample article in <c>mathbook/examples/webwork/</c>. A <tag>statement</tag>, <em>optional</em> <tag>hint</tag>, and <em>optional</em> <tag>solution</tag> follow.</p>
             <pre>
             <![CDATA[
             <webwork>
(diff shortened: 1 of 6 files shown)


rbeezer:
*  Do you mean that `\x85` is a **verbatim** delimiter?  Not sure I understand what a verbose delimiter is, or even exactly what is going on here.  Where and how does the substitution of a tilde change the output?  All I can infer so far is that `\x85` is unwanted.
* Can you add one sentence about how one string is text and the other is math and you have to eventually wrap one in the `m` tag?  It was not too hard to figure out, but I did wonder what was going on as I read it.

Visual review only.  I can try to test later today, but that may be my last easy opportunity.

Alex-Jordan:
Yes, "verbatim", sorry. That's what happens when I tidy up too late at night. 

Davide C dealt with "dangerous" characters in string answers printed in math mode in the following way. Put them in `\verbhere`. You can see this at line 598 and then at line 602 here:
https://github.com/openwebwork/pg/blob/8a089edceb5d3b36500bac47ef3c2daeec10e0e4/macros/PGstringevaluators.pl

The Python further down complains about encoding if we get this character. I'm hoping to avoid rebuilding all that Python to make it OK with characters in that range, all just for this one thing.

Instead of a blanket change of \x85 to tilde, I will do a more targeted regex replace on `\verbstuff`.  And I'll try to design it to use tilde only if "stuff" has no tilde. Maybe I'll just grab the lowest printable character not appearing in "stuff".


Is your second comment about the problem I added that shows the dangerous XML characters?

rbeezer:
Ok, I get it now (I think).  Let's definitely target `\verb` so we don't impact anything else.

What about iterating through a list of "unlikely" characters, until one is not present?  Like `~ | ?`, then try more "normal" characters at the end of the list (a-z, A-Z).  Not fool-proof, but close enough?

I think Python is now much pickier about strings and Unicode.  A fix could be just being careful about specifying an encoding, though `\x85` sounds like a hack, so maybe best to get rid of it.

I was talking about teh difference between having `correct_ans` and `correct_ans_latex_string`.  One gets an `m` tag, the other doesn't.  Tell the reader of the Python **beforehand** that is the purpose of code leading up to new line 545.

Alex-Jordan:
And I just tossed out semicolon as a delimiter too. I wondered about things like `\verb;M&amp;M;`. No testing if that would arise or not work out, but it couldn't hurt to be safe.

rbeezer:
Verbatim text will drive you crazy, no?  I wonder if we will ever see
```text
!A verbatim string contained all ASCII characters from 33 through 126!
```

;-)

Alex-Jordan:
I'm off to a meeting, won't be working on this for the rest of the day. You can see the exercise at line 635 of the sample chapter where the XML char test is. It should be straightforward to add more answers that test more characters, if you get to a point where you want to finish this.

If you can't finish today and/or discover more issues, I don't think there's a rush for this.

Alex-Jordan:
Squeezed in an example while I'm waiting for this meeting to start. (Also fixed a small bug in that example that would not have been noticed until the problem was used while logged in to the WW server.)

rbeezer:
* In the extraction file, with the new exercise, I see
```xml
The first answer is: <less /><greater /><ampersand />'&quot;;
```
I should think the single quote mark should be `&apos;` on the WW server side?  I don't think we will get bit by this, if all attributes coming back from the SimpleXML routine are wrapped in double-quotes.  But maybe WW wants/needs a fix?

* The `answer` for the second question is
```xml
<p><m>\verb[!#$%()*+,-./0123456789:=?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[</m></p>
```
The delimiters don't match.  Is something off-by-one?  I can't see just why this would be happening.

Alex-Jordan:
Are you expecting `\verb[...]`? In this context, it's supposed to be `\verb[...[`

rbeezer:
Right.  Confused...

rbeezer:
So an author puts "dangerous" characters within `pg-code` and they migrate to a WW answer.  As a string, it gets wrapped in a LaTeX `\verb` construction.  If it is part of a math answer it gets wrapped in a PTX `m` tag as an eventual `static/exercise/answer`.

We don't exercise too much control over the contents of `m`, so authors are free to make stuff that does not compile.  Not much we can do here.

In this situation, the `\verb` means the text is rendered in a monospace font - certainly in PDF, and apparently in live HTML.

Would it not be better to wrap the whole thing in `\text{}` and escape characters like `#, &, %`, etc.  Isn't that what we would tell an author to do with a chunk of text inside an `m` element?

It feels to me like this `\verb` device is just an end-run around doing a little bit of work to manage characters authored in Perl migrating to their proper equivalents in LaTeX?  Would `\verb` suffice to signal where this care should be taken?  Am I mis-understanding something (again)?

Alex-Jordan:
> In the extraction file, with the new exercise, I see
>> The first answer is: `<less /><greater /><ampersand />'&quot;;`

> I should think the single quote mark should be &apos; on the WW server side? I don't think we will get bit by this, if all attributes coming back from the SimpleXML routine are wrapped in double-quotes. But maybe WW wants/needs a fix?

This part is handled by the WW server. There are two steps. One is when the PGML parser sees things like a quotes pair and makes a note to make these smart quotes. Probably I have that set to come out as `q` in PTX output. Then after that, there is a cleanup that turns things like `<` into `<less />` because we used to need that. With your recent work, I can tone down this cleanup step.

Davide C intentionally changed `\text{stuff}` to `\verb�stuff` at one point:
https://github.com/openwebwork/pg/commit/7ae713871238e2452f333937f79a0685848df123
So I don't want to mess with that at its source. We could look at the mbx script changing that `\verb` to `\text` and experimenting with all the escaping of characters. For some reason `\text{}` was insufficient, and we'd be looking to see if escaping was enough to counter that (in both tex and MathJax).

Alex-Jordan:
Let's not merge yet. There are at least two other closely related things I'd like to get right first/at the same time. One is in a post I just made about verbatim, pre, and code. I'll post the other thing too now. And there isn't a rush. This can wait several weeks or whatever it takes.

Alex-Jordan:
I keep finding something or other that tells me that `\verb` is necessary. Then I think harder, and I find a way out of it and make progress with `\text`. But now I've hit an issue I can't escape. 

This whole thing with either `\verb` or `\text` only comes up within math. That is important, because I think none of your tools for treating naked characters differently in different output modes can be applied.

Now, with MathJax, `\text{}` does not expand macros inside the argument, with a few exceptions. For example, one of the exceptions is that `\text{\}}` will print a `}`. But `\text{\%}` really will print `\%`. OK, if you just want the percent sign, `\text{%}` will print `%`. But if we have that in PTX source, then it will break the PDF production.

There's no way around this with making some new macro like `\percent` because, again, in MathJax that macro will not be expanded. And again, since this is all inside math, we can't use tools that would turn `\text{%}` in source into one thing for HTML and another for PDF.

So we have a situation where MathJax is not faithful to the original LaTeX. This stems from MathJax wanting to support `\text` but not to the extent that LaTeX supports it. Either we can

1. Leave it as `\verb`. Or
2. Convert it to `\text` but throw an error/warning  if the string contains certain characters. `%` is one, but the issue is similar with `#`, `^`, and probably more.
3. Search the string for the dangerous characters and only convert to `\text` if they are absent.

I can imagine real answers that are text strings that actually use `%`, so 2 seems like the worst choice. I can't decide between the consistency and simplicity of 1 and the aesthetics of 3.

Alex-Jordan:
I'm pretty happy with how things are now (with a lot of unpushed changes). 

There are a few changes to webwork2 and pg to go along with this, and I just opened PRs for them to their develop branches. While I wait for those to be merged, I'll work on documenting all of this in the author guide, etc. 

Then I'll recommit all the work here, plus unpushed work, plus documentation, and open a new PR with the commits more sensibly organized.

It will remain for you to intercept bare apostrophes in author source and replace them with `[$APOS]`. But `[$APOS]` will be there waiting for you once the pg PR is merged.

Alex-Jordan:
OK, I force pushed again after reorganizing commits. Known issues (that I think are outside the scope of this PR):

1. Anything with stage~d problems.
2. Issue posted about in the forums with ``\char`\{`` eating up a space that follows (a `mathbook-latex.xsl` problem)
3. Issue posted about in the forums with `[$APOS]`. Seems to be no good way to implement that, and we could live with seeing some `<rsq/>`s in the returned PTX. (There are certainly other things there that are clearly not written by a human.

rbeezer:
> 3. Issue posted about in the forums with [$APOS]. Seems to be no good way to implement that, and we could live with seeing some `<rsq/>`'s in the returned PTX. (There are certainly other things there that are clearly not written by a human.

At 30c94eb764fcb3b0250c4dee4abe422237c4a4f2 apostrophes in text are being trapped and made in `[$APOS]`.  Suprisingly few of those actually within a problem itself in sample chapter, and coincidentally, only in the formatting test.  I did not rebuild samples at website.

The apostrophes (which were `<rsq/>`) are now missing in `static`.  Hopefully I didn't do things out-of-order.  I'll review PR soon.

rbeezer:
> 2. Issue posted about in the forums with `\char\`\{` eating up a space that follows (a mathbook-latex.xsl problem)

That's been resolved at 953c6b6ade08390d367d5f925b121dbe213b47c5, it was only for verbatim text and I think the switch to `\texttt{}` made it unnecessary.

Alex-Jordan:
I need to use this branch for making an answers manual for ORCCA, so I rebased from dev. I went ahead and force-pushed, so just a heads up about that.

Alex-Jordan:
Found one actual issue to fix (that goes back to a386c52) and one improvement in functionality of when we abort ww extraction. So two more small commits here.

rbeezer:
Finally getting around to reviewing this.  Sorry for the long wait.

*  Adding new problems seems to change the default seed, then diffs get filled with noise.  Same problem we have with the internal id (which permid should help alleviate, plus reduced extraneous use).  Could we add seeds to each problem to make testing easier?  That can go on a new issue/PR, not a showstopper here.
* One problem - building any conversion of the sample-chapter from the merged file gives a bunch of PTX errors/warnings for XML elements inside of verbatim elements.  Culprit is the 12 x 2 = 24 test characters in the "PGML Formatting" exercise inside `cd` and `cline`.
  -  authored fine, as bare characters
  -  `extract-pg.xsl` and `extract-pg-ptx.xsl`, run by themselves, produce bare characters, since basically the text elements just go through the `"sanitize-text"` utility template (verified)
  -  Am I right to conclude that the WW server is doing a more general transform on this text, when it shouldn't be?
  -  Note that it is no longer necessary for PTX source to use `<dollar/>` and friends.  So we could kill two birds with one stone:  100% drop the conversion of the 10 LaTeX characters (and more).  These are not deprecated yet, but careful code is in place and has not caused any problems (yet).
  - 25 errors being reported.  I've been seeing a `<dollar/>` behaving similarly during builds for the website, so perhaps that will be addressed here.

Alex-Jordan:
On the first point, yes, let's add seeds to a future PR.

On the second, before I think too hard, did you change the server to -dev? I sort of expect what you describe with -ptx. I would want to coordinate you merging this PR with me bringing the upgrades from -dev to -ptx.

rbeezer:
Sorry - yes, I was on `-ptx`.  Lost track of what I was doing.  I'll try another run tomorrow.

rbeezer:
OK, finally got this configured right for testing.  Looks good.  (But let's get seeds to not change when content is added!)

*  `<!-- PTX:WARNING: PGML wanted to indent here. -->`  Are these necessary?  There are tons of them and I'm not sure about the necessity.  Also, better to not have them floating around inside `p` as they could easily disrupt some things like whitespace handling if the code is a bit too loose (they could be matched inadvertently).
*  LaTeX -> PDF and I get lots of `\'` in the apostrophe testing.  Seems to originate in "static"?

rbeezer:
Seeds can change now or later.  It won't hold up this, but it will increase the 
reliability of testing before/after.  I understand not doing it early.  But 
maybe it can start quickly, and include a disclaimer?  The sample performs two 
contradictory roles: documentation and testing.  We can't always have both. 
Maybe we should "permid" it as well, just to get stable ids.

I'll argue that indentation of a "p" is presentation.  So I'd vote for dropping 
that one entirely.  If the remaining are all from OPL, then that makes a lot of 
sense to me.  However they were authored is beyond any thought of PTX, so it 
seems right to record an author's intent that we can't recover properly.

Any extraneous nodes at start of "p" or end of "p", or next to display items 
(math, lists, and "cd"), are places where delicate stuff happens, and eventually 
I should audit all the preceding/following constructions.  So at teh start 
(inside) makes me nervous.  For now, the best thing to do is place it just 
before a "p" starts, perhaps on its own line.  That interstitial space is safer.

I'll pursue  30c94eb  confusion in a bit.

What is the timing/stategy for this?  Should it merge ASAP, or do we need to 
coordinate with  -ptx  server, or a general WW release?

I will likely combine this and break it out (like the seemingly distinct work on 
answer blank width, and sample chapter edits, etc) and then put your name on teh 
pieces.  So it will be a small, but one-time only, job to reorganize.

Rob

Alex-Jordan:
>  What is the timing/strategy for this?

I think like this.

1. I make some changes to -dev to drop noting the indentation, and move the other notes outside the p.
2. I could also do seeds in the sample-chapter.
3. You review and signal me that things are ready to merge.
4. At roughly the same time, within a few hours of each other, you merge, and I port the -dev edits over to -ptx.

Now that leaves both -dev and -ptx on WeBWorK "develop" branches, plus a few extra changes that will be in an open pull request to WeBWorK's "develop". It will mean that if someone does not want to use the AIM servers, they will need to put their server in about the same state. Before we had the AIM servers, I did not like the idea of having features that didn't work with the current "master" of WeBWorK. But the AIM servers mitigate the concerns, and waiting for WW 2.15 for releasing this feature strikes me as far too long a wait.

So I think I will manage -ptx to be on the WW "develop" branch, plus a few extra things that are in the pipeline to join the WW "develop" branch. From time to time, "develop" branch features will have bugs. It is very unlikely that we will care, because the bugs are very likely to be with other aspects of WW that we do not use, such as file management within an actual WW course.

>I will likely combine this and break it out (like the seemingly distinct work on answer blank width, and sample chapter edits, etc) and then put your name on teh pieces. So it will be a small, but one-time only, job to reorganize.

OK, but do you see that it is already in 8 separate commits? For example the answer blank simplification is in its own commit.

rbeezer:
Perfect.  Do we need to announce "use -ptx server while we evolve 
functionality"?  (Maybe we did?)

I only see 8 commits right now.  Not how each one is organized.  ;-)

I have about 8 conversations going on right now and want to patrol title-ending 
periods in HTML while I have a chunk of time.  ;-)

If certain commits are atomic, that will make consolidation much, much easier.

So we just need to sort out 30c94eb?  Ihave that on a sticky note.  I.e., high 
priority.

Rob

Alex-Jordan:
With 30c94eb, it would be OK to just manually undo it as a new commit. Except in one place you fixed a typo in a comment, so that could stay. But it's a very small commit. You'd revert to just not singling out an apostrophe as a character to do special management with.

We will announce for sure that answers are now harvested automatically.

I'm unsure if announcing about using -ptx is the right thing, or if it should just be in one of the guides.

rbeezer:
Alex - I'm going to revert the one commit.  Will un-revert later, and just let typo come and go, I think.  Haven't looked.  Then see what consolidation to do on the branch.

I'm concerned about folks using their *own* server.  I'd like to tell them where they need to be on -announce.

I plan to be ready to merge, but can sit tight until you are ready.  Then we can coordinate.  Seeds can be independent.

Rob

Alex-Jordan:
Dear Rob,

I have made the PTX:WARNING edits to the -dev server. I will wait to do seeds. If it's OK with you, when this is all done I'll just set seeds without regard to whatever seed is used now for problems. So with that one commit, the diff of output might be large. But the diff of PTX source will just be seed-setting

1. Try processing the sample-chapter again now, remembering to set the server to -dev. The only differences from last time should be PTX:WARNING~s gone for indentation, and the ones about center alignment should now be outside of any p.

1. Try processing the sample-chapter with the server variable being `SERVER = "(https://math.webwork.rochester.edu,daemon_course,daemon,daemon,daemon)"`. That is Mike's server. It is version out-of-the-box 2.14. You won't get answers, but you still get an extraction. See if it is acceptable. This is to check that merging this PR will not break things for people not using an AIM server.

When I use Mike's server, I can make the PDF. I get no errors, but lots of warnings about using things like `<dollar>` inside code. And there is corresponding funkiness in the output. It's hard to imagine this mattering in practice. How rare are these WW problems with code inside that has special characters? Ask that question for PTX-authored. And then again for OPL, but with OPL it will only matter when the problem is using PGML.

There are other differences from doing this with Mike's server, because he has not configured the host course to be used by PTX. So for example you see notes about an instructor preview. The "standard" setup for a host course to process PTX does something to nullify these.

To my chagrin, I see now that we have already made things incompatible with 2.13. Maybe I knew this at some point in the past, but I had forgotten. You can try the above with `SERVER = http://math.mc.edu` to use JT's server where he has set up an "anonymous" PTX processing host course. Even on dev branch without this PR, it fails. Because with 2.13, for example the "ptx" outputformat hasn't been put in place. And probably many other things. Probably I said goodby to 2.13 when we switched from the tex-based method to getting WW to output actual PTX.

After you review this all, either signal me you are ready to merge (and we live with the `<dollar>` stuff for people using their own server until 2.15 comes out) or we can keep talking and see if we can work around that.

rbeezer:
> I have made the PTX:WARNING edits to the -dev server.

Great!

> I'll just set seeds without regard to whatever seed is used now for problems.

Sounds good.

> When I use Mike's server, I can make the PDF. I get no errors, but lots of warnings about using things like <dollar> inside code. And there is corresponding funkiness in the output. It's hard to imagine this mattering in practice. How rare are these WW problems with code inside that has special characters? Ask that question for PTX-authored. And then again for OPL, but with OPL it will only matter when the problem is using PGML.

Well, my additude is not how rare something might be, but instead how bulletproof we can be.  Just a few mess-ups lead to low opinions.  OPL is OPL, nothing we can do and we have to communicate that fact.

> There are other differences from doing this with Mike's server, because he has not configured the host course to be used by PTX. So for example you see notes about an instructor preview. The "standard" setup for a host course to process PTX does something to nullify these.

So why are we testing this if it is not configured?  I'm missing the point here, it seems.

>  To my chagrin, I see now that we have already made things incompatible with 2.13. Maybe I knew this at some point in the past, but I had forgotten. You can try the above with SERVER = http://math.mc.edu to use JT's server where he has set up an "anonymous" PTX processing host course. Even on dev branch without this PR, it fails. Because with 2.13, for example the "ptx" outputformat hasn't been put in place. And probably many other things. Probably I said goodby to 2.13 when we switched from the tex-based method to getting WW to output actual PTX.

Is it possible for PTX output to be built on one server (AIM) but with hosting talking to a different server (Mississippi)?

rbeezer:
Alright, tested, and ready to merge.  Thanks for your patience waiting for this one.  Let me know about timing.  Likely any evening (including Wednesday) would be a good time if we need to coordinate.

*  When you redo seeds, can we move the `SERVER` variable into `Makefile.paths`?  That'd make switching branches easier when testing.  You could include as much documentation as necessary.
*  Semi-related.  You saw all the deprecations this morning?  :-(  As a PreTeXt producer, you'll want to react to those changes.  Sorry.  ;-)

Alex-Jordan:
OK, great. @davidfarmer just installed a needed perl module on -ptx. I believe I've made -ptx (close enough to) identical to -dev. If you are ready, you could do one quick test by switching the SERVER to -ptx. I'm not in a position to test myself or I would, sorry.

If it passes the test, merge into dev, and I can post to -announce about answers being available now. (And also mention the issues with using a 2.14 server.)

rbeezer:
I've run the sample chapter on `-ptx` and compared it to the same run earlier this afternoon, both with the PR active.  Diff on the "extraction" file shows the answers are not there.  Those are the **only** significant differences, so I think I must be using all the right combinations.  But I've been confused before.

Alex-Jordan:
It looks like I did not correctly restart apache. It is difficult-ish to restart apache on that server, because of the permissions I have.

I was looking at raw output and not seeing the answers. I restarted apache "the right way" and now I do see the answers. So I'm optimistic it will work now :)

rbeezer:
That's it.  Merging now.

rbeezer:
Monumental.  Thanks for all the work on this one.  A big step forward.

Go ahead and announce.  Keep it concise and high-level, but do try to explain the server and versions situation.  Maybe point to new documentation.  Invite discussion on -dev and/or -support as you think appropriate.

Thanks!

PR #984, opened 2018-12-18
Add {} after \textbar

mitchkeller:
This is necessary so that an author can write Wolfram|Alpha and not wind up with Wolfram\textbarAlpha throwing errors at them.


rbeezer:
Thanks, I missed that one.

rbeezer:
Thanks very much, Mitch.  I should add a test to the sample article which smashes all these macros up tight to other characters, so this does not happen again.  I'll do that as part of some upcoming (general) work on characters like this.

PR #986, opened 2018-12-23
Move percent sign to after optional valign argument.

oscarlevin:
This fixes sidebyside panels not respecting valign arguments in latex.  The issue was that the optional [center] and [bottom] commands were being put behind a `%`.


rbeezer:
Thanks!  I'll merge in a few days.

rbeezer:
Thanks again, Oscar, for the catch and the fix.  The bug and the correction are amply illustrated in the sample article, especially the current Subsections 22.2 and 22.5.  Should have noticed that!

PR #988, opened 2019-01-08
slight changes for cli newbies

kcrisman:
I think these changes, or something analogous to them you approve of, would be helpful.  Interestingly, I expected `mbx -v` to give me the version when applied without arguments!


rbeezer:
I adjusted the path in the latter part, so `user` would be in the right place.  Pushed, but have not updated website.

Thanks!

PR #997, opened 2019-02-11
ToC as list, along with js and css reorganization

davidfarmer:
The ToC is a nested list, with no `h2`s.

Javascript and CSS now taken from pretextbook.org.  Each also has version numbers.
(So, `html.js.server` is no longer deprecated.)

The attributes `knowl` and `refid` are now prefixed with `data-`, along with corresponding
javascript and CSS changes.

Colors are specified by `html.css.colorfile`, but `colors_default.css` is the only option
currently available.  Consequently, `html.css.file` is deprecated.

The class `mathbook-content` has been replaced by `pretext-content`, and CSS updated.
This may break custom css implemented by publishers.

The new default colors are not quite the same, but I can fix that later.
(Will not require a pull request.)

I will prepare an announcement describing these changes.

### DIFF
--- a/xsl/mathbook-common.xsl
+++ b/xsl/mathbook-common.xsl
@@ -9959,7 +9959,7 @@ http://andrewmccarthy.ie/2014/11/06/swung-dash-in-latex/
         <xsl:with-param name="date-string" select="'2018-07-10'" />
         <xsl:with-param name="message" select="'the obsolete  html.css.file  parameter has been removed, please use html.css.colorfile to choose a color scheme'" />
             <xsl:with-param name="incorrect-use" select="($html.css.file != '')" />
-</xsl:call-template>
+    </xsl:call-template>
     <!--  -->
     <!-- 2018-09-26  appendix subdivision confusion resolved -->
     <xsl:call-template name="deprecation-message">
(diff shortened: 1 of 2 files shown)


rbeezer:
Visual review looks good.  I'll test tomorrow morning first thing - another snow 
day.

Let me know about timing for a merge and push.  You might have submitted a few 
minor issues, so this would be PR #1000.  A double-milestone.  ;-)

Rob

davidfarmer:
Currently the colon and space between a title and subtitle (as in the
sample artice) is added by CSS.

I took that out of the CSS, but did not add it to the xsl because it did
not seem completely trivial.

If you add that, then all is good.

If you don't, then I'll put back the CSS (which is outside this
pull request).

rbeezer:
> If you add that, then all is good.

I have subtitles fixed on a branch.  Everthing else is fine as well.  Knowls not behaving in Jupyter, but that is perhaps my fault.  More testing to do.  HTML validator is much happier the `data-` attributes, so still worthwhile.  I understand "all is good" meaning I'm clear to push, which I will do, unless you catch me in the next few minutes.  A bit more testing to do...

davidfarmer:
There are three other xsl files where "aimath.org" is still hard-coded.
One of them is  mathbook-jupyter.xsl
So that is not the correct CSS, but I don't think it is simple
to fix that.  I suggest taking care of that later.

rbeezer:
Pushed.  Minor editorial work on commit messages.  Melded "clean up" commit, and moved deprecation message into chronological order.

Nice work on the XSL, and I know there is a significant amount of CSS and JS backing all this.  **Thank-you.**

PR #998, opened 2019-02-13
Give ARIA attributes to fillin

Alex-Jordan:
This puts `role="img"` and `aria-label="10-character blank"` (or some other number) on the spans for fillins that are not inside math. This makes them accessible. Note I considered other roles, but nothing available seems to be a great fit.

For a fillin inside math, there is nothing to do right now.  Volker says it can't be made accessible right now. But it is on his roadmap to do something, as a similar issue has come to his attention with Ximera.

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -6753,6 +6753,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     <xsl:value-of select="5 * $characters div 11" />
                     <xsl:text>em;</xsl:text>
                 </xsl:attribute>
+                <xsl:attribute name="role">
+                    <xsl:text>img</xsl:text>
+                </xsl:attribute>
+                <xsl:attribute name="aria-label">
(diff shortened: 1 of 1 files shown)


rbeezer:
Yes, the ARIA roles are not always clear.  Good to be getting some experience with them, though.  I'll have look if/when power comes back.  :-(

Alex-Jordan:
>if/when power comes back.

We had snow Friday night, enough so that I couldn't drive up my hilly driveway Saturday evening. Then it rained Sunday. And Monday. And Tuesday. Supposedly we will get a few hours of not-rain Wednesday morning.

rbeezer:
Thanks - good start on a full suite of ARIA attributes!  ;-)

I've been advised its better to "hardcode" the output HTML in the template, and you can still add some additional attributes programatically.  This is a great example.  So your code was great, but I've been slowly changing over as I am in the neighborhood.  This was a perfect example.  So a new commit on top, migrating.  Have a look - just to see the difference.  At a706823e61ea8163735438d311c51a1bae217e3f

Thanks for this one!

PR #1007, opened 2019-02-24
Hide LaTeX macro definitions in HTML WW

Alex-Jordan:
See https://pretextbook.org/examples/webwork/sample-chapter/html/section-4.html#exercises-1, exercise "Every Continuous Function has an Antiderivative". You should see a blank line at the top of the webwork. This is where LaTeX macros are defined. 

This PR puts that in a hidden div (styled inline, since PTX styling doesn't affect WW problems) so after the PR, that blank line is not visible.

### DIFF
--- a/xsl/extract-pg-common.xsl
+++ b/xsl/extract-pg-common.xsl
@@ -370,9 +370,11 @@
         </xsl:variable>
         <xsl:if test="$macros != ''">
             <xsl:text>loadMacros("PCCmacros.pl");&#xa;</xsl:text>
-            <xsl:text>TEXT(KeyboardInstructions(q@\(</xsl:text>
+            <!-- @ character is unlikely to appear in latex macro definitions, -->
+            <!-- so it makes a good delimiter, better than single quote.       -->
+            <xsl:text>TEXT(MODES(PTX=>'',HTML=>q@&lt;div style="display:none;">\(</xsl:text>
             <xsl:value-of select="$macros" />
-            <xsl:text>\)@));&#xa;</xsl:text>
+            <xsl:text>\)&lt;/div>@));&#xa;</xsl:text>
         </xsl:if>
     </xsl:if>
     <xsl:if test="$b-verbose">


davidfarmer:
One of the few times when inline styling is good, is when you
want something to be hidden.  Usually hidden forever, but occasionally
hidden until some later time.

Example of hidden forever: LaTeX macros.

Example of hidden initially:  GeoGebra calculator.

In particular, "absolutely no inline style" is not an achievable
goal.  But we still want to eliminate most inline styling, of course.

rbeezer:
>  @ character is unlikely to appear

makes me mildly nervous.  I understand there may be no better choice.  See #1009 for eventual warning.

Merged and pushed with no testing.  ;-)  Thanks for working on the math bits.

Alex-Jordan:
Of course, as soon as you said:

>makes me mildly nervous. I understand there may be no better choice.

I realized there is a better choice :)

I could make it use a Perl here-doc. Like:
```
HTML => <<'END_MACROS';
\newcommand{\mymacro}{...}
\newcommand{\myothermacro}{...}
END_MACROS
```

I'll probably do this. Sometime when I am less deep into wine time.

rbeezer:
> I realized there is a better choice

Excellent, that'd save me #1009.

> Sometime when I am less deep into wine time.

Yes, I'm well along being done for the night.

Alex-Jordan:
Continuing conversation here just a bit more before trying a new PR.

Backslash nearly always has special meaning.  Moreso in PG, which mixes LaTeX and Perl with their own meanings for backslash. Part of how PG deals with this is by almost immediately converting `\` to `\\`. It often gets converted back later. But sometimes you really really don't want that first conversion to happen at all. For that, PG has `~~`, which becomes `\` immediately after `\` becomes `\\`.

Without getting into details, to use a here-doc as I described earlier, it turns out that `~~` cannot show up in the LaTeX macros. So far I cannot think of a way around this (without changing WW). Either we pick a single character (currently `@`) or we pick `~~`. The single character could also be some "untypable" ASCII if we wanted. Whatever we pick needs to be banned from the LaTeX macros.

Do you see a "winner"?

rbeezer:
Without getting into details, can we sanitize macros to have ~{}~ instead (which 
I guess would break the function of tilde?  Or maybe not?

No matter what we do, we can use a Schematron rule to catch and warn.

Alex-Jordan:
You are taxing my imagination :)

```
$\def\^^#1{{\tilde{\tilde{#1}}}}$

$\^^t$

$\^{}^t$
```

"works" in MathJax, in that it successfully defines a command, applies it to make a double-hatted "t", and then fails on line 3.

In regular LaTeX that macro definition fails. But it's this kind of thing that could break the empty node insertion as a solution.

PR #1008, opened 2019-02-24
Support xref in mrow in webwork

Alex-Jordan:
Thanks to ff244d4, now it seems like xref is behaving well in WW. This issue had been the nbsp character used in xref text. 

(We don't actually link from a WW problem, since the problem could have a life of its own. But we would like to print the text at least. Authors should be careful that a reference to "Theorem 2.1" may have meaning in the context their book, but not elsewhere.)

This change gets mrow to actually process its xref children, and puts an example in the sample chapter. After processing, look at the solution to checkpoint 1.4 for an xref in an mrow, and in part 1 of checkpoint 2.2 for an xref not in an mrow. Working, both in HTML and PDF.

All with no changes to WeBWorK :)

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -263,8 +263,8 @@
                     </statement>
 
                     <solution>
-                        <p>We <em>add</em> the exponents as follows:<md>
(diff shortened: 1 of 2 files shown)


rbeezer:
Looks good.  Merged and pushed on sight.  ;-)

Issue #960, opened 2018-09-17
Showcase Document


rbeezer:
Copied from `pretext-support` post on "Documentation Work" on 2018-04-04.
https://groups.google.com/d/topic/pretext-support/adGS5sv0dx4/discussion

An idea for the documentation.

The "sample article" began life as exactly that, an example of what could be done.  But it quickly morphed into a testing document and a collection of as many possibilities as we had the patience to write.

I think we need a "Showcase Article."  A good example or two of every main feature, structured in a rational way that shows off how a document can look and behave.  Never trivial (though lots of Lorem Ipsum, or *all* Lorem Ipsum, would be fine).  Never broken or incomplete.  But not too long.  Model formatting in the source.  Always validates.  And so on.  But also always up-to-date with the latest (stable) additions.

It'd be a prominent link off the main web page for PTX, in as many formats as possible (once conversions are similarly mature).  For newcomers that are interested it'd be, well, a showcase.  And for those getting started it might serve better than the sample article to answer questions like, "How do I build a complex table?"

And it is a perfect activity for a volunteer.    Or volunteers.  It could live in a separate GitHub repo, and every time I update the other samples (as I do with every push of the code), my script would just pull and re-build (like I do with Jessica K's tutorial).  So someone could take the lead as owner of the repository, and others could make pull requests as they were interested in tackling a newly-mature feature.

Any takers?  It'd be a big help and you'd collaterally be a PTX pro.

Alex-Jordan:
What amount of auxiliary work do you want to build it? Thinking about latex-images, webwork, youtube thumbnails, etc.

rbeezer:
Any amount.  ;-)  It is not for newcomers to experiment with.  It is a 
demonstration of every great feature at least once (YouTube videos) without 
being too cluttered.  So an expert can scrape the thumbnails and commit them, etc.

A newcomer might look at the source, and might eventually try building it for 
practice.  Maybe a documented build script would be a good companion.

rbeezer:
Initiated at #1029.  Lots more to do, so will leave this issue around for a while.

Alex-Jordan:
#1207 will add WeBWorK exercises to the showcase article.

rbeezer:
[...]

PR #1011, opened 2019-02-25
Ggb calculator

Alex-Jordan:
This gets the calculator going. Just a GGB one, but room for other things later.

I thought about documenting use somewhere. I think I might prefer to wait a bit. After completing work on ggb slates, I'd like to return to this and allow an author to customize the calculator with API commands just like the plan for slates. Any write up of this feature will include a description of that customization option.

There are some accessibility issues with the calculator, as discussed in email. Mainly,

1. We don't know yet if a screen reader user can get much use out of it. There is perceptional accessibility and operational accessibility. It seems to be to have operational accessibility, but perceptional is a question mark to me without some AT user testing.

2. While it's not a full on keyboard trap, getting out of the calculator and back into main content currently requires some special doing. Maybe we can improve upon this with time.

With both issues, especially the first, it may be that GGB itself needs to evolve. In any case, we can give warnings in the documentation about accessibility concerns that can be introduced by activating this feature.

Once this feature is merged, will it be demonstrated in one on of the samples? I wonder which is best. Maybe the sample-chapter, since that already has accessibility issues with the WW iframes? Just trying not to corrupt the accessibility of the other two.

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -140,7 +140,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <xsl:param name="html.css.server" select="'https://pretextbook.org'" />
 <xsl:param name="html.css.version" select="'0.2'" />
 <xsl:param name="html.js.server" select="'https://pretextbook.org'" />
-<xsl:param name="html.js.version" select="'0.1'" />
+<xsl:param name="html.js.version" select="'0.11'" />
 <xsl:param name="html.css.colorfile" select="''" />
 <xsl:variable name="html-css-colorfile">
     <xsl:choose>
@@ -7577,6 +7577,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
        <xsl:text>var role = 'student';&#xa;</xsl:text>
        <xsl:text>var guest_access = true;&#xa;</xsl:text>
        <xsl:text>var login_required = false;&#xa;</xsl:text>
+       <xsl:text>var js_version = </xsl:text><xsl:value-of select='$html.js.version'/><xsl:text>;&#xa;</xsl:text>
     </script>
 </xsl:template>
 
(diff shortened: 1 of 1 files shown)


davidfarmer:
I think the sample article HTML should have the calculator.

rbeezer:
> I think the sample article HTML should have the calculator.

I agree.  Sample article has one of everything, is mainly for testing, then for reverse-engineering undocumented items (like this will be to start).  It has lots of bad things going on, so bad-accessibility is not a deterrent.  If we had a showcase document (#960), then we would only want good examples of accessibility there.

Can you make a commit (here, or elsewhere) to add it to the sample article?

Alex-Jordan:
> Can you make a commit (here, or elsewhere) to add it to the sample article?

You activate it with a stringparam `html.calculator geogebra-classic` whan building HTML. Is there a place in the repo for such things, like an XSLT layer for the sample article? I was thinking you would need to put it into your build script.

rbeezer:
> you would need to put it into your build script

Right!  Sorry for the noise.

rbeezer:
I was going to ask David to check-off on this, but it seems straightforward with regard to the JS, so I'll procede (soon) unless David tries to put a hold on this.

davidfarmer:
If the JS needs attention, I'll do that after the sample article is 
rebuilt.

rbeezer:
Excellent!  I'm going to update website in the next hour, with calculator enabled on sample article.

@Alex-Jordan - you can do the announcement on `-announce`.  Please prefix the subject line with `[NEW]`.

I added a commit to edit the XSL lightly.  At 4ccfd5cb2a2b5875a598bff52f82a960123b9d1f.  Didn't try to fix it into the other three, given mix of ownership.  Two comments:

*  Greater use of `xsl:text` and hard-coded new lines makes the HTML much more compact.  Neat and tidy.
* This is one of those places where a named template **is** the right call.  ;-)  If the result tree only depends on variables, and not on context/elements, then it is like a parameter-less procedure call.

Nice work, Alex and David.

Alex-Jordan:
Should I announce now, or wait until I complete planned enhancements? My guess is another day for slate, plus back and forth as you review it. Once we feel good about that, it will not take much to give the calculator the same functionality.

If it's not clear, what I mean is letting an author put something in docinfo/html with some API commands that prime the calculator. For instance, that would be an avenue to get the spreadsheet view running. Or if you really want a major and minor grid on the graph. Or if you want the graph to always start out viewing [0,10]x[0,10] or whatever because your book covers real applications without negative numbers.

rbeezer:
Calculator works, let's announce.  Improvements will be incremental.

Enhancing a publisher option via source will require discussion.

PR #1015, opened 2019-02-28
More idx entries, and an aside in a project

davidfarmer:
I added index entries for some things I was looking for, and deleted some backtics that
were not needed.

Also added an aside in Project 4.1, between the prelude and the statement.
When I make the HTML there are no error messages, but the aside does not appear.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -820,6 +820,11 @@ the xsltproc executable.
                             <p>This is quite the activity upcoming.  This is a <c>prelude</c> authored within the <c>activity</c> element, but visually just prior.</p>
                         </prelude>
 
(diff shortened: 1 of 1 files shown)


rbeezer:
> Also added an aside in Project 4.1, between the prelude and the statement.
> When I make the HTML there are no error messages, but the aside does not appear.

That could be a good place for it.  But just like footnotes (`fn`), we'll need to decide just where they go.  Implications for the schema, and for the code (as you have discovered).

Let's do that discussion on `-dev` not here.

davidfarmer:
In tomorrow's newsletter (which I should finish and send to you
for comment) I will ask for examples of Asides.  I can do some
poking around before bringing this up on -dev.

PR #1028, opened 2019-03-24
switch latex.print hyperref setting to hide links instead of deactivate

oscarlevin:
Added "hidelinks" option in both latex.print=yes (true) and latex.print=no (false) to allow a publisher to quickly switch inside the .tex file.  Modified latex comments accordingly.

### DIFF
--- a/xsl/mathbook-latex.xsl
+++ b/xsl/mathbook-latex.xsl
@@ -2025,11 +2025,11 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     </xsl:if>
     <xsl:if test="$latex.print='no'">
         <xsl:text>%% Hyperlinking active in electronic PDFs, all links solid and blue&#xa;</xsl:text>
-        <xsl:text>\hypersetup{colorlinks=true,linkcolor=blue,citecolor=blue,filecolor=blue,urlcolor=blue}&#xa;</xsl:text>
+        <xsl:text>\hypersetup{hidelinks=false,colorlinks=true,linkcolor=blue,citecolor=blue,filecolor=blue,urlcolor=blue}&#xa;</xsl:text>
     </xsl:if>
     <xsl:if test="$latex.print='yes'">
-        <xsl:text>%% latex.print parameter set to 'yes', all hyperlinks black and inactive&#xa;</xsl:text>
-        <xsl:text>\hypersetup{draft}&#xa;</xsl:text>
+        <xsl:text>%% latex.print parameter set to 'yes', all hyperlinks "hidden", set to textcolor (but still active to preserve spacing)&#xa;</xsl:text>
+        <xsl:text>\hypersetup{hidelinks=true,colorlinks=true,linkcolor=blue,citecolor=blue,filecolor=blue,urlcolor=blue}&#xa;</xsl:text>
     </xsl:if>
     <xsl:text>\hypersetup{pdftitle={</xsl:text>
     <xsl:apply-templates select="." mode="title-short" />


rbeezer:
**Really late** getting back to this one.  Sorry!

The hyperref manual says that "draft" implies "all hypertext options are turned off".

The intent is to make a PDF that will not get rejected by print-on-demand, Amazon, etc for having banned constructions.  So far nobody seems to have had any problems, so I'm reluctant to change when it is not broken.

Authors should not be editing their LaTeX file.  ;-)  With the CLI, they should never even see it.

We could write the unused version as a comment (in both cases) so the alternative is there, without actually changing the commands.

Some interesting alternatives here:
https://tex.stackexchange.com/questions/53300/how-to-disable-links-completely-using-hyperref-package

oscarlevin:
It's been a while, but I believe the reason for my suggested solution was that using the draft option actually messed with some spacing or page breaks.  That meant that the print and non-print versions would not have the same content on the same pages.

In terms of the other concerns:
1. Amazon always rejects my files and I have to post-process them through adobe anyway.
2. Because LaTeX is bad at doing good page breaking, the *publisher* needs to go through the pdf and latex source carefully to fix things by hand.  This is really unavoidable.  But then the same changes need to be applied to both print and digital pdfs, so it makes much more sense to be working from a single latex file and split the file at the very end to turn off color and hyperlinks.

mitchkeller:
I'm just showing up to agree 1000% with @oscarlevin. Unless a publisher wants to unleash a garbage print version, they will have to edit the LaTeX source; there are always going to be things that cannot be handled by global styling. With my git-driven workflow for production of the final PDF/print-on-demand versions, I don't really see a need for a `hidelinks` option, but I'm not opposed to it.

sean-fitzpatrick:
I'm not sure if this is related to the issue @oscarlevin is reporting, but a discovery from today that might be relevant:

For print, I'm using the two sided printing option. I just discovered (after an hour or so of being confused about things) that
`\documentclass[oneside]{book}` uses `\raggedbottom` for page layout, while `\documentclass[twoside]{book}` uses `\flushbottom`.

Consequences of `\flushbottom` include:
- some page breaks will be different
- to try to fit the page with the flush bottom, LaTeX will sometimes squish things, which can result in adjacent tcolorboxes overlapping. (I noticed this with Sage input/output boxes, which I have set up to fit together.)
- if you use `\pagebreak` mid page (maybe you shouldn't, but if you do...), with \flushbottom, the page content will get stretched. So you could have one paragraph at the top, and one at the bottom.

rbeezer:
Had to duplicate this, since I'd left it for so long (XSL files all got renamed, so rebasing goes wild).  Has Oscar's name on it.

Still got blue links in the print version, which print faintly.  So I got to messing around.  I think the adjusted code does the right thing, but subtleties or testing is welcome, and I won't let it sit for years.

I get identical page breaks in the sample article (at least halfway through), **if** I set `latex.pageref` to `no` for the print version.  (That should become a publisher switch).

Pinging @sean-fitzpatrick  since he is another page-fitter.  Updating website examples just now.

sean-fitzpatrick:
I don't know that I do too much page fitting, but I do like having the page breaks align for my print and colour versions of the PDF. (So if I tell a student, "It's on page 53" they don't have to ask, "which version?")

The two things that affect where pages break for me are the page references (since this adds text) and the flush vs. ragged bottom. I've never noticed any difference with hyperref set to draft or not. Do @oscarlevin or @mitchkeller still notice a difference once the change to ragged bottom is made?

I like the aesthetic of the flush bottom, and I would have been willing to default to that in both versions, if not for the cases where two adjacent tcolorboxes end up on top of one another.

sean-fitzpatrick:
Which I guess is equivalent to saying that if flush bottom were to be chosen as a publisher option, there would have to also be an adjustment to all the tcolorbox styles to force a bit of vertical space before and after.

PR #1029, opened 2019-04-02
Showcase

Alex-Jordan:
Here is what it looks like now in PDF and HTML:
http://spot.pcc.edu/~ajordan/temp/showcase/


Alex-Jordan:
I can commit the image files if they look good to you. Also, do we want the showcase to offer the "archive" download options?

rbeezer:
Looks great!  Yes, go ahead and commit images.  I will commit this, this 
afternoon, so either do it right now, or do it on a new PR.

Did you design these images?  If not, do you want to put source citations into 
the source?

Alex-Jordan:
Commit all of .tex, .pdf, .png, .svg, .eps fo the archive feature?

Alex-Jordan:
Force pushed with changes.

1. Swapped out one image so that...
2. I could be sure to not violate someone's copyright, and put comments attributing the image code
3. Activated archival image links
4. Gave images xml:id and committed all five image files for each one
5. Applied `valign="bottom"` to two of the sidebysides (but see #1031)

rbeezer:
Sorry, forgot that part, in haste to get to class.

Your call - I'd maybe do two of the most useful/distinctive?  Or one-each of 
source (*.tex), vector (*.svg), and raster (*.png).

Definitely show the archive feature, but it does not have to be 
overdone/comprehensive.  We want authors to discover what is possible (and how), 
and then run to the documentation for full details.

This would be a place to drop some sort of comment in the relevant docinfo 
incantation that makes this possible.  And a place for me to maybe write some 
documentation!

Maybe the showcase will drive the documentation....

Rob

Alex-Jordan:
Archival image files can be local to an image, is that right? It does not have to be global for the document? How about a future PR could move the archive lines of code to just these images, or a subset.

rbeezer:
On 4/2/19 10:36 AM, Alex Jordan wrote:
> Archival image files can be local to an image, is that right?

I forget.  ;-)

> How about a future PR could move the archive 
> lines of code to just these images, or a subset.

Yes, generally we can iterate.  It is not like the XSL where we do not want to 
release broken code, or avoid the need to deprecate mis-informed decisions.  So 
if we decide to work the Showcase differently, we can just change it.  But 
thanks for getting the seed right, that is important if we get contributors.

Corollary: I'm likely to merge blindly.  ;-)

Rob

rbeezer:
Looks great!  Thanks for starting this.  Forget to add PR number, so very quickly did a forced push after editing commits.

rbeezer:
Appears that all the versions of the "image-parametrized-surface" graphic did not make it into the commit?

````bash
$ git ls-files examples/showcase/images/*.svg
examples/showcase/images/image-building-parabola.svg
examples/showcase/images/image-gaussian-sum-proof.svg
examples/showcase/images/image-implicitly-defined-curve.svg
examples/showcase/images/image-rainbow-perception.svg
examples/showcase/images/image-raindrop-refraction-and-reflection.svg
```

Alex-Jordan:
Oops. If you look at where I gave it an xml:id, I gave the description an xml:id instead of the image. Will fix later tonight.

rbeezer:
Thanks, I'll merge in the AM. I've pretty much got the build script working.

Alex-Jordan:
Perhaps you should use a draft watermark for the showcase. It looks a bit awkward if you arrive at the showcase but there is only one feature on display.

There will be other features that happen at process time to consider. For now, the calculator is all that comes to mind.

PR #1030, opened 2019-04-02
Look for stack presence when determining width percentage of video, e…

Alex-Jordan:
…tc. in sbs

I had a video and some p's in a stack, in a sbs. (For a course [calendar](http://spot.pcc.edu/~ajordan/MTH65Spring2019/section-calendar-mw.html).) When I went to make PDF, it wants to make the qrcodes, but to do so it needs to calculate the video's width. It came back as NaN because the way this template was not considering the possibility that the video could be in a stack.


rbeezer:
Thanks for the fix.  Pushed, but website not updated (yet).  Only cursory testing, so keep an eye on your calendar, etc.

PR #1032, opened 2019-04-03
Give xml:id to Showcase image

Alex-Jordan:
This is to follow up on #1029. Fixing a bad placement of an xml:id.


rbeezer:
Merged, forgot the PR number.  Again.  Won't fix that.  ;-)

Thanks.

PR #1033, opened 2019-04-03
Add a principle of accessibility

davidfarmer:
I wrote the principle as simply and broadly as possible:

   11. PreTeXt recognizes the inherent value in producing material
        that is accessible to all users.

This covers multiple meanings of "accessible".

I don't think the statement of the principle is the place to explain what the
principle means.  Do most people understand this phrase from the 1st principle?
       ...captures the structure... .
In my experience, the people who most need to understand the implications of
that phrase, are the farthest from understanding it.  But it still is useful as something
to point to when it influences a decision, and repeated exposure helps people
grasp the meaning.

I expect the 11th principle to serve a similar purpose.

Not completely thrilled with the word "recognizes", but various synonyms did not strike
me as superior:  endorses, acknowledges, supports,... .  

But "recognizes" is a pretty good word because it conveys the idea of awareness,
and raising awareness is a key step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,4 +40,4 @@ interaction of authors, publishers, scholars, instructors,
 students, and readers, with each group having its own needs
 and goals.
 11. PreTeXt recognizes the inherent value in producing material
-that is accessible to all users.
+that is accessible to everyone.
--- a/README.md
+++ b/README.md
@@ -39,3 +39,5 @@ LaTeX and then developed using standard LaTeX tools.
 interaction of authors, publishers, scholars, instructors,
 students, and readers, with each group having its own needs
 and goals.
+11. PreTeXt recognizes the inherent value in producing material
+that is accessible to all users.


rbeezer:
recognizes -> values?

davidfarmer:
I like the phrase "inherent value".  I like the "inherent" part, because
it conveys the idea of doing what is right, and that doing what is right
will have a lot of benefits.  You don't need to use the known benefits
as justification.

And I think "value" is an even harder word to replace.

rbeezer:
> I like the phrase "inherent value".

OK.

Since I seem to be in a word-smithing mood:  we've manged to avoid "user" in the principles, which I always associate with software.  And is a weak word.

How about recycling the consumers from Principle 10:

all users -> scholars, instructors, students, and readers

davidfarmer:
> all users -> scholars, instructors, students, and readers

How about keep the "all" and rearrange?

all students, instructors, scholars, and readers

rbeezer:
Sure!

> How about keep the "all" and rearrange?
> 
> all students, instructors, scholars, and readers

davidfarmer:
Now I don't like how that reads immediately after principle 10.

How about:

PreTeXt recognizes the inherent value in producing material
that is accessible to everyone.

rbeezer:
> accessible to everyone

Sure!  (And I think that indirectly implies a sort of universality.)

@Alex-Jordan ?

Alex-Jordan:
This is great, let's merge it.

rbeezer:
Will do.  Later today.

@davidfarmer Would you like to announce?  If so, please preface subject line with `[NEW]`.  You can add lots of verbose explanations of what it all means.  ;-)

rbeezer:
Added (verbatim) to the lists in the Author's Guide and the Publisher's Guide.  Smooshed all three commits into one.  Merged and pushed.  Website update needs to wait until a few hours from now, once onto home machine.

rbeezer:
Live now on website, so can be announced.

PR #1045, opened 2019-04-22
Add stacked slates in an "interactive" to sample article.

rickroesler:
Rob: I've completed the work on the "stacked" sidebyside. I updated sample-article.xml per your instructions and included a new code file (code/roesler/kinematics.js) and preview file (images/kinematics.png). 

NOTE: I had to fiddle quite a bit with the <interactive> width and the aspect ratios to get the html to look ok. Don't know if that's typical.

NOTE: When I built the pdf, the preview image is very small, and there are QR codes all over the doc which I don't previously recall seeing.


rbeezer:
Nicely done, Rick.  And that's a great demo.

Comments:
* QR codes in the PDF are normal.
* Print representation of every interactive is wide and short.  Your demo is skinny and tall (as it should be).  So your preview image looks really small.  Not a problem.
* 6:1 + 6:3 + 6:3 + 6:3  suggests to me an overall 6:10, which looks pretty good to my eye.  If you want that change, I can just do it instantly, no need for a PR.

Two nits:
*  Two new blank lines had twelve leading spaces.  I cleaned them out.
*  I took the period off the "caption".  Not sure if that is right or not, but for short ones, I've not used periods.  One day we will have functional "title" also and we can get that sorted.

I rebased your branch on my `dev`, and edited the commit message with `PR #1045` so that GitHub will link them.  That makes your branch obsolete.  I will send instructuions just now on our email thread.

Thanks very much for the contribution.

Rob

rickroesler:
Thanks for the feedback. I  now see where the blank lines have spaces; an artifact of the Microsoft VS Code editor, I think. Is that something that I need to be concerned about?

When I made the aspect ration > 6:7, the caption seemed far too far away from the figure. Even 6:8 seemed wrong.

Curiosity: What are the QR codes for?

Rick

davidfarmer:
> When I made the aspect ration > 6:7, the caption seemed far too far away from the figure. Even 6:8
> seemed wrong.

Distance from figure to caption should not be a consideration
when choosing aspect ratio.  Can you point me to an example
where it looks bad and I'll see if I can adjust?

David

rickroesler:
I just recompiled the sample-article locally changing the aspect ratio of my example to 6:10 (Line 4198). The larger the aspect ratio, the farther away the caption is.

[cid:image002.png@01D4F884.A4835930]


________________________________

Alex-Jordan:
> Curiosity: What are the QR codes for?

When everything is uploaded in the right place, a reader who scans the QR codes with their phone will be taken to a place like this one: https://pretextbook.org/examples/sample-article/html/kinematics.html

rickroesler:
Well, that is quite clever.

________________________________

rbeezer:
I've only been offline for an hour or so.  ;-)

I think it is my problem, height and width are fixed in pixels.  Perhaps due to the use of "stack".  Look at Rick's source and generated PTX.

rbeezer:
Extra whitespace can be a small problem, but not so much right here.  I just like to be neat and tidy all the time.  ;-)

davidfarmer:
> I think it is my problem, height and width are fixed in pixels. Perhaps due to the use of "stack".
> Look at Rick's source and generated PTX.

I'm not sure what (if anything) I am supposed to look at at this point.
Please post HTML when there is something for me to look at.

David

rbeezer:
I've finally been able to revisit the layout and sizing of this.

I think the root problem was the `sidebyside` had a 70% width and 5% margins.  I fiddled around a lot, setting aspect ratios on a 60 x 70 grid for the whole interactive.  Eventually, it all came back to setting the margins to 15%.  The four `slate` "add up" to a 6:10 aspect ratio, but I guess the 70% width makes that a 6:7 aspect on the overall `interactive`?  In any event the three graphs are now  420px x 210px  which seems to be Rick's original intent when designing these?

In any event, no change to the code, and all the HTML widths and heights seem to behave properly.

* Change made and pushed, sample article updated at website.  At f625d250f062cce4daab17438d57bbcb9a895abc

* The layout parameters should not have escaped summing to less than 100% without warning!  See #1051.

PR #1047, opened 2019-04-29
Add math examples to Showcase Article

Alex-Jordan:
I couldn't stand the Showcase Article going on without something more in it, especially at least demonstrating some math content. So I added these things while I sit through 3_ hours of ORCCA compiling. Still getting a feel for what tone you'd like the SC to have, so feel free to criticize.

@davidfarmer may want to make use of semantic macros here.

Unless the accessibility issues are a detractor, I think that the build script for the HTML should toss in the GeoGebra calculator.

It would be good to show off the index too. But there needs to be some things in there worth indexing. Enough to have knowls that show several types of blocks, as well as index links to several types of division.

### DIFF
--- a/examples/showcase/mathematics.ptx
+++ b/examples/showcase/mathematics.ptx
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<!-- This file is part of the PreTeXt Showcase article. -->
+<!--                                                    -->
+<!-- Copyright (©) 2019 The PreTeXt Organization.       -->
+<!-- See the README file for contributing guidelines.   -->
+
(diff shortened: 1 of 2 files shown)


rbeezer:
Nice!  I'll try to get the merge in this afternoon, and the calculator.

rbeezer:
Merged.  Thanks!  I like the style.

Fixed one typo, and put nearly-full SE URL into a footnote (which I like for print, so good to demonstrate here).

Also added calculator and "draft" watermark, as requested on #1029.

rbeezer:
> It would be good to show off the index too. 

This suggestion has moved off this completed pull request to a new issue, #1049.

PR #1053, opened 2019-05-07
Index advice1

davidfarmer:
Terminology, principles, and advice on indexing.


rbeezer:
Removed some whitespace.  Used title-case on the subsection, which is the style so far.

Thanks!  Good to have this material expanded.

There are now **two** forward references from the overview to the topics.  Did you mean there to be two?  Should we drop one?  (And/or split out "Notation"?)

davidfarmer:
I did not intend two forward references.

I'll fix that, and also some typos, in another pull request.

rbeezer:
> I'll fix that, and also some typos, in another pull request.

See #1054

PR #1064, opened 2019-05-31
Add cc unit

Alex-Jordan:


rbeezer:
Post here when you are really, positively, absolutely done!  ;-)  **Then** I'll merge.

Thanks for improving the units features!

Rob

Alex-Jordan:
We have units like `mileperhour` with abbreviations `mph` because it's kind of its own thing, separate from `mi/hr`.

I'm adding `bit` and `byte`, so what do you think about `bitpersecond` and `bytepersecond`? Transfer rates are often expressed in `Mbps`, so it's a common abbreviation.

I don't see a future where we automate away these compound units. Meaning, I don't think we would enable `<unit base="mile"/><per base="hour"/>` to become `mph` instead of `mi/hr`. Because there will be a need for things like `mph/s` and it would be awkward to handle that without an explicit `mileperhour`. 

This is all my way of saying I think `bitpersecond` and friends are good for the long haul.

(Note that I think we _will_ some day enable `<unit base="mile"/><per base="hour"/>` to come out as `mi`&nbsp;`hr^{-1}` instead of `mi/hr`.)

rbeezer:
Should I merge this now?  You can always start a new PR if you want to revisit this.

Alex-Jordan:
Should be ready to decide in a few days. Almost done with the new units section of ORCCA.

rbeezer:
OK, no rush at all - just wanted to make sure ball was not in my court.

Alex-Jordan:
OK, I think this is ready. I force pushed so it is one commit. Sorry, there is a lot of white space realignment included.

Alex-Jordan:
This should be ready for review. I believe @lightheartw is hoping to use these soon.

rbeezer:
Its in!  See commit added 3 days ago.  Mea culpa - I forgot to close it.  Running too hard here.

PR #1067, opened 2019-06-05
Fix failure of mbx script on certain WW exercises

mitchkeller:
Hopefully this is the sort of thing that @Alex-Jordan had in mind.

### DIFF
--- a/script/mbx
+++ b/script/mbx
@@ -593,7 +593,13 @@ def webwork_to_xml(xslt_exec, ptx_xsl_dir, xml_source, server_params, dest_dir):
             # By reversing the string, we can just target first match
             reverse = static[problem][::-1]
             parts = re.split(r"(\n>tnemetats/<|\n>tnih/<)",reverse, 1)
-            static[problem] = parts[2][::-1] + parts[1][::-1] + answer + parts[0][::-1]
+            try:
+                static[problem] = parts[2][::-1] + parts[1][::-1] + answer + parts[0][::-1]
+            except:
+                msg = "PTX:WARNING: It appears that {} is missing <statement>. "
+                msg = msg + "Continuing, but extraction likely not valid PTX."
+                print(msg.format(problem))
+                static[problem] = parts[0][::-1] + answer
 
         # nice to know what seed was used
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
The previous line is looking for the last closing `statement` or `hint` tag, so that it can insert the answer hash data right there. (There may be a `solution` that follows.)

I think this is the line to work on. Or, really, even earlier. If we are specifically testing for presence of a statement tag, let's devote a check to that all by itself, independent of where the answer is spliced in.

Much earlier, the checks are run to see if the file was empty, did not compile, or produced invalid XML. This check should be added in that area.

(And now I understand why this didn't happen to you before. The new bit is that we get the WW answers automatically, and this choking is happening when answers are attempted to be inserted.)

rbeezer:
When one of your proposes a final fix, the other can give it a "positive 
review", and I'll merge, likely as-is.

Thanks for the prompt work on this one.

mitchkeller:
I am, on some level, way out of my depths in trying to do any serious modifications to the WW processing. I typically take it as a black box. Given that a lot of the XML used in the extraction files isn't specified as part of the PTX schema, it's hard for me to infer what's supposed to be going on at any particular place in the code. This change gives me something that lets me get an extraction file that I can modify on only four problems and then it all works for what I need. Thus, I'm not going to invest too terribly much into coming up with a more robust solution here.

That said, I'm killing time before the indexing call, and so I'll give a quick rundown of something that I think I could do to see if @Alex-Jordan says it's in line with what he's thinking:

- In the block of code beginning with the comment `# Now if we are not aborting upon recoverable errors`, do something if there's no `statement` element around the time the place where the `webwork` tags are changed to `static`.
- How egregious would it be of us to just wrap `<statement>` around the text that occurs between `<static>` and the first of `<statement>` and `<hint>`? I'm not familiar enough with what comes back from the server to know for sure what might be kosher. I'm sure we don't want to help "bad" OPL problems keep hobbling along, but it would be good to do something sensible and write out a warning to the console to tell whoever is running the script that they should look at the extraction for that problem, as it might not make sense, but we've done our best to cobble it together.

mitchkeller:
Close because better resolved by #1069

PR #1069, opened 2019-06-06
Check WW extraction for statement tag

Alex-Jordan:
This is to replace PR #1067 .

After we know we have valid XML, check to see if the root element has a `<statement>` child. And then there are the consequences. Report the error. Maybe terminate depending on -a flag. Otherwise make place holder shell problems.

Also caught an error in the existing code with the place where `url_tag` was defined.

@mitchkeller Can you try this branch and see what happens? I did minimal testing, adding things the WW mini example. If it works for you, we can close #1067. Sorry for the amateur Python. [But I am an amateur after all :)]

### DIFF
--- a/script/mbx
+++ b/script/mbx
@@ -553,7 +553,7 @@ def webwork_to_xml(xslt_exec, ptx_xsl_dir, xml_source, server_params, dest_dir):
             static[problem] = "<static failure='xml'>\n<statement>\n  <p>\n    " + bad_xml_msg.format(problem_identifier, seed[problem],'    Use -a to halt with returned content') + "  </p>\n</statement>\n</static>"
         elif no_statement:
             print(no_statement_msg.format(problem_identifier, seed[problem],'  Use -a to halt with returned content'))
-            static[problem] = "<static failure='xml'>\n<statement>\n  <p>\n    " + no_statement_msg.format(problem_identifier, seed[problem],'    Use -a to halt with returned content') + "  </p>\n</statement>\n</static>"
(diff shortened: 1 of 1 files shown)


script/mbx
```diff
@@ -544,6 +551,9 @@ def webwork_to_xml(xslt_exec, ptx_xsl_dir, xml_source, server_params, dest_dir):
         elif bad_xml:
             print(bad_xml_msg.format(problem_identifier, seed[problem],'  Use -a to halt with returned content'))
             static[problem] = "<static failure='xml'>\n<statement>\n  <p>\n    " + bad_xml_msg.format(problem_identifier, seed[problem],'    Use -a to halt with returned content') + "  </p>\n</statement>\n</static>"
+        elif no_statement:
+            print(no_statement_msg.format(problem_identifier, seed[problem],'  Use -a to halt with returned content'))
+            static[problem] = "<static failure='xml'>\n<statement>\n  <p>\n    " + no_statement_msg.format(problem_identifier, seed[problem],'    Use -a to halt with returned content') + "  </p>\n</statement>\n</static>"
```
mitchkeller: Should this be @failure="no_statement" here? I was out of the office yesterday and haven't had a chance to test, but thought I'd look this over this morning and get to testing a bit later today.

Alex-Jordan: Good catch!

Alex-Jordan: I made it `failure='statement'` to keep it one word like the others. They don't all read well as "It failed because of _____" anyway.  So far, I believe there is nothing that uses these `@failure` attributes, but it seemed like they might come in handy some day.

rbeezer:
Thanks, I'll get it in a couple days, not online right now.  Thanks for the work on this!

rbeezer:
Sorry for the delay.  Merged and updating website examples now.

Thanks for working this one out.

Issue #1086, opened 2019-06-20
Geogebra app draws focus when loaded


dbrianwalton:
Current behavior can be seen in Sample Article HTML when viewing Section 14 (Interactive Elements, Authored). Once the page finishes loading, the view immediately jumps to Figure 14.23 (Geogebra: from scratch). I have confirmed this in both Chrome and Safari.

In a modified version of the section, if I rearrange the figures, the focus is always drawn to the same Figure. If the figure is removed from the source, there is no jumping.

Why it happens: The GGBApplet seems actively requests the page focus, which bubbles through the iFrame. It only appears to be happening when the GeoGebra object definitions are established through javascript commands (hence, the from scratch).

Solution: Add another default parameter for the GGBApplet to prevent focus.
(pull-request is being prepared)

kcrisman:
> Current behavior can be seen in Sample Article HTML when viewing Section
> 14 (Interactive Elements, Authored). Once the page finishes loading, the
> view immediately jumps to Figure 14.23 (Geogebra: from scratch). I have
> confirmed this in both Chrome and Safari.
>

Thank you for discovering that; I had been noticing that all week, but
failed to see the significance - I thought I had been accidentally clicking
on a too-sensitive touchpad!

dbrianwalton:
Unfortunately, the focus attribute may not be enough to fix this. (My page was just taking quite a bit longer for the focus to jump, so I thought I had fixed it.)

dbrianwalton:
This question on the GeoGebra forums suggests that maybe the source is a zoomability on the interactive.

[https://help.geogebra.org/topic/scrolling-through-a-web-page-with-geogebra-applets-geogebra-captures-focus]

dbrianwalton:
Okay, the offending command is the GeoGebra command `setPerspective`, which is the first command in the from-scratch javascript GGB interactive. The change of page focus appears to be an unintentional side effect by GeoGebra (it does not respect the focus flag) and seems to be out of our control.

On the other hand, there appears to be no loss of functionality by dropping that particular command from the script. The view for this interactive shows the graphical view by default.

rbeezer:
Thanks, Brian!  Maybe @Alex-Jordan has more to add?

Alex-Jordan:
[...]

PR #1087, opened 2019-06-20
Prevent javascript geogebra apps from taking focus as default behavior

dbrianwalton:
Resolution for Issue #1086.
Adds a default "preventFocus: true" parameter to the GGBApplet.


dbrianwalton:
Hold off on this. It's still pulling focus (but was taking longer so I missed it).

PR #1089, opened 2019-06-20
Improvements to EPUB generation

mitchkeller:
This cleans up three issues with the EPUB conversion. One impacts the CSS by ensuring that it no longer adds periods after codenumbers that now have them added by the XSLT. The other two impact build.sh by making it cross-platform (fixing the `sed -i` issue) and getting us closer to validating by only including images declared in the manifest (via the `images-list.txt` file that the XSLT generates).


rbeezer:
Thanks, Mitch!  These are great improvements!

PR #1090, opened 2019-06-20
Revise WeBWorK documentation

Alex-Jordan:
Following some feedback from @nordstromjf today. Also while I was in there, made a few corrections, moved the existing idx tags to a better place, and reformatted source more to my liking.


rbeezer:
Thanks @Alex-Jordan  and @nordstromjf!

PR #1092, opened 2019-06-20
Fix threejs splines example in sample-article update error

dbrianwalton:
In May, the splines.js used in sample-article was edited to remove stats window. This introduced an error when the object (now gone) was asked to do an update call.

Commented this offending line out so that there is no longer an error.


rbeezer:
Thanks, Brian!  Nice catch.

PR #1107, opened 2019-06-21
Can capture pretext hash without submodules

jjrsylvestre:
Something you said this morning (Fri 21 Jun) made me realize you can capture the pretext commit hash without all the submodule stuff. (And the submodule idea has a flaw I didn't realize until after we had talked on Thu 20 Jun in Portland.)

Here is a very simple approach to grab the dev commit hash and place it in the header. The drawback is that if fails if the user is compiling without the -xinclude flag. (But I account for this in the xsl code edit.)

### DIFF
--- a/pretext.hash
+++ b/pretext.hash
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="ASCII" ?>
+
+<pretexthash xmlns:xi="http://www.w3.org/2001/XInclude">
+<xi:include href=".git/refs/heads/dev" parse="text" />
+</pretexthash>
--- a/xsl/mathbook-common.xsl
+++ b/xsl/mathbook-common.xsl
@@ -37,7 +37,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
 <!-- Identify as a stylesheet -->
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0"
-    xmlns:xml="http://www.w3.org/XML/1998/namespace" 
+    xmlns:xml="http://www.w3.org/XML/1998/namespace"
     xmlns:date="http://exslt.org/dates-and-times"
     xmlns:exsl="http://exslt.org/common"
     xmlns:str="http://exslt.org/strings"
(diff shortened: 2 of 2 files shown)


jjrsylvestre:
If using this, you will need to make sure that other means of obtaining PreTeXt (zip file, Dave R.'s installer, etc.) include a file .git/refs/heads/dev, whether including the actual commit hash used to bundle up PreTeXt, or perhaps just empty.

rbeezer:
Excellent!  Do you think a "git post-commit hook" could reliablly construct `pretext.hash` without needing the `xi:include` part?  Maybe just `echo` the XML tags and `cat` the file containing the hash?  Then a `document()` can read it in as it does now?

https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks

jjrsylvestre:
That's what I was doing with the submodule approach (but with a pre-commit hook), so yes it can be done but then you're back to a chicken-and-egg situation. The post-commit hook would only update the hash in the committer's local repo, and then another commit would be needed to push the new hash to the public repo, and then that new hash inside pretext.hash would be immediately out of date.

I guess another approach would be to adopt a "always double commit when modifying xsl" policy:

> modify xsl
> commit
> manually store this latest commit hash in pretext.hash
> commit *only* the change to pretext.hash

For any commits that don't touch mathbook/xsl/, don't bother to update pretext.hash. Then you can always compare the hash stored in users' output headers with the hash in pretext.hash, to see what update to mathbook/xsl they are using.

(This is essentially equivalent to the submodule approach, but without the extra complications of actually using submodules...)

rbeezer:
Yep, I realized early this morning I was back to the chicken and the egg again.  :-(

So then I found the "entity technique" near the end of

https://stackoverflow.com/questions/20109910/how-to-read-plain-text-content-with-xslt-1-0

to read in a "plain text" file.

I have not yet tested, but I suspect if the XSL gets separated from the `.git` directory, then this entity business will raise a big (fatal) error.  Which will cause more problems than it solves.

There have been requests for tagging the repository regularly at stages where the code is reliable.  Maybe the most reasonable course is to manually copy the adjacent commit hash as part of that process.  Then the double-commit routine will just happen on a semi-regular schedule (and is not a permanent fixture in daily workflow).  For example, the header could look like:

```
Most recent stable commit, June 2019 
  002c0f48cf8a103d69962b351e08446d8b409713
```

jjrsylvestre:
Would probably want to do that on a separate branch dev-stable (or just
master ;)) and ask end users to use that branch. Otherwise, what gets
written to the header from pretext.hash will get out of sync with what was
actually used to process a document. (But then users using the dev branch
will have the same problem.)

rbeezer:
The CLI is now the way to go for non-developers.  And the now oft-requested `pretext support` will include the "core" commit being employed (in addition to the CLI version number).

So we are now getting good information from authors whose problem is that they have not updated recenty.  ;-)

@jjrsylvestre - I appreciate all the excellent technical work here, but I think we've got what we need through the move to the CLI.  Thanks very much, all the same.

PR #1116, opened 2019-06-26
Couple of small EPUB changes

mitchkeller:
Change to the sed script fixes a validation issue. Change to the XSLT solves both a validation issue and a content issue as author biographies are now both put in the manifest and on the spine to be readable.


rbeezer:
Thanks, Mitch!  

git nit:  Since I edit commit messages, the "old" commits on this branch are stale, and the "official" ones have new commit hashes.  In other projects, what you have done would work well.  I was able to `fetch` this branch and `cherry-pick` the tip, which was easy.  Best practice for you would be to delete a branch once I've merged it.

I'm going to clean up my work from yesterday right now.

PR #1124, opened 2019-07-01
sample cross-reference

kcrisman:
I'm sure this is not exactly what is desired (or even the right syntax!), but just checking as I would like to improve these as I read through/use these documents, now that they are linked.

Unfortunately, the sample article is still probably the best resource for nearly everything in terms of detail, and I don't think that can be xref'ed - nor would it work that well for that purpose, given its structure.

### DIFF
--- a/doc/guide/basics/basics-part.xml
+++ b/doc/guide/basics/basics-part.xml
@@ -129,7 +129,7 @@
             <p>There are two common types of lists: ordered and unordered.
             <idx><h>list</h><h>ordered</h></idx>
             <idx><h>ordered list</h></idx>
-            (There's also the description list. See the documentation for more information on it.) As in <init>HTML</init>, an ordered list is produced with <c>ol</c> and an unordered list with <c>ul</c>. The items of your list are structured inside <c>li</c> tags.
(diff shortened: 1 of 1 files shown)


kcrisman:
> I see very little value in linking to overview-lists instead of topic-lists here. I am adding such an xref right now on my branch.

Okay, I just picked what seemed to be a likely link.

> We're going to have a call for contributions soon, and that will include instructions on how to do things in a standardized way. 

Hmm, perhaps it shouldn't have been added to the "official" branch here then quite yet.  I'm not sure what you mean by standardized but I assume that will be clear when you put that info on pretext-dev.

mitchkeller:
> > We're going to have a call for contributions soon, and that will include instructions on how to do things in a standardized way.
> 
> Hmm, perhaps it shouldn't have been added to the "official" branch here then quite yet. I'm not sure what you mean by standardized but I assume that will be clear when you put that info on pretext-dev.

We've got at least three people's efforts involved in just getting The Guide merged, and so the steps happen when those individuals can make them a priority. Furthermore, a few of the things related to issuing the call for contributions required seeing the merged Guide and figuring out best practices. Standardized refers mainly to index entries (three of us, not identical to the three mentioned earlier, will be fully indexing the Basics Reference in the coming weeks), xml:id/xref style, and code snippet style. Enthusiasm is appreciated and welcomed, but sometimes patience must also be part of things.

PR #1127, opened 2019-07-01
Suggestion as to earlier xref mention

kcrisman:
I'm sure you won't use this verbatim, or even in anything like its format, but I highly suggest a link to the xref section this early, so that people don't lose energy before getting there and then have to go back and start `xml:id`ing at that point.


rbeezer:
This needs to be rebased, since Mitch just updated the Guide.  KDC - a forced push here is fine.  Thanks.

PR #1129, opened 2019-07-01
Second method for Font Book

kcrisman:
I'm sure you'll want to reorganize this, but I recommend inclusion of this second method, because I found it easier than messing with trying to get stuff from texlive into the Font Book using the GUI - and it kept me from the madness of downloading new versions of these fonts.


rbeezer:
Thanks!  Reformatted slightly.  Public soon.

PR #1131, opened 2019-07-01
found source for Greek quote

kcrisman:
Apparently that quote is ALSO ancient Greek.  So go figure.  Modern Greek only has two diacritics.  You can remove the word "modern" various places if you confirm.  Not sure why I had so much trouble with this earlier - I think it was the missing link to the CM fonts.


rbeezer:
Aaa-ha!  Thanks!

PR #1134, opened 2019-07-02
trivial typo

kcrisman:


rbeezer:
Combined into one commit.  Thanks.  Will become public shortly.

PR #1137, opened 2019-07-03
Add list of character elements

whaynes:
I saw your call for contributions to the documentation.  

I was looking for this information and it wasn't where I thought it should be, so I added it.

Will


rbeezer:
@whaynes - very nice, thanks!  Two nits:

The space near the end of an empty tag was something I saw an expert advise online.  Then an expert I paid as a consultant told me the opposite, which is what I try to do now.  ;-) This would be a good place to model that early on.  And we have `<tage>` for the empty version of `<tag>`.

I can quickly make those changes and have just one commit with your name on it.  Are those changes OK with you?

whaynes:
Absolutely, go right ahead.

rbeezer:
Thanks, Will!  That's a great addition for folks just getting started.

I had to make `<icon>` actually create one as the demonstration part.

Delete your branch, and pull, since the commits have changed due to my edits and manipulations.

Thanks for contribution - welcome to PreTeXt development!  I hope there will be more.

Rob

PR #1141, opened 2019-07-03
Improve rendering of units in quantities

whaynes:
This tweaks the html rendering of units by putting a non-breaking space between the quantity and the units, and a non-breaking hyphen between consecutive units in the numerator or denominator. 

Example:

<p> <quantity>
      <mag>100</mag>
      <unit prefix="kilo" base="newton" />
      <unit base="meter" />
</quantity></p>

produces 100 kN-m instead of 100kNm

You may prefer to use a centered dot, but I don't know how to prevent line-breaking at that character.


whaynes:
Sorry, i didn't escape the pretext

```
    <p> <quantity>
      <mag>100</mag>
      <unit prefix="kilo" base="newton" />
      <unit base="meter" />
    </quantity></p>
```

Alex-Jordan:
Thanks William, some observations:

You added the nbsp within the mag template. Earlier at line 6731, after the mag template is applied, a thin-space is used iff there was a mag and there will be some kind of unit. would it be appropriate to have the new nbsp replace that thin space instead? (And is there such a thing as a non-breaking thin space?)

Maybe a thing like `variable name="non-breaking-hypen"` should be named less specific, like `variable name="inter-unit-separator"`. In the future we could make that something a user can define.

The "thin space" between units is bad if it can break. But it was chosen to follow what happens in the .tex side, where we just get siunitx defaults. I don't personally object to the styling of units diverging in HTML and print, but maybe that's worth thinking about.

Side thoughts: I think there is a nonbreaking thin space character. Should we use it (if we were not already)? Should units just be wrapped in a span that prevents line breaking, and then we can forget about using individual "no break" characters?

whaynes:
> On Jul 3, 2019, at 8:13 PM, Alex Jordan <notifications@github.com> wrote:
> 
> Thanks William, some observations:
> 
> You added the nbsp within the mag template. Earlier at line 6731, after the mag template is applied, a thin-space is used iff there was a mag and there will be some kind of unit. would it be appropriate to have the new nbsp replace that thin space instead? (And is there such a thing as a non-breaking thin space?)
> 
Yes, that is a better place for it.
> Maybe a thing like variable name="non-breaking-hypen" should be named less specific, like variable name="inter-unit-separator". In the future we could make that something a user can define.
> 
That’s a good idea. I just checked the docs and siunitx provides this as an option called inter-unit-product.  There are several reasonable choices including space, dash and center dot and the choice should be the user’s. I suppose it’s possible to feed the selection to siunitx to keep the html and pdf appearance synchronized, but I wouldn’t know how to do that.
> Side thoughts: I think there is a nonbreaking thin space character. Should we use it (if we were not already)? Should units just be wrapped in a span that prevents line breaking, and then we can forget about using individual "no break" characters?
> 
A span sounds like the best solution.  As I understand it, magnitudes and units form an indivisible item and shouldn’t break across lines.  siunitx permits breaks between the value and unit, but its an option which is off by default. As it is now in the html, I think units can break at the fraction-slash, which would be wrong. If the entire quantity is wrapped with a non-breaking span then that would fix the slash problem and any choice for inter-unit-separator would work even if unicode doesn’t provide a non-breaking version.

What’s my next step?  (Sorry, but I’m new here.)

Alex-Jordan:
> What’s my next step?

We need @davidfarmer to confirm it would be possible to have a "nonbreaking span". If so, put that around the unit output (with something like `<span class="unit">...</span>`). Then clean away all the non-breaking characters, replacing with the right things. Might be best to start off going back to thinspace to introduce the span in one pull request, then work in the product options somehow in a separate PR that covers both HTML and print at the same time.

Any new commits can stay on this branch and Rob will clean that up before merging.

>  (Sorry, but I’m new here.)

New is great! We're grateful for more contributors.

davidfarmer:
Now if you wrap the content in a span with class "unit", then you
will not get a line break after the "/" of m/s^2 .
(The specific CSS for that class is
     white-space: nowrap;
)

Maybe this could also help pronunciation, if the screen reader
was somehow told that inside .unit the "m" is pronounced "meter" ?

Currently you put a non-breaking space between "9.8" and "m/s^2".
That space, in the version of the sample article I am looking at,
is narrower than a regular space.  And I think it looks better than
a regular space.

What I did was, inside .unit, made regular spaces smaller by:
     word-spacing: -0.25ex;

To my eye, that looks the same as you had previously.

So, if you wrap in .unit, then you can just use regular spaces
between the number and its units.

All this is backward-compatible with current practice.

Tested on Firefox, Chrome, and Safari.  (On Safari the thin spaces
you were using did not look good.  This new approach looks better.
But Safari still has other issues, such as no space between kg and m
in  kg m⁄s^2 .)

On my own I would have used "units" instead of "unit" for the class,
but I recall a long discussion about plurals for units, so I went
with what Alex wrote.

David

rbeezer:
Nice work!  (By everybody!)

rbeezer:
And I just thought - maybe this is all-around better for accessibility - a span 
to identify content, and just generic spaces, etc for the screen-readers.

whaynes:
Is .unit a new css rule?  If it is, maybe .quantity would be a better name for it since it will wrap both the value and the associated units.

davidfarmer:
I added those new CSS rules.

I just added the same rules for .quantity.

Once we decide which one is official, I will delete the other.

whaynes:
I've implemented the various suggestions.  Here's what's changed:

In mathbook-html.xsl

1. Quantity is wrapped with <span class='quantity'> to prevent line breaking within.
2. Ordinary space is used between Value and Unit
3. References to thin space and non-breaking hypen were removed.
4. variable "inter-unit-product" created, which contains unicode MIDDLE-DOT
   Presumably this could be overridden to use a different separator.

In mathbook-latex.xsl
5. Set siunitx preference to use \cdot as inter-unit-product for consistency with html.

I would suggest that a small bit of padding-right be added to the CSS because
periods and commas are uncomfortably close to the unit in the sample document.

davidfarmer:
I added some margin on the right to move the punctuation away
slightly.  It was tricky to move it enough to make a difference but
not so much that the punctuation seemed disconnected from the
sentence.

I'll let you think about why I did it with margin and not padding :)

rbeezer:
Thanks, Will.  Merged and pushed, and will update website later tonight.  To straighten up some whitespace I ended up smashing the two commits together.  Just a big fiddle otherwise.

@davidfarmer:  `span.quantity` is the only new CSS.

Two nits:

`.DStore` file ended up in commit.  Maybe a too liberal `git add`?

I have used the `xsl:element` instruction in far too many places and am slowly eradication it when it is unnecessary..  So here, a bare
```html
<span class="quantity">
...
</span>
```
turns out to be preferable.

PR #1145, opened 2019-07-08
Guide: Basics Reference improvements

mitchkeller:
This is the pretty-printing of the Basics Reference plus taking care of the index cleanup based on the discussions in Portland. I also fixed some `xref` issues in other parts of The Guide.

### DIFF
--- a/doc/guide/appendices/glossary.xml
+++ b/doc/guide/appendices/glossary.xml
@@ -313,8 +313,8 @@
         </defined-term>
 
         <defined-term>
-                <idx>origin branch</idx>
-            <title>origin branch</title>
+                <idx>origin repository</idx>
+            <title>origin repository</title>
             <p>In git, the default name for the remote repository from which the local repository was cloned.</p>
         </defined-term>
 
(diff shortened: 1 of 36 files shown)


rbeezer:
Good timing!  Working through several of these.  Absolutely no changes, thanks for following house style.

Therefore I broke some habits.  Rebased onto the tip, so commits have changed.  But then did a fast-forward merge but with an explicit merge commit quoting the PR number.

So, the PR is a bit cleaner, the history is not as cluttered with multiple PR references, and GH has a reconstructable record.  But you will still need to delete your branch.

Thanks very much for your work on this!  I'll update website before night is out.  If you want to announce contributors guide, or anything else, on `pretext-announce`, just send it to me and I'll post it.

mitchkeller:
Matt, Oscar, and I are doing the indexing on the Basics Reference and I will send one comprehensive PR once that is done (hopefully soon, because we're doing this to get in sync for a team indexing of our books). Once the indexing is done, I will send you something to send out to `pretext-announce`, as then everything should be in place for contributors to have really good models to base their contributions off of.

rbeezer:
Sounds great.  Thanks.

If not traveling, I can handle well-formed PRs quite quickly.

PR #1148, opened 2019-07-12
EPUB: More simple test elements in sampler, small CSS edit, and TOC change for backmatter

mitchkeller:
* Add an `aside`, a couple of `idx`, an `appendix`, an `index`, an `introduction`, and a `conclusion` to the EPUB sampler for better testing.
* Rearrange the TOC slightly so that it comes where eBook stores prefer on the spine, remove the half-title, and treat backmatter items more correctly for the TOC. This involved adding a new CSS selector so that some `li` in the TOC could be unnumbered.

### DIFF
--- a/examples/epub/epub-sampler.xml
+++ b/examples/epub/epub-sampler.xml
@@ -46,18 +46,31 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
         <chapter>
             <title>Two</title>
-
+            <introduction><p>This is the introduction to the chapter.</p></introduction>
+            <section>
+                <title>A silly section.</title>
+                <p>This section just exists so that we can add <tag>introduction</tag> and <tag>conclusion</tag> tags to this chapter.
+                </p>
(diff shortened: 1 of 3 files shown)


rbeezer:
Thanks, Mitch.
* So what looks like a title page is now coming from the `frontmatter` element.  Right?
* There are other back matter elements that render as appendices, specifically `references` and `solutions`.  So on the next pass maybe you can add those in.
* Should we let HTML assign the letters for the appendix?  Or ask PreTeXt for them and hardcode them?  Can we get away with the latter?

mitchkeller:
> Thanks, Mitch.
> 
> * So what looks like a title page is now coming from the `frontmatter` element.  Right?

Yes. The earlier title page was a real mess on a multi-author book like mine. It also seemed redundant to have more than one when the one generated by `frontmatter` was so nice.

> * There are other back matter elements that render as appendices, specifically `references` and `solutions`.  So on the next pass maybe you can add those in.

Yes. What does the notation list render as? Is there anything else that winds up in backmatter to be aware of?

> * Should we let HTML assign the letters for the appendix?  Or ask PreTeXt for them and hardcode them?  Can we get away with the latter?

I think the ideal solution is to have PreTeXt hardcode all the numbers for the TOC, but I need to do some exploring to see what's valid. Right now, none of the `frontmatter` things (`preface`, `foreword`, etc.) appear in the EPUB TOC. The number on an XHTML `li` can definitely be suppressed via CSS in EPUB, but I'm not sure if it will happily let us assign all of the labels manually. Probably would be good to see how things convert to Kindle on that front, too, before getting too wedded to anything in particular.

rbeezer:
> The earlier title page was a real mess on a multi-author book like

on Judson's AATA, which I tested and it looked good as currently implemented.

> What does the notation list render as?

You put the `<notation-list/>` element into an `appendix` so that is covered.

> Is there anything else that winds up in backmatter to be aware of?

Forgot an overall `glossary`.

> I think the ideal solution is to have PreTeXt hardcode all the numbers for the TOC

Sounds good.

PR #1149, opened 2019-07-13
Fix filename error

sean-fitzpatrick:
Fixing a file reference to .bashrc that should have pointed to .bash-profile.


rbeezer:
Thanks, Sean!  It'll go live with the next website update.

PR #1151, opened 2019-07-17
Account for some schema-compliant division reps

Alex-Jordan:
Lane's ORCCA has an `appendix` with an `exercises` child, akin to a "section". In the rest of the book, `exercises` are children of a `section`, so akin to a `subsection`. What was happening is the `exercises-section` environment was not being defined (only `exercises-subsection`), causing a LaTeX compile error.

This PR adds a check for `exercises` as a child of `appendix`. While I was there, I looked for other schema-compliant corner cases that were not accounted for.

I didn't touch `worksheet`, `reading-questions`, or `glossary` because they are not in the schema, so it's unclear when they can be a peer of `backmatter`, a child of `backmatter`, or a child of an `appendix`.

### DIFF
--- a/xsl/mathbook-latex.xsl
+++ b/xsl/mathbook-latex.xsl
@@ -863,12 +863,12 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         ($document-root//subsubsection)[1]|
         ($root/book/backmatter/appendix|$root/article/backmatter/appendix)[1]|
         ($document-root//index)[1]|
-        ($document-root//chapter/exercises|$root/article/exercises)[1]|
-        ($document-root//section/exercises)[1]|
+        ($document-root//chapter/exercises|$root/book/backmatter/appendix/exercises|$root/article/exercises)[1]|
+        ($document-root//section/exercises|$root/article/backmatter/appendix/exercises)[1]|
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
@lightheartw, when this is resolved, then you will be able to include the appendices in a PDF build. Pinging you so you get notifications about this PR.

rbeezer:
Merged and pushed, but no testing beyond making sure the stylesheet still works.

@lightheartw - it'd be great if you could confirm the fix.  Thanks, Alex, for tacking this one down, I think you original post got lost in the flood - sorry.

Alex-Jordan:
I tested Lane's ORCCA with only section 1.1 and the appendix. Didn't work before this change, and did work after it.

> I think you original post got lost in the flood

This PR _was_ the original post :)  This only came up yesterday and I thought I could just fix it instead of bugging you. Hoping to do that more often once ORCCA edition 2 is off my back.

rbeezer:
> This PR was the original post :)

I'll swear I just got something recently about LaTeX not compiling since one of these environments was not being built.  But I can't find it now, so who knows!

Forgot to say that this little bit of code (to keep the preamble clean) has been wrong several times already.  Thanks for the expert fix!  (Looking forward to more.  ;-) )

PR #1161, opened 2019-08-07
Upgrade Javascript to version 0.12

davidfarmer:
Changes to interactive reading questions required incompatible changes to
the Javascript.  Only difference for PreTeXt is changing Javascript version from
0.11 to 0.12.


rbeezer:
Thanks - updating website right now.

Issue #1120, opened 2019-06-28
check for statement inside stage in extracted WW


Alex-Jordan:
In #1069 we started looking for a statement tag in extracted WW. Because my knowledge of tools for analyzing XML in Python is weak, I just found a way to check if there was a `statement` child of the root `webwork` and called it a day. But this neglected to account for when there is a `stage`, with the `statement` inside the stage, making the `statement` a grandchild of the root, not a child.

So we need a better way.

rbeezer:
I like "lxml" but it does not seem to be part of standard Python.  Sounds like ElementTree is not as powerful/complete.

While poking around, I found this, which sounded helpful for this particular problem, but I won't claim to have studied it very carefully for this particular case:

https://stackoverflow.com/questions/5373902/python-xml-query-get-parent

mitchkeller:
The python documentation makes me think that 

`if problem_root.find('statement') is None:`

can become

`if problem_root.find('.//statement') is None:`

since `//` from XPATH is supported and the example in the documentation says about `//`:

> Selects all subelements, on all levels beneath the current element. For example, .//egg selects all egg elements in the entire tree.

rbeezer:
That sounds right.

rbeezer:
Fied at #1164

PR #1164, opened 2019-08-13
Extend statement search in extracted WW

Alex-Jordan:
This addresses #1120. Enrique ran up against this and reported it to me, so with an affected user  it floated to the top of my to-do list. Super easy. Thanks @mitchkeller .

@rbeezer, should your build script for the sample chapter use `-a`? It would have aborted on the problem there with a `stage` and we would have seen this a long time ago. I must not have been testing with the sample chapter when I made the commit to look for `statement`.

### DIFF
--- a/script/mbx
+++ b/script/mbx
@@ -474,7 +474,7 @@ def webwork_to_xml(xslt_exec, ptx_xsl_dir, xml_source, server_params, dest_dir):
         except:
             bad_xml = True
         if not bad_xml:
-            if problem_root.find('statement') is None:
+            if problem_root.find('.//statement') is None:
                 no_statement = True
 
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, merged, and website being updated right now.

I added `-a` to the build script.  Tested with this commit, no problems.  We will only catch it if I am paying attention.  Script does not halt on errors.  You have to watch it.  ;-)

Alex-Jordan:
> Script does not halt on errors. You have to watch it. ;-)

In this case if you use `-a`, the webwork-extraction.xml file doesn't fully form and is invalid XML. Then the script will attempt to use the merge stylesheet and will explode with issues. Might not halt your script, but it will be a big mess on the screen for a bit.

rbeezer:
Good, I am likely to catch it out of the corner of my eye.

PR #1174, opened 2019-08-29
Add automatic displaystyle for lists of math items

sean-fitzpatrick:


sean-fitzpatrick:
Changes to mathbook-common do what they're supposed to do on my list examples.
Running the sample article through it, the only things that turn up running a git-diff are the list items containing math (these get \displaystyle as intended) , and the time stamp.
It's possible that this is not the most elegant implementation of xsl but it seems to work. :-)

davidfarmer:
If the source had displaystyle in it, then now the HTML has it twice?

davidfarmer:
I think it would be good to include (some of) your test cases in the sample article,
plus another example as in the next paragraph.

If you had the following as a list item
`<li><m>x^2=5</m></li>`
then I am guessing the output will be wrong because there will not be a space
between the `\displaystyle` and the `x`.
That issue is not revealed when the math content starts with a backslash.

rbeezer:
> the output will be wrong because there will not be a space

Right.  I use an empty group (`{}`) elsewhere, but I think a space is proper and best here.

@sean-fitzpatrick I'll likely make some edits, but leave your name on this.  Get back to course prep.  ;-)

rbeezer:
> I think it would be good to include (some of) your test cases in the sample article,
plus another example as in the next paragraph.

I also have a few on a commit I was using to test this.

> If the source had displaystyle in it, then now the HTML has it twice?

Undesirable, but does it break anything?

sean-fitzpatrick:
Forgot to include the updated sample article. It now has the list testing in it.
Added a space after displaystyle in xsl:text which fixes the issue David pointed out (good catch!)
Tested, and LaTeX does not care how many times \displaystyle appears in an equation.

rbeezer:
I was able to make this happen without the `choose` and the duplicated code.  Identical performance on the sample article.  I put it on a new commit - so we know who to blame (me).

@sean-fitzpatrick Your editor may be stripping trailing whitespace, which is a good thing.  But there were two spurious changes in all this (my fault!).  Could have been much, much worse.  I made a new whitespace-only commit.  Because I know how you like that sort of thing.

Thanks - more on the lists.

sean-fitzpatrick:
Yes, I think Atom strips trailing whitespace by default.
Thanks for getting this in.

PR #1175, opened 2019-08-29
add test to only show 'chapter' in heading when it gets a number

oscarlevin:
This fixes the following bug: 

If numbering.maximum.level is set to 0, not even chapters get numbers.  In HTML this results in the heading for the chapter still having the word "Chapter" followed by the title of the chapter.  

Fix adds $numbering.maximum.level!=0 to the test for applying "heading" as the class of the title.

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -773,7 +773,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:element name="{$html-heading}">
         <xsl:attribute name="class">
             <xsl:choose>
-                <xsl:when test="self::chapter">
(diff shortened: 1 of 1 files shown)


rbeezer:
Generally, author "dotted" variables become internal "dashed" variables after some sanitization.  So I changed the test to use the internal variable, and so then to test as integers, not strings.

I tested that this does not break existing behavior.  But I hit this bug, #1178, while testing the fix.  Could you test that this fix still works for you?

I strongly suspect there will be other numbering surprises with a zero maximum level, so keep your eyes open!  Please.

Thanks very much for the PR on this one.

oscarlevin:
Looking good on my end.  I have no footnotes yet.  I'll stay vigilant.

PR #1177, opened 2019-08-30
Correct stringparam name for 1-side vs 2-side

dbrianwalton:
The stringparam in xsl/mathbook-latex.xsl that controls sidedness of PDF is "latex-sides".
The documentation in the publisher's guide says "latex-sided". 
The pull request is to make the publisher's guide match the XSL.


rbeezer:
Thanks, Brian!  I had a little debate with myself on the two names for this, but I guess I did not clean up thoroughly after it was decided.

PR #1181, opened 2019-09-06
Give sample book frontmatter new xml:id

Alex-Jordan:
In the sample-book, both the `book` and `frontmatter` elements have xml:id "sample-book". I was seeing a duplicate xml:id error. This changes the `frontmatter` id to "frontmatter".


rbeezer:
Thanks!  No website update yet.

PR #1189, opened 2019-09-26
Schema says statement is not mandatory in an exercise....

jjrsylvestre:
but basics guide says otherwise.

Also changed some `<c>` tags to `<tag>` tags.

### DIFF
--- a/doc/guide/basics/exercise.ptx
+++ b/doc/guide/basics/exercise.ptx
@@ -1,7 +1,10 @@
 <exercise>
   <statement>
     <p>
-      The <c>statement</c> is mandatory.
+      The <tag>statement</tag> is mandatory when any of
+      <tag>hint</tag>, <tag>answer</tag>, or <tag>solution</tag>
+      is included as a child of <tag>exercise</tag>,
+      otherwise it may be omitted.
     </p>
   </statement>
   <hint>
--- a/doc/guide/basics/exercise.ptx
+++ b/doc/guide/basics/exercise.ptx
@@ -1,6 +1,8 @@
 <exercise>
   <statement>
     <p>
+      <idx><h sortby="statement"><tag>statement</tag></h><h sortby="of an exercise">of an <tag>exercise</tag></h></idx>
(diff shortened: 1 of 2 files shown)


mitchkeller:
Good catch. I would suggest adding some index entries for hint/answer/solution in exercisegroup.ptx to mirror what is already there for `statement`.

I have mixed feelings about removing the `statement` tags from the other `exercise` tags in exercisegroup.ptx. I imagine the unfortunate situation where someone copies and pastes this, puts their content in it, and then remembers that they can add `hint` and tries to put `hint` after the final `p`. At best, they have invalid PTX. At worst, they get mangled output. I don't see much cost to having template code that has the `statement` tag in there even when not strictly necessary, since I'd say it's a best practice to author with `statement` tags in there even without any hint/answer/solution.

davidfarmer:
I have always been against tags that can be inferred by a mechanical 
process.  I understand that we have them because it makes it easier
to write the xsl, but it would be nice to decide that, in principle,
it is desirable to make those tags optional.

I suggest that 'statement' is an example.  When a definition, theorem,
or exercise only contains a statement, then there is no logical need
for the 'statement' tag.

But what if there is a hint, answer, or solution?  As long as those
are required to come after the statement, then one can infer that
the statement is everything up to the first hint, answer, or solution.

Is there some situation I am missing?

jjrsylvestre:
Actually, looking at the output, all the new idx's muddy things up.

Should any of the code snippets have idx at all in them? (Other than demonstrating the use of the <idx> tag, of course.)

Probably should just put any useful idx in as child of the <program> or <listing> that contains the code snippet. You want code snippets that are easy to copy-and-paste-and-use, not snippets that require copy-and-paste-and-delete-a-bunch-of-crap-and-use.

rbeezer:
Finally!  Discussion here gave me pause.  A long pause.  I'm happy with what is here now, as reflecting best advice **given current XSL implementation**.  If we need new advice, we need (substantial) new code.  Thanks, @jjrsylvestre for your patience (and everyone's input).

PR #1196, opened 2019-10-30
Wsl update

mdoob:
This updates the appendix section on Windows Subsystem for Linux. It incorporates:
1. Git is included in Ubuntu by default.
2. WSL as improved interactions with Windows10 software. In particular, it is possible to
    display a local html file in Internet Explorer.


rbeezer:
Thanks, Michael!  Great to see the progress on this.

25 commits here, so I just grabbed the WSL one.  Only.  Removed some tabs and some extra trailing whitespace.  Otherwise, as-is.

There is an Asymptote/EPS commit not mentioned.  If you think that should go in, make a new PR and just point back here - I can grab it in the same way.

PR #1198, opened 2019-11-09
Fix "missing number treated as zero" bug

oscarlevin:
When a `sidebyside` is not in a figure, and latex is compiled using TeXLive 2019, an error "missing number treated as zero" is thrown.  Changing the value of `raster before skip` to a single number instead of `{2.0ex plus 0.5ex}` resolves this issue.  

Still works with TeXLive 2018 compile.  Limited testing does not reveal any noticeable differences.


rbeezer:
Merged and pushed.  More discussion on `pretext-dev`.

rbeezer:
Oops, discussion on `pretext-support`.

rbeezer:
All four affected projects report this fix does the job, so I'll close this.  Thanks very much, Oscar!

Follow-up on `pretext-support` thread.

PR #1200, opened 2019-11-16
Add description list for revealjs

Alex-Jordan:
Here's a simple `dl` implementation. For example, it made [slide 2](https://spot.pcc.edu/~ajordan/AMATYC2019ORCCA/#/2).

### DIFF
--- a/xsl/pretext-revealjs.xsl
+++ b/xsl/pretext-revealjs.xsl
@@ -276,6 +276,13 @@ Reveal.initialize({
   </ol>
 </xsl:template>
 
+<xsl:template match="dl">
+  <dl>
+    <xsl:apply-templates select="li"/>
+  </dl>
+</xsl:template>
+
+
 <xsl:template match="li">
   <li>
     <xsl:if test="parent::*/@slide-step = 'true'">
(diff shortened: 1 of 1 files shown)


rbeezer:
Good timing, I may look at `slideshow` later today.

Any reason not to support the current `@slide-step`? (Which will be renamed `@pause`.)

Alex-Jordan:
> Any reason not to support the current @slide-step? (Which will be renamed @pause.)

No reason. I hadn't even thought about that or noticed it in the other `li` template. Would it progress to reveal title _and_ content at the same time? Allow one step to reveal title, another to reveal content?

rbeezer:
Good.  ;-)  I'll peek and perhaps do an add-on commit.  Yes, I'd think **title and content** should be an atomic unit.  So an attribute of `li` (without looking).  Anybody who thinks different can ask for it.

rbeezer:
Thanks!   Lets keep incrementally improving this.  Nothing like real use to suggest necessary bits.

* You can group the `dt` and `dt` into a `div`.  Precisely so you can style them in unison, such as making the pair a `reveal.js` `fragment`.  So **title and content** together.  Nice.  (https://www.stefanjudis.com/today-i-learned/divs-are-valid-elements-inside-of-a-definition-list/)
* Made the match on `ol/li` and `ul/li` more precise, as you did for `dl/li`.  With a better idea of what we are doing, we can afford to make the XSL less permissive when it is easy to do so.
* All titles get killed by a default template, and then are queried when needed with modal templates like `title-full`, so I removed the `not(title)` filter.

More elsewhere.

PR #1202, opened 2019-11-20
Separate CSS files and switch to CSS version 0.31

davidfarmer:
To help develop more HTML styles, now there are separate CSS files for the TOC, banner,
knowls, and the main body style.  This is accompanied to a CSS version change to 0.31
(skipping 0.3 because it was used briefly for development).


rbeezer:
Thanks for these improvements!  I should try the "body" style with the Jupyter conversion.

Issue #1121, opened 2019-06-28
latex stylesheet swallows idx in exercise


kcrisman:
The schema supports `<idx>` elements in `<exercise>` elements (here, in an `<exercises>` division), but somehow the LaTeX conversion just eats them up.
```
    <exercise xml:id="exercise-cryptography-el-gamal">
      <idx><h>encryption</h><h>El-Gamal</h></idx>
      <statement><p>Learn about the El-Gamal public key encryption method.  How is it implemented?  What mathematics used there is similar to what is used in this chapter?  What is different?</p></statement>
    </exercise>
```

works fine in html, but the reference is gone in pdf.  The `.tex` file has
```
\begin{divisionexercise}{12}{}{}{x:exercise:exercise-cryptography-el-gamal}%
Learn about the El-Gamal public key encryption method.  How is it implemented?  What mathematics used there is similar to what is used in this chapter?  What is different?%
\end{division exercise}%
```

However, if one places the `<idx>` _inside_ the `<statement>`, then all is well:
```
    <exercise xml:id="exercise-quadratic-reciprocity-hard-legendre-jacobi">
      <statement><p>Make up and compute some Legendre symbols that seem pretty hard by using the Jacobi<idx><h>Jacobi symbol</h></idx> symbol instead.</p></statement>
    </exercise>
Make up and compute some Legendre symbols that seem pretty hard by using the Jacobi\index{Jacobi symbol} symbol instead.%
```

So I wonder whether there should be something added [here](https://github.com/rbeezer/mathbook/blob/dev/xsl/mathbook-latex.xsl#L5166):
```
        <xsl:otherwise>
            <xsl:apply-templates select="." mode="exercise-components">
                <xsl:with-param name="b-original" select="true()" />
                <xsl:with-param name="b-has-statement" select="$b-has-statement" />
                <xsl:with-param name="b-has-hint"      select="$b-has-hint" />
                <xsl:with-param name="b-has-answer"    select="$b-has-answer" />
                <xsl:with-param name="b-has-solution"  select="$b-has-solution" />
            </xsl:apply-templates>
        </xsl:otherwise>
```
or somewhere nearby.  

----------

I admit I don't really understand the recursive nature of these stylesheets, so I hesitate to waste a lot of time adding
```
<xsl:apply-templates select="idx"/>
```
at random spots to try it out.  But if this is the right sort of thing to do, I'm willing to test a fix if I can be given guidance on placement.

rbeezer:
[...]

PR #1203, opened 2019-11-21
LaTeX: include index entries from exercises

Alex-Jordan:
An `idx` inside an `exercise` currently doesn't do anything for LaTeX output. Index entries simply go missing from print. This is one way to address that, although I'm not sure if it's the right way.

### DIFF
--- a/xsl/mathbook-latex.xsl
+++ b/xsl/mathbook-latex.xsl
@@ -5036,6 +5036,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:text>\begin{inlineexercise}</xsl:text>
     <xsl:apply-templates select="." mode="block-options"/>
     <xsl:text>%&#xa;</xsl:text>
+    <xsl:apply-templates select="idx"/>
     <!-- Allow a webwork or myopenmath exercise to introduce/connect    -->
     <!-- a problem (especially from server) to the text in various ways -->
     <xsl:if test="webwork-reps|myopenmath">


rbeezer:
Yep, that is where I was going before I peeked at the commit.  Since we carefully parse the structure, we need to explicitly address `idx`.  I think another one needs to go at about Line 5557 of the LaTeX conversion.

I can look this afternoon, perhaps, and build on your PR.

rbeezer:
I think I recall the holdup on this.  Exercises get duplicated (e.g., for solutions in the back of the book), so we need to be sure not to make **two** or more index entries with the same label.  I think there are enough `original` flags moving around, but it'll take a bit more care...

kcrisman:
Yeah, I think that is one of the things I ran into when I was messing around with indexing stuff last year.  I did end up having to be very careful about entries in the exercises, and this syntax seems to have worked (see also http://math.gordon.edu/ntic/ntic2019/index-1.html and search for "pirates"):

```
    <exercise xml:id="exercise-linear-congruences-pirates">
      <idx><h>monkeys</h><see>pirates</see></idx>
      <idx>pirates</idx>
      <statement><p>Find a problem on the internet about pirates quarreling ove$
    </exercise>
```

Alex-Jordan:
So does that mean we are also losing `idx` within a `p` within an `exercise`?

rbeezer:
No, I think we might be getting it twice - in the original and in a duplicate, 
like in a hint in the back of the book.  I have a plan but it needs more 
elaborate testing before I'm confident about it.

kcrisman:
Ignore my earlier comment; the reason that code works is because I had some trial code I mentioned in #1121, but if this all gets resolved here that is fine. I closed that issue, but Rob can reopen it if he feels they are sufficiently different.

kcrisman:
Why did you close this only to open the new one?

Alex-Jordan:
@kcrisman This was closing a pull request. Then I opened an issue.

kcrisman:
Okay, that makes sense.  I guess we'll have to wait for @rbeezer then - presumably my solution is too naive.

PR #1204, opened 2019-11-22
Sample article "mdash" correction

mdoob:
There was a `<emdash>` in the sample article.


rbeezer:
Nicely done!  Thanks.

PR #1207, opened 2019-11-24
Showcase webwork

Alex-Jordan:
This adds a section to the Showcase Article with several WeBWorK problems.

In doing so, I wanted to add a few macro libraries to `extract-pg-common.xsl`, and while I was there I realized it could be shortened by 50 lines or so by using a named template. So I made that change.

Also, it came up that the answer to a WW problem can have newline characters, which leads to a small change in the mbx script.


Alex-Jordan:
I should have said that this is sort of independent from the "fillinarray" discussion. I've added "fillinarray" to the AIM -ptx WW server, and that makes it so that this showcase article section will work when you use the AIM -ptx server. But only that server, until "fillinarray" is settled and I can officially add it to WW.

Related, if you make the pdf for this, then what you currently see for the column vector answer array is just two fillin answer blanks, one after the other, horizontally. The "fillinarray" is passed through with no explicit template and the two "fillin"s within are processed.

rbeezer:
I've had this on a back burner while we sorted out security certificates.  Once we're certain about that, I'll bring this to a boil.

rbeezer:
OK, finally, some comments.  Sorry for the delay on this big contribution.

* I want public examples to work with public code.  When I build the website, I pull from public repositories, just to make sure none of my "experimental" stuff is used instead.  So I'd like the Showcase Article to build properly with the most recent general-release version of the WW server.  Can we save-off the fillin" bits for later.  I'll need to fiddle with website build script to get this integrated properly, which will require another short delay.
* Nice to see the parser clean-up!  But I don't want named templates to assume a context.  I believe that a modal template, in this case with a `match="webwork"`, will be stricter about `setup/pg-code` being correct within.  And no need for the leading dot, I think.  Looks like adjusting `name="pg-macros"` similarly may be the only necessary change.  (But I think there are others about if you wanted to do some wholesale clean-up.)
* I could merge the newline commit (once on a better position to test), and maybe the other two should go on their own PRs?

rbeezer:
Thanks, sorry this took so long.  Merged as-is.  Next step for me: expand website build script to do all the WW constructions.

PR #1208, opened 2019-11-24
Add option to use custom css style

oscarlevin:
Implemented as both a stringparam and element of publisher.xml options file.  String param is given preference (this might not be needed, but was good practice to make backwards compatible publisher.xml file options in the future).

I opted to include a new child of `publication/html` called `css` rather than including `@css-style` as an attribute of `publication/html`.  This seemed natural, but is certainly open to debate.  See publication.xml in sample-book.  As an example, I put `@colors` as another attribute of the `css` element, but this is not yet implemented (I could easily add this to the pull request, but it would take a small refactor of the html-css-colorfile xsl variable code).

### DIFF
--- a/examples/sample-book/publication.xml
+++ b/examples/sample-book/publication.xml
@@ -28,6 +28,11 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
         <!-- default behavior, but we do this as a test and as an  -->
         <!-- illustration                                          -->
         <index-page ref="sample-book"/>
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks, Oscar - that is perfect.  I think `html/css` is the right way to go.  Website update sometime later tonight.

I removed one line that had only been commented out.  Since we can always recove stuff from teh git history, I'm reluctant to save too much old code in comments (and it has messed me up a few times).

PR #1210, opened 2019-11-29
Add vi section to editors.xml

mdoob:


mdoob:
1. Adds new material about vi to editors.xml.
2. Deletes link.
3. Contains one asymptote image. Might need to use asy -f pdf and
   then convert to svg.

rbeezer:
Looks great.  Can you complete this by adding the SVG necessary for HTML output?  You can `git add images/foo.svg` and then commit that change, on the same local branch where everything else is.  Then push that branch to your fork again.  GitHub will know what to do.  It'll show up here.  When I merge I can combine the commits (don't try that now).  Thanks.  -Rob

mdoob:
I thought of leaving the pdf and svg files, but shouldn't those be generated by the user using
mbx? I left the comment about using asy for your benefit rather than for general information.
In some sense it would be a better check for you (as the user) to do this since it would
validate the asymptote code.

Having said all of this, I'll add the pdf and svg files to mathbook/doc/guide/images if
if makes things better.

Cheers,
Michael

rbeezer:
I don't rebuild images for the guide every time, the website build script is long enough as-is.  So images get built once stable and added into the repo.  The sample article is what I use for testing.  Thanks.

mdoob:
The extra files are added: /mathbook/doc/images/vimodes.{pdf,svg}.
Cheers,
Michael

rbeezer:
Thanks, Michael.  Nicely written.  I learned a few things.

Nits:  A pile of "blank" lines that were really 12 space characters.  Those can really make git crazy once they get loose.  I removed them all.

I deleted the link you recommended.  Just go ahead and do that - it'll be seen in the review.  If it needs more comment, do that in the discussion part of GitHub.  We can un-delete it if discussion indicates.

There was a merge commit.  Better for my workflow to leave the branch dangling.

I consolidated all the above down to one commit.  Merged and pushed.  Should go live in a few hours, at most.

Thanks for putting this together.

PR #1211, opened 2019-12-05
Wwpatch

Alex-Jordan:
Two commits here. One to remove that last `verify=False`. The other addresses the issue @jjrsylvestre encountered, which is a Python 3 issue. Tested the sample chapter. Both Python 2 and 3. Both to HTML and to PDF.


rbeezer:
Working well with the website build script.  Thanks again for getting this one squared away!  Live in a few minutes.

PR #1212, opened 2019-12-05
Add pubfile options for css colors, knowls, toc, and banner

oscarlevin:
Added variables for the other *_default.css files, that take attributes from the publisher xml file.


rbeezer:
@davidfarmer - I'm assuming the intent is that all four files can be replaced/set by the publisher?   I've not had my head in these, so just wanted to check.

davidfarmer:
Eventually all four can be replaced, perhaps independently,
but at the moment only one of then could have a non-default
version

rbeezer:
Thanks for the extensions, Oscar.

I'm going to be insistent that the publisher file is documented correctly.  So that will need to be done along with future additions.  See follow-on commit with an example - both narrative material and reference material, with index entries.

Thanks!

Issue #1136, opened 2019-07-02
regex issue with WeBWorK extraction


Alex-Jordan:
In #1135, @coltongrainger describes an issue with regex use in the webwork extraction. Look into this and see about making things better.

rbeezer:
I think #1213 must have addressed this.  If not, please reopen.

PR #1213, opened 2019-12-06
WeBWorK: refactor image file extraction

Alex-Jordan:
This might address #1136. I've tested that it does no harm to the common workflow. (On Python 2 and 3, it still makes the WW sample chapter PDF with the image files for me.)

It changes to only use regex to graph the image file path, not to do any further parsing. Further parsing is done by the urllib.parse module and the os.path module, which can be better trusted to work as intended.

@coltongrainger, may I ask you to test this? If I understand correctly, you would return to a situation where you are using a WW server in such a way as to get image URLs like: `http://127.0.0.1:8080/wwtmp/anonymous//gif/df899ba6-b336-33a5-9603-8b672b91ccc7___6bb688ad-bad2-32ef-ad88-fc9d44e14581.png` and see if the image extraction succeeds now.

I recommend not merging this until Colton or someone can test as described above.

### DIFF
--- a/script/mbx
+++ b/script/mbx
@@ -312,8 +312,10 @@ def webwork_to_xml(xslt_exec, ptx_xsl_dir, xml_source, server_params, dest_dir):
 
     if PY3:
         import urllib.parse
+        from urllib.parse import urlparse
     else:
         import urllib
+        from urlparse import urlparse
 
     # at least on Mac installations, requests module is not standard
     try:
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I rebased this to current `dev`. Then I extracted the sample chapter under 4 conditions: branches dev/wwpatch and python 2/3. The diffs on the webwork-extraction folders from the four runs are empty, except for timestamps. So I think that this does no harm.

Just looking at the code, I think it is mostly an improvement. The regex is much simpler. And the tearing down of the URL has some surgical precision now.

I am unable to test if this fixes the issue that was reported, but I can say that it ought to, because of the simpler regex and the better URL handling. I guess I recommend merging before you get too deep into mbx script refactoring that would impact this.

rbeezer:
1.  Thanks for the attention to this one before the Python 3 work.  Greatly appreciated.
2.  Current work rebased fine on top of this.  Thanks again.
2.  I think I got it in my head this was obsolete.  Sorry for the delay.
3.  Merged as-is.  Will update website sometime soon.

Issue #873, opened 2018-05-17
webwork title and idx


Alex-Jordan:
The schema says a `webwork` can have a `title` and `idx` children. I no longer see a need for either. We could remove from the schema and patrol the style sheets to confirm they do not employ these.

rbeezer:
Working on this one, and likely too hard to patrol just which `title` and `idx` might be in play.  So I made one of each in the sample chapter, with distinctive content, and then grep'ed for that.

* Both came through in the `authored` version in the extraction file.  No surprise there.
* The `title` came through in the `static` version in teh extraction file.  Maybe that suggests removing some code from the WW server?

Nothing in HTML, PDF, or problem sets.  So in any event, a reader will never see it.  So I think we are good.

Alex-Jordan:
#1214 addressed this.

PR #1214, opened 2019-12-07
WeBWorK: remove title from samples, mbx script

Alex-Jordan:
This follows through with #873, removing `title` and `idx` as schema-compliant children of a `webwork`. To go along with that, `title`s are removed from the sample-chapter and minimal webwork examples. And lastly, there was a section in the `mbx` script that was intentionally copying authored `title`s into the static representations, and this removes that.

I think I am doing the right thing, in only editing `schema/pretext.xml` in the schema folder. And I think that means you generate the other schema folder files before making the "real" commit for this.

### DIFF
--- a/examples/webwork/minimal/mini.xml
+++ b/examples/webwork/minimal/mini.xml
@@ -37,7 +37,6 @@
       <title><webwork /> Minimal Section</title>
       <exercise>
         <webwork>
-          <title>Integer Addition</title>
           <setup>
             <pg-code>
               $a = Compute(random(1, 9, 1));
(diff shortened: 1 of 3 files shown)


Alex-Jordan:
Now this PR removes the instances of `title` from the WW samples, and removes a bit of code from `mbx` that copied titles from `authored` into `static`. (Once in `static', nothing was being done with them anyway. A separate PR will remove `title` (and `idx`) from the schema for a WW.

rbeezer:
Done.  Thanks.

PR #1216, opened 2019-12-07
Add note about -xinclude flag to xsltproc in the quick start Processing section in the guide

kyp44:
After having some problems compiling my document that David Farmer converted from LaTeX, I added a note about the -xinclude flag in the Processing section of the quick start chapter in the guide. Adding this was recommended by David.


rbeezer:
Thanks, Dan.  Looks great.

I'm not at a good place right now to merge this in, but I'll get to it in a bit.

rbeezer:
Thanks again, Dan.  Merged and pushed, but may be a delay making it live.

Combined commits, made tabs into spaces, and lightly edited commit message.  So you'll want to trash your branch with this and pick up the new commit as part of a usual update.

Welcome aboard!

PR #1217, opened 2019-12-09
add environments section to showcase

oscarlevin:
This adds a new section to the showcase article, demonstrating the different environments pretext offers.  To illustrate how an environment can be renamed, "fact" is renamed in docinfo.

### DIFF
--- a/examples/showcase/environments.ptx
+++ b/examples/showcase/environments.ptx
@@ -1,0 +1,457 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<!-- This file is part of the PreTeXt Showcase article. -->
+<!--                                                    -->
+<!-- Copyright (©) 2019 The PreTeXt Organization.       -->
+<!-- See the README file for contributing guidelines.   -->
+
+<section xml:id="environments">
+  <title>Environments</title>
+  <author>Oscar Levin</author>
+  <introduction>
+    <p>
+      This section will demonstrate some of the <q>environments</q> available in <pretext />.  In addition to cataloging the options an author has to break up content, this will be a nice place to play with various styles that are under development. 
+    </p>
+    <p>
+      Maybe a better way to describe what this section will do is by specifying our objectives.  Luckily there is an environment for that.
+    </p>
+    <objectives>
+      <ul>
+        <li>
+          <p>
+            Illustrate the various environments (blocks) available in <pretext />.
+          </p>
+        </li>
+        <li>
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks, Oscar.  The master file conflicted with Alex's recent WW section, so I made this one the second section.  Going to take a bit to get the website build script to work properly with the WW stuff.

davidfarmer:
I agree.  I imagined the showcase article would look like a
random assemblage of good stuff from other documents.

Alex-Jordan:
I'll try to clarify in the README (which can evolve) and add a comment in the main source file for contributors to review the README. I'll put the comment where contributors are sure to see it, in the place where sections are included. Oscar, I hope this doesn't sound like criticism. Thanks for this section, and only if/when you have time, it could be revised. Since I may be in there in the very near future, I may go ahead and edit to fix the schema violation, but nothing else.

oscarlevin:
I think I agree in principle, except where you say "shouldn't" I might say "needn't".    In particular, for the section we are discussing, I thought it was important to create a document with examples of as many environments as possible, and just found that writing about PreTeXt was easier than writing about graph theory.  I also figured it wouldn't hurt to let readers know what the environments were (although certainly this shouldn't be a requirement of the document).  

With regards to styling, I thought the purpose of the sample article was to show off how the features of PreTeXt look, so having something that could be styled would be interesting.  I envisioned that eventually, when we get a few distinct styles, we would compile a few different versions of the sample article to show these off.  So not for testing styling, but for showing it off.

Absolutely it should be schema compliant; that was an oversight.  Feel free to fix if it is convenient.  If others have a strong distaste for the meta-ness of the chapter, I could try to scale that back at some point.

rbeezer:
Thanks, Alex, for raising this.  I think the README reflects my intention on this one, but feel free to amplify and add source code comments - those changes would be welcome.

*  I intend for this to be an advertisement for PreTeXt, showing what can be accomplished, but without getting into edge cases unnecessarily or pushing to the breaking point or trying to be overly comprehensive.
*  In particular, I think it should be instructive *only* if an author compares output with source.
*  The narrative should be realistic, but does not need to be consistent.  A definition about fiber bundles can be followed by a theorem about Hamiltonian paths.
*  Long-term, I think we should apply *some* comprehensive exemplar of styling, so long as it does not obscure the other purposes (but rather, complements them).

Thanks for everybody's work on this one.

PR #1219, opened 2019-12-09
WeBWorK: remove title and idx from schema

Alex-Jordan:
As mentioned in #1214, this removes `title` and `idx` from the WW portion of the schema.


rbeezer:
Performed the conversions to the other formats, and fixed into your commit.

Online schema browser build is busted (dang Java) so that'll need repair (soon, not now).

PR #1222, opened 2019-12-11
Sample Article: Correct src url for threejs example

drewyoungren:
Seems they moves some source for that example to a subdirectory.


rbeezer:
Aah-ha.  Thanks!  I'll catch this in a bit.

rbeezer:
Thanks, Drew!  Change pushed.  Trash your local branch, I rebased it and edited the commit message.  It'll be a few days before I can get the website examples rebuilt properly.

PR #1223, opened 2019-12-11
Fix <kbd> to work identically in LaTeX and html

mdoob:
For mathbook-common.xsl:

1. Filled out entries in "kbdkey-rtf" to include the three reserved character
   for xml and the problematic LaTeX characters. 
      a. The ampersand, less than and greater than seem like a really good idea.
         The others are less essential but might be nice.
      b. The unicode settings for openbrace and closebrace are incorrect (the
         values of &#007C should be &#007B and &#007D respectively). The correct
         values need to be escaped so that they won't throw an error. I'm not
         sure of the best way to do this.
      c. The sample file has a table of the entries displayed from both literal
         and named entries.

For mathbook-latex.xsl:

1. The kbd template has been modified to use escape-text-to-latex.
   This should be checked (think of me as a xslt noobie). What I
   have written works, but is, perhaps, suboptimal.

2. The LaTeX kbd macro now has an extra set of braces in its definition.
   This is because of one character: the comma. It is a separation character
   in the menukeys package, and so with no change the comma is eaten up and 
   doesn't appear on the key; the addition of extra braces saves it.
      a. The menukeys package will take \keys{a,b} call and expand it to 
         \keys(a}+\keys{b} to indicate the simultaneous pressing of keys.
         My additional braces kills this expansion. I consider this good since
         the expansion is anithetical to our philosophy of separating markup and 
         presentation. Surely it would be better to have something like
         ``<simulkeys>
             <kbd>a</kbd>
             <kbd>b</kbd>
         </simulkeys>``
      b. The use of the plus sign to indicate simultaneous pressing of keys
         is really common. We surely don't want this built in. There are
         other opitons that could make sense. The could be a horizontal list
         of keys with simultaeous one stacked, or simultaneous keys could
         be put into an fbox. I would think it better to kick the presentation
         over to David to include in the css files. 
      c. The sample file has tables for all the keyboard keys (upper and lower case).
         It works for both html and latex. 
[kbdsamples.xml.txt](https://github.com/rbeezer/mathbook/files/3952564/kbdsamples.xml.txt)

### DIFF
--- a/xsl/mathbook-common.xsl
+++ b/xsl/mathbook-common.xsl
@@ -4542,7 +4542,7 @@ Neither: A structural node that is simply a (visual) subdivision of a chunk
 <!-- ############# -->
 
 <!-- Comments are Unicode names, from fileformat.info            -->
-<!-- @latex values are macros in teh menukeys package specifying -->
+<!-- @latex values are macros in the menukeys package specifying -->
 <!-- keyboard keys that are typically labeled with graphics      -->
 <xsl:variable name="kbdkey-rtf">
     <kbdkeyinfo name="left"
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks, Michael.  I'll get the samples integrated into the test article, as part of testing.  Needs to be a few days before I can sit down with this.

One question:  can `\char"26` be `\&amp;`?  I'd like to have the LaTeX be free of more technical constructions that a casual user might wonder about.

mdoob:
No problem with the change to \&amp;. The same thing has already been done with
lessthan and greaterthan.

rbeezer:
Sorry this took me a while to get back to, but it needed a bit more work than just merging.  Everything below has your name on it, in two commits.  Merged and pushed, and hoping to update website tonight.

* The braces were being confused for part of an XSL "Abstract Value Template" (AVT).  Solution: double each.  I learned something.
* Some trailing whitespace to clean up, and a few tabs replaced by spaces.
* I put the tables into the sample article.  Once they settle down, maybe they need to be part of The Guide.

I renamed a few, mostly to use internal names, and element names of some now-deprecated empty elements.  Cosmetic.

* `lessthan` -> `less`
* `greaterthan` -> `greater`
* `octothorpe` -> `hash`
* `tie` -> `tilde`

Thanks for the extra braces to protect the comma, that might have stumped me!

(Dang - forgot to put a PR # on the commits!  See f3fab2444a0c644f32bf900281aff7591c90ed92 and 879a577f5bceeaf01bec220d874116bea85b0514.)

PR #1224, opened 2019-12-12
Javascript013

davidfarmer:
I needed a new Javascript version for the new reading questions interface.

Since I was in the neighborhood, I added some index entries to the sample article.


rbeezer:
Pushed as-is, but it'll be 48 hours until website is rebuilt.

PR #1226, opened 2019-12-15
LaTeX: move into publication file: font-size, geometry, preamble-earl…

Alex-Jordan:
…y, preamble-late

I'm putting together a sample syllabus. Long story short, it led to these things. Swapping out `latex.font.size`, `latex.geometry`, `latex.preamble.early`, `latex.preamble.late` for `publication/latex/font-size`, `publication/latex/geometry`, `publication/latex/preamble-early`, `publication/latex/preamble-late`.

The intention is the stringparams are still functional, but deprecated. Limited testing bears that out.

### DIFF
--- a/xsl/extract-latex-image.xsl
+++ b/xsl/extract-latex-image.xsl
@@ -47,16 +47,49 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- job of ensuring they remain in-sync, please      -->
 <!-- coordinate the two sets of templates by hand     -->
 
-<!-- ######################################### -->
+<!-- DEPRECATED: 2019-12-14, do not use, any value -->
+<!-- besides an empty string will raise a warning  -->
+<xsl:param name="latex.font.size" select="''" />
+<xsl:param name="latex.geometry" select="''"/>
+
+
+
+
+<!-- ######################### -->
+<!-- LaTeX-Publication Options -->
+<!-- ######################### -->
+
+<!-- The $publication variable comes from -common and is the result -->
+<!-- of a command-line string parameter pointing to an XML file of  -->
+<!-- various options.                                               -->
+<!-- Elements and attributes of this file are meant to influence    -->
+<!-- decisions taken *after* an author is completed writing.  In    -->
+<!-- limited cases a command-line string parameter may be used to   -->
+<!-- override these settings (especially for testing purposes).     -->
+<!-- In other cases, deprecated string parameters may be consulted  -->
+<!-- secondarily, for a limited time.                               -->
+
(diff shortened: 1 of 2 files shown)


rbeezer:
Dear Alex,

I've not given too much thought to the LaTeX portion of the publisher file.

*  I want to be more deliberate about this than with `docinfo` and stringparam.  For example, does the font size need `pt` suffix on all of them?
*  I'm pretty sure I want `\geometry{}` commands to be part of the page-style templates in the LaTeX styling scheme.
*  It'd be nice if `latex.preamble.*` was not abused.  So maybe de-emphasized, I don't know.
*  I'm pretty sure I want font size to be part of style.

It is a publisher's job to decide on style, so a style file will be a publisher's choice.  I can only think (right now) of one-sided/two-sided and print/electronic as publisher file items.

There is a lot of work here.  Maybe we can transition it easily?  Maybe publisher file vs. style file needs discussion on `pretext-dev`?

Alex-Jordan:
What motivated this is that I'm about ready with a sample syllabus "article". If it comes out in regular article class, the margins are too wide for a syllabus. So I'm left thinking how will I add this to the repo complete with instructions on how to use it? Options that occurred to me:

1. The `README` includes a long `xsltproc` with the right `stringparam`.
2. Include a local layer xsl file with the `stringparam`s hard-coded.
3. Include a production script or Makefile.
4. Get the important pieces into the publication file.

Numbers 2, 3 are bad for at least two reasons. One reason is they promote customization. The other is that it complicates instructions since the user would have to get their paths straightened out first.

I went with 4. I could go with 1. It just seemed a shame since these `stringparam`s are supposed to be going away. And updating instructions will be one more thing to take care of when they do go away.

What do you think?

rbeezer:
More to consider, but let's discuss margins first, since I think they cover the principles at hand.

If default should be skinny margins, then maybe you are building a new PTX type:  book, article, memo, letter, syllabus.  Then there should also be an HTML realization.

Or you are customizing an article with a new style.  Then style an article.  That's an XSL layer.  Make one everybody can use.  It doesn't look any different on the LaTeX side:
```bash
xsltproc xsl/pretext-latex-syllabus.xsl alex/syllabi/2020-spring-math-8945.ptx
```

I definitely think `\geometry{}` commands are part of the "page style" and are a style item.  The package is loaded already.  Study the templates with:
```xsl
mode="titleps-style"
```

We can work through the other options, perhaps deciding between style and publisher and de-emphasized.

Alex-Jordan:
A new PTX type doesn't feel like a good balance between work required and priorities, so I just sent Sandi some files directly (as you probably saw, cc'd). It was good for me to see what will be involved with any other `publication` items. This branch will always be here for reference, and I can close the PR.

PR #1227, opened 2019-12-16
Add javascript for instructor features

davidfarmer:
Most of the code is in place for implementing the HTML instructor version.

This pull request adds a line that reads in that code.

Inserting the instructor resources into the HTML will be explained in an upcoming
post to pretext-dev.


davidfarmer:
Undocumented feature: login as "instructor" with password "instructor" to be a guest instructor.

PR #1239, opened 2020-01-21
goggle->google typo fixes

jjrsylvestre:
Found some places in the publisher's guide where google was written "goggle."

Also found an instance of "goggle" inside mathbook-html.xsl.


rbeezer:
Thanks, Jeremy!  That's a common mistake of mine.  :-(

Merged, will update website soon.

PR #1241, opened 2020-01-25
lots of beamer fixes

oscarlevin:
A handful of improvements to the pretext-beamer stylesheet.  Most of these will be non-controversial: added to preamble to allow images, sidebyside, etc; added example environment, allowed for section or no sections, better handling of title page information, maybe some others.

Some style choices I made: improved font for math, removed navigation buttons, added space between paragraphs with \pause between them.

Still more work to do (in particular, the sample-slideshow doesn't work because of some sage stuff there).  Should work okay for simple slideshows though.

### DIFF
--- a/xsl/pretext-beamer.xsl
+++ b/xsl/pretext-beamer.xsl
@@ -1,5 +1,24 @@
 <?xml version='1.0'?>
 
+<!--********************************************************************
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, Oscar!  Corrected one mis-spelling of `@encoding` that happened to jump out at me.

Merged and pushed with zero testing.  ;-)  If anything breaks, we/I can update/correct quickly.

Issue #1048, opened 2019-04-30
Index headnote


rbeezer:
The arrays in an index may be preceded by a "headnote".

*  An `introduction` serves this purpose.  Rename?  Remove the companion `conclusion`?
* Introduction does not seem to be implemented in LaTeX.  Our fault or a pure-LaTeX shortcoming?

davidfarmer:
According to the "Advice on Indexing" in the Author's Guide, the headnote is only appropriate
if "the index has some unusual features".

If it is unlikely that the LaTeX index would have an unusual feature (due to lack of options),
then maybe it is reasonable to have the `introduction` become the headnote in the HTML
and be ignored in the LaTeX?

Alex-Jordan:
What does our index do with things like:
$\chi$-squared test
?

You can Latinize and order by "chi..." or you could put the Greek letters at the start or the end in their natural order. Or maybe something else? If Greek letters and/or other math-specific terms that don't start with a Latin letter are indexed, what is the plan? And perhaps an explanation in the headnote is appropriate.

rbeezer:
@sortby="chi"

davidfarmer:
> I know about @sortby. What I mean is, should an author always (only) use Latin spelling for such things? If an index term
> starts with \nabla, should it appear with "@sortby=nabla..."? With "@sortby=gradient..." I think both, with
> "@see="gradient...". But what about a reader who has no idea how to pronounce it? Just let them scan until they find it?

If there are a lot of symbols in the index, then there should be a place
(often at the beginning) where all the symbols appear.  The specific area
of mathematics should guide you to the right order to put those symbols.
If the spaces L^1, L^2, L^p, and L^\infty all appear, then they probably
should all be together, in that order.  If the little-l spaces also
appear, then presumably those go immediately before the big-L spaces.
(In particular, not alphabetized under \ell, so you have to over-ride the
alphabetical order coming from the LaTeX markup.) (May also need to
over-ride the default of capital letters coming before lower case.)

I think the key point is that the index requires thoughtful work, after
the text is mostly complete.  Without knowing how long the list of symbols
is, and what is the relationship between the symbols, you can't decide
on a good order for them.

And for the alphabetical part of the index, there is nothing wrong with
having the upside-down triangle appear in the location of 'grad...' and
also 'nabla...'.  As you note, a judicious "see" will provide additional
information to the reader.

rbeezer:
Since this issue has been hijacked, I'll pile on.

[...]

PR #1242, opened 2020-01-27
Showcase exercises

Alex-Jordan:
Main commit here adds an exercises section to the Showcase Article. Two other commits clear trailing whitespace and schema violations from the environments section.

The Showcase Article currently has these schema violations, which may just be signs that the schema is ready to catch up:

- It doesn't like `index` having an `introduction`.
- It doesn't recognize `document-id` and `html` from `docinfo`.
- It doesn't recognize `reading-questions`.
- It doesn't recognize `subexercises`.

The latter two are introduced by this PR. Unsure if you'd want to wait to merge this until the schema has some of these things.

### DIFF
--- a/examples/showcase/environments.ptx
+++ b/examples/showcase/environments.ptx
@@ -25,7 +25,7 @@
 
   <introduction>
     <p>
-      This section will demonstrate some of the <q>environments</q> available in <pretext />.  In addition to cataloging the options an author has to break up content, this will be a nice place to play with various styles that are under development. 
+      This section will demonstrate some of the <q>environments</q> available in <pretext />.  In addition to cataloging the options an author has to break up content, this will be a nice place to play with various styles that are under development.
     </p>
     <p>
       Maybe a better way to describe what this section will do is by specifying our objectives.  Luckily there is an environment for that.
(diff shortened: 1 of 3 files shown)


rbeezer:
Thanks!  The Sylow theorem topic is great!

* I added titles to the exanple and checkpoints that render as knowls.  I think it is good to model that knowled content works better if there is a title.  If you don't like them (I didn't think too hard) feel free to change them.  I figured it was easiest to just get this in.
* Merged without changes since the commits were all perfect.
* My fault for not noticing, but it seems this branch was built off some commit from August?  If possible, rebasing to something more current would help.  I just don't like having real complicated history with real merges going back many commits.  ;-)
* I'll see what I can accomplish with the schema, between now and the end of Wednesday.
* Except `docinfo` - some of that will migrate to the publisher file, so I'm not going to clean that up until later.  Don't let that stop you from more work on the SA!

Alex-Jordan:
Alright, thanks. Good call on titles. Sorry about the old launch point, although how could be that old if it has Oscar's environments section from roughly December? But I'm still migrating from old laptop to new laptop, and I probably got confused which one I had just updated.

rbeezer:
> how could it be that old

Hmm, not sure.  Oh, I got it.  Branch is only from Jan 1 (but still many commits back).  Right after that are braille commits that I had on a branch for a **long** time, and they carry August dates, but were only comitted over winter break.  Sorry for the mild confusion.

rbeezer:
*  `subexercises` are now in the schema, as discussed
*  `reading-questions` will be part of a bigger schema project
*  `index/introduction` will be obsoleted by #1048
*  `docinfo` will just need to wait for gradual introduction of publisher switches

PR #1247, opened 2020-02-03
Add small section on editing etiquette

mdoob:
Better editing without tab and end of line spaces (Doctor, heal thyself).

### DIFF
--- a/doc/guide/appendices/editors.xml
+++ b/doc/guide/appendices/editors.xml
@@ -392,7 +392,7 @@
                 Usually vi starts in normal mode.
             </li>
             <li> <em>Insert mode: </em> This mode is for inserting new text into the file.
-                Typing <c>The galloping gertie leapt into the air</c> causes that text
+                Typing <c>Galloping Gertie leapt into the air.</c> causes that text
                 to be inserted at the current position in the file.
             </li>
             <li> <em>Command mode: </em> This mode is to execute commands.
@@ -428,7 +428,7 @@
             <p>
             <xref ref="vimodesfigure" />  show the keys used to move between modes.
             </p>
-            
+
             <figure xml:id="vimodesfigure">
             <caption>Keys for moving between different modes in vi</caption>
                 <image xml:id="vimodes">
@@ -531,6 +531,10 @@
                     <cell>Set the display size of each line</cell>
                 </row>
                 <row>
+                    <cell right="minor">:set list</cell>
+                    <cell>Show tabs and carraige returns </cell>
+                </row>
+                <row>
                     <cell right="minor">:set tabstops<m>=</m>n</cell>
                     <cell><kbd>Tab</kbd> inserts <m>n</m> spaces </cell>
                 </row>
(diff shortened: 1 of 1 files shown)


davidfarmer:
In the discussion about removing end of line spaces, you could point out that
`
:1,$s/ $//
`
will remove an end of line space.  Doing that repeatedly will remove all end of
line spaces.

rbeezer:
Thanks, Michael!  Looks grreat.  I'll let this sit while you consider David's comment.  But not as long as I delayed on the last PR!

mdoob:
Thanks for adding to the discussion, David. Even better would be
the one-shot solution:
`:%s/\s\+$//` 
which deletes multiple spaces before the EOL for every
line of the file.

I have avoided even hinting at regular expressions as being 
a bridge too far for this discussion. I think the visual cues from 
`:set list`  plus the searching hints are enough for a user needing
this introduction.

In passing: (1) You'll be amused to know that this is a case of
"physician, heal thyself", as Rob as has been polite enough not
to point out. (2) I would have never discovered the bug in the
keyboard macro if I hadn't noticed that `<kbd>$</kbd>` 
gave different output for for the html and pdf versions.

rbeezer:
@davidfarmer - would you like to write a short section for Appendix C about regular expressions?  You could include the right `vi` commands and discuss `grep` and `sed`.

I use grep **all the time** on PTX source to locate problems, check for problems, count elements (how many exercises), etc.  And in SublimeTeXt you can do sed-like `s/foo/bar/` with regular expressions via the standard interface with a one-button click to turn on regex syntax.

We could retitle Appendix C "Text Editing, Spell Check"

I'd collaborate on such a write-up.  We could try a jointly edited PR? (I need to figure that one out.)

rbeezer:
> as Rob as has been polite enough not to point out

Those who live in glass houses...

I **did** point it out.  I just didn't yell about it.  ;-)  Thanks for the additions here.

rbeezer:
Merged.  Website output will go up on next refresh.  Thanks again.

davidfarmer:
I would be happy to add something about regular expressions.
I will submit an issue.

mdoob:
Regular expressions is a huge  and esoteric topic. I come across stuff that can't understand all the time. I'm happy to contribute, but exactly what do you envision as the appropriate topics?

Alex-Jordan:
A nice thing might just be a short list of useful regex for editing PTX files. Like something to remove trailing whitespace, something to replace tabs, etc.

Could probably crowdsource a few other useful examples.

Watch out for different implementations though. Like sometimes `+` means 1 or more, and sometimes you need to escape and use `\+`. Depends on the program/utility.

davidfarmer:
That is why, when I submit my pull request, for vi I will
recommend
`
1,$s/ $//
`
as the way to remove trailing spaces.  Apply repeatedly until
nothing changes.

A few prepackaged examples will be sufficient for the Guide.
People can go elsewhere to learn more.

rbeezer:
An even really simple stuff like
```bash
grep -R "<tikz" * --color=auto
```
can be really useful when chasing down things like deprecations.  No syntax-confusion in that one, I think.

Or
```bash
grep -R "<exercise" * | wc
```
which produces 7,000 for ORCCA.

Yes, keep it light and folks can go elsewhere for details.  But PTX-specific tasks is best.

I really like online testing sites like  https://regex101.com/  Maybe a reference to a better one would be useful.

sean-fitzpatrick:
My only concern with documenting regex is that there are subtle changes from one editor to the next.
I thought I finally had it figured out and then I switched from Sublime to Atom and had to figure it out all over again. (And my LaTeX editor has its own regex that is completely mysterious and I eventually gave up on it.)

mdoob:
> That is why, when I submit my pull request, for vi I will recommend ` 1,$s/ $// ` as the way to remove trailing spaces. Apply repeatedly until nothing changes. A few prepackaged examples will be sufficient for the Guide. People can go elsewhere to learn more.

Are we still talking about vi or more general regular expressions? 

If a user in vi is working with the default settings, a tab will expand to four spaces,
and if that user has an xml file with lines having three indents, they will be 
replaced by 12 spaces. A blank line will then have 12 spaces in front of the EOL character.
I don't think a typical user is willing run a regular expression 12 times.  

In addition, it is conceivable that after some editing a line will emerge ending in
`<space><tab><space><eol>` and the space deletion procedure won't work. 
In other words, the expansion of tabs and 
the deletion of trailing spaces are now noncommutative.

If we are actually talking about vi here, then we can use what is in vim
(since every distribution uses that now) to get:
`%s/\s\+$\\`.

`\s` matches both spaces and tabs and `\+` means one or more occurrences.
Honestly, in the vi world this is really standard. In addition, you need run it one
time only.

davidfarmer:
Actually, in my vi (which has no special setting for tab),
a tab is a literal Tab and has the same width as 8 spaces.
So, I never type a Tab in vi.

The point of the vi command I will put in the docs is that
it is perfectly adequate for catching the occasional trailing
space, and avoids using syntax which is ambiguous or
confusing in terms of different ways to write a regular expression.

Anyone who types Tabs and has them expand to spaces can handle
figuring out how to fix the problem they created.  We can focus
on the naive user who is unaware of the power of the tools available.

PR #1253, opened 2020-02-20
Add support for abstracts in pretext/slideshow.

### LINKED ISSUE #1252: PreTeXt slideshow needs to support abstracts
Right now putting an `<abstract/>` in the `<frontmatter/>` of a PreTeXt slideshow doesn't do anything. It should create a slide below the titlepage slide with the abstract.

I hope to submit a PR to address this later this week.

StevenClontz:
Closes #1252

### DIFF
--- a/doc/guide/author/topics.xml
+++ b/doc/guide/author/topics.xml
@@ -2598,6 +2598,9 @@ Following is a somewhat minimal working example.
         &lt;event&gt;PreTeXt Workshop 20XX&lt;/event&gt;
         &lt;date&gt;Month 20XX&lt;/date&gt;
       &lt;/titlepage&gt;
+      &lt;abstract&gt;
+        &lt;Your wonderful abstract.&/lt;
+      &lt;/abstract&gt;
     &lt;/frontmatter&gt;
     &lt;section&gt;
       &lt;title&gt;Title of Section&lt;/title&gt;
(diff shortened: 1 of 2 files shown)


rbeezer:
Looks good, visually.  I'll test and merge shortly.  A couple suggestions, which I can implement while testing.
* I suspect the new mode is not necessary, and just leftover from some previous code copied from elsewhere?
* The `<h3>Abstract</h3>` will (I think) dump a couple newlines and lots of whitespace into the HTML?  We could (a) tighten it up, or (b) wrap it in `<xsl:text>` to preserve the look in the XSL.
* We'll want an abstract in the sample slideshow.  Adding examples as features are added is critical for informative testing.

StevenClontz:
- Yeah I was using CDD (copypaste-driven development) for this feature, which was modified from Andrew's work. Please feel free to improve/simplify.
- I wrote the HTML output the way I'd write it personally if I was authoring HTML directly, but please feel free to improve the styling to better match PreTeXt conventions.
- I'll add an abstract to the documentation once this is merged.

StevenClontz:
Actually I'll push an abstract to the documentation in a second so it'll be included with this PR.

rbeezer:
*  I added a new commit to put a small (two-paragraph) `abstract` into the sample slideshow, for effective testing.
*  And it ended up being centered, which I thought looked weird.  So added a `@align="left"`.  Holler if you think it should be otherwise.
*  But removing the mode had no adverse effect, so I took that out.  (I still love CDD!)
*  The escaped angle brackets in the sample in The Guide were a bit twisted up, so I corrected those.
*  Smashed the two commits and minor changes all together.

Thanks for the contribution!  Appreciate the improvements to the XSL.

PR #1255, opened 2020-03-08
revealJs simplistic support for <term>s

StevenClontz:
Slideshow terms are wrapped in `<b/>` when converted to RevealJs.

### DIFF
--- a/xsl/pretext-revealjs.xsl
+++ b/xsl/pretext-revealjs.xsl
@@ -378,6 +378,10 @@ Reveal.initialize({
   </dl>
 </xsl:template>
 
+<xsl:template match="term">
+  <b class="term"><xsl:apply-templates/></b>
+</xsl:template>
+
 
 <xsl:template match="ul/li|ol/li">
   <li>


rbeezer:
This made me wonder if we were not importing the HTML conversion?  I see from #1257 we are.  Can you check that `term` produces `dfn.terminology` in the output?  If so, (1) search HTML converter for `<dfn` or `="term"`  to increase your PTX/XSL knowledge, and (2) hack some in-page CSS style do the right thing?  Eventually we should import stock PTX CSS and override the necessary parts.  I'm doing this with EPUB right now, going very well, based on work by @davidfarmer .

It'd be great to have a new slide in the sample slideshow which tested some of this.  Maybe just an `ol` with a bunch of nonsense sentences with a lot of "intra-paragraph" sort of stuff like this.  See Author's Guide for ideas.  Might discover more situations - which could take a quick fix, or wait for comprehensive CSS.  I can send experimental ideas for pulling in stock CSS.

StevenClontz:
Yeah it was already wrapping terms in `<dfn/>` tags, so I'm closing this and doing the right thing by injecting CSS styling for that tag.

PR #1256, opened 2020-03-09
Add frontmatter/backmatter classes to TOC

Alex-Jordan:
This puts a class onto the TOC list item for the frontmatter and backmatter, so that they may be styled. 

Personally, I imagine styling the backmatter item in the same manner as a "part", but that can be discussed separately once there is a class to play with.

For reference, I don't love how "Backmatter", "Index", and "Colophon" appear to be peers here in the [Showcase Article](https://pretextbook.org/examples/showcase/html/frontmatter.html). And similarly with other projects.

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -10483,9 +10483,16 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     <xsl:apply-templates select="." mode="url"/>
                </xsl:variable>
                <!-- text of anchor's class, active if a match, otherwise plain -->
-               <!-- Based on node-set union size, "part" is for styling        -->
+               <!-- Based on node-set union size; "frontmatter", "backmatter", -->
+               <!-- "part" are for styling                                     -->
                <xsl:variable name="class">
                     <xsl:text>link</xsl:text>
+                    <xsl:if test="self::frontmatter">
+                        <xsl:text> frontmatter</xsl:text>
+                    </xsl:if>
+                    <xsl:if test="self::backmatter">
+                        <xsl:text> backmatter</xsl:text>
+                    </xsl:if>
                     <xsl:if test="self::part">
                         <xsl:text> part</xsl:text>
                     </xsl:if>


rbeezer:
I'd like @davidfarmer to check-off on this one.  There was recently a very swift reaction when I gave the mistaken impression I was adding HTML structure.  ;-)  (Yes, I know these are **just** classes.)

davidfarmer:
I am okay with this:  new classes are no problem.

Note that these new classes will have no immediate effect.  Suggestions for how
to change the styling are welcome.

Alex-Jordan:
> Suggestions for how to change the styling are welcome.

I suggest styling these (at least backmatter) just like `part` is styled. Yes, even when the book/article has no parts.

If you ignore the custom coloring job I did, you can see what I mean with [ORCCA](https://spot.pcc.edu/math/orcca/ed2/html/frontmatter.html). Actually I only did this with the backmatter (titled "Appendices"). With backmatter it is more clear (to me, at least) that this would be good. You have `appendix` elements and peers, and each `appendix` can have `section`s. So it is natural to style each appendix as a chapter and each section as a section. And that leads to the parent (the `backmatter`) being like a part.

With frontmatter, you have `preface` and peers. `preface` is being treated on the same level as `section`, as far as styling goes. Either

1. that doesn't change, or
2. upgrade `preface` and peers to be styled like chapters

I think item 2 might make the frontmatter appear too "heavy", so I'm on the fence about styling the `frontmatter` as a part.

But as long as I was going to advocate for this change to backmatter styling, it seemed wrong not to give frontmatter a corresponding class too.

rbeezer:
Merged.  Will update website later tonight, and then the different versions of the sample book might be good for tests (w/, w/o parts, etc.).

* I might have written the three (obviously) mutually exclusive `if` as a `choose/when` with an empty `otherwise`.  Here it makes little difference for readability and little difference (I guess) for execution.
* Can we put the styling discussion onto `pretext-dev` since it will affect folks builds immediately?  The advisability of having these classes is fine here, but styling changes like this perhaps deserve a more public exposure (even if it is just the three of us talking).  You could just cut/paste the above pretty much.  Assuming that is OK, I'm going to close this.

Alex-Jordan:
Will do. 

I did use `choose` at first. That makes the code two lines longer, assuming the same line breaking with `when`s as with `if`s. Since it seemed like 6 and one half dozen for readability and execution, I went with what made the code shorter.

PR #1258, opened 2020-03-09
WeBWorK: support vars containing PGML syntax

Alex-Jordan:
Every so often, you are coding the setup portion of a WW problem, and you need to store something like:

``$solution = "The answer is [`\frac{1}{2}`].";``

Then in the text (`statement`, `solution`, etc.) you need to call on that variable:

`<var name="$solution"/>`

As things stand, you will literally get ``The answer is [`\frac{1}{2}`].`` in the output that the end user sees. But what we really wanted is PGML to actually process that PGML syntax with the bracw-backtick pairs. This PR makes it possible for that to happen. The `var` needs to be called on like:

`<var name="$solution" pgml="yes"/>`

where `@pgml` is indicative of the `var` being a string of PGML syntax.

Note that my example above is simple to communicate the idea, but too simple to actually need this device. This device is needed when there is randomized math content and regular text that changes significantly based on the problem's random seed.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -663,6 +663,30 @@
                 </webwork>
             </exercise>
 
+            <p>
+                Occasionally (and probably very rarely) your PG code will include a string variable where the content is PGML syntax.
+                If this is put into the exercise statement, it will be inserted <emph>after</emph> PGML processing is done,
+                and therefore it won't render as you might expect. To insert this content and also have it processed as PGML,
+                use <c>data="pgml"</c>. (This is not in the <pretext/> schema as of 3/26/2020, and is subject to change.)
+            </p>
+
+            <exercise>
+                <webwork>
+                    <setup>
+                        <pg-code>
+                            $statement = 'some PGML math: [`\frac{1}{2}+\frac{3}{2}=2`]; and some *bold text*';
+                        </pg-code>
+                    </setup>
+                    <statement>
+                        <p>
+                            <var name="$statement"/> makes:
+                        </p>
+                        <p>
+                            <var name="$statement" data="pgml"/>
+                        </p>
+                    </statement>
+                </webwork>
+            </exercise>
         </section>
 
         <section>
--- a/xsl/extract-pg-common.xsl
+++ b/xsl/extract-pg-common.xsl
@@ -989,6 +989,10 @@
         <xsl:text>->correct_ans()</xsl:text>
     </xsl:if>
     <xsl:text>]</xsl:text>
+    <!-- if the variable is a string of PGML syntax to be processed -->
+    <xsl:if test="@data='pgml'">
+        <xsl:text>**</xsl:text>
+    </xsl:if>
 </xsl:template>
 
 <!-- An image description may depend on the value of a simple scalar var   -->


rbeezer:
Thanks, Alex!  Even in the schema.

Let's imagine `var` is general-purpose, like for other methods/syntax of randomization (I'm thinkinking of Brian Walton's Javascript from recent-ish AIM workshop).  `yes|no` attributes are easy to outgrow and then we have a mess on our hands with deprecations and/or additions.

Would it be feasible/logical to have something more generic like
```xml
<var language="pgml"/>
```
along with some explicit default/no-op (like `none`)?

Alex-Jordan:
OK, generalizing sounds good. Can we think of a better attribute name than "language"? My quibble with it is that the variable is a variable in Perl, not PGML. So it's a Perl variable storing a string that is meant to be processed as a line of PGML markup.

How about `content="pgml"`? Or something along those lines?

Alex-Jordan:
I just force pushed this with two changes.

1. Now uses `content="pgml"`.

1. And there is an example of this in the sample chapter.

rbeezer:
Sorry, behind on this one.  `@language` was not proposed with much thought.  `@content` will confuse me since "content" has a very precise meaning when describing parts of XML (stuff between the tags).  (A lot of things got clearer for me when I realized it was not a generic term.)

`@syntax`, `@code`?  Something whose values describe what sort of processing is necessary on the variable's value?

Alex-Jordan:
This feels like it should be easy, but it's not (for me anyway, right now). An _abstract_ `var` is first of all, of some data type. The data type options depend on the language. For example Perl has `scalar`, `array`, `hash`, and `code`. And then `scalar` can be interpreted in one of three "contexts": `string`, `number`, or `reference`. Now think for a moment about other abstract `var`s in other languages and all the data types they permit.

Right now we are trying to say this thing contains a string of PGML. In the Perl arena, it's a `scalar` intended to be read in `string` context. That's pretty specific.

I pause with `@syntax` or `@code` because they sound like they only make sense for _string_-type variables. A string of syntax or a string of code. If the goal is abstraction, it doesn't abstract far enough. So I pause, because I wonder if there is an attribute name that could also be used to tell you that the `var` is an integer. Or a floating point rational. Or that it is a 2D array of booleans. Etc.

I don't love it, but how do you feel about `@data`? `@data="pgml"`?

rbeezer:
Let's run with @data for now.  Keep it undocumented and unannounced for a while.

With experience, you might settle on something different?  Then we (you!) can 
make changes without any widespread pain?  Or if @data sticks, we can announce 
and document that.

I can merge this afternoon.  I have not looked - let me know if you think it 
would be easy for me to make the (cosmetic) changes, should that be easier for you?

Alex-Jordan:
Hmm, it looks like GitHub thought you were tagging a user named DaTa. I presume the camelcase is not intended? I rebased, changed to "data", and am about to force push.

Alex-Jordan:
Wait, when you say "keep it undocumented" do you mean I should remove the sample chapter example and its commentary? And what about schema?

rbeezer:
Was commenting by email.  Rushed here.  Camelcase was not me.  ;-)

Keep in sample article, so we can test.  Warn about a possible change (if you 
want, or not).

Remove from schema - that is our contract with authors.

Alex-Jordan:
OK, done and done.

rbeezer:
Merged and pushed.  No website update yet.  Thanks for the naming discussion on this one.

I should make a careful style guide.  I think I like spaces around logical operators (`=`) in `@test`.  ;-)

Alex-Jordan:
I'm gearing up to actually use this in a bunch of APEX problems. Could run with `@data`.

I'm less down on `@syntax` now. Presence of `@syntax` makes it implicit that the variable is a string data type, and that could just be understood. My objection was that it would be an attribute that would never have a place with a totally generic `var`, but now I think, who cares?

Synonyms for `@content`: `@composition`, `@substance`.

Will run with `@data` in APEX and it's easy to search and replace. Probably don't want to let this go beyond APEX before making any change.

Alex-Jordan:
> I should make a careful style guide. I think I like spaces around logical operators (=) in @test. ;-)

I find 487 instances of `test="[^"]*(?<! )=.*?"` (so spaces around `=`) in the current code, and 511 instances of `test="[^"]* =.*?"` (with spaces). A style guide would be good, but the de facto style guide is looking at what is already there. Maybe one PR for this?

PR #1259, opened 2020-03-10
WeBWorK description

Alex-Jordan:
This allows for a short description of your WW problem, which ends up as a comment at the top of the .pg file. (Which is not ideal, but at present that is the best practice for placing a description in a .pg file.)

### DIFF
--- a/schema/pretext.xml
+++ b/schema/pretext.xml
@@ -1592,6 +1592,7 @@
         WebWorkAuthored =
             element webwork {
                 attribute seed {xsd:integer}?,
+                element description {TextShort}?
                 WWMacros?,
                 WWSetup?,
                 (
--- a/xsl/extract-pg-common.xsl
+++ b/xsl/extract-pg-common.xsl
@@ -117,6 +117,9 @@
     <xsl:param name="b-hint" select="true()" />
     <xsl:param name="b-solution" select="true()" />
     <xsl:param name="b-verbose" />
+    <xsl:apply-templates select="." mode="description">
+        <xsl:with-param name="b-verbose" select="$b-verbose" />
+    </xsl:apply-templates>
     <xsl:call-template   name="begin-problem">
         <xsl:with-param name="b-verbose" select="$b-verbose" />
     </xsl:call-template>
@@ -152,6 +155,9 @@
     <xsl:param name="b-hint" select="true()" />
     <xsl:param name="b-solution" select="true()" />
     <xsl:param name="b-verbose" />
+    <xsl:apply-templates select="." mode="description">
+        <xsl:with-param name="b-verbose" select="$b-verbose" />
+    </xsl:apply-templates>
     <xsl:call-template   name="begin-problem" >
         <xsl:with-param name="b-verbose" select="$b-verbose" />
     </xsl:call-template>
@@ -190,6 +196,16 @@
     <xsl:call-template   name="end-problem">
         <xsl:with-param name="b-verbose" select="$b-verbose" />
     </xsl:call-template>
+</xsl:template>
+
+<xsl:template match="webwork" mode="description">
+    <xsl:param name="b-verbose" />
+    <xsl:if test="description and $b-verbose">
+        <xsl:text># </xsl:text>
+        <!-- value-of instead of apply-templates, so that character escaping is not applied -->
+        <xsl:value-of select="description"/>
+        <xsl:text>&#xa;&#xa;</xsl:text>
+    </xsl:if>
 </xsl:template>
 
 <!-- The setup element formerly had more internal structure. Now it only   -->


sean-fitzpatrick:
If I'm reading this correctly, this is for adding comments visible in the PG code.
So that would include things like `##MO(1)` that tell the library browser that the question uses MathObjects. (Which is good.)

When we were writing problems, we wanted to acknowledge that most instructors are not interested in reading the source file to find out details about the problem.

So we used comments like the following, which we placed after the PGML solution:
```COMMENT('
    Randomization provides 14 different possible versions of this question. 
    Includes a solution set.<BR>
    Made from a ULETH template.<BR>
    ');
```
The advantage of this (as you probably know) is that it produces a comment that's visible in the library browser, so instructors know details like "the matrix is always invertible", or "randomization produces 34,000 versions of this question".

Can we do this type of comment as well?
(At minimum it might be good advertising to have "Authored in PreTeXt" showing up when people are browsing the Open Problem Library. Maybe this already happens and I've never noticed.)

Alex-Jordan:
Is your post about using the `COMMENT` device? I think so, but maybe a line of your example was cut?

Short answer: yes, we could put a `description` into the `COMMENT`.

Currently we use that to state `Authored in PreTeXt`. It could say both. My real opinion is that the official metadata tags should have `## DESCRIPTION` too, and a `description` from PTX should go there. The library browser could display this at the top when browsing. 

The `COMMENT` device doesn't have a clear purpose if you ask me. (You can see how working with PreTeXt has affected how I think about these things.)

I would not use `description` or `COMMENT` in quite the same way as your example. For example, presence of a solution is just that; it's there or it's not. Perhaps the Library Browser should have another fancy icon to signify that. But it would just be based on if there is a solution there, not any tagging.

I'm working on rebuilding how we write metadata into PTX-built PG problems. It's using "bad" XSLT, and it doesn't take advantage of all the things that could go into that metadata. I'll refactor how `COMMENT` is used as part of that work, but it will take more time. For now my real goal is actually just getting `description` into the schema.

rbeezer:
Comments.  Or a description?  ;-)

* Great idea.
* A non-trivial example (nonsense is OK) needs to go into something like the sample chapter.  Too much potential for mistakes if we don't have a test case or two.
* Is a modal template necessary?  The schema part is pretty restrictive.  Can you just match on `webwork/description` and select on the child `description` where you want it to land?  Or maybe just pick it up as part of a general `apply-templates` since it is authored first and output first?  I'm just reading the diff here, so it is possible I'm not considering enough context.  Maybe it is like a `title` but maybe not?
* Would you ever want to structure this text as a sequence of `line` elements?  That's pretty easy to accomodate.
* Is every element of `TextShort` covered with what we have already?  I'd think so.  Put a few into the test case?
* Documentation?

sean-fitzpatrick:
Ah, yes - sorry. That should have begun with `COMMENT ('`

We find this useful for instructors who are using the library browser to build a homework set.
For example, knowing how many versions are created by randomization informs whether the problem can be used with Show Me Another.
Or we might want to know that a function has three critical points, etc.

But this is still from the point of view where problems live on the WeBWorK server and instructors build homework sets there.
I could see how much of what we thought was important information might be less so if we were building homework sets in PreTeXt and then exporting.

(But it was a useful exercise to make students run through all versions created by randomization in a spreadsheet, to see how many they got, and whether some seed might create an error.)

Alex-Jordan:
>  For example, knowing how many versions are created by randomization informs whether the problem can be used with Show Me Another.

I totally get the usefulness of communicating that to instructors. I think the _right_ way to do that is to have a new metadata tag like `## RandomVersions(12)` though, and the Library Browser displays that integer in some way. So add that to my list of new WW features to work on.

> Or we might want to know that a function has three critical points, etc.

This is more what I am thinking this will be for. Something that describes the problem from a pedagogical angle.

Alex-Jordan:
> Documentation?

I've learned not to get into documentation or a sample until getting past at least one round of constructive criticism. Too much time rewriting stuff otherwise ;)

> Is a modal template necessary? 

Using a template was because (1) the code gets used twice, and (2) everything else in that top-level template is calling other templates, so it didn't look like clean code.

You mention `apply-templates`. Within a WW text elements like `This problem computes b^2 - 4ac from a given quadratic function.` gets converted to `This problem computes b^2 \- 4ac from a given quadratic function.` That is because the `-` has special PGML meaning (as do many other things). I used `value-of` instead of `apply-templates` to sidestep this, but if you can tell me how to `apply-templates` and ignore the usual WW character replacements, I'd happily go that way.

> Would you ever want to structure this text as a sequence of line elements?

We have `description` for `image`. _That_ description needs to be plain old ASCII since it ends up in HTML alt attributes. So I was thinking to keep `description` as plain old ASCII across the board. If you are up for multiple flavors of `description` in the schema though, then yes, multiline would be nice.

rbeezer:
OK, let's indicate PRs are up for comment and not final, and once final, then various bits (test cases, documentation, schema) all are present?  That'll mean less wasted work on both ends.

To use `value-of` then I think the schema should just say `text`, not `TextShort`.  And yes, I think `image/description` is way too liberal in the schema.

I'll look closer at templates with more context.  I think "used twice" just argues for **some** template.  But maybe modal makes sense at this level.

Alex-Jordan:
It sounds like an important thing to work out is whether `description` should be allowed to have lines. I didn't see that coming and I could open a forum thread if this has grown too big for here.

After thinking more, I think even an `image/description` could maybe be allowed `line` children. In alt text, there would be no line break. Maybe just an inserted double space between lines or something like that. But in principle the `image/description` could use lines in other places. Say if it comes out as title text when you hover over the image in a PDF.

Alex-Jordan:
I have a larger "WW metadata" edit on a branch that incorporates an evolution of this. It uses @sean-fitzpatrick's suggestion to use `COMMENT()`. Closing this, and will open the metadata PR once #1258 and #1261 are processed.

sean-fitzpatrick:
Thanks. Sorry for dropping off of this thread. Things are sort of blowing up here, like a lot of places, I guess.

Issue #710, opened 2017-09-04
deprecate setup in webwork


Alex-Jordan:
With the forthcoming scheme for webwork elements, there is no need for var/static as children of setup. This means setup will have only one child: pg-code. So we can just eliminate setup.

Fixing deprecations would mean deleting all setup/var and then moving setup/pg-code to just pg-code, deleting setup. Children of a webwork would be:
title?,pg-macros?,pg-code?,statement,hint*,solution*

This is just a reminder to do this at some point.

rbeezer:
Will the change be backward-compatible?  In other words can the XSL look for `pg-code` both ways?  I can add a warning for the old way.

Note to myself:  `fix-deprecations.xsl` will go something like
```xsl
<xsl:template match="setup">
  <xsl:apply-templates select="pg-code" />
</xsl:template>
```

Alex-Jordan:
I think it's easy enough to make the pg-code part backward compatible. But I don't want to make backward compatible the part about using author-declared values for static variables.

rbeezer:
Right - the old setup/var can just vanish, as I understand it.

You'll just have to test for "setup" being present and get "pg-code" from there 
in that case, otherwise get it from new arrangement.

Alex-Jordan:
See #1261.

rbeezer:
Addressed on #1261.

PR #1261, opened 2020-03-14
WeBWorK: deprecate setup

Alex-Jordan:
This deprecates the `setup` element in a WeBWorK. Once upon a time, `setup` was a wrapper for a few things. Over time it became the wrapper for `pg-code` only, and it's out of place.

* Removed from all examples
* Update the author guide
* Edit the XSLT to behave with or without `setup`
* Add deprecation message
* Adjust schema

### DIFF
--- a/doc/guide/author/webwork.xml
+++ b/doc/guide/author/webwork.xml
@@ -138,11 +138,9 @@
         <![CDATA[
         <exercise>
           <webwork>
-            <setup>
-              <pg-code>
-                Context("ImplicitPlane");
-              </pg-code>
-            </setup>
+            <pg-code>
+              Context("ImplicitPlane");
+            </pg-code>
             <statement>
               <p>The answer is <m>x+y=1</m>.</p>
               <p><var name="'x+y=1'" width="8" /></p>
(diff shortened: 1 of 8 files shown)


Alex-Jordan:
One thing that maybe should be done is adding the right thing to `fix-deprecations.xsl`. I would like to leave that to Rob, if I may. Just because I do not understand that well. In #710 you (@rbeezer ) made a note about this.

rbeezer:
I added a test case of the old style into the `sample-errors-and-warnings.xml` example, plus an automatic fix in the `fix-deprecations.xsl` file.  Just added into the rest.  Merged and pushed.  Finally doing website rebuild promised for last night.

rbeezer:
The `.//pg-code` make me nervous.  ;-)  They'll allow some slop to sneak through.  No need to justify - I understand why.

* A fix would be to get the `pg-code` element with a variable, like
```xsl
<xsl:variable name="the-code" select="setup/pg-code|pg-code"/>
```
and use that in the various `contains()`, etc.
* Better would be to just fix the source with the new "assembly" pre-processor.  But first, let's get that integrated properly with the main WW pipeline, and then go back and fix these up.  I'll make an issue and point back here.

rbeezer:
Forgot to say: do you want to announce deprecation on -announce?

Prefix subject with "[DEPRECATE]", make it clear it is not urgent, and "you will get warnings".

Or tell me to do it.

rbeezer:
I never updated the various versions of the schema, and hence missed a missing comma.  Mea culpa.

All fixed up at e5a0155ab17b770ce22e1435c689cdf081823ea5

Issue #1199, opened 2019-11-10
Rename slideshow attribute for gradually revealed content


rbeezer:
We've talked about changing `slide/@slide-step` to `slide/@pause`.  This is just a marker for that project, since I think I've run out of time to get to it today.  Entirely cosmetic.  Not sure if it needs a `pretext-announce` warning or deprecation message.

@sean-fitzpatrick @StevenClontz

sean-fitzpatrick:
Just realized (the hard way) that the documentation still needs updating on this. I'd forgotten about the change. Did `slide-step="true"`, to no effect. Tried it in a few places. Then remembered this. Changed to `pause="true"`. Finally figured out that it's now `pause="yes"`.

I can push the documentation change unless @StevenClontz beats me to it. But for now I'm being weak and going to bed. (Or keeping up my strength for when the pandemic finally finds me.)

StevenClontz:
I wrote a MWE in the PreTeXt guide, but I think the correct solution is to scrub that and link to https://github.com/rbeezer/mathbook/blob/dev/examples/sample-slideshow/sample-slideshow.xml instead, since that actually gets "tested" against new features/changes and issues like @sean-fitzpatrick found won't happen as easily.

rbeezer:
Came to close this, and saw that #1263 needed one more fix.  Done at 8b5dc534d7c3460464fb12014d24817fecff97df

Will update website version of the Guide later tonight.  Thanks, for reporting and working through this one.  I think this batch of related issure/PRs is done, but let me know if anything is left dangling.

PR #1263, opened 2020-03-17
scrub MWE for slides and replace with link to examples in repo

StevenClontz:
Fixes documentation issue noted in #1199.


rbeezer:
Thanks, Steven!  I added a footnote for the URL so that print has something usable.  And especially since the `rbeezer/mathbook` part should change.

Merged, will refresh website later tonight.

PR #1264, opened 2020-03-24
add early/late preamble options

oscarlevin:
Implement preamble.early and preamble.late string params for beamer stylesheet.


rbeezer:
Thanks again, Oscar.  Merged with minimal testing.

@sean-fitzpatrick - holler if this does not work.

PR #1267, opened 2020-04-09
Schema: distinguish interactive exercise

Alex-Jordan:
This is me attempting to edit the schema in a more nontrivial way than I've tried so far. For consideration/discussion.

My goal is to make it so that an exericsegroup either has non-WW exercises or it has WW exercises, but never a mix.

In the future, MOM exercises and other flavors might be established. Sort of trying to do this in a way that isn't WW-centric. (But of course, I also think it would be bad to mix WW and MOM exercises in an exercisegroup.)

### DIFF
--- a/schema/pretext.xml
+++ b/schema/pretext.xml
@@ -1504,20 +1504,35 @@
             element statement { ExerciseBody }
         Exercise =
             element exercise {
+                (
+                NoninteractiveExercise |
+                InteractiveExercise
+                )
+            }
+        NoninteractiveExercise =
+            element exercise {
                 MetaDataTitleOptional,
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Closing this. The goal here was to put limitations in place so that I could make narrow assumptions about an exercisegroup, so that I could handle their introductions better. 

But we are now going to:

1) Always bring the exercisegroup introduction into the webwork when building the archival .pg file.
2) Consequently, advise authors to be careful with wording in an exercisegroup.

Note that we will not be bringing the introduction into the problem when making static or when making the live version URL for embedding. In those situations, the introduction stays where it normally lives.

PR #1268, opened 2020-04-17
Extract asymptote addition

mdoob:
The ``extract-asymptote.xsl`` file  makes a last guess at determining 3D files with (essentially) a grep search for "import graph3;". Many 3D fiiles use "import three;" instead. An additional check for this case is added.

### DIFF
--- a/doc/guide/appendices/editors.xml
+++ b/doc/guide/appendices/editors.xml
@@ -392,7 +392,7 @@
                 Usually vi starts in normal mode.
             </li>
             <li> <em>Insert mode: </em> This mode is for inserting new text into the file.
-                Typing <c>The galloping gertie leapt into the air</c> causes that text
+                Typing <c>Galloping Gertie leapt into the air.</c> causes that text
                 to be inserted at the current position in the file.
             </li>
             <li> <em>Command mode: </em> This mode is to execute commands.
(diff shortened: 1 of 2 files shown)


mdoob:
Small change to recognize 3D files that use "import three;".

rbeezer:
This pull request has four commits, and I was uncertain where the oldest ws rooted, since it is from an older pull request, and was changed in the process of merging it.  So I didn't think I could isolate it safely.

Instead, I just copied the new code into a new commit and referenced this pull request.

Thanks for the contribution!

sean-fitzpatrick:
I haven't followed carefully (still in the middle of assigning grades and contacting cheaters) but if I understand the recent correspondence between John Bowman and us, we might not need the ability to detect 3D graphs: `asy -f svg -render=4` will produce decent SVG from any Asymptote file.

But it might come with the usual caveats about having up to date versions of everything.

mdoob:
Yes, Sean, I think that is correct. The default value is render=-1,
so it has to be changed. The value render=4 is probably overkill
for most situations, but we could experiment.

As I surmised and John confirmed, older implementations will cause
problems. Should we throw a warning?

Cheers,
Michael

------------------------------------------------------------------
Michael Doob                    Telephone: (204) 474-9796
Department of Mathematics       email: Michael.Doob@UManitoba.ca
University of Manitoba
Winnipeg, MB, Canada R3T 2N2
------------------------------------------------------------------

sean-fitzpatrick:
Yes -- I think Rob already added a warning that shows up if you run `mbx` with the `-v` option. We can edit it to give minimum version recommendations for `asy`, `dvigsm`, etc.

PR #1269, opened 2020-04-17
Remove WW-specific fillin template

Alex-Jordan:
I am not able to see the purpose of this template. Maybe it used to matter, or evolved into something that we don't really need. There is the "regular" fillin template at line 8135. And that one is good for this too. In fact, what led me here is that this WW-specific one is making blanks that are far too long. They look nicer after I delete this.


rbeezer:
Merged as-is.  Thanks for the code clean-up.

Argument to the LaTeX macro is `em` while PTX source has `@characters` attribute.  The conversion is `5/11 characters/em`.  The WW template was not doing this conversion, hence blanks were about twice as long as they should have been.

PR #1270, opened 2020-04-21
Suppress knowls in pretext-revealjs; bold #dfn in pretext-revealjs

### LINKED ISSUE #1257: Math in RevealJs proofs
Right now, the black magic being used from (I think) `mathbook-common.xsl` is causing MathJax to ignore math markup found in a theorem's proof:

image

There's obviously some cruft with the `hidden-content` that isn't appropriate here, which is probably related. But the specific culprit is `tex2jax_ignore` - once that class is removed then the math renders as desired.

StevenClontz:
I created a stringparam `html.knowl.all` which can be used to override all knowl defaults. I use this to quickly kill all the knowls in pretext-revealjs. This closes #1257.

I also added styling for `#dfn` tags in our pretext-revealjs output, so terms will appear in bold. This is the correct way to fix what I attempted to fix in #1255.

### DIFF
--- a/xsl/mathbook-html.xsl
+++ b/xsl/mathbook-html.xsl
@@ -464,23 +464,109 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- PROJECT-LIKE gets own switch here           -->
 <!-- "example" are set to 'yes' by default       -->
 <!-- so new authors know that knowls exist       -->
-<xsl:param name="html.knowl.theorem" select="'no'" />
-<xsl:param name="html.knowl.proof" select="'yes'" />
-<xsl:param name="html.knowl.definition" select="'no'" />
-<xsl:param name="html.knowl.example" select="'yes'" />
-<xsl:param name="html.knowl.project" select="'no'" />
-<xsl:param name="html.knowl.task" select="'no'" />
-<xsl:param name="html.knowl.list" select="'no'" />
-<xsl:param name="html.knowl.remark" select="'no'" />
-<xsl:param name="html.knowl.objectives" select="'no'" />
-<xsl:param name="html.knowl.outcomes" select="'no'" />
-<xsl:param name="html.knowl.figure" select="'no'" />
-<xsl:param name="html.knowl.table" select="'no'" />
-<xsl:param name="html.knowl.listing" select="'no'" />
-<xsl:param name="html.knowl.exercise.inline" select="'yes'" />
-<xsl:param name="html.knowl.exercise.sectional" select="'no'" />
-<xsl:param name="html.knowl.exercise.worksheet" select="'no'" />
-<xsl:param name="html.knowl.exercise.readingquestion" select="'no'" />
+<xsl:param name="html.knowl.all" select="'default'" />
+<xsl:param name="html.knowl.theorem">
+  <xsl:choose>
+    <xsl:when test="$html.knowl.all = 'default'">no</xsl:when>
+    <xsl:otherwise><xsl:value-of select="$html.knowl.all" /></xsl:otherwise>
+  </xsl:choose>
+</xsl:param>
+<xsl:param name="html.knowl.proof">
+  <xsl:choose>
+    <xsl:when test="$html.knowl.all = 'default'">yes</xsl:when>
+    <xsl:otherwise><xsl:value-of select="$html.knowl.all" /></xsl:otherwise>
+  </xsl:choose>
+</xsl:param>
+<xsl:param name="html.knowl.definition">
+  <xsl:choose>
+    <xsl:when test="$html.knowl.all = 'default'">no</xsl:when>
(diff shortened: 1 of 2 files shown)


davidfarmer:
I'd like a discussion about the need for inline styling.

Is it necessary for revealjs?

That discussion can occur after this pull request is accepted.
But let's not forget.

StevenClontz:
@davidfarmer do you mean https://github.com/rbeezer/mathbook/blob/71c25637c2299f2921551780fc73bd10bddfb9d8/xsl/pretext-revealjs.xsl#L206 ? It might be more appropriate to have a external `pretext-revealjs.css` that's imported, similar to how it seems https://pretextbook.org/css/0.2/pretext.css is used by HTML output. I think all the CSS work for RevealJS output so far has been hacked together (mostly by Andrew in Portland last summer), so there's not been much discussion on the topic before.

rbeezer:
Styling:  Yes, there is CSS in the `<head>` of each slideshow.  That makes more sense in a file of its own.  What is there now is definitely a temporary thing that should be discussed and squared away before it goes too far.

Our Reveal slideshows have a switch to allow online or offline use.  I think this is an important feature for (a) conference presentations with unreliable connections, (b) archiving talks online.  Note that we make only one output file for an entire slideshow.

Is there a **disadvantage** to putting CSS into the HEAD just once?

Related - should we, could we, use our stock CSS and just override parts of it?  I would think things like `dfn` within a sentence would "just work".  But blocks, like a Definition, might have an entirely different look?  Could all the color scheme work just migrate over (with care)?

As David suggests, I'm inclined to merge this now and it'll get fixed as part of a bigger project.

rbeezer:
Knowl-killing:  Those "dotted" parameters should mostly migrate to new publisher switches (a semi-complicated, semi-tedious project).  So I'd rather not have a new one down in the HTML conversion.  We could

(a) set each parameter as an override in the Reveal conversion.  An author could try to override it while building and  get what they deserve, so not fool-proof.

OR

(b) prevent knowlization at a lower level, such as is done presently for the conversion to braille:

https://github.com/rbeezer/mathbook/blob/15fcdc709f7e81940ef23f0af5388b6a23816b09/xsl/pretext-braille.xsl#L283-L287

StevenClontz:
Re: styling. There's a temptation to actually make both css styling AND JavaScript inline for the offline version, so we produce a single HTML file that just works™, and let users define filepaths to local js/css if they want an offline version with their own copy of revealjs.

Of course this depends on the license of revealjs, which I'll look at now.

StevenClontz:
(GitHub had server issues when I was posting, but it's back and I'm back.)

Anyway, RevealJs is MIT, which I believe we can package as a part of PreTeXt (which I believe is GPL). So we could freeze a copy of RevealJs Javascript/CSS to use in producing a single-file HTML slideshow. If we want.

rbeezer:
I'd be reluctant to maintain a copy, unless the updating was automatic.  That's why we try to use services that come from CDNs and don't require authors to host (updated) software on servers.

BUT, if we had a single-step Python/mbx command to build a slideshow (like I'm making for EPUB right now), then we **could** scrape the Reveal site/repo and bundle everything up, every time, as an optional archive/offline version.  Without burdening authors nor developers.

StevenClontz:
I'm probably just a troglodyte, but my first instinct is to avoid mucking up the RevealJS conversion with additional dependencies. Then again, I already wrote a hacky build script to convert some Tikz into PNGs for my most recent slideshow, so I probably should give up on that goal.

Something that would make all of this less sketchy for me is if we had a commandline tool for Linux to manage dependencies. I remember @oscarlevin and others working on that in Portland last summer but don't recall where it ended up.

rbeezer:
Troglodyte-ism is good.  ;-)  This would not be additional - it'd be a replacement.  **Instead** of `xsltproc` there would be a Python script.  Advantages:

1.  Hopefully much more mainstream on Windows.
1. The possibility of a GUI tool (with Python bindings) to drive Python routines.

I thought the latter was exactly what @oscarlevin , et al., were working on in PDX.  it is definitely their work that has me moving in this direction.

rbeezer:
I merged the `term/dfn` styling.  But no real testing.

@StevenClontz - do you want to do the knowl-killing?  If not, let me know.  I'm going to close this, so that could go on a new PR.

Thanks!

StevenClontz:
Yeah, a Python-based command-line PreTeXt application not only applies XSLT, but also other tasks like pulling a copy of Reveal.js, seems like a worthwhile project.

I'll do the knowl-killing in a new PR.

oscarlevin:
I'm still hoping to work on this, as soon as the semester ends.  No plans for a GUI just yet, but a command line tool ala latexmk but for pretext seems like a reasonable goal and something that could be plugged into modern text editors or an eventual GUI.

rbeezer:
I'm guessing it would be better to split the mbx script into the command-line-interpretation portion, and a Python `module` containing the functions that do the real work?  With good docstrings and stable argument lists, etc.

PR #1271, opened 2020-04-28
Kill knowls in revealjs conversion

### LINKED ISSUE #1257: Math in RevealJs proofs

StevenClontz:
As discussed in #1270 this kills knowls without introducing new stringparams.

Closes #1257

### DIFF
--- a/xsl/pretext-revealjs.xsl
+++ b/xsl/pretext-revealjs.xsl
@@ -170,6 +170,11 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <!-- conversion (we just build one monolithic page) -->
 <xsl:variable name="html-index-page" select="/.."/>
 
+<!-- Kill knowl-ing of various environments -->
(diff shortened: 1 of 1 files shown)


rbeezer:
I suspect you ran into trouble with the entities.  I added the arcane incantation for that to the top of the file, and then replaced the `match`.

Otherwise, we'd be chasing every new corollary, lemma, fact, observation,....

Please test?  I've done zero testing.

Thanks for putting this one together!

StevenClontz:
Thanks - I'm not sure how much I saved your time with these PRs, but I'm learning things at least. :-)

I did do a quick test on the file I've been using and it looks good.

PR #1272, opened 2020-04-30
remove "setup" instance

Alex-Jordan:
I think the PR that deprecated `setup` and the PR that added PGML syntax in a `var` overlapped chronologically without a rebase, and this `setup` snuck back in.


rbeezer:
Tsk, tsk.  ;-)

Thanks for the clean-up.  Merged, but not yet live on the website - next big update.

PR #1273, opened 2020-05-03
WeBWorK: randomize inline exercises

Alex-Jordan:
For code review and merge if it appears OK.

https://github.com/PreTeXtBook/JS_core/pull/28 will have to be approved before this will actually work.

Taking the opportunity to remove some illegal attributes to the iframe. I see no bad consequences in output.


rbeezer:
Visual scan looks great.  I'll merge in a bit.

rbeezer:
Merged as-is.  Just for kicks, I built the sample-chapter with the temporary switch.  Everything seemed fine.  Thanks for the quick work on this.  Peter and Ruby will be thrilled.

I'll let you announce whatever you think is appropriate on -dev.

PR #1274, opened 2020-05-07
Add caution about punctuation in lists of math items

sean-fitzpatrick:
In both the sample article and the documentation, emphasize that *any* text outside of `m` tags in a `li` will disable automatic display math, including punctuation.
In the documentation, also mention that CMoS says that punctuation shouldn't be there anyway.


rbeezer:
All merged in, but no output updated yet at website.

You editor corrected a bunch of traling whitespace, thank-you very much.  I split it out into its own commit and took responsibility.

sean-fitzpatrick:
I'm using Atom. I think whitespace removal is a default setting

PR #1276, opened 2020-05-12
WIP: pip installable mbx script

cbm755:
Here's a early WIP attempt at making a `pretext-mbx` command installable via pip.

I discussed this earlier with @arechnitzer: perhaps we can iterate a bit on this.  Certainly not mergeable yet!


rbeezer:
Dear Colin,

Thanks for all of this.  Sorry for the delay - I just got final course grades in yesterday.  ;-)

I'm upgrading to Python 3 right now, along with more transparent handling of Windows pathnames.  So your timing is good.  (All that is being tested right now, so not yet public.)

Miscellaneous comments:

1.  I should rearrange `main()` as you have done.  I cleaned up the (problematic) global imports.
1.  I'm trying to target Python 3.4 (necessary for `pathlib` and `lxml` modules).
1.  I'm trying to replace `os.path` functions with analogs for `pathlib.Path()`.
1.  Generally (finally), I'm trying to replace "MathBook XML" ("mbx") with "PreTeXt".  So maybe the script should become `pretext.py`?  Is the `.py` strictly necessary?  I'd thought having `pretext` as the filename, along with a shebang would be as good (better?).

Long-term, I intend for this file to just be the CLI stuff, and put the routines into a module that can be used with some nicer front-end program/interface.

Versioning at #1277 next.

Rob

setup.py
```diff
@@ -0,0 +1,66 @@
...
+    classifiers=[
+        "Programming Language :: Python :: 3",
+        "License :: OSI Approved :: GNU General Public License v2 or GNU General Public License v3",
+        "Development Status :: 3 - Alpha",
+        "Operating System :: OS Independent",
+        "Topic :: Text Processing",
+    ],
+    entry_points={
+        "console_scripts": [
+            "pretext-mbx=script.mbx:main",
```
cbm755, line 41: > So maybe the script should become pretext.py? Is the .py strictly necessary? I'd thought having pretext as the filename, along with a shebang would be as good (better?).

The important point here from user's point of view is this line.  Upon install this will create
`/usr/local/bin/pretext-mbx`
or
`${HOME}/.local/bin/pretext-mbx` or something else appropriate.

The content of that autogenerated file can point to any Python module's main() function.

It sounds like perhaps this entry point should be called `pretext`?

cbm755:
> Is the .py strictly necessary? 

AFAIK yes.  What we really have here is a Python module called `script`, this is the purpose of `script/__init__.py`.  Probably we want `s/script/pretext/` at some point.  As I explained above, none of this matters for end-users.  (It does however matter for people doing `import script` in say `iPython` which is why it should probably be `import pretext`).

rbeezer:
Thanks, Colin (@cbm755) for your contributions here.  Oscar Levin and Steven Clontz are well along on a very Pythonic CLI that will be versioned, pip-installable, etc.

https://github.com/PreTeXtBook/pretext-cli

So I think these have been usurped.  Appreciate your interest and help.

PR #1278, opened 2020-05-14
WW: when producing verbose PG, include any introduction from a parent…

Alex-Jordan:
… exercisegroup

If there is math in the introduction and the math uses author LaTeX macros, then we need to pass `b-verbose` to the introduction so that the LaTeX macros get defined within the PG file. So that leads to creating an `introduction` template in `extract-pg-common`.

### DIFF
--- a/doc/guide/author/webwork.xml
+++ b/doc/guide/author/webwork.xml
@@ -262,6 +262,25 @@
     </subsection>
 
     <subsection>
+      <title><webwork/> in an <tag>exercisegroup</tag></title>
+      <p>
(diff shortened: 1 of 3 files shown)


rbeezer:
Looks good.  But I only find **one** `exercisegroup` in the sample chapter.

The existing one is pretty simple, which is good.  Can we add one which tests this introduction-migration, including the necessity of including macros?

Alex-Jordan:
Now with an `exercisegroup` in the antiderivatives section.  Uses the author LaTeX macro for definiteintegral. Tested on HTML, PDF, and PG output, and all seems well.

rbeezer:
Very good.  Thanks.

PR #1279, opened 2020-05-15
WW: load essay macros if explanation_box is used

Alex-Jordan:
`explanation_box` is a variant from the essay macros. It is intended to follow a quantitative answer box, where the student explains their work. (Like find the derivative: ____ Show that you used the definition of the derivative: [explanation box].)

The reason to separate this from an essay is so that you can turn it off globally. Maybe you want to use the question, but now burden yourself with grading the essays. Or if you want to use it during a WW quiz, you really want to not use essay questions, because quiz mode is not designed to handle them well.

This PR detects your desire to use `explanation_box` from the `pg-code` block, and then loads the necessary macro library for you.

### DIFF
--- a/xsl/extract-pg-common.xsl
+++ b/xsl/extract-pg-common.xsl
@@ -513,7 +513,7 @@
             </xsl:choose>
         </xsl:if>
         <!-- essay answers -->
-        <xsl:if test=".//var[@form='essay']">
+        <xsl:if test=".//var[@form='essay'] or contains(.//pg-code,'explanation_box')">
             <xsl:choose>
                 <xsl:when test="$b-verbose">
                     <xsl:text>  "PGessaymacros.pl",&#xa;</xsl:text>


rbeezer:
Merged.  Thanks.

Would it be good to have an example in teh sample chapter that exercised this option?

Alex-Jordan:
> Would it be good to have an example in the sample chapter that exercised this option?

I considered it. Two things:

- There are lots of other tools that are automatically included in the same way. We don't demonstrate them all in the sample chapter.
- This one could be factored into source markup someday, like I mentioned [here](https://groups.google.com/d/msg/pretext-dev/CvfVSyvGk7M/EO6yaZVaBAAJ). Using this now requires knowing and using some very special things about WW and PTX. So, on the one hand, good to explicitly demonstrate. But I'm unsure if it is worth the undoing in people's source that would be needed if this goes into PTX markup someday.

Incidentally, I've started using this in APEX.

PR #1281, opened 2020-05-15
WeBWorK: code cleanup

Alex-Jordan:
Recent work in `extract-pg-common` revealed opportunities for code cleanup. Instead of including this in other recent commits, I separated it into this one. Together these edits cut  away 70 lines of code.

I ran the sample chapter extraction pre- and post-. The only difference comes from the change to when `$refreshCachedImages`. First, an unnecessary space character is removed. Second, we should really only be setting this to 1 when we are making the live problem embedded in HTML or when we are making the image files. But when we are making "verbose" versions for people to use on their WW server as problems, then we should leave this to be its default, 0. So the diff was what I would want it to be.

Not sure how you feel about named templates. This introduces one as a way to not have to repeat PG macro names. For example, the string `niceTables.pl` is used once only, and the `macro-padding` named template wraps quotes around it and gives it a trailing comma. The payoff is that it handles the "verbose" and "not verbose" cases together, using spaces and a line break (or not).

### DIFF
--- a/xsl/extract-pg-common.xsl
+++ b/xsl/extract-pg-common.xsl
@@ -426,18 +426,18 @@
 
     <!-- three standard macros always, order and placement is critical -->
     <xsl:variable name="standard-macros">
-        <xsl:choose>
-            <xsl:when test="$b-verbose">
-                <xsl:text>  "PGstandard.pl",&#xa;</xsl:text>
-                <xsl:text>  "MathObjects.pl",&#xa;</xsl:text>
-                <xsl:text>  "PGML.pl",&#xa;</xsl:text>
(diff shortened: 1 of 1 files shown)


rbeezer:
Named templates are fine, with one proviso.  They should not depend on context.  I think those could be hard to deal with when debugging in certain scenarios.  So what you have here looks great: it is parameterized (clearly) and does not have any dependence on the active node when it is employed.  (I think there **are** some rogue named templates about - they could switch to `apply-templates` rather than `call-template`).  Merging now...

rbeezer:
Great!  I love code cleanup.  ;-)  Will likely now update website.

PR #1282, opened 2020-05-16
Add `--version` to mbx script and version file

cbm755:
Other tools can grab the version number from this file too.  Since there
is stuff here other than Python, maybe the central version number should
not be stored in a py file: that could be changed later.

### DIFF
--- a/script/version.py
+++ b/script/version.py
@@ -1,0 +1,4 @@
+# Copyright (C) 2020 Colin B. Macdonald
+# SPDX-License-Identifier: FSFAP
+
+__version__ = "0.0.1.dev"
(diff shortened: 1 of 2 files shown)


script/version.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (C) 2020 Colin B. Macdonald
+# SPDX-License-Identifier: FSFAP
```
cbm755, line 2: @rbeezer I'm old fashioned and usually try for per-file copyright notices.  For a trivial file like this I typically use the "FSF All Permissive" which is pretty much public domain.

If you don't care about any of this, you can take my name off, or remove the copyright notice altogether --- it is afterall "all permissive" :-)

PR #1285, opened 2020-05-20
Sample Article: Fix the Sage interact

mitchkeller:
Sage's move to python3 has caused the Sage interact in the sample article to break. I have fixed the code so that it runs (doubling backslashes and adding parens with `print`), adjusted the width of the image produced so that no horizontal scrolling is required, and updated the `@aspect` on the `interactive`.


rbeezer:
Thanks, Mitch!  Merged, but not pushed to website.  I'm clocking out - long day of too much Python!  More in the morning.

PR #1286, opened 2020-05-21
Asymptote corrections

sean-fitzpatrick:
Update to mbx script for svg output. Tested against svg, pdf, and html output successfully.

- I did not add the option `-tex xelatex` to the `asy` executable but this should be considered in the future: I can install `asy 2.66` on my 18.04 laptop (by compiling from source) but not `dvisvgm 2.9.1` (need more recent `libc6`) and this is needed to test this option.
- With current options for `svg` the subroutine to copy over `png` files is needed. Adding the `-tex xelatex` option might eliminate the need for this: in all examples I've tested, I get a standalone `svg`.
- Also updated the `workcone` example in the sample article to include the improvements from John Bowman.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -2809,50 +2809,50 @@ the xsltproc executable.
                     <image xml:id="asymptote-workcone" width="50%">
                         <!-- aspect ratio from canvas in produced HTML -->
                         <asymptote aspect="139:151">
-                          import solids;
-                          size(0,150);
-                          currentprojection=orthographic(0,-30,5);
+                        import solids;
+                        size(0,150);
+                        currentprojection=orthographic(0,-30,5);
 
-                          real r=4;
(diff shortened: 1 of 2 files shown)


rbeezer:
Hah!  I just added the work cone.  ;-)  As separate commits, I just dropped them and kept the asy/png/svg fix.  Thanks very much for the fix, and the organization as logical commits.

Git Tip:  it could have been easy to "add in" the whitespace changes to the previous commit on your end before shipping up to GitHub.  See how `git commit --amend` works (after a `git add`).  Even more fun, and much more dangerous, is an interactive rebase, `git rebase -i`, which I do **all the time** to clean up in-progress messes before merging/pushing.

sean-fitzpatrick:
In this case the whitespace change was a mistake by me that I didn't notice until GitHub gave me the diff on the pull request. Text editor added an extra indentation when I pasted from the .asy file.

PR #1287, opened 2020-05-23
Mathbook to pretext

StevenClontz:
Leaving in some `mathbook-*.xsl` shims for now, I've updated everything to use `pretext-*.xsl` that didn't already.


rbeezer:
Thanks, Steven!  Looks great.  I'll need to apply this and test how my existing branches will all behave, I'm optimistic about that.  And cleanse the documentation.  Today or tomorrow.

rbeezer:
Thanks for the report.  I see it under `pdflatex` as well.

For a `sidebyside` we have `tcolorbox` `sbsstyle` which has `raster before skip = 2.0ex`.  The manual says this "glue" is "discardable".  That word is only used twice in the manual, the other time right next to this one.

Maybe we need some "real" TeX glue?  Setting the `2.0ex` to an absurdly large amount had no effect.

So, I tried editing the actual `sidebyside` environment

```tex
\NewDocumentEnvironment{sidebyside}{mmmm}
  {\vspace*{10ex}\begin{tcbraster}
    [sbsstyle,raster columns=#1,
    raster left skip=#2\linewidth,raster right skip=#3\linewidth,raster column skip=#4\linewidth]}
  {\end{tcbraster}}
```
That `10ex` **was** effective.  But I think that gets us into the whole `\leavevmode` quagmire if we are not careful.

1.  We could try glue.
2.  We could compare `tcolorbox` version numbers, plus I will upgrade any day now.
```
Package: tcolorbox 2018/01/12 version 4.12 text color boxes
```
3.  Test adding vertical space before the enviroment (last resort) with sample article, etc.
4.  We could file a report upstream.

sean-fitzpatrick:
@rbeezer was that last comment in reply to #1288 but accidentally landed here?

rbeezer:
Right.  Thanks.  I'll move it.

rbeezer:
Dear Steven,

I added some new work last night to `mathbook-common.xsl`, not really thinking about this.  So now that file has a massive conflict when I try to merge, since the shim (a great idea) has the same name as the file with changes.  I think (but have not tested), that a `git mv` to make the re-naming would allow the merge to figure out where to go looking for the right place for the changes.

So...I tried the following

* new branch
* revert last night's work (temporarily)
* `cherry-pick` the two commits here
* `cherry-pick` the original commit that was reverted

Basically the same problem on the last step.  I think this means that a lot of in-progress work is not going to merge/rebase cleanly.  This is why this hasn't been done sooner.

So I think we need to coordinate.  And maybe there is a `git mv` commit, a commit to change references, massive rebasing (hopefully following moves/renames) in an announced 24-hour window, and **then** shims go in.  Maybe.

I'll do some research/testing.  Question for now: would you by up for doing this again when the repo is at a state more amenable?  I know this was a lot of tedious work, and I know it needs to be done, but I fear losing a lot of work, or tedious, error-prone "by hand" work to reclaim stuff.

StevenClontz:
Sure -- it wasn't more than an hour of work (mostly just search/replacing `mathbook-` with `pretext-` in the `xsl` folder). Ping me when I can pitch in.

rbeezer:
Sounds good.  Thanks for understanding.  I've got a tough rebase for EPUB work interrupted by all the Python work, which includes new XSL, so I'll get to that, and I'll test rebasing following moves/renames.

rbeezer:
I made a significant (test) change to an XSL file, on a branch.  Made a new branch elsewhere, and did a `git mv` on the same file.  Then I cherry-picked the commit with a change.  It followed the file and there was no conflict.

So I think we need to:

* Pause changes on `dev`
* Add a commit with **only** `git mv` changes
* 24 hours to allow rebasing in-progress work that will follow new filenames
* Adjust references to fix everything, and add shim files under old names.

But let me sleep on that plan.  ;-)  Might be ready for this early next week.  Are there good times for you to work on either end of a 24-hour period (two evenings?).  I've got nothing much going.

StevenClontz:
I'm flexible -- Wednesday looks to be the more restrictive on my end but I don't anticipate needing a lot of time to replay the work I did for this PR.

rbeezer:
Go ahead and make a PR that just does a `git mv` on affected files.  No references, no stub.  I'll merge tonight.

Please leave `xsl/mathbook-epub.xsl` alone.  I have plans for that and will catch up.  Thanks.

Announcements coming.

rbeezer:
First half completed on #1290.

PR #1289, opened 2020-05-25
Enable enhanced privacy for embedded video

sean-fitzpatrick:
Here is the first crack at getting "enhanced privacy mode" working for YouTube.

Possibly some adjustments needed to address questions and concerns as noted [on pretext-support](https://groups.google.com/d/msg/pretext-support/YMukQdYuwPQ/GLk1OTRZAAAJ).

- Created xsl variable `$embedded-video-privacy` (around line 400)
- Depends on element `publication/html/video/@privacy` in publisher file
- Determines xsl boolean `$b-video-privacy`
- Boolean used to set YouTube URL to `https://www.youtube-nocookie.com/embed` if true and `https://www.youtube.com/embed` if false (around line 6800)
- Default value is true.

I might have things backwards as far as the default value and boolean logic. (Maybe some `not`s are needed?) Discussed on the forum.

Documentation changes as well.

I didn't change any of the publisher files in the examples, but can if you want.

### DIFF
--- a/doc/guide/publisher/online-html.xml
+++ b/doc/guide/publisher/online-html.xml
@@ -36,6 +36,24 @@
 
             <p>The default is to first have <c>index.html</c> redirect to a page for the <tag>frontmatter</tag>, and if this is not possible, then it will redirect to a page for the top-level of your content.  If your document is short or simple, you may just have a single web page.  You do not need to distribute the <c>index.html</c> file and may just wish to use a concise and descriptive <attr>xml:id</attr> for your top-level element.</p>
         </paragraphs>
+
+        <paragraphs xml:id="video-embedding">
+            <title>Privacy options for video embedding</title>
+            <idx><h>video</h><h>HTML</h></idx>
+            <idx><h>privacy</h><h>video</h></idx>
+
+            <p>When videos are embedded in <init>HTML</init> from sites like <url href="https://www.youtube.com">YouTube</url> or <url href="https://vimeo.com/">Vimeo</url>,
+            they come with whatever tracking cookies these sites want to include. Some of these can be helpful; for example, to let the viewer keep track of what they have watched.
+            Others are designed to target advertising, and load when the page loads, rather than when the video plays, which can increase the time it takes for your book to load.</p>
+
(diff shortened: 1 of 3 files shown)


sean-fitzpatrick:
I should add: testing was against Chapter 1 of APEX on my videos branch.
This includes about 50 videos, most of which are solo, but some are in (pretext-generated) playlists.

Tested with the following:
- `<video privacy="yes"/>`
- `<video privacy="no"/>`
- `<video privacy="true"/>` (to confirm correct function if something other than `yes` or `no` is used
- no `video` element in publisher file.

With `@privacy="no"` I get the existing results from PreTeXt (with tracking enabled). All other tests resulted in enhanced privacy mode.

rbeezer:
OK, I've got this in my local repo (as a branch still), but rebased so now applying to `xsl/pretext-html`.  So it is "safe" for the renaming exercise.  You could (before dinner Tuesday):

* pull official
* rebase your local branch on `dev`
* force push (-f) to this PR

That'd let you get your last few changes in?  I'm happy to also do them to what I have right now and blend them in to your commit.  Sorry you got caught in teh middle of this.  There is never a good time for these big changes.

sean-fitzpatrick:
Done. But double-check the last change, since I can't test at the moment.
Wrapped the warning in `<xsl:when>` as suggested. My understanding of the logic was to put this *inside* the `<xsl:otherwise>`.
(When `yes`, blah... when `no`, blah... *Otherwise*, it isn't yes, and it isn't no, and by the third test, it isn't empty, so print the warning.)

Git is magic.

rbeezer:
Thanks, I should get this merged in today and can test, since I'll merge #1291 locally first.

If you want to write an announcement for `pretext-announce` send it to me off-list and I can post.  If not, let me know here and I'll do it.

magic = "that which we don't understand"  ;-)

rbeezer:
I merged the resolution of the stylesheet-naming cutover and this is working well on top of that.  It'll go live around dinnertime, once the window closes.

rbeezer:
OK, all done.  Thanks!

A lot of overhead to vary just one line, no?  Nothing is ever as easy as it seems...  ;-)

PR #1291, opened 2020-05-26
Un-break pretext-*.xsl changes and add mathbook-*.xsl shims

StevenClontz:
This should complete the `pretext-`ification of the historical `mathbook-` XSLTs.

At some point we can have the `mathbook-*.xsl` shims stop importing `pretext-*.xsl` and raise an error rather than a warning, before eventually removing them altogether.

### DIFF
--- a/xsl/README.md
+++ b/xsl/README.md
@@ -7,19 +7,17 @@ simply called "conversions."  Here we list **some** of the available
 conversions, the list is not exhaustive.  See _The PreTeXt Guide_
 for detailed documentation of use, in chapters of the part
 titled _Publisher's Guide_.
-(The `mathbook-` prefixes are historical.)
 
-
-* `mathbook-latex.xsl` - conversion to LaTeX, which can then
+* `pretext-latex.xsl` - conversion to LaTeX, which can then
 be converted to PDF, in print or electronic flavors.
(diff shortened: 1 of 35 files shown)


rbeezer:
Nicely done!  I like the clever use of `apply-imports`.  Perfect.

Merged locally and tested with my (massive) website build script and did not notice any problems.  I'm going to add a small commit to expand the warnings to include an instruction about the new/correct way to proceed.

I've got a huge documentation patch ready to go, and will release that tonight.

Thaks very much for the small push to get this finally done, and expecially doing it on a coordinated schedule.  I've always been nervous about the shock waves from this.

rbeezer:
All done.  I added a couple of follow-on commits.  Thanks again!

kcrisman:
woohoo!

kcrisman:
Not sure where to put this, but https://pretextbook.org/quickstart.html needs to be updated for this change too.  I don't think the repo for that stuff is here, though?  Sorry if this is the wrong place.

rbeezer:
Totally private, and on/off my mind the past few days.

Literally, just in time.  Fixed and now running website build script.  THANKS!

kcrisman:
>
> Literally, just in time. Fixed and now running website build script.
> THANKS!
>

Nice! I will try to use `pretext-*` today.

PR #1294, opened 2020-05-28
some Spanish translations

julian-upc:
first PR :)


kcrisman:
Thanks for contributing it!  Now we just need someone else fluent in Spanish mathematics to confirm these ... by the way, is there no way to distinguish between assumption and hypothesis en español?  Does it not matter?  This is one of the troubles with localization - maybe the distinction only matters in English.

rbeezer:
Perfect!  Thank-you very much, Julian.

> first PR

We'll look forward to more!  ;-)

kcrisman:
> We'll look forward to more! ;-)
>
>
Based on your name, perhaps contributing a de-DE file would be possible?
:wink:

>

PR #1298, opened 2020-05-30
Add first attempt at German localization

kcrisman:


kcrisman:
This is probably not so great, but at least is not 100% naive.  There are a few things in exercise and remark like objects where translation is pretty much impossible to be one-to-one.  I also don't like the two-letter abbreviations for navigation.  But at least a nontrivial small article compiles without warnings now.

Also convenient that the default for references seems to be capital letters - since in German _all_ nouns must be capitalized :wink:

rbeezer:
Thanks, KDC!  Merged, but I forgot to wrap this one up.

kcrisman:
Ok, thanks!  Let's hope actual users are kind - I already had a discussion with a heavy Sage cell user about the translation of "Evaluate", which I made more like "Run this" rather than "Evaluate this".

Issue #1299, opened 2020-05-30
Better doc for localizations


kcrisman:
Currently the readme for localizations is the only place where they are documented (see https://groups.google.com/forum/?hl=en#!topic/pretext-dev/VpXRmjgGbgw).  Tasks to fix this, according to @rbeezer:

1.  New subsection in "Overview" that is for the casual reader (curious, newbie, 
etc.) 

2.  Careful usage details for an author in "Topics" 

3.  Cross-reference (1) and (2) both ways 

4.  Developer part:  mention that contributions by email (send me entire edited 
file, so I can make an easy commit) are welcome, if a PR is too daunting. 
Pointing to the README, and the  en-US  file as a template/explanation, should 
be sufficient for anybody who goes this far.  I don't want to duplicate much of 
what is already in README and then have two copies to maintain and sync.

kcrisman:
(Yes, I'm volunteering.)

rbeezer:
> (Yes, I'm volunteering.)

Excellent!  I just **assigned** this issue to you.  That's a first.  ;-)

Off to work on #1298

kcrisman:
Hah, I didn't know that was possible.  Oh what have I gotten myself into!

kcrisman:
Possibly relevant: https://www.w3.org/International/questions/qa-i18n on which word to use.  Since the directory is already called l10n and since those files are the actual implementations as opposed to the infrastructure enabling their development, probably let's stick with that word.

kcrisman:
See #1300

rbeezer:
Implemented at #1300

PR #1300, opened 2020-06-01
Documentation for localization

kcrisman:


kcrisman:
I assume you will want to do a fair amount of editing, as the "style" does not seem Beezeresque.   But I think I hit everything, and ended up not really having that much to say for the developer side other than pointing a little more to the fact that you are happy to accept emailed files in the Appendix.

rbeezer:
Excellent!  Very Beezeresque.  A year from now I'll think I wrote it and won't believe `git blame`.  Really minimal editing.

Nits:

*  An expert online strongly suggested in an old post that the space toward the end of an empty tag was a great idea.  Another expert I was paying said that was crazy, even though he knows the first expert.  I've migrated to no space.
*  We have new elements `tag`, `tage` (empty), and `attr` that will do monospace formatting, etc.
*  I have hijacked the `note` element for a "Best Practices" series.  I think those paragraphs are fine without it.

Thanks for fulfilling your assignment on #1299.  ;-)  You'll be KDC, COD soon.  (Chief of Documentation.)  It'll go live on next push to website.

kcrisman:
I don't think I was aware of the `<cd>` and `<cline>` tags, so I'll have to update NTIC for that.  Also I note you didn't put `<p>` tags in your line items, which I remember having to very carefully and consistently do - maybe because I often have sublists.  Oh well.  Glad you liked it and glad to make a contribution _other_ than fixing your typos 😜 

Speaking of which, you added a confusing (to me) clause:

> which indicates one should use the <c>en-US</c> as a template, and for futher instruction. 

How does the last bit fit into that sentence, grammatically?  "Brief instructions [are here], which indicates one should use ... , and for further instruction."  Maybe instead "Brief instructions [are here], which indicates one should use ... both as a template, and for further instruction."  It really confused me until I said it backwards, and then Rob Beezer was sent back to the 5th dimension for a minimum of 90 days.

kcrisman:
Also, it's spelled "further". :wink:

rbeezer:
I think that sentence didn't make sense to me at first.  But since I can't spell, clearly wasn't thinking straight.

Will you make a PR to fix the typo I introduced into your PR?  JK.

rbeezer:
Patched up at 0b3387d257b4e7bbe37bb47e569cad64158d176c

PR #1301, opened 2020-06-02
W wmetadata

Alex-Jordan:
In a few branches, I have need for a utility that can identify an unused character in a string, and then use it as a delimiter for some WeBWorK/PG/Perl purpose. One of these commits adds such a utility. The other is so that we may use `$apos` in XSL attribute strings (sort of) for the apostrophe character. Like `contat('L',$apos,'Hospital')`. The two things are related, since the apostrophe is one of the first desired options for a delimiter character.


Alex-Jordan:
Now with a third commit that uses the new tool.

xsl/pretext-common.xsl
```diff
@@ -10934,6 +10943,34 @@ http://andrewmccarthy.ie/2014/11/06/swung-dash-in-latex/
...
+<!-- ################# -->
+
+<!-- Find a delimiter: find a character that can be wrapped around a string -->
+<!-- Take a string as one input and a list of characters as another input   -->
+<!-- Return the first character from the list that is not in the string     -->
+<xsl:template name="find-unused-character">
+    <xsl:param name="string"/>
+    <xsl:param name="characters" select="concat($apos,'&quot;|/!?=~')"/>
+    <xsl:choose>
+        <xsl:when test="string-length($characters) &lt; 1">
```
rbeezer, line 10957: Could this be `$characters = ''`?

xsl/pretext-common.xsl
```diff
@@ -10934,6 +10943,34 @@ http://andrewmccarthy.ie/2014/11/06/swung-dash-in-latex/
...
+            <xsl:message>PTX WARNING:  unable to find a character that is not used in "<xsl:value-of select="$string" />" from list provided. Proceeding with default '~'</xsl:message>
```
rbeezer, line 10958: You can put a `&#xa;` in here to get it to split lines when output.  Or not.

Do you want to save initial value of `$characters` in a variable and then repeat it in the warning, rather than just "list provided'?

rbeezer: Lines selected are teh last ones shown, preceding lines are automatic from GitHub.

xsl/pretext-common.xsl
Alex-Jordan: Yes! Also, it looks like it could even just be `test="not($characters)"`, because `$characters` is defined as a string based on the `select` in its `xsl:param`, and the empty string is "false" as a boolean. If testing verifies that is the case, do you have a preference?

xsl/pretext-common.xsl
Alex-Jordan: "list provided" is a remnant from an earlier version, before I gave a default for `$characters`. So yes, I'll make that change too.

xsl/pretext-common.xsl
rbeezer: I believe you are entirely correct about the simple `test="not($characters)"`.  I've made a few mistakes with exactly how "empty" is false (node sets, strings, etc), so I have shied away from it mostly.  So I prefer the comparison with the empty string, and it also reads nicely, i think.

rbeezer:
Do you like the line-by-line review?  I "resolved" both "conversations.  Better than a bullet list?

Alex-Jordan:
> Do you like the line-by-line review? I "resolved" both "conversations. Better than a bullet list?

Either works for me. Might work best to go one way or the other only, but decide per PR. Some PR's may need more pointing to specific lines of code than others.

If we fail to find an unused character, I'd like to make that a fatal error instead of failing gracefully.  In practice, the first place this will be used is when extracting WW exercises. Say you have 1000 of them and the 432nd is one where the warning would pop up. Unless we have a way to aggregate errors and print them at the end of the extraction, it will never get noticed. And it seems justified, because if you are trying to wrap a string with a delimiter character for some code purpose and it doesn't really do that, it very well could be a fatal error at the next step (PG, LaTeX, etc.). What do you think?

rbeezer:
Did you see this?

https://pretextbook.org/doc/guide/html/section-175.html

New.  Very new.  ;-)  I got rid of some `terminate="yes"` before I wrote that.  The "there will be a disaster later" rationale **is** a reason to do this.  The length of that potential-delimiter string makes me think this would be a pretty rare event?

Your call.

Can't recall:  did you use the "location-report" template?  I've got plans to make it even more useful.

Alex-Jordan:
>Did you see this?
>
>https://pretextbook.org/doc/guide/html/section-175.html

No, but thanks. Good time for me to review the developer's guide. Just one week left to go with teaching this term.

> The length of that potential-delimiter string makes me think this would be a pretty rare event?

Yes, for this Perl situation. I can see this utility being used for LaTeX `\verb` or something and maybe the list will be shorter for other uses.

> Can't recall: did you use the "location-report" template? I've got plans to make it even more useful.

 No but I will look into that.

Did you see the introduction of `$apos`? Similarly, I am now looking at making `$alphanumeric` be `0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz`, since it shows up in three .xsl files. Or would it be appropriate to use an entity? I don't quite understand what those are, but I have a sense that this could fit there. (Does `$apos` belong there too?)

rbeezer:
There are upper- and lower-case entities.  Mimic those?  They are like 100% drop-in aliases/shortcuts.  For things you don't want to type over and over and which you don't even want to look at.  So entirely appropriate ans consistent here.  Do a separate commit as part of anything else.

I saw `$apos`.  I suspect an entity will not do the side-step you are accomplishing.  Of two minds about `$apos` being global.

Alex-Jordan:
Turns out that `entities.ent` already has:

```
<!ENTITY DIGIT "'0123456789'">
<!ENTITY SIMPLECHAR "'0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'">
```

So no need to add something. Except I notice that the Braille conversion uses `abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ` without the digits. While I am thinking about this, I could add an entity for that. Like `&ALPHABET;` or `&ASCIIALPHABET;`. Or maybe not worth it until there are at least two uses.

I understand entities better now. And I agree that something for `apos` is unlikely to work. Any more thoughts about that as a variable? Maybe with capitals, so `$APOS`?

rbeezer:
I like adding "ALPHABET" by either name.  Then you will have a braille commit (if you don't already).  ;-)

Let's leave `$apos;` as-is, we can always revisit later.

Alex-Jordan:
Force pushed. Commits are now:

- `$apos`
- `&ALPHABET;`
- `find-unused-character`
- Use `find-unused-character` in WW, for wrapping "instruction" text. (Formerly it was wrapped with `?` all the time, occasionally a problem if `?` is in the "instruction" text.

Alex-Jordan:
Just letting you know that another commit is ready with "description" elements for a "webwork". No rush. It relies on this stuff here, and I don't want to push and make this PR too large for assessment. So when this is resolved, it will come soon after.

And next up after that are updates to the OPL metadata (not counting the taxonomy bit that is still in discussion limbo somewhere.) That will complete the stuff I have on this branch.

rbeezer:
I didn't know you could use a `concat()` function in an entity definition.  ;-)

Thanks for the improvements.  Keep 'em coming!

Alex-Jordan:
> I didn't know you could use a concat() function in an entity definition. ;-)

I didn't either :) Tried a few things just to see where it would go. `"&NUMERIC;$ALPHABET;"` did not work out. It made `'0123456789''ABCDE...xyz'` with extra quotes in the middle.

PR #1304, opened 2020-06-04
WeBWorK: description element

Alex-Jordan:
Allows a `webwork` to have a `description` element. If the `webwork` is PTX-authored, this goes into a `COMMENT()` in the archival (aka verbose) version of the PG file. (OPL problems tend to have a description like this as perl comments at the top of the file, but we feel that is less useful than in a `COMMENT()`. If there is user feedback, we could duplicate the description in the "traditional" way too.)

If the source is a problem file on the server, nothing is done with any `description`. This seems appropriate, since we assume server problems are managed separately. Anyway it would be unusual for a "webwork" with "@source" to not be a self-closing tag and have child elements. 

I did a file diff on the webwork extraction files, and the only changes are the good ones.

Please take a closer look at where I edited the schema for this. In particular I'm unsure if I put "SimpleLine" in a good place.

### DIFF
--- a/doc/guide/author/webwork.xml
+++ b/doc/guide/author/webwork.xml
@@ -280,6 +280,53 @@
       </p>
     </subsection>
 
+    <subsection xml:id="webwork-metadata">
+      <title>Metadata</title>
+      <p>
+        A <tag>webwork</tag> <em>without a source attribute</em> can have a plain text <tag>description</tag>.
+        This should be a summary of what the exercise asks a user to do,
+        including any relevant pedagogical details of the exercise. For example:
+      </p>
+      <pre>
+        <![CDATA[
+        <webwork>
+
+          <description>
+            Add two fractions with distinct one-digit prime denominators.
+          </description>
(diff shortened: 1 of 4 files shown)


rbeezer:
Schema question (all I've looked at).  I'll presume `TextShort` cannot substitute for `TextSimple`.  If correct, I think `SimpleLine` should be defined just before the definitions of `ShortLine` and `LongLine`.  Do not ask me why **those** are where they are.  ;-)

Alex-Jordan:
About TextShort. I'd prefer (for now) to only accept TextSimple. If there is a good reason, it could expand to TextShort later. But I believe extra templates would have to be made for things in TextShort \setminus TextSimple to come out right in a WW description field. And so in the documentation, I have it saying to use "plain text".

Alex-Jordan:
Force pushed with `SimpleLine` moved.

rbeezer:
Merged as-is.  Thanks.

PR #1310, opened 2020-06-10
WeBWorK: detect server version

Alex-Jordan:
This establishes the WW version and will give helpful error messages.

Try with `SERVER = https://webwork-ptx.aimath.org` and all should go smoothly.

Try with `SERVER = https://www.pcc.edu` and (since there is no WW server there) you will get an error message.

Try with `SERVER = https://secure.willamette.edu` and (since the version is 2.10) you will get an error message that the version is too old.

The real reason I wanted to have this is so that the new json approach to WW exercises can be conditioned on using version 2.16 or above. (Or maybe 2.17 if I'm not quick enough.) I will change the label for the AIM development server to 2.16 to keep going.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -423,7 +423,36 @@ def webwork_to_xml(xml_source, abort_early, server_params, dest_dir):
         password = 'anonymous'
         course_password = 'anonymous'
 
-    wwurl = server_url + "webwork2/html2xml"
+    ww2url = server_url + "webwork2/"
+    wwurl = ww2url + "html2xml"
+
+    # Establish WeBWorK version
+    try:
+        landing_page = requests.get(ww2url)
+    except Exception as e:
+        root_cause = str(e)
+        msg = ("PTX:ERROR:   There was a problem contacting the WeBWorK server.\n" +
+               "             Is there a WeBWorK landing page at {}?\n")
+        raise ValueError(msg.format(ww2url) + root_cause)
+
+    landing_page_text = landing_page.text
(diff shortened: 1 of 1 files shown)


rbeezer:
In the "no server" case, it seems that `ww_version_match` is `None` and then the subsequent `.group(1)` throws an `AttributeError`.  The question in `msg` then does not make much sense.  Can the `except` condition on the distinction between the `AttributeError` and the `ValueError` and provide a better message in the "no server" case?

Alex-Jordan:
Force pushed. Is this better? Still a newb python programmer here and I'm unsure about error handling routines.

rbeezer:
Looks good.  I think I like the `ValueError` better for the `None`.  I was thinking of multiple `except`, but an `AttributeError` would be a little weird.  Maybe you could catch the `AttributeError` and raise a `ValueError`?  Anyway, this is what I was thinking of:

https://stackoverflow.com/questions/6095717/python-one-try-multiple-except

I can merge as-is, which is fine.  Or you can try the above if you want more Python-Fu.  Just let me know.  I can merge before online UTMOST workshop starts on Monday, and I've not forgotten the other PR outstanding.

Alex-Jordan:
Force pushed again. Still don't feel like I know what I'm doing wrt errors and exceptions. Maybe I should sign up for a summer Python course :)

Why are we raising "ValueError" so much, even when it's a situation where it's not a "value error"?

rbeezer:
Thanks, merged as-is.

In my experience `ValueError` shows up a lot.  For example the problem here is that an author used the wrong **value** of the server name.  The choices are all pretty system oriented.

https://docs.python.org/3/library/exceptions.html#exception-hierarchy

PR #1330, opened 2020-07-06
WeBWorK: consolidate PG extraction

Alex-Jordan:
The main thing here is as the commit says: bring the three PG extraction files together finally.

The largest reason they were separated was to make it easy for `authored` to persist into the representations file. I'm ready to let that go. (Even if we want it someday, it now seems like something that assembly could handle.)

Another thing is that I was feeling that the labeling of "verbose" PG versus non-verbose was not capturing what was happening. The separation is really about making pretty, human readable PG versus making optimized PG that helps keep the base64 string short so we don't hit URL string length limits.

I tested by building the representations file in dev and this branch and doing a diff. The only changes were expected ones: time stamps, absences of "authored", and a change in certain base64 strings that has to do with cutting an unnecessary line from the macros insertion.

Maybe do the same test? I'm on a Mac, so maybe something different happens with the python on Linux.

Once this is cleared, I can start fine-tuning the work on `python.py` and `pretext-html` that will use the json manipulation approach (but only for WW servers 2.16+). I have submitted a tidy PR to WeBWorK to support that.


rbeezer:
Yahoo!  Visual inspection looks good.

I've forgotten all the reasons, but I've taken to import whole modules, not just names from within (`lxml.etree.ElementTree`).  I might test a small change there when I do the testing you suggest.

Alex-Jordan:
I remembered that from a thread somewhere. I think David gave a clear reason why. But my attempt to import the whole module failed. If you make a change like that and it works, maybe send it to me first so I can see if it causes an issue on a Mac.

rbeezer:
Very good!  Very happy to consolidate those files.  ;-)  I was wrong about the whole `import`business.  I had not recalled the nature of the `ElementTree` class correctly.  So merged with no changes.  Updated website versions right now...

PR #1335, opened 2020-07-29
Wwpatch

Alex-Jordan:
Two commits here.

- I think it is safer to treat radio button and checkbox answers as if they do not by themselves start a new paragraph. Most authors will _want_ them to start a new paragraph, and they can easily use a `p` for that. But this way more closely mimics what WW itself does with these kinds of answers, where it's possible to have the first item start within the preceding paragraph if you really want that.

- I addressed the `webwork-domain` issue, but not using `b-has-webwork-reps`. I don't think the warning about finding a value for `$webwork-domain` is necessary. If there is a problem with that, the `pretext` script should have hit it. At the same time as making this change, I consolidated all WW top-level stuff (global params, variables) to one place. And cleaned up a thing or two with the templates that load webwork js.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -2547,6 +2547,18 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     <xsl:text> original</xsl:text>
                 </xsl:otherwise>
             </xsl:choose>
+            <!-- reliable way to distinguish adjacent hint/answer/solution knowls -->
+            <xsl:choose>
+                <xsl:when test="self::hint">
+                    <xsl:text> hint</xsl:text>
+                </xsl:when>
+                <xsl:when test="self::answer">
+                    <xsl:text> answer</xsl:text>
+                </xsl:when>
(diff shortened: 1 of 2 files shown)


rbeezer:
Looks good in visual scan.

Those variables will just test as empty/null/false which should work out fine.  At some point we should advise authors to regenerate their representations if they've been hoarding them?

Alex-Jordan:
Third commit adds "hint", "answer", or "solution" class to knowls for such things, as discussed in the forum. @davidfarmer  signed off on using a class for this purpose. Feedback welcome if you'd like a different approach.

rbeezer:
I was partway through a review of the first two commits, so I just merged those two now.  I think I can review the third OK as it sits, so no need to make a new PR.  But I'll need to pop a few more things off the stack first.

rbeezer:
The third commit has been superseded by 1f7072fdd9b437a2abbd9a60bfeca3e8cd694ca3.

Discussion at

https://groups.google.com/d/topic/pretext-dev/1aV-iRLywsg/discussion

PR #1338, opened 2020-08-05
Integrates New MathJax 3 script

zorkow:
* Adds the new script under `pretext/pretext.js`
* Updates the `pretext.cfg` accordingly
* Adds `package.json` and `update-sre` script for `npm` handling
* Integrates the new script into the `pretext.py` script. 

### How to run 
* Install with
```bash
./update-sre
```
in `pretext` folder.
* Then run as usual.

### Tests and Problems

I believe that I get the correct output for the braille test: 
```
./pretext/pretext -vv -c all -f braille examples/braille/braille-test-book.xml
```
However, I got errors when attempting conversion with `epub-kindle`, `epub-speech`, although the `mj-output-kind.html` and `mj-output-speech.html` files looked correct. I assume there is some post-processing that needs to be adapted.

### Future Considerations

* The MathJax output is now in a `mj-data` tag that could be used for processing.
* `mj_tag` variable contains the actual inner tag for the requested output format in the form of `mj-braille`, `mj-speech` etc.
* Not sure whether the currently handling of `&nbsp;` is still correct or necessary.

### DIFF
--- a/pretext/package.json
+++ b/pretext/package.json
@@ -1,0 +1,7 @@
+{
+  "dependencies": {
+    "mathjax-full": "^3.0.5",
+    "speech-rule-engine": "^3.1.0-beta.4",
+    "yargs": "^15.4.1"
+  }
+}
--- a/pretext/pretext.cfg
+++ b/pretext/pretext.cfg
@@ -53,5 +53,5 @@ pdfeps = pdftops
 pdfcrop = pdf-crop-margins
 pageres = pageres
 # Following is a suggestion, node.js modules don't seem to be on path
-mjpage = /home/user-name-here/node_modules/mathjax-node-page/bin/mjpage
+mjpage = pretext/pretext.js
 liblouis = file2brl
--- a/pretext/pretext.js
+++ b/pretext/pretext.js
@@ -122,7 +122,12 @@ function action(state, code, setup = null) {
     const adaptor = doc.adaptor;
     setup && setup();
     for (const math of doc.math) {
-      code(math, doc, adaptor);
+      try {
+        code(math, doc, adaptor);
+      } catch (err) {
+        const id = adaptor.getAttribute(adaptor.parent(math.start.node), 'id');
+        console.error('Error on item ' + id + ': ' + err.message);
+      }
     }
   }];
 }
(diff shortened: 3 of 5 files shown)


rbeezer:
Cool.  Looks great.  Sorta like a JS programmer tried to make Python look like JS.  ;-)

This will be really helpful.  All the new MJ container names mean the post-processing XSL will need to be adjusted slightly.  But this will save a huge chunk of time and do it right.   Thanks.

zorkow:
> Cool. Looks great. Sorta like a JS programmer tried to make Python look like JS. ;-)

Thanks! It's been a few years since I've done anything larger than a script in Python. But I remembered immediately, why I don't like it: All those colons and necessary whitespace. It's like programming Cobol!


> This will be really helpful. All the new MJ container names mean the post-processing XSL will need to be adjusted slightly. But this will save a huge chunk of time and do it right. Thanks.

Talking about efficiency: You might want to consider replacing some of your case analyses by dictionary lookup. In interpreted languages they tend to be faster in particular when repeated often in loops (constant vs linear).

rbeezer:
>  You might want to consider replacing some of your case analyses by dictionary lookup.

Spot on.  That's exactly what caught my eye.  I'm writing Python like an XSL programmer.  ;-)  Lots of case statements, and I can't even think how to fake a dictionary.

rbeezer:
Line 96 of `pretext.js` is

```javascript
const {sreReady} = (needsSRE ? require('mathjax-full/js/a11y/sre.js') : {sreReady: Promise.resolve()});
```
which I think means I am not getting the newest SRE until `mathjax-full` says so (and it is stuck on a much older version)?  I tried to "require" my `beta.5` version of SRE but was not able to hack that together.

I can't see my way through to how the `SRE` constant is even defined.  All my attempts at changes leave it undefined (it appears).

zorkow:
MathJax always comes with a fixed version of SRE, so SRE updates won't break it. `npm` is smart enough to retain that dependency by moving the old version to a local `node_modules` directory, when you try to override it by installing the latest version of SRE with `npm install`.

I have provided the `update-sre` script that does all the manipulations necessary to update SRE to the latest version so MathJax 3 picks it up without any changes to the `.js` script.

Have a look if that works for you.

zorkow:
The [README of Davide's gist](https://gist.github.com/dpvc/386e8aac18c010361ef362b9237c71e9#file-readme-md) also explains that. My `update-sre` script effectively does the `npm install` and removal of the local `node_modules` directory in one go.

rbeezer:
OK, I was being a bit **daft**.  (Is that part of the Queen's English?)

I saw the bash script, and mis-understood its purpose.  And had forgotten David's instructions.  Sorry for the noise.

Working now - I just need to adjust to receive things like `mjx-data`, etc.

rbeezer:
Thanks very much, @zorkow  and @dpvc!

I consolidate these four commits into two commits, then integrated the results into subsequent processing with 8e24e9ebb681b4f6010b4a085bc49a92898bd54f, and then did some cosmetic rearrangements on 98f21740783f166a773df4dc83cab5293ab63a4a.

So now in great shape it seems and can move forward!  Thanks again

dpvc:
Glad the script is in place.  I hope it speeds things up for you.  Nice to see you again ash AIM, if only virtually.

PR #1339, opened 2020-08-09
LaTeX: corral leavevmode and apply to paragraphs/sidebyside

Alex-Jordan:
This addresses an issue reported in the [forums](https://groups.google.com/d/msg/pretext-support/oxbk1KJ1PxQ/ZY62v2mxCAAJ) where a `sidebyside` that is the first child of a `paragraphs` causes an issue with the title of the paragraphs.

I also took the liberty of moving most instances of `\leavevmode` to the same area. 

The approach I took was to tie the `\leavevmode` to the thing it precedes, not to the parent of that thing, an alternative which was considered. 

Aside from expected diffs (timestamps and different generated id~s) I don't see any diff on the sample article or sample chapter.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -8165,7 +8165,7 @@ the xsltproc executable.
             <subsection>
                 <title>Testing a Side-By-Side First</title>
 
-                <p>A <tag>sidebyside</tag> that appears first within some other container can wreak havoc in <latex /> output.  Below we have this situation twice, once in an <tag>activity</tag>, then in an <tag>example</tag>.</p>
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Force pushed after splitting this up in two commits.

rbeezer:
I built a test case for the sample article, where the `paragraphs` has a `title` (as it would normally) and the fix did not make it in.  We should include tests like this, so that delicate matters don't break again later.  Or more precisely, when they do break **somebody** might notice.  ;-)  Easier for me to review if the test is on the first commit so new content can stay in place as code comes/goes and ids, etc don't change.

I think the `preceding` condition needs the "metadata" filter in use on the "sage" situation.

Alex-Jordan:
I swear it worked with my minimal example file, but I must have changed the match near the end and not really re-ran a test.

OK, now with example in the sample chapter and the right filter to make it work.

rbeezer:
Great!  Thanks for working the fix AND getting all these exceptions consolidated.

PR #1346, opened 2020-08-12
Pythonwarning

Alex-Jordan:
First commit addresses something @mitchkeller noticed. WW representations build was not closing a file nor the requests session, and that leads to resource usage warnings if python is set to show all such warnings.

Test by changing Makefile from `$(MB)/pretext/pretext -v -c webwork -d $(WWOUT) -s $(SERVER) $(SMPCHP)` to `PYTHONWARNINGS=module $(MB)/pretext/pretext -v -c webwork -d $(WWOUT) -s $(SERVER) $(SMPCHP)`, same for sample chapter. You will see the warnings at the end of reps building, but not after this commit. Output reps file has no meaningful diff.

Second commit is just that I happened to notice sessions being initiated in other subroutines that appear to never be used.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -343,6 +343,7 @@ def webwork_to_xml(xml_source, pub_file, stringparams, abort_early, server_param
     # "run" an assignment for the list of triples of strings
     ww_file = open(ww_filename, 'r')
     problem_dictionaries = ww_file.read()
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, @Alex-Jordan and @mitchkeller!

PR #1349, opened 2020-08-14
publisher file works on windows

oscarlevin:
The absolute path to the publisher file must use forward slashes.  On windows, os.path.abspath() produces the correct path with backslashes.  This fix ensures the separator for the path to the publisher file is a forward slash.

Since a user can specify the publisher file as a string param (using -x), this is checked for and fixed as well.

### DIFF
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -198,12 +198,20 @@ def main():
     # publisher file is optional, so allow for None here
     if args.publisher_file:
         publisher_file = verify_input_file(args.publisher_file)
+        #ensure forward slash to pass to xsltproc:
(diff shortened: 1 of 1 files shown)


rbeezer:
OK, I see where you are going with this.  Ignore Windows-specific comment on https://github.com/PreTeXtBook/pretext-cli/issues/34

rbeezer:
Thnaks for catching this and fixing it!  And sneaking in a `publisher` parameter with the other string parameters is not something I'd ever think of happening.

I added one condition: if a publisher file is declared two ways (`-p` and `-x`), then `-p` wins.  In other words, catching the `-x` version will now not overwrite the `-p` version.

Thanks!

PR #1350, opened 2020-08-14
clarified pretext script documentation; added xref to webwork section

oscarlevin:
I had some trouble getting the pretext script to accept string params, before I realized that -x had to not be the last option.  I've added an example to the guild.

Also, cross-ref to webwork-publisher from the alert in the webwork section.


rbeezer:
Very good.  That `-x` at the end has bitten me before.  Thanks.

PR #1351, opened 2020-08-15
Removes MathML enrichment.

zorkow:
This fixes [SRE issue 394[(https://github.com/zorkow/speech-rule-engine/issues/394).

The reason that the separator showed up was that an already enriched expression was sent to SRE. I.e., SRE processed the expression twice, which led to the added invisible separator being treated as an explicit one.

@dpvc was there a reason to call the enrichment? After removing it, I can't see a change of behaviour on `--svg` or `--mathml` output, so I don't think we need that there.

### DIFF
--- a/script/mjsre/mj-sre-page.js
+++ b/script/mjsre/mj-sre-page.js
@@ -153,7 +153,7 @@ const renderActions = {
   pretext: action(STATE.PRETEXT, (math, doc, adaptor) => {
     math.outputData.pretext = [adaptor.text('\n')];
     if (needsSRE) {
-      math.outputData.mml = SRE.toEnriched(toMathML(math.root)).toString();
+      math.outputData.mml = toMathML(math.root).toString();
     }
   }),
   //


dpvc:
> was there a reason to call the enrichment?

I thought SRE needed the MathML to be enriched.  Does `toSpeech()` do that itself already?  if so, then your edit is good.

It might be good to have SRE check to see if there are already `data-semantic` tags and not enrich again if there are.  Since people sometimes save MathJax output, it is possible that already enriched MathML may be sent to SRE again.

zorkow:
> > was there a reason to call the enrichment?
> 
> I thought SRE needed the MathML to be enriched. Does `toSpeech()` do that itself already? if so, then your edit is good.
> 

We just need the enriched structure on the actual content in the page (SVG, CHTML) to be able to walk it. SRE can recompute its semantic tree from that and at this point there is no more MathML involved. The enriched MathML is just a trick to get MathJax to put the `data-semantic` attributes into the page.

SRE computes speech/nemeth etc. without  having to compute the enrichment first.

> It might be good to have SRE check to see if there are already `data-semantic` tags and not enrich again if there are. Since people sometimes save MathJax output, it is possible that already enriched MathML may be sent to SRE again.

True. I'll make a feature request issue for that.

rbeezer:
Thanks very much, @zorkow and @dpvc!

With Alexei, I'm making good progress on some test-suites at https://github.com/PreTeXtBook/a11y, so I'll get those in place and then merge this, and move to SRE 3.1.  Thanks for all the updates.  Our test with a **real** book and a **real** student is going pretty well.

rbeezer:
Thanks, again.  Interestingly, no change in output for the test suite from abstract algebra, and seemingly only changes in differentials (the original symptom) in the calculus suites.  SRE 3.1 next.

PR #1352, opened 2020-08-16
Only throw PTX:WARNING on nonempty <webwork>

daverosoff:
The empty <webwork/> element is unrelated to the presence of WeBWorK problems in the source and should not throw this warning.


rbeezer:
Thanks - that was a bit sloppy of me!

Let me guess - you have a syllabus talking about WW but without any actual problems?

daverosoff:
> Thanks - that was a bit sloppy of me!
> 
> Let me guess - you have a syllabus talking about WW but without any actual problems?

Got it in one ;)

PR #1353, opened 2020-08-16
Asymptote files added to examples/showcase

mdoob:
Added asymptote samples:
1) Three short flag samples side by side
2) A five-ellipse Venn diagram
3) A 3d file that can be manipulated with WebGL commands.


Alex-Jordan:
Very nice!

My two cents: it would be entirely appropriate for asymptote images to get their own page separate from the latex-image page.

rbeezer:
Beautiful!  But why is the Maple Leaf first?  You've seen the 17-set Venn diagrams?

We need `*.pdf` to go with LaTeX - can you add those?  A second commit would be great - I'd combine them.  Otherwise, this looks real good.

What do you think of putting something in the Death Star caption that indicates it is "rotatable"?  I almost forgot to have fun with it and only knew to because of your description here.

mdoob:
I started to make an extra section, Alex, but then I thought that for a first go I would make changes to as few files as I could get away with (as it turns out, one!), and so I just put them into subsections. I have no strong preference either way.

Rob: I put the flags in alphabetical order. No big political message there (:-)).

With the deathstar, I put those little yellow balls in place so you can zoom in and see the reflected light sources. Asymptote, you're so sweet!

A further note: I made the  text files containing the asymptote code by cutting and pasting from the ptx file. It really is the actual source.  Would it make sense to read the asymptote source from the text file in the images subdirectory? Can this be done with xsltproc? 

I guess either the pretext script or John's wonderful internet site could make the pdfs. I'll get at it when I have a few minutes.

I wrote a note once to answer the following question: You want to make a Venn diagram with n closed curves in the plane that pairwise intersect at k points.  How big must k be? I don't remember the details, but it was all trivial counting. For n=5, it turns out that the curves must intersect in at least k=4 points, so ellipses jump out at the right thing to use. Asymptote gives us a proof by example.

rbeezer:
Alphabetical.  Right.  ;-)

The "real" procedure is that the source should be generated from the PreTeXt via the script.  How it is, however you did it, is fine, so long as they match.  Let's not get fancy about getting the source from individual files.  Follow Alex's lead (as you have done) on putting archive files in the `images` directory.

You know [this work](https://www.combinatorics.org/files/Surveys/ds5/ds5v3-2005/VennEJC.html)?  Another Canadian resident.

mdoob:
Some good news and some bad news. I made some small changes along the lines we discussed, and
created pdfs using the pretext script. All but the fancy 3d example ran with no problem. 

Running asy -f pdf on the 3d example works when not using the -noprc option. When the option is
used an error  message is thrown:

malloc(): unsorted double linked list corrupted

This may take a bit of time to resolve. I need to get a minimal working example.
You might remember that I found a tiny asymptote bug during one of our Tuesday
sessions. I reported and it is fixed now, but it was three weeks before anyone looked
at the PR. 

I'll send a new PR with the little bits added to the material from this morning.
In light of the missing deathstar.pdf, you'll have to decide how much to make public.

mdoob:
Did I do that right? Can you see both commits now?

rbeezer:
I believe so!  But on fone, so can't inspect too carefully just now.

rbeezer:
A. Remove Death Star entirely (save your work somewhere!), I merge now, you make a second follow-on PR later.
B.  Let this sit as-is, add PDF when sorted, and I merge then.

Your call, immaterial to me.

mdoob:
OK --- I pecked away got something that might help.

Imitating the pretext script
asy -f pdf -noprc -iconify -batchMask Deathstar.asy 
fails but
asy -f pdf -noprc  -batchMask Deathstar.asy 
succeeds. 

Do we really need the iconify option?

mdoob:
More news, this time better. I was thinking about the unusual error messages and thought that it must be coming for one of the linked libraries. In fact ldd tells me there are 68 of them. Then I remembered that I was going down this rabbit hole for something else and found that PixelBook with the google libraries worked easily. So I tried things there (oh git, you make my life better) and there were no problems. The 3D file reads with evince, so I think we're now OK. I pushed a third (!) commit and  we seem to be home now.

rbeezer:
Excellent!  Thanks, this is a great contribution.  It'll go live on the next website update.

* White stripes on US flag are transparent.  Maybe Maple Leaf too?  Make a new PR if you think they should really be white.
* Diff on PTX source is really big!  Not your fault, just unexpected.  Not certain, but maybe a small commit for subsection structure, then a big add for new content in Subsection 2 would have been cleaner.  **NOT** a problem - I just learned something,

mdoob:
Good point about the flags. I did the Canadian and US ones some time ago by going to the government web sites to get the official specifications. Colours are given in both Pantone and RGB values. I added the EU flag from their web site on Saturday. I'll check back to see if white is specified in each case. It might be slightly off white, but in any case it is a trivial fix. I can do this after dinner tonight (I don't have kids aged three and five anymore (:-)).
Observations:
1. Each flag can be done in  just a few lines of code. 
2. Even if you don't know asymptote, it is so {C, C++, java}-like that anyone can read it.
3. There is nary a < or > or & to be seen in the code.

PR #1355, opened 2020-08-18
Asymptote showcase examples

mdoob:
This commit fixes the background to the Canadian and US flag, and adds one more example to show a plot using asymptote. Please use git log to see if I've messed up the merge. Sorry.


rbeezer:
Yep, that was a bit of a mess.  ;-)  Since I rebased your previous work, all the commit hashes change, and you need to clean-up your fork, not build further onto old branch.  It is now useless.  See "Cleaning Up" at the bottom of https://pretextbook.org/doc/guide/html/section-185.html, and especially note that I added new work on `dev` between your two pull requests.

Flags look great, and the area example is very nice.  Thanks.  Merged, but have not re-built website yet.

PR #1356, opened 2020-08-18
another typo

kcrisman:


rbeezer:
Oops - I lost track of this one.  And that paragraph got edited in the interim, and the typo fixed collaterally.  So nothing to do here.  Sorry!

Thanks all the same!

PR #1362, opened 2020-09-01
Separate 'narrow' from 'description-list'

davidfarmer:
For a narrow description list, it is better to have
   class="description-list narrow"
than
   class="description-list-narrow"
to avoid having to repeat selectors for behavior which is common to the two cases.

The CSS has been updated and I checked that this change has no effect on the
appearance of  lists.html#subsection-43  in the sample article.


rbeezer:
Merged, and pushed. Website being rebuilt right now, give it a few.  Gotta love those single-character commits!  Thanks.

PR #1364, opened 2020-09-05
Pretext: Allow command-line options in pretext.cfg

jjrsylvestre:
Allow users to specify extra command line options for specified commands in pretext.cfg.

For example, I recently needed to pass the threshold option to pdf-crop-margins, as some images that were cropped fine by old pdfcrop were being more aggressively cropped by new pdf-crop-margins.

(That's the only case I tested this code on, but since all the code edits are essentially the same, it *should* all work fine for the other commands. *fingers-crossed*)

### DIFF
--- a/pretext/pretext.cfg
+++ b/pretext/pretext.cfg
@@ -30,6 +30,12 @@
 # Only the options needing changes need to be listed in custom version
 # Once you make a copy, you will need to add new configuration items
 # "by hand", which is why there is a history of changes below
+#
+# Note that you may include command-line options for your commands, like so:
+#   pdfcrop = pdf-crop-margins --threshold 200
+# Sometimes command-line options require an equals sign. No problem:
+#   pdfcrop = pdf-crop-margins --threshold=200
+#
 # ***********************************************************************
 
 # History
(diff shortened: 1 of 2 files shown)


rbeezer:
A visual scan of the code looks good.  I wasn't sure how you were going to accomplish that -- I like it in the config file.

If there was **any** good documentation, I'd suggest you add to it.  :-(

jjrsylvestre:
OK, added some rudimentary documentation to the stock pretext.cfg itself.

The documentation for configparser says that occurrences of delimiters in values are allowed, and sure enough it works fine. Though to avoid confusion you might consider changing the delimiter in pretext.cfg to : instead of =.

rbeezer:
@jjrsylvestre Sorry - I lost track of this one until you mentioned it a while back.  I have some serious Python work to do soon, so I'm saving this until my head is into that work.  Soon!  I really don't want to forget again!  ;-)

rbeezer:
Thanks, Jeremy.  Merged this just now, even though I think I've made a mess for myself with some other pending work of mine.  ;-)

Why?  You'll see a need on `pretext-support`.  I'm posting there now.  Perhaps you can have a look?

Nice work.  And timely.  Thanks again.

rbeezer:
@jjrsylvestre I think this needed one little fix-up.  You mind taking a peek and make sure I did it right?  Seems to function just fine.  Thanks.

See 14889567d57a9c863b1596166ca9af911a571bed

jjrsylvestre:
Looks good, not sure how I missed that one. Hope it didn't cause extra
traffic on -support.

rbeezer:
Looks like there were other adjustments there, since it was a bit more complicated than the others.

No problem - I hit it while testing some image management work, and got it fixed pretty quickly.  My own personal support list.

Thanks for the quick check!

PR #1372, opened 2020-09-26
WeBWorK: suppress 'Make Interactive' button when there are no answer …

Alex-Jordan:
…input fields

Some PG files either never ask a question, or they only ask for an essay response. At present, making these interactive is at best silly, at worst disappointing to the reader. This commit causes any such problem to be rendered in only its static form in HTML.

For example, problems 2 and 3 here:
https://pretextbook.org/examples/webwork/sample-chapter/html/section-4.html
will now just come out static.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -9307,7 +9307,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                                           (ancestor::worksheet and ($webwork.worksheet.static = 'yes')) or
                                           (not(ancestor::exercises or ancestor::reading-questions or ancestor::worksheet) and ($webwork.inline.static = 'yes'))"/>
     <xsl:choose>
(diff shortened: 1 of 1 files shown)


sean-fitzpatrick:
This is a nice improvement! (Of course I just finished building 4 versions of APEX HTML ;-) )

Alex-Jordan:
@sean-fitzpatrick It never ends :)

rbeezer:
Looks good.  I'll see about merging later today.

rbeezer:
I added parentheses on the first part of condition.  Perhaps not necessary, but I got caught out once in a similar situation, and now it is habit.

Merged un-tested.  ;-)

sean-fitzpatrick:
New builds up, [with videos](https://opentext.uleth.ca/apex-standard/) and [without videos](https://opentext.uleth.ca/apex-accelerated/). The ones being used here for teaching get built first.
Tomorrow I'll build two more versions that don't have the Parts divisions.

PR #1373, opened 2020-10-02
LaTeX: separate out exercise styling

Alex-Jordan:
This does some reorganization of styling for divisional exercises. The purpose is to allow you to do things like style `worksheet` exercises independently from `exercises` exercises.

For example, a worksheet exercise could revert to default `tcolorbox` styling with:
```
--stringparam latex.preamble.late '\tcbset{ worksheet/.style={} }'
```

It replaces environments like `divisionexerciseegcol`, etc., all under the common umbrella of `divisionexercise`. And `divisionexercise` now gets an additional argument that identifies which type of exericse it is. At present, nine types: (exercises, worksheet, reading-questions) x (regular, exercisegroup with one column, exercisegroup with more columns). Styles for individual exercises like `worksheet` and `exercises-egcol`. This approach should scale up if more kinds of divisional exercises come into existence. For example, a decision to style `subexercises` differently. Or if `exercise` were to become a schema-compliant child of an `activity`. Etc.

Parallel changes to the divisional exercise solutions environments.

### DIFF
--- a/xsl/pretext-latex.xsl
+++ b/xsl/pretext-latex.xsl
@@ -1421,47 +1421,40 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <xsl:apply-templates select="$instance" mode="environment"/>
         </xsl:if>
     </xsl:if>
+    <!-- Divisional Exercises -->
+    <!-- Three types: within exercises, worksheet, reading-questions    -->
+    <!-- Three types: not in exercisegroup, in eg with 1 col, with more -->
+    <xsl:variable name="exercise-reps" select="
+        ($document-root//exercises//exercise[not(ancestor::exercisegroup)])[1]|
+        ($document-root//exercises//exercisegroup[not(@cols) or (@cols = 1)]/exercise)[1]|
+        ($document-root//exercises//exercisegroup[@cols > 1]/exercise)[1]|
+        ($document-root//worksheet//exercise[not(ancestor::exercisegroup)])[1]|
+        ($document-root//worksheet//exercisegroup[not(@cols) or (@cols = 1)]/exercise)[1]|
+        ($document-root//worksheet//exercisegroup[@cols > 1]/exercise)[1]|
+        ($document-root//reading-questions//exercise[not(ancestor::exercisegroup)])[1]|
+        ($document-root//reading-questions//exercisegroup[not(@cols) or (@cols = 1)]/exercise)[1]|
+        ($document-root//reading-questions//exercisegroup[@cols > 1]/exercise)[1]"/>
+    <xsl:if test="$exercise-reps">
+        <xsl:text>%%&#xa;</xsl:text>
+        <xsl:text>%% tcolorbox, with styles, for divisional exercise environments&#xa;</xsl:text>
+        <xsl:text>%%&#xa;</xsl:text>
+    </xsl:if>
+    <xsl:for-each select="$exercise-reps">
+        <xsl:apply-templates select="." mode="environment"/>
+    </xsl:for-each>
+    <!-- EXERCISE AND PROJECT_LIKE SOLUTIONS -->
     <xsl:if test="$document-root//solutions or $b-needs-solution-styles">
         <xsl:text>%% begin: environments for duplicates in solutions divisions&#xa;</xsl:text>
         <!-- Solutions present, check for exercise types     -->
         <!-- This may have false positives, but no real harm -->
-        <!--  -->
-        <!-- solutions to inline exercises -->
-        <xsl:if test="$document-root//exercise[boolean(&INLINE-EXERCISE-FILTER;)]">
-        <xsl:text>%% Solutions to inline exercises, style and environment&#xa;</xsl:text>
-            <xsl:text>\tcbset{ inlineexercisesolutionstyle/.style={bwminimalstyle, runintitlestyle, exercisespacingstyle, after title={\space}, breakable, parbox=false } }&#xa;</xsl:text>
-            <xsl:text>\newtcolorbox{inlineexercisesolution}[3]</xsl:text>
-            <xsl:text>{inlineexercisesolutionstyle, title={\hyperref[#3]{</xsl:text>
-            <!-- Hardcode "name" of an inline exercise in the environment -->
-            <xsl:call-template name="type-name">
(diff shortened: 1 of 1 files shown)


rbeezer:
OK, I think I understand this one mostly.  `tcbset` via `latex.late.preamble` will not be the way to effect styling.  Rather, there are hooks so that XSL templates can augment (and overwrite) existing tcolorbox styles.  Is this PR ultimately about supporting styles?  (Can't tell if it about changing uses of tcolorbox styles for "structural" properties, or for visual appearance).

In either case, I need to understand how/what can be styled in a manner consistent with the existing modal `tcb-style` templates.  What will the `match` be?  What will be affected?  Can you hack a quick commit (which might be disposable) that will put some exercises into a purple box, just as a demonstration?

I'm mostly uncertain about the new parameter to the environments and how that affects the above (for better or worse).

Nits:

* Structural or visual, rightly or wrongly, anything that is a tcolorboxstyle has a name that ends with "style", no punctuation.  Only way I can keep track of what's what.
* Again, rightly or wrongly, I set environment names without punctuation, since I expected them to be internal.  Above might influence that.
* With new options for the number of columns in an `exercisegroup` the preamble will make more environments than necessary.  In other words, presence/absence of `@cols` will not control any environment creation.  (`@cols=1` is now legitmate.)  Simply presence of `exercisegroup` will cause all variants to be built.  I'm avoiding making that change on current work, so we do not have conflicts here.  So you could make this part of your changes, I think.

Alex-Jordan:
>  tcbset via latex.late.preamble will not be the way to effect styling.

Right, sorry for the mislead. My point was more like "someone can redefine the `worksheet` style (or other styles) very esaily". It would be a style writer doing the `\tcbset`.

> In either case, I need to understand how/what can be styled in a manner consistent with the existing modal tcb-style templates. What will the match be? What will be affected? Can you hack a quick commit (which might be disposable) that will put some exercises into a purple box, just as a demonstration?

I'm not sure what the sanctioned way is for making new styling. But something like:
```
\tcbset{worksheet/.style={divisional, colupper=purple}}
```
would put purple font in `worksheet/exercise`, but not in `exercises/exercise`. The use of `divisional` there keeps the default styling/tcbstructure and just tacks on purple font.

Somewhere  in between style and structure are matters of spacing. An indentation width that differs from the default might be a more realistic use for this than purple font.  

Another real use may be a font change. Something considered less formal.

I did not know about the modal `tcb-style` templates. This commit is compatible with moving the exercise and solution styling to using modal `tcb-style` templates too, if that's desirable.

>I'm mostly uncertain about the new parameter to the environments and how that affects the above (for better or worse).

The new parameter only goes to the `divisionexercise` and `divisionsolution` environments, which do not appear to be using modal `tcb-style` templates. 

> Structural or visual, rightly or wrongly, anything that is a tcolorboxstyle has a name that ends with "style", no punctuation. Only way I can keep track of what's what.

OK. I downsized because the verbosity was getting high. But that's no problem.

> Again, rightly or wrongly, I set environment names without punctuation, since I expected them to be internal. Above might influence that.

Also OK. Naming convention standards are good. Can `eg` become `exercisegroup`? One in particular that was causing my eyes to cross was `divisionexerciseegcol`.

> With new options for the number of columns in an exercisegroup the preamble will make more environments than necessary. In other words, presence/absence of @cols will not control any environment creation. (@cols=1 is now legitmate.) Simply presence of exercisegroup will cause all variants to be built. I'm avoiding making that change on current work, so we do not have conflicts here. So you could make this part of your changes, I think.

Should I make a modal utility template for cols that returns "1" when @cols is missing? Would that be used elsewhere?

rbeezer:
> I'm not sure what the sanctioned way is for making new styling.

> I did not know about the modal tcb-style templates.

That's all the point of asking for a demonstration.  ;-)

> which do not appear to be using modal tcb-style templates

Exactly.  Just not yet, but perhaps we do not want to close-out on that possibility.

> Should I make a modal utility template for cols that returns "1" when @cols is missing?

Done.  Maybe last night.

Alex-Jordan:
> That's all the point of asking for a demonstration. ;-)

You're going to have to spell things out for me more. I'm just not following this part of the thread.

rbeezer:
Well, I did a bad thing and refactored a bunch of the code this PR sits on.  Sorry.  But in the win column, it removed about 3% of the code for the LaTeX conversion.  I think there is another refactoring to make, but I need to digest this one first.

All the changes affect constructing the **body** of the LaTeX file, not the **preamble** and only introduce extra % that were not there before (except one little rename).  Within lines 5257-5836 and 6486-6786 as displayed here (I think).  So the changes here for preamble construction might be just fine.  And parts of the body could be a lot easier, or highly similar.

There's been far too much evolution and addition for exercises, so it is imperative that I get things consolidated and organized, before it all falls apart.  In particular, I can't think about greater flexibility in the context of styling yet.  So I'd like to wait on this.

I want to first accomodate reading questions, coding exercises, multiple choice, etc, in the context of Runestone hosting.  That's the reason for this now.

If there is something really wrong for worksheets we can discuss that.

Alex-Jordan:
IIRC, this was triggered by things I was working on with `worksheet` before I started ODT. I'm probably forgetting things, but two things that I think would be good:

* Allow worksheet exercises to have styling that differs from exercises exercises in LaTeX output. More broadly, enable a `worksheet` to stand out when flipping through a book.
* Enable `subexercises` inside a `worksheet`, with possibility for horizontal progression. I'm concerned it's not going to fly with ODT to use `sidebyside/exercise` without hard-coding exercise numbers. [And hard-coding numbers would run against the point of making ODT: to make something that is easy for the lay-instructor to make their own edits to.]

On that last point, we'll see. Sidebyside is coming up on my ODT list, after I finish math, tabular, and image. You've observed that getting something to work in LaTeX and HTML has been good for abstracting templates. With a worksheet, HTML is not a factor. The two print mechanisms will be LaTeX and ODT, and settling on markup that works for both of these is sort of the parallel.

For now, no need to try to salvage what was here. At least I learned better how you have structured LaTeX output. [Or maybe you just changed all that!  ;)  ]

rbeezer:
Got it.  I'll close tnis, but it'll still be available if we want to go back and mine it.

I hope you can get `sidebyside` sorted - then moving a whole document to ODT might start to sound feasible.

PR #1375, opened 2020-10-09
Update pretext-common.xsl

Alex-Jordan:
An oversight?


rbeezer:
Thanks - definitely some editing mistake.

PR #1377, opened 2020-10-14
Fix plusminus handling

Alex-Jordan:
I suspect (but am not sure!) that these are mistakes. Maybe one mistake, then a copy-paste into the other stylesheet.

Of course I am only looking at this as I crib stuff over into the odt stylesheet.


rbeezer:
Thanks - definitely some mistake as I flip-flopped on exactly how to handle these.

PR #1381, opened 2020-10-20
Odt

Alex-Jordan:
Probably not for serious merge consideration. If you would like to idly browse code with your morning coffee, any XSLT observations you have are appreciated.

This is basic structures for an .odt file, and enough templates to get a worksheet with an introduction, some exercises, and `p` content. And everything that can go inside a `p` except math.

Thanks for the boost the other day about making math. I've hit an obstacle I am working on. Might start a discussion or wait until Friday. It feels like it may cause a significant delay, so I thought to get this much up for now.

### DIFF
--- a/examples/workbook/README.md
+++ b/examples/workbook/README.md
@@ -1,0 +1,8 @@
+# Workbook Example
+
+To extract the worksheets from this PreTeXt book into .odt (Open Office) files,
+use the pretext-worksheet-odt.xsl stylesheet. This will build a folder tree with
+root `worksheets/` where the leaves are file components of an Open Office file.
+These component files must be zipped into a .odt file to be opened by, say, 
+LibreOffice. The command is:
+`zip -r -X filename.odt mimetype content.xml settings.xml meta.xml style.xml META-INF`
--- a/examples/workbook/workbook.xml
+++ b/examples/workbook/workbook.xml
@@ -127,6 +127,56 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                         </p>
                     </statement>
                 </exercise>
+                <exercise>
+                    <statement>
+                        <p>
+                            Verbatim characters: <c>~`!@#$%^&amp;*()_-+=[]{}|\;:'",&lt;.>?/</c>. And with leading and trailing whitespace: <c>  Two before.   Three in between.   Four after.    </c>
+                        </p>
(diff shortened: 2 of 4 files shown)


rbeezer:
That's looking good.  A real tour-de-force.  High-level organization and infrastructure looks solid.  Perhaps some more detail-oriented comments in a bit.

rbeezer:
Some XSL discussion.  I miss starting a new conversion with a clean slate.  ;-)  No backward compatibility constraints or jitters.

* Some trailing whitespace about (its bright red in my `git diff`).  Otherwise use a regex in your editor to locate?
* Any reason to have the assembly stylesheet turned off?  It should not be a performance problem (its very fast).
* Is "Latin Modern" available by default?
* `boolean(parent::*/title or parent::statement/parent::*/title)` should be a boolean due to the "or" if you drop the `boolean`.  If you were only probing one path, you'd need it since you don't want the nodeset it would represent.  Try dropping these later when you can do before/after testing.
* `position()` can be very dangerous.  I hardly ever use it anymore, but it is tempting at first.  For example `boolean(*[not(&METADATA-FILTER;)][position() = 1][not(self::p)])`.  Not sure if `[1]` is still correct here.

Alex-Jordan:
Thanks, I applied the more trivial edits.

> Any reason to have the assembly stylesheet turned off? It should not be a performance problem (its very fast).

I just haven't looked into understanding that part yet. Like, does it affect how I write the entry template?

> Is "Latin Modern" available by default?

Unsure. LibreOffice will default to something acceptable though if it cannot find a font. (With different defaults for paragraph text and header text.) No different than if an author with Latin Modern writes an ODT file using LibreOffice, then sends it to someone else who may not have Latin Modern.

> For example boolean(*[not(&METADATA-FILTER;)][position() = 1][not(self::p)]).

This is a construction where I need to know if the first child of something is a `p`. But not counting metadata children like `title`. So for example, if the element opens with a `sidebyside` instead of a `p`. A variant of this with `[1]` did not work, because `[not(&METADATA-FILTER;)][1]` is already an element, not a node set, so tacking on the `[not(self::p)]` didn't work. Any ideas?

Alex-Jordan:
Closing this. I'm experiencing some issues with my laptop so I'm pushing all my work to GitHub just in case, including "temporary commits". So this is about to look like a mess.

PR #1385, opened 2020-11-13
Add support for an asymptote preamble

sean-fitzpatrick:
Added a short blurb in the guide, but didn't have time to touch the sample article

### DIFF
--- a/doc/guide/author/topics.xml
+++ b/doc/guide/author/topics.xml
@@ -1830,7 +1830,7 @@
 
                     draw(graph(Gaussian,min(a),max(a)),blue);
 
-                    // Optionally calculate "optimal" number 
+                    // Optionally calculate "optimal" number
                     // of bins a la Shimazaki and Shinomoto.
                     int N=bins(a);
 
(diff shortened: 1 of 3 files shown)


rbeezer:
I don't have a newer `asy` executable, and I didn't even see a good item to move into a preamble for the sample article, so I just left that undone.

Moved out some trailing whitespace (my fault!) that your editor was fixing, onto a whitespace-only commit with my name to blame.

I use `<note>` for "Best Practices" - pretty major pieces of advice.  So I moved your note on colors/pens into the list of "Notes" just prior.

Thanks for putting this together.

sean-fitzpatrick:
Ok, great! As far as I can tell there's no need to have xsl for this anywhere other than the Asymptote extraction template.
I suppose (for now) I've added a new schema violation but I'm all right with that :-)

rbeezer:
Right.  I was going to add it to the schema but forgot.  Soon.

rbeezer:
In the schema:  d8206fc749987c5a6c6b0e282897b811428db13e

sean-fitzpatrick:
Awesome, thanks. Turns out asymptote segfaults on PDF output in Ubuntu 20.10. I've just set John Bowman up with ssh access to my laptop so he can try debugging. Workaround is to add `-render=0` to the executable but that might have unintended quality consequences. I'll let you know if we need to update the executable in the `pretext` script.

PR #1395, opened 2020-12-14
Oddanswers

Alex-Jordan:
Two commits here: the "filter" business is the main one. And then cleaning up the sample book a bit.

There is a third thing to do, but it is not really connected to filtering, and I'm waiting for David's response to a forum post. Basically I noticed that no solutions were coming out in the HTML backmatter of the WW sample chapter, even though there is a `solutions` there. It turned out that there was no "solutions" modal template for WW exercises. (It was not delving down into `exercises/webwork-reps/static`.) The filter commit addresses that, because it was part of testing that filtering does what it should be doing. But there is one loose end about how to demark the "Part 1", "Part 2", etc when a WW problem in a solutions is staged. That can be part of a separate commit.


rbeezer:
* Let's **remove** the commented-out `hint`s.  I've been confused more than once when I grep source to debug some problem and the stuff I'm locating is in a commented-out section.  We can always get stuff back from the commit.
* I'm also on a mission to reserve named templates for just the simplest purposes.  I'd like templates to explicitly rely on a context "passed in", as a more robust form.  The `determine-admission` template can match/select on an `exercise`.  Then the template can inspect (compute) the serial number of the exercise, instead of passing it as a parameter.  You'll still need to pass `$admit` though (I'd guess passing the invoking `solutions` down for later computations could also clean up a lot of code, but only thought of that just now).
* `admit` is a good alternative to `include`, but still sounds vague/unintuitive to me.  `parity` is limited to mod 2, I guess.  But I don't have a good suggestion for an alternative.

Alex-Jordan:
> Let's remove the commented-out hints. I've been confused more than once when I grep source to debug some problem and the stuff I'm locating is in a commented-out section. We can always get stuff back from the commit.

Got it.

>I'm also on a mission to reserve named templates for just the simplest purposes. I'd like templates to explicitly rely on a context "passed in", as a more robust form. The determine-admission template can match/select on an exercise. Then the template can inspect (compute) the serial number of the exercise, instead of passing it as a parameter. You'll still need to pass $admit though

Got it.

> (I'd guess passing the invoking solutions down for later computations could also clean up a lot of code, but only thought of that just now)

I count 23 things that are passed from the invoking `solutions` to the `solutions-generator`. Three text, 20 boolean. Would it be worth putting it all into an rtf to pass around? And reconstitute the bits that are needed?

I count 21 instances (-common, -latex) of passing `purpose`. And 32 instances (-common, -latex, -html) of passing `admit`. Hard to count the booleans. 

If you pass the invoking `solutions` as a variable, I don't fully understand. Does that variable know its ancestors, or just its descendants?

> admit is a good alternative to include, but still sounds vague/unintuitive to me. parity is limited to mod 2, I guess. But I don't have a good suggestion for an alternative.

With `admit`, I started thinking of it like a bouncer at a fancy night club. Not letting you (an exercise) in unless you are on the list. I guess with changing to `parity`, it comes down to how much we want to commit to just the two options now. I'll leave it as `admit` for now, awaiting further instruction.

How's linear algebra coming?

Alex-Jordan:
>> (I'd guess passing the invoking solutions down for later computations could also clean up a lot of code, but only thought of that just now)

> I count 23 things that are passed from the invoking solutions to the solutions-generator. Three text, 20 boolean. Would it be worth putting it all into an rtf to pass around? And reconstitute the bits that are needed?
> I count 21 instances (-common, -latex) of passing purpose. And 32 instances (-common, -latex, -html) of passing admit. Hard to count the booleans.
> If you pass the invoking solutions as a variable, I don't fully understand. Does that variable know its ancestors, or just its descendants?

Was looking at this more. Doing this in a way that would clean the code up would be such a massive diff that I think it is worth saving for a different PR. What do you think? I'm going to commit the other things, and push without rebasing and force pushing. Can always squash later.

rbeezer:
Second evening's push to get this organized and debugged, and not done yet.  Please don't add any more on this PR.

rbeezer:
Three or four different things going on here, and very hard to unravel for testing and a clean history.  Three evenings' worth of discretionary coding time.

* `@admit` merged.  Needed to add `PROJECT-LIKE` to the `determine-admission` template.  Only barely caught this because I added a test to the sample article, which should be part of testing any new feature.  If we keep the sample article up-to-date it will continue to have one of everything and prevent bugs like this from getting out.  Only the commits for this feature got tagged with the PR number.  Would you like the pleasure of announcing on `-announce`?
*  Corrections of (1) typos, and (2) sample book schema violations, have been preserved on their commits.
*  None of the `webwork/stage` bits have been merged.  I have a modified version on a branch.
(a) "Part" should be localized.
(b) The heading created via an exceptional `if` in "exercise-components" feels out of place.  Please look at how `stage` is handled on the LaTeX side.  First, it is missing a heading, so that needs fixing.  Second, it does a dry-run, so no empty stages get their headings output with no content.  Third, with a `for-each`, I think the `heading-full` template can be located more naturally.  The HTML routines for exercises may need the same sort of re-factoring that the LateX routines recently saw.  So this could be a first step in that direction.

Thanks for finally getting odd answers in!  Gotta keep our author-instructors happy.

Rob

Alex-Jordan:
Sorry, I didn't realize that's where you were. I had planned to squash the two `oddanswers` commits and force push after another casual eyeing over the code in bc4eebe and any follow-up on the most recent thread discussion. It was supposed to be easy to just grab that once commit once those things were together.

> the sample article, which should be part of testing any new feature.

OK. I thought I was doing that by adding to the sample _book_. But I did not think about project-like.

About `webwork/stage`. What is really needed is for `stage` to become essentially the same as `task`. That will be a complete overhaul from the webwork extraction, through PG processing in a future WW version, back to translation stylesheets. I may not give stage any serious attention until that is happening, because it would be wasted effort.

rbeezer:
I think I got confused by

> I'm going to commit the other things, and push without rebasing and force pushing. Can always squash later.

and thought you were not going to force push anymore and I needed to squash.

Let's agree that I will not merge *anything* until you give me an explicit green light.  And then let's not promise others easy tangential bits and pieces put on in-progress PRs?

Alex-Jordan:
Something isn't working. I tried to do odd answers for ORCCA as part of revealing to some people that they can do this now, but it only gives emptiness where each odd answer should be. It may be isolated to WW questions. I'll keep looking.

Alex-Jordan:
Mea culpa on 'teh', for sure. It was late and I was burning through that issue review, and not thinking clearly.

xsl/pretext-html.xsl
```diff
@@ -3807,6 +3839,26 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
...
+                    <xsl:apply-templates select="webwork-reps/static/stage" mode="exercise-components">
+                        <xsl:with-param name="b-original" select="false()" />
+                        <xsl:with-param name="b-has-statement" select="$b-has-statement" />
+                        <xsl:with-param name="b-has-hint"      select="$b-has-hint" />
+                        <xsl:with-param name="b-has-answer"    select="$b-has-answer" />
+                        <xsl:with-param name="b-has-solution"  select="$b-has-solution" />
+                    </xsl:apply-templates>
+                </xsl:when>
+                <!-- webwork without stages          -->
+                <xsl:when test="webwork-reps">
```
Alex-Jordan, line 3853: This bit with `test="webwork-reps"` was cut along with the `stage` stuff. If you leave this part in, it still won't do good things for when there is a `stage`, but it will work the other 99.9% of the time.

Alex-Jordan: I'm happy to open a new PR that puts this back in. Before doing that, I wanted to understand if you believe that is the right thing to do (adding back in this chunk). Or if you would prefer to wait for stage to have its refactor at the same time.

rbeezer:
The "exercise-components" bit for `webwork-reps` is now merged.  Website rebuild will wait until I have more changes, like tonight or tomorrow.  (I did update the website with the "sidebyside" stuff.)

With your name, and the PR #.

I lengthened the second `@test` to be more restrictive and match the `@select` immediately following it.  The original would have been correct, I think, but the change makes it harder for bad things to slip through, so we'll catch them sooner.

PR #1399, opened 2020-12-21
Code cleanup: use select when possible for variables

Alex-Jordan:
The bigger picture here is I'm trying to automate comparing the output before and after some changes. I'm not there yet, but what we have here is the result of my first experiment. I thought it was worth saving, so for your consideration...

This changes all `xsl:variable` declarations that just had a `xsl:value-of` child to use a `@select` instead.

My half-baked automated testing found no differences between pre- and post- versions of the sample article and WW sample chapter. At first, in all cases, the `@select` from the `value-of` simply migrated to become the `@select` for the `xsl:variable`. Then my testing revealed that in the case of the `document-id` variable in `-common`, this wasn't quite right. For that one, it was necessary and appropriate to wrap a `string()` around the `select` expression.

Two small improvements here:
1. It's a net reduction of 70 lines of code.
2. More importantly, it's good to use `@select` when possible. As we recently discussed, this allows the variable to be explicitly string, number, or boolean. When not using `@select`, the variable is an rtf, and using it downstream doesn't always match with what you naively expect to happen. So making this change, there are fewer "bad examples" for a code contributor to mimic.

### DIFF
--- a/xsl/extract-pg.xsl
+++ b/xsl/extract-pg.xsl
@@ -1418,12 +1418,8 @@
             <xsl:value-of select="substring-before($trimmed-start, '}')"/>
         </xsl:if>
     </xsl:variable>
-    <xsl:variable name="macro-command">
-        <xsl:value-of select="substring-before($macros, '&#xa;')"/>
-    </xsl:variable>
-    <xsl:variable name="next-lines">
-        <xsl:value-of select="substring-after($macros, '&#xa;')"/>
-    </xsl:variable>
+    <xsl:variable name="macro-command" select="substring-before($macros, '&#xa;')"/>
+    <xsl:variable name="next-lines" select="substring-after($macros, '&#xa;')"/>
     <xsl:if test="contains(., $macro-name)">
         <xsl:value-of select="normalize-space($macro-command)"/>
     </xsl:if>
@@ -1808,9 +1804,7 @@
     </xsl:if>
     <!-- prefix is optional -->
     <xsl:if test="@prefix">
-        <xsl:variable name="prefix">
-            <xsl:value-of select="@prefix" />
-        </xsl:variable>
+        <xsl:variable name="prefix" select="string(@prefix)"/>
         <xsl:variable name="short">
             <xsl:for-each select="document('pretext-units.xsl')">
                 <xsl:value-of select="key('prefix-key',concat('prefixes',$prefix))/@short"/>
@@ -1821,9 +1815,7 @@
     <!-- base unit is *mandatory* so check to see if it has been provided -->
     <xsl:choose>
         <xsl:when test="@base">
-            <xsl:variable name="base">
-                <xsl:value-of select="@base" />
-            </xsl:variable>
+            <xsl:variable name="base" select="string(@base)"/>
             <xsl:variable name="short">
                 <xsl:for-each select="document('pretext-units.xsl')">
                     <xsl:value-of select="key('base-key',concat('bases',$base))/@short"/>
--- a/xsl/pretext-beamer.xsl
+++ b/xsl/pretext-beamer.xsl
@@ -634,12 +634,8 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
   <xsl:value-of select="count(*)"/>
   <xsl:text>, raster equal height=rows, raster force size=false, raster column skip=0ex] &#xa;</xsl:text>
 
-  <xsl:variable name="columnCount">
-    <xsl:value-of select="count(*)"/>
-  </xsl:variable>
-  <xsl:variable name="widthFraction">
-    <xsl:value-of select="1 div $columnCount" />
-  </xsl:variable>
+  <xsl:variable name="columnCount" select="count(*)"/>
+  <xsl:variable name="widthFraction" select="1 div $columnCount"/>
 
   <xsl:for-each select="*">
     <xsl:if test="parent::*/@pause = 'yes'">
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -350,17 +350,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- This is set (temporarily) in docinfo, which will change           -->
 <!-- We do no special error-checking here since this will change       -->
 <!-- The variable will be empty if not set                             -->
-<xsl:variable name="numbering-exercises">
-    <xsl:value-of select="$docinfo/numbering/exercises/@level"/>
-</xsl:variable>
+<xsl:variable name="numbering-exercises" select="number($docinfo/numbering/exercises/@level)"/>
 
 <!-- Figure-Like can optionally run on their own numbering scheme      -->
 <!-- This is set (temporarily) in docinfo, which will change           -->
 <!-- We do no special error-checking here since this will change       -->
 <!-- The variable will be empty if not set                             -->
-<xsl:variable name="numbering-figures">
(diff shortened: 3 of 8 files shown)


rbeezer:
Very interesting.  I'm sure the functions (e.g. `substring-before()`) will behave.  And you think the attributes become strings with their values?

`mag` makes me nervous.  I think the variable is now a nodeset (a very real one, not an RTF).  I did not even look see now it is used.

In both cases, maybe things like
```
<xsl:message><xsl:value-of select="count($foo/*)"/></xsl:message>
```
will confirm or deny some suspicions (or not).

Alex-Jordan:
I'm assuming that `<xsl:variable name="foo" select="@bar"/>` is no worse than `<xsl:variable name="foo"><xsl:value-of select="@bar"/></xsl;variable>`. I know the latter produces an rtf and I guess the first case produces a node set, but I am unsure. The testing I did looked at the sample article (HTML and PDF) and the sample chapter (HTML, PDF, and PG) and found no differences, so that's the extend of the evidence I have.

`mag` is only used as the first argument to a `str:replace()`. I guess that first argument gets stringified well, since it's been working and there was no diff. But I wrapped `string()` around it now which is seeming to me like it may
be a good habit for situations like this.

There were two other `select="."` coming from the Jupyter conversion. I wrapped them in `string()` as well, which seemed appropriate. Again in both cases, the variable is only used as the first argument to a `str:replace()`.

Force pushed with the above changes.

This can be academic, as I appreciate the mantra "if it ain't broke, don't fix it". I'm just happy to be improving my understanding of these things for future XSLT coding.

rbeezer:
Reading/consulting Lenz.

`xsl:value-of` will always create a string.  (And has the odd property that if it gets a node-set, it just deals with the first node.)

And as you note, the `xsl:variable` constructions that you are replacing create an RTF.  The new versions create a node-set (or a single node).

A typical use of these variables is to retrive their values with a `value-of`, hence ultimately a string.  So I ask myself, what is the value of an RTF?  While looking for something else related to this, Lenz p. 90 says: 

> Tip - "When being converted to another data type, RTFs behave exactly the same as a node-set containing a single root node."

So currently, an attribute value will go into an RTF as a `text` node, then when used/accessed with a `value-of` it becomes the string value of the first node of the RTF, which is the desired value.  Changes here save the attribute as a node-set (uncertain exactly what this is, an attribute all by itself?  an attribute of what?), but its `value-of` will be the attribute value.  I think.

Lenz also says (p. 90):

> Whenever it is possible for the XSLT processor to convert a given value to a value of the expected data type, it does so automatically.

So currently there are many places where a string (or node-set?) gets converted automatically to be used as a number.  Since it just "works".

I think I'd find this more interesting if each `xsl:variable` explicitly created the desired data-type from a node-set right away, rather than relying on conversions (from RTF now, node-sets as proposed).  This would mean some amount of adding `number()` and `string()` primarily, I'd guess.  In other words, I'd be much more excited if explicit data-type conversions survived the before/after testing.  (And then this would become the custom.)  I would not even be surprised if some bug/mismatch turns up.

Besides the hassle factor, what do you think?

Rob

Alex-Jordan:
About to head out for the day/afternoon. Real quick: one of the two things that triggered this was a case where a variable was _not_ behaving as expected. I'll have to think if I can remember what it was while I'm out walking. Probably some recent WW thing, ODT thing, or something with `@admit`. (The other thing that triggered it was a small test of my first stab at automated testing of a diff.)

rbeezer:
Thanks.  No big rush - I'm letting this one percolate, as you can tell.  I'm sure a long walk will jog your memory.  I seem to recall something as well.  Maybe the Seahawks will jog my memory....

Alex-Jordan:
OK, here's what it was. (Can't credit the walk; I had to look up my own stackoverflow posts.) It was an ODT thing. I'll try to give a minimal example rather than the actual original issue that arose.

Suppose the context is a `cell` in the second column, and I want to get the `@halign` attribute of the corresponding `col`. For this minimal example, pretend `@colspan` is not a consideration. If you do:

```
    <xsl:variable name="nthcol">
        <xsl:value-of select="preceding-sibling::cell + 1"/>
    </xsl:variable>
    <!-- Now $nthcol is an RTF with the _string_ '2' -->
    <!-- And if we do something like: -->
    <xsl:value-of select="ancestor::tabular/col[$nthcol]/@halign"/>
    <!-- I'm naively expecting it to look into the 2nd col and get its @halign -->
    <!-- Instead, the _text_ string $nthcol is simply evaluated as true(), -->
    <!-- So the above might as well be "ancestor::tabular/col/@halign" -->
    <!-- and I've actually got a whole node set of all the @haligns from -->
    <!-- all of the col~s. And lastly, the `value-of` just stringifies the -->
    <!-- _first_ one, which as you note, is an odd design choice. -->
    <!-- So I'm expecting the 2nd @halign, but I got the first. -->
```

Had `$nthcol` been a true number variable in the first place, I wouldn't have spent significant time trying to figure out why `@halign` on a `col` was not working.

So, that's just the background on how I came to appreciate making a variable actually be of the type you want it to be instead of relying on it being interpreted right. You quoted:

> Whenever it is possible for the XSLT processor to convert a given value to a value of the expected data type, it does so automatically.

and I think this is a situation where that fails, sort of. It would be reasonable for the predicate evaluation to look to see if the "string" it has been passed actually qualifies as an integer.

If I understand you right, I agree that more explicit use of `number()`, `boolean()`, and `string()` would be helpful. It may be a massive diff to put them everywhere where they "belong".  I suppose it could go along with something else that came up in a recent thread: using prefixes on most variables to indicate type. Like `b-` (already in use) but also `n-` (number), `s-` (string), and `ns-` (node set). And no prefix implies an RTF, which may be the most common kind since you get that from variables that do not use `@select`. But maybe this is too much...

rbeezer:
Right.  I remember now.  Without a walk.

I think your example is missing a `count()`?  So my suggestion, which your comments suggest is:
```xsl
<xsl:variable name="nthcol" select="number(count(preceding-sibling::cell) + 1)"/>
```

> It may be a massive diff...

I'm only suggesting **this PR** would be improved by doing it in these 30 or so instances.

> using prefixes on most variables to indicate type

Yes, this strikes me as overkill.  It took me a while to discover the design pattern where a big complicated template does a bunch of logic to *always* produce `yes` or `no` (or similar), and **then** make a new `$b-foo` variable that is really a boolean.  So there is old code where strings are being passed around as binary options, a crude boolean.  So the `$b-` device is to help keep those sorted out.

Alex-Jordan:
Just to update this, I put `string()` and `number()` around the instances where there was just a plain node expression.

To recap, I have run into two situations where something unexpected happens.

(1) Using something you expect is a number as a predicate like `ancestor::tabular/col[$nthcol]/@halign` but it might be interpreted as "true" instead.

(2) when there is no `docinfo/document-id`, then using `<xsl:variable name="document-id" select="$docinfo/document-id"/>` led to the expression `$document-id = ''` being evaluated as false, which is unexpected. Wrapping the expression in `string()` makes it evaluate as true. [Also, not using `@select` at all, and using templates to set the variable, also makes it evaluate to true, which confuses me even more.]

If it were only item (1), that feels like an XSLT nuance that an experienced XSLT code writer becomes familiar with, and should just know how to avoid the issue. But then there are things like (2) that I still don't understand. And I wonder how many other things there might be like that which remain undiscovered (by me/us).

rbeezer:
Perhaps (2) is what ahs led me to routinely use the construction
```xsl
<xsl:if test="not($foo = '')"/>
```
rather than a not-equal operator?  I've still not fully digested all of this.

PR #1400, opened 2020-12-24
WeBWorK: tunnel human readability param through groupings

Alex-Jordan:
APEX had a `...<q><m>\dx</m></q>` where `\dx` is one of Greg's macros (`\Delta x`). What _should_ have been happening is that the `m` receives a `b-human-readable` parameter, and if that is true, it puts the macro definition right there in the PG:

```
..."[`\newcommand{\dx}{\Delta x}\dx`]"
```

But the `q` template from -common was catching this, which does not receive `b-human-readable`, let alone pass it along to the `m`.

So this PR adds a template for `q` (and a few other things) to the extract-pg stylesheet.

### DIFF
--- a/xsl/extract-pg.xsl
+++ b/xsl/extract-pg.xsl
@@ -1694,6 +1694,51 @@
     <xsl:text>&#xa;</xsl:text>
 </xsl:template>
 
+<!-- ######### -->
+<!-- Groupings -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks very much for keeping this maintained!

PR #1403, opened 2021-01-07
Catch failed latex-image pdf build

Alex-Jordan:
If something goes wrong with compiling a latex-image .tex file and the pdf is not built, there was nothing in the log (even under -vv) to report this.


rbeezer:
Great idea.  I added `ERROR` to the prefix on the "failed compilation" string.

We should definitely do more of this sort of thing, it is totally obvious without any `-v`, `-vv`, what the problem is, and **it** is a frequent problem.

I don't want to report how long it took me to **break** an image from the sample article so I could test.  ;-)  Pushed.

PR #1405, opened 2021-01-08
WeBWorK: warn if content is missing

Alex-Jordan:
Prior to this commit, there are situations where WW reps building can fail without a good explanation. The references below to `last[0]` kill python when `last` is an empty list. (Which can happen in silly ways, like literally putting nothing into a `statement`, but also when using an OPL problem that is not PTX compatible.) Now if that happens, you get a warning that something was empty, and processing otherwise proceeds. And I put one triggering example in the sample chapter.

This branch started out as its name implies, to cleanse sidebyside (gracefully) from within a webwork. But one thing leads to another in ways that feel too intertwined to do on separate branches. So I'll try to isolate meaningful changes one commit at a time, and await review and merging. (No rush!)

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -1745,6 +1745,19 @@
                     </statement>
                 </webwork>
             </exercise>
+
+            <exercise>
+                <introduction>
+                    <p>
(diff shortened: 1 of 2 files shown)


rbeezer:
Looks great.  Minimal testing, pushed, no website update.

Thanks!

PR #1406, opened 2021-01-08
WeBWorK sample chapter: move xml:ids from 'webwork' to 'exercise'

Alex-Jordan:
Here's the next one. Trivial, I think. The Sample-chapter had xml:id~s on webwork elements, but we don't allow that. I moved them all to the enveloping exercise.


rbeezer:
I noticed that once recently.  Thanks.  Will handle soon.

rbeezer:
Thanks for the maintenance!

PR #1408, opened 2021-01-08
typo fix

kcrisman:
(Unless this was in the original and the intent is to preserve that, but I assume that's not the case.)


rbeezer:
Thanks very much!

PR #1409, opened 2021-01-08
WeBWorK sample chapter and showcase article: cleanup

Alex-Jordan:
Another trivial one. Titles on some stages (and making them match across the sample chapter and the showcase article) and removing an outdated comment about hints.

The next one will really remove sbs from webwork.

After that, the only thing left maybe controversial: adding the `data` attribute to the schema.


rbeezer:
Very good!  Thanks.  Website update later tnhis evening.

PR #1410, opened 2021-01-10
WeBWorK: remove sidebyside

Alex-Jordan:
This finalizes making it so `sidebyside` is something you should never author within a `webwork`.

A WW server will still return PTX where `image` and `tabular` elements are wrapped in a `sidebyside`. This is harmless; it's like an author using a `sidebyside` around a lone `image` while _not_ inside a `webwork`.

But I willl make edits to WW so that WW 2.16 won't do that any longer.

### DIFF
--- a/examples/showcase/webwork.ptx
+++ b/examples/showcase/webwork.ptx
@@ -225,9 +225,7 @@
         <p>
           This Venn Diagram groups animals by certain characteristics.
         </p>
-        <sidebyside widths="50%">
-          <image pg-name="$vd"/>
-        </sidebyside>
+        <image pg-name="$vd" width="50%"/>
         <p>
           Name an animal that belongs in the center region. Spelling counts!
         </p>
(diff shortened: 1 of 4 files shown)


rbeezer:
Good riddance!  ;-)  Thanks.  Updating website now.

Let's do schema changes on their own commit, right after relevant code changes (prefix "Schema:").  I have to run a script to make new versions of the four formats, and that is easier to package all up and fix into a different commit.  Too lazy here to break this up and still preserve your name on it.  ;-)

I'm experimenting with support for different versions, and using `component` as an attribute on source elements.  If you ever think that your `mode="components` is beig confused with that, just do a global search/replace.

Do you want to announce on `-dev`?  I don't think `-announce` needs it again -- this would just be so developers are aware of what is changing and can make careful reports of any problems.  Sometimes people just automatically suspect whatever is being worked on and it is an easy fix.  ;-) ;-) ;-)

Rob

PR #1411, opened 2021-01-11
WeBWorK: perl string variables

Alex-Jordan:
If you were writing a .pg problem from scratch, in the middle of the statement, you might put some "raw" perl code. This would be for the rare occasion that you need to do something that you know can be done across the various WeBWorK output formats, but PGML has no macro to do it. So you could use PGML like `[code goes here]*`. The brackets tell PGML you are making a substitution. The star tells PGML not to escape any of the characters inside.

So this commit adds the option `data="perl"` for this situation. We already have `data="pgml"` for a similar situation. In that case, it makes two stars, which tells PGML not only to hold back on escaping the characters inside, but also to process the code inside as PGML syntax, not regular "raw" perl.

The need: the MHCC project has GeoGebra applets they have put into WeBWorK problems. This meets the criteria of being something PGML has no macro for, but the WW output formats can handle it. In the distant past, this was doable without `data="perl"`, because PreTeXt itself allowed them to type a command like `[code goes here]*` directly into a `p`. But today, PTX itself escapes the brackets and star, assuming the author wanted those literally.

So this PR adds `data="perl"` alongside `data="pgml"`.

### DIFF
--- a/xsl/extract-pg.xsl
+++ b/xsl/extract-pg.xsl
@@ -1152,6 +1152,10 @@
         <xsl:text>->correct_ans()</xsl:text>
     </xsl:if>
     <xsl:text>]</xsl:text>
+    <!-- if the variable is a string of perl code to be executed -->
+    <xsl:if test="@data='perl'">
+        <xsl:text>*</xsl:text>
+    </xsl:if>
     <!-- if the variable is a string of PGML syntax to be processed -->
     <xsl:if test="@data='pgml'">
         <xsl:text>**</xsl:text>


rbeezer:
Time to think about names, before this goes in the schema.  `data` doesn't really speak to me.  Is this really even a `var`-like thing?  I understand `var` to be like an answer blank of sorts.  Is this just injecting "raw" code which may not necessarily be soliciting an answer?  (Can it be abused?)

Would something new like `<code syntax="pgml">....</code>` be more accurate?  (Except we already have `code` for something else.)  Perhaps it would be easier to isolate as WW-specific within the schema, though I suspect it is meant to be used inline within a `p`?

Alex-Jordan:
Since submitting this, I found a workaround that uses the existing `data="pgml"`. So I don't need an extension right now. Probably will close this if there's no more to discuss more generally about `var`.


> I understand var to be like an answer blank of sorts.

That's not its primary purpose. The primary purpose is that something like `$f` is a randomized expression, and you want it printed in the statement. So your statement might be like:

```
<p>
    Differentiate <var name="$f"/>.
</p>
```

So it's dropping the value of the perl variable `$f` somewhere. That is the commonality with its other uses. Other uses currently are:

1. To make an answer blank. If you've arranged for `$df` to be the derivative of `$f$` then you could follow the above with:
```
<p>
    <var name="$df" width="20"/>
</p>
```

2. To drop in a string that you expect to be processed as PGML. We have one situation where I regularly use this. It is to take advantage of a WW feature for problems where there is a quantitative answer part (just type in the derivative) and also an explanation essay part (show the steps you used, applying the definition of the derivative). Hard to explain typing in a thread, maybe at next drop-in session?

Thing 1 could be accomplished with `<fillin width="20" something="$f"/>`. We did not want to use `answer` before because `answer` is not generally a descendant of `statement`. And it was just so much easier to use `var` rather than invent a new element. But since then, we have `fillin`...

The end result of thing 2 is important to me. Without that, lots of APEX exercises can't really function as WW exercises in the way a calculus teacher wants them to. I'm open to working out an alternative to `data="pgml"` but so far it's a very niche thing that feels hard to abstractify.

About the name `data`. I'm not a huge fan either but we couldn't find a good one earlier. I liked `content` because we are talking about a variable in a programming language, and want to say something that describes what is being stored in that variable (here, pgml markup; maybe more abstractly you could have `string`, `real`, `boolean`, `ptx` (markup string), etc.). But `content` is a loaded term.

rbeezer:
Thanks for continuing to explain this to me.  I guess it would help me to think of thing 0 and thing 2 like the `/var` directory on a Linux machine.  Something that is **different**, or varies, from instance to instance.

Do you think it is feasible to use `fillin` or something else (not `answer`) for a place where the reader does some interaction/entry, and then reserve `var` for thing 0?  I'd see the `@data` use case then more consitent with other uses of `var`.

In your examples, someone who reads the PreTeXt (and not the `pg-code`) can't really tell that `$df` is the answer?  The `@width` might be the only clue (and it is optional?)?  In other words, someplace, in WW-speak, `$f` and `$df`, are fundamentally different?

I'm trying to get to syntax that can be WW-independent for really simple problems, and I feel like I've never really grokked `var` (mostly out of ignorance).  And maybe other authors who are not very experienced with WW  would not understand it either

Alex-Jordan:
Is it clear that the answer varies as well, from instance to instance?

Here is a quick example, easy to compile in your head:
```
<exercise>
    <webwork>
        <pg-code>
            $n = list_random(2,3);
            $f = Formula("x^$n");
            $df = $f->D('x');
        </pg-code>
        <statement>
            <p>
                Differentiate <m><var name="$f"/></m>.
            </p>
            <p>
                <var name="$df" width="20"/>.
                <!-- @width and certain other attributes for popup, etc,
                are what triggers this as an answer input field.
                But tied to $df as the expected answer. -->
            </p>
        </statement>
        <solution>
            <p>
                Use the power rule to differentiate <m><var name="$f"/></m> and you get <m><var name="$df"/></m>.
                <!-- Both instances above are just printing the value, not making an input field. -->
            </p>            
        </solution>
    </webwork>
</exercise>
```

I think the thing that makes an input field really should become `fillin` with some attribute that gets the `$df` or whatever. It's a lingering mistake from our early days with all this. It will take some time and planning to make the change gracefully.

PR #1414, opened 2021-01-12
Stab at automatic lists stub

kcrisman:
This probably has some typos, and probably isn't how you would have said it (though I did copy some wording straight from the sample article).  I like this feature a lot, so I figured writing this would help me use it better, and I think it has.  Hope it's helpful.

On a related note, the fact that the overview (as opposed to topic list) in the author guide has "back matter", "front matter", but also "index and notation entries", was a little confusing to me, as was the topic-front-back-matter I see online (https://pretextbook.org/doc/author-guide/html/topic-front-back-matter.html) but where the code of the author guide is (seemingly) split up into two parts now.

### DIFF
--- a/doc/guide/author/topics.xml
+++ b/doc/guide/author/topics.xml
@@ -2701,12 +2701,12 @@
             <p></p>
         </subsection>
 
-        <subsection>
-            <title>(*) Automatic Lists</title>
+        <subsection xml:id="automatic-lists">
+            <title>Automatic Lists</title>
 
             <p></p>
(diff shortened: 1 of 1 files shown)


kcrisman:
I already see a typo - well, an entire section/line of typos!  I had tested this in the earlier subtopic but I thought I had removed those line changes.  Obviously you should and will remove that even if you like the main event.

rbeezer:
That is beautifully concise and accurate.  Thanks.

PR #1415, opened 2021-01-12
Important typo for reference type

kcrisman:
The types need to correspond here, and this is a very useful example.


rbeezer:
That section needs a lot of work, but that needed fixing right away.  Thanks.

kcrisman:
Yeah, it was really confusing until I carefully compared the source!  That annotated version of the sample article VERY useful.  Unfortunately it doesn't have a search mechanism.

PR #1416, opened 2021-01-14
add a reference to pretext script chapter in sageplot basics

kcrisman:
A suggestion since I found it confusing just now to have to loop back when I really only wanted to know about sageplot, not latex images.


rbeezer:
Done.  Thanks.

kcrisman:
👍

Issue #1302, opened 2020-06-02
Index entries in exercise go missing from LaTeX


Alex-Jordan:
An `idx` inside an `exercise` currently doesn't do anything for LaTeX output. Index entries simply go missing from print. Some discussion at #1203.

kcrisman:
Similarly for `exercisegroup`, I believe.  

You also had some sample code there; here is what I use as a workaround:
```
diff --git a/xsl/pretext-latex.xsl b/xsl/pretext-latex.xsl
index 96d0b55..2f98d16 100644
--- a/xsl/pretext-latex.xsl
+++ b/xsl/pretext-latex.xsl
@@ -5426,6 +5426,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <!-- depending on what type of division the "exercise" lives in. -->
     <!-- For each, exactly one "ancestor" is true, and then the      -->
     <!-- expression will evaluate to the corresponding global switch -->
+    <xsl:apply-templates select="idx"/>
     <xsl:variable name="b-has-statement" select="true()"/>
     <xsl:variable name="b-has-hint"
         select="(ancestor::exercises and $b-has-divisional-hint) or
```

rbeezer:
I've got a pile of index-related work to do, tagged now so I'll see it then.

kcrisman:
Hopefully resolved by #1419 also exercise groups.

rbeezer:
Fixed at #1419

PR #1419, opened 2021-01-15
Get index elements in exercises and exercisegroups

kcrisman:
I found that for some reason I had to put these in just before the beginning of the environment; I got nasty errors otherwise. This works well for me, with quite a few of each in my source.

### DIFF
--- a/xsl/pretext-latex.xsl
+++ b/xsl/pretext-latex.xsl
@@ -5283,6 +5283,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:apply-templates select="." mode="label"/>
     </xsl:if>
     <xsl:text>%&#xa;</xsl:text>
+    <xsl:apply-templates select="idx"/>
     <xsl:apply-templates select="introduction" />
     <xsl:choose>
         <xsl:when test="$ncols = 1">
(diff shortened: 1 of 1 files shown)


kcrisman:
Should resolve #1302 - let's hope!

rbeezer:
I suspect you were not placing the `\index{}` far enough below the `\begin`, it has several arguments/parameters that need to be built before this can appear.

I've moved both slightly, so they are **inside** the relevant environment.  Less chance the locator ends up pointing to the wrong page?  Holler if you see bad behavior - we can adjust.

HTML is not so smart about `exercisegroup`, but not totally incorrect.  See #1424

Always best to include some tests, much like a Sage doctest, to the sample article.  Then in A/B testing I'll see injuries later with other (related) changes.  See 402be333ff8ec3fae71c3f5196fc9ecd70b3169f.

Thanks for going after this one.

kcrisman:
You may be right - I just put it after the `\end` but maybe there was something else in the way.  Sorry about not including examples - it was my first "real" PR on this.   But obviously that would be needed 😄

rbeezer:
> it was my first "real" PR

YES!  I meant to say it was really nice to see more than a typo!  ;-)

kcrisman:
>
>
> it was my first "real" PR
>
> YES! I meant to say it was really nice to see more than a typo! ;-)
>
>
Haha, that's how I got started with Sage too ...

PR #1430, opened 2021-01-21
Add important pause examples to slideshow

kcrisman:
This documents (at least a little) heretofore undocumented behavior of exactly how `pause` and `subslide` work.  I recognize that this may not be the best place for it (since the "samples" are supposed to be "minimal") but I feel like it is a good idea to have it documented somehow, and obviously the sample article is not the place for slideshow stuff.  I feel like this is easier to understand than the author guide.

------

@StevenClontz who wrote all this and I hope I'm not stepping on toes.

### DIFF
--- a/examples/sample-slideshow/sample-slideshow.xml
+++ b/examples/sample-slideshow/sample-slideshow.xml
@@ -220,5 +220,36 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                 <p>And some more mathematics.  An identity due to Ramanujan:<me>\frac{1}{\Bigl(\sqrt{\phi \sqrt{5}}-\phi\Bigr) e^{\frac25 \pi}} = 1+\frac{e^{-2\pi}} {1+\frac{e^{-4\pi}} {1+\frac{e^{-6\pi}} {1+\frac{e^{-8\pi}} {1+\ldots} } } }</me>.</p>
             </slide>
         </section>
+
+        <section>
+            <title>Other features</title>
+
+            <slide>
+                <title>Pausing</title>
+
+                <p>Pausing doesn't always act how you expect from <latex />.  The first and third lines are always visible.</p>
+
+                <p pause="yes">But this one needs you to advance the slide.</p>
+
+                <p>(Visible from the outset.)</p>
+            </slide>
+
+            <slide>
+                <title>Subslides</title>
+
+                <p>Subslides are another way to achieve pausing without lists.</p>
+
+                <subslide>
+                    <p>This is a subslide.  It always comes after a pause.</p>
+                    <p><ul pause="yes">
+                        <li>You can also pause within it.</li>
+                        <li>Of course!</li>
+                    </ul></p>
+                </subslide>
+
+                <p>But this is still visible from the outset.</p>
+            </slide>
+
+        </section>
     </slideshow>
 </pretext>


kcrisman:
Warning!  I only tested this initially on revealjs - in LaTeX it seems to perhaps behave like, well, LaTeX.   Also, Beamer seems to look rather different - e.g. section doesn't give titles or anything, figures don't work right, ...  But then again it says

> Conversion to Beamer presentations/slideshows is experimental and needs improvements

so maybe that's okay for now.

StevenClontz:
For clarity, @arechnitzer wrote most of PreTeXt-to-reveal/beamer code; my work has focused on the PreTeXt/Book-to-PreTeXt/Slideshow extraction not currently implemented officially anywhere.

arechnitzer:
Yup - the beamer conversion is **very** rudimentary and hasn't been worked on in quite a while. When I built PDF slides for my class last term, I used the pretext->html and then [decktape](https://github.com/astefanutti/decktape) to convert that to PDF (which I then annotated using xournal++).

kcrisman:
Oh!  I don't think I realized that. Sorry, @arechnitzer !

kcrisman:
I'm not interested in beamer (for now), just reveal.js, though maybe that is overly confident.  Maybe @rbeezer would suggest a "common" slides stylesheet for the most obvious stuff first?  I will probably have some (very rudimentary and probably wrong) improvements for the reveal version, but I won't have time to do anything on the beamer this term.

kcrisman:
Also (and now this is just gratuitous), totally waiting on @StevenClontz to make a very robust `<book>` to slideshow!  But I don't currently need that, thankfully.

StevenClontz:
For posterity: book-to-slideshow WIP is here https://github.com/TeamBasedInquiryLearning/linear-algebra/tree/main/slides and showcased here https://teambasedinquirylearning.github.io/linear-algebra/slideshow-link.html

sean-fitzpatrick:
Beamer worked fine for me. Sections don't get a title page, if that's what you mean. If you mean that it doesn't display the section at the top of the slide, that's just the design of the default Boadilla theme. Changing themes will get sections to show up.

Last year I was using slides because I was teaching back-to-back-to-back and I needed something to remind me what I was talking about. Students asked for PDF because they found the reval.js inconvenient to view later on.

It worked find for what I was using it for.

kcrisman:
Sorry, I didn't mean to critique the beamer overall.  What I meant was that, among other things, it doesn't seem to have the same capabilities as the pdf (e.g. w.r.t. figures) *and* it also handles pauses differently than the reveal.js currently does.  I don't know if that is a bug in one of the conversions, or neither.

arechnitzer:
Not at all - the beamer conversion was very preliminary and hasn't been hacked much in quite some time. The reveal conversion got more attention and works far better. If you want PDFs then I recommend using decktape (see comment above) to convert reveal -> PDF. I built 40+ slideshows this way for my class last term.

rbeezer:
Thanks - great to have this in the sample slideshow.

The Beamer conversion should probably be furloughed until we find a maintainer.

StevenClontz:
> The Beamer conversion should probably be furloughed until we find a maintainer.

Strong agree - particularly with the trickiness of getting content to fit in both formats, it's better to just support the HTML format (where the viewport can be resized as needed when content overflows).

arechnitzer:
Also agree.
Is relatively easy to convert reveal.js to PDF using tools like [decktape](https://github.com/astefanutti/decktape)

PR #1432, opened 2021-01-21
Add internal link for guide regarding slides

kcrisman:
This one needs a link for ease of navigation for new users


rbeezer:
Consolidated into #1463.

PR #1433, opened 2021-01-26
typo "URL"

kcrisman:
I think this is a typo, anyway, though there are SO many acronyms out there.  (I mean initialisms!)


rbeezer:
Consolidated into #1463.

PR #1439, opened 2021-01-30
typo "project"

kcrisman:


Alex-Jordan:
Thanks for finding these. It might be helpful if you would maintain a "typos" branch that accumulates all of these and then you only periodically submit a PR once they accumulate to a dozen or so. You could even maintain that branch just within GitHub, like these patch branches. It maybe only takes two minutes for each one to be examined, approved, and merged, but it also would only take two minutes if a dozen of them came as one package.

Alex-Jordan:
Even better: a "typos" branch that you maintain could just have an open PR. When you find a new typo, correct it on that branch, and the PR will just grow by one commit. So you'd not need to decide when it's time to open a PR.

kcrisman:
I hear you, but the problem is the overhead for me on that.  I definitely don't have the capacity to maintain it locally, and I haven't figured out how to do so on GH (yet).  If you have a link that explains that clearly I would try it.  (But don't look for it special!)

TL; DR For now, if I don't submit it immediately, it won't happen - and then I will get annoyed when I read that section and see that no one corrected it, especially if that no one is me.  

This is actually (in my opinion) a problem with the DVCS paradigm with a very large number of contributors.  Because it does make more work for everyone, you are right.  I find it so much more easy when someone using my book has just written a list of typos - and it's easier for them, too.  But having to maintain branches and pull requests and tigers, oh my! is a lot of cognitive load.  For everyone ☹️

Alex-Jordan:
For this particular project, PreTeXt, the overhead for the primary maintainers is a greater concern than your overhead for pull requests that have negligible value (as is the case with most typos). Please accept my suggestion as part of how the three primary maintainers are trying to keep this project sustainable without us three burning out.

kcrisman:
Yes, I read between those lines.  My point is that it's not possible to contain those things in an open source project - this is a discussion I had many, many times in Sage development.  If the primary maintainers find that any given PR has negligible value, they are free to ignore them, as is practice in open development of this type.

----

(I don't think typo fixes are negligible at all, personally; they are incredibly distracting to many readers and have other negative aspects, but I understand that there will be disagreement on this point among well-meaning individuals.)

davidfarmer:
I may be repeating, but my understanding of the suggesiton is that
this pull request can stay open for a while.  More commits can
be pushed to the same branch, and those will pile up on this
pull request until they reach a critical mass and Rob squashes and
merges.

I happen to have a local branch where I also am accumulating typos.
When I have something more substantial to contribute, a put that
contribution on the same branch and submit a pull request with two
commits: one typos, and one content-based.

I am just conveying a point of interest and am not suggesting
that others do what I am doing.  I consider a typos branch with an
open pull request to be perfectly fine.

In fact, what do you think of me making a pull request to Karl-Dieter's
branch, which can be used to manage typos for a while until we are
ready for Rob to accept it?

Alex-Jordan:
You've taken more time to comment on this thread than it would take in "overhead" to put all the typos on this one branch. I will take the time now to walk you through it.

Click `krisman:patch-8` up above (if you are viewing this on GH). Now you are on that branch that has one typo fix. Navigate to some file that has some other file. Click to edit it and correct the typo. Save/commit using the tool at the bottom of the edit screen.

Voila, you've added a second typo fix to this same branch, under this same PR. All future typo corrections could go to the same place. You could even poke around and rename `patch-8` to `typos` for long term recognition.

davidfarmer:
I just realized that there are multiple small typo pull requests.
That is not appropriate use of git/GitHub.

Once there was one open pull request with a typo, that same branch
(which still exists, because it has not been merged) should have
been used for the other typos.  It makes significant extra work
(for everyone) to use multiple branches.

Also, the names of the branches (patch-N) are not good.  Be semantic,
like typo-N.

I cannot implement the plan in my previous comment (about me sending
a pull request to a kcrisman branch), because that branch is not
well-defined.

rbeezer:
Dear Karl,

Now all three lead maintainers have objected to a proliferation of one-typo-per-PR commits.  Alex thinks it takes me two minutes, when it is more like five minutes to process each one, which includes adding a proper commit message.

If you have the capacity to *use* PreTeXt, then you do also have the capacity to make a local clone of a fork and properly manage a branch that can accumulate multiple typos.  It takes me about 30 seconds to combine multiple commits from one PR into one commit, even preserving attribution.  Please learn how to do this.

Please read about commit messages:
https://pretextbook.org/doc/guide/html/section-195.html

Help forming pull requests:
https://pretextbook.org/doc/guide/html/section-196.html

Please consolidate your open pull requests into such a branch and form a new pull request once you have something like 10 or 12 ready to go.

Thanks,
Rob

PR #1441, opened 2021-02-09
Typos and script instructions

davidfarmer:
Corrected typos in the sample article, and clarified the script instructions to indicate
that the script is run on the entire document, not on individual files.

Also had to fix a merge conflict because I had corrected the typos some time ago.


rbeezer:
Thanks!

The "typo" commit already existed, so when I fetched, it meant the branch started a long time ago.  Then the merge commit had **lots** of changes in it.  My attempt to remove it lead to a gitastrophe, which I was able to recover from.

Perhaps better would have been if you made a new branch at the tip, and then you could have cherry-picked the documentation change onto that empty branch.  Then the PR of that branch would not have had an existing commit and an unnecessary merge commit (unnecessary on my end).

I think.  ;-)

davidfarmer:
Yeah, in the middle of doing it, I realized that I failed in my attempt
to submit typos at a small marginal cost to you.  I don't have a good
process for maintaining long-running branches.  (That is not a request for
instructions or suggestions!  Explaining more git to me now would be a 
waste of your time!)

rbeezer:
No problem.  I learned something.  (I shoulda cherry-picked myself sooner, rather than later.)

Issue #1376, opened 2020-10-13
Waring about "alt text" missing from images


rbeezer:
From @jhpalmieri.  A warning, or some other low-level message, could be provided by the `schema/pretext-validation-plus.xsl` stylesheet when a `image` element does not have a `description` element with content ("alt text").  Perhaps there are other accessibility features that require an author's participation, and can be checked automatically.

kcrisman:
👍 very good idea

Alex-Jordan:
PR #1444 would address this, although it is stronger than validation plus, putting the check into the -common generic warnings.

rbeezer:
Implemented in schema/pretext-validation-plus.xsl, as proposed: an `image` with neither a `shortdescription` nor a `description`, and not declared `@decorative`, draws the image-description-missing report. Closing.

*Claude Fable 5, acting as a review assistant for Rob Beezer*

PR #1444, opened 2021-02-13
Add image/@decorative attribute

Alex-Jordan:
Catches each `image` without a `description`. Presently all WW images lack a description, and this is out of the author's control. So they are exempted for now.

Waiting to hear feedback about the basic approach here. Once that is settled, I'll do another commit that adds `description` as needed to all the repo's examples and documentation files.


Alex-Jordan:
Just a heads up that this will clash with #1445, assuming both PRs still put a new generic-warning into `-common`.

rbeezer:
Maybe it was me who interrupted our discussion of this.  If there's going to be a flood of warnings on every build, then authors are just going to ignore them, and new/critical warnings will never be seen.  So I'd definitely like to see the `for-each` loop in the validation-plus stylesheet, as suggested on #1376.  So when an author is ready, they can locate them.  The more we can do in validation, and the less at run-time, the better.

I'd compromise on a single message with a count, even if I don't really want to see it myself (like for the sample article!).  The one warning could suggest the validation-plus worksheet for more detail.

Nits:
* Could `[not(ancestor::webwork-reps)][not(description)]` become `[not(ancestor::webwork-reps|description)]'?
* I'm willing to let empty `description` slide for one big warning, and not in validation-plus?
* The match really only needs to be on `pretext` since the pre-processor should be "fixing" `mathbook`.

Alex-Jordan:
I'll sort this out after #1445 is resolved.

Alex-Jordan:
Force pushed. New summary:

* add a `@decorative` attribute for `image` elements
* in the schema as "yes" or "no"
* put to work in `pretext-html` and `extract-webwork`
* add a paragraph to author guide about `image/description` and `image/@decorative`
* add 3 checks to validation-plus stylesheet: (1) no description and no use of `@decorative="yes"` (2) description present and  `@decorative="yes"` (3) description longer than 125 characters
* add examples to sample article

Alex-Jordan:
Hmm, you will have to generate the downriver schema files if this gets merged. I think you asked me to submit the schema edit on a separate pull request so you will remember. But I forgot. So this is a reminder :)

rbeezer:
Thanks for putting this together - completely.

* `<attr>` only wraps the attribute itself, not an assigment to a value.  `...setting <attr>decorative</attr> to the value <c>yes</c>...`
* Every time the named `svg-wrapper` is called, the `description` param is passed in.  So I removed the default value from the definition.  Generally, I don't supply a default if I never expect it to be used.
*  No testing of validation+.  ;-)
* Just an observation for the record, which I am sure is what you intended.  This **removes** any "accidental" empty `@alt` attributes.

XSL-fu.  Following replacement has identical effect (but I had to check!).

```xsl
<xsl:when test="$decorative = 'yes'">
    <xsl:attribute name="alt">
        <xsl:value-of select="''" />
    </xsl:attribute>
</xsl:when>
```

```xsl
<xsl:when test="$decorative = 'yes'">
    <xsl:attribute name="alt"/>
</xsl:when>
```

git-fu

Easy to reorganize your commits/work to have the **final** commit have **only** schema changes.  When I do this on my end I have to work harder to keep your name on the commit as the author.  If rearranged this way when I get it, I can easily manufacture and combine derivative schema prooducts into the last commit.

```bash
git reset HEAD~1 (all work unstaged)
git add <files>
git add -p <file>
git commit ...
```

rbeezer:
The `sbsgroup` with the images in the sample article looked broken in the LaTeX output, though it was only that there was zero space between panels.  I spread it out a bit at 2c8346b31c8c6f31ff999dbfed95a7f1d5195173, which made some bad linebreaks with the `p` but that isn't really the point anyway.  ;-)  Not yet updated online.

PR #1453, opened 2021-03-08
Update the czech translation

jirilebl:


rbeezer:
That was fast!  Thanks very much for keeping this one up-to-date.  I hope you liked the less-crazy format of these files.

jirilebl:
Actually I didn't notice the change in the format, though I didn't really think about it that much.  I did notice that the file name changed :)

PR #1457, opened 2021-03-17
Worksheet addition, with new page structure

davidfarmer:
Added @daverosoff 's two-column worksheet, converted all worksheets
to use "page" tags instead of "pagebreak", and set workspace attributes
to use absolute minimum measurements ('in' or 'cm').

The new worksheet has attributes which currently are ignored but may
play a role in the future.

HTML compiles with minor complaints:
   An object (page) lacks a serial number
   An object (page) lacks a structure number

The title on a worksheet introduction may need attention.


rbeezer:
GitHub sez `projection1.png` is an empty file.  Can you **attach** a fresh copy here?  Don't make a new commit or anything, I've got stuff built on this branch already.  More comprhensive news soon.

davidfarmer:
Sorry.  Not sure how it got empty.

projection1

rbeezer:
Thanks.

rbeezer:
Thanks!  Forgot to close this.

PR #1458, opened 2021-03-19
Add instructions on building

bnmnetp:
This just adds some instructions on building the runestone sample as well as installing it in a Runestone server container.


rbeezer:
Thanks - that'll help us all stay sane!  I added a "dot" in the `rm` command, which is just a convenience I use for testing so I don't get confused with multiple runs.

PR #1459, opened 2021-03-20
WeBWorK: fix exercisegroup xrefs

Alex-Jordan:
I'm not sure why this was here. I think that it was needed at one time, but -common evolved nad now it is resposible for seeing `[NUM]` in output. After removing this, output is good.


rbeezer:
Thanks for the maintenance!

Issue #1160, opened 2019-08-04
Style info on HTML img elments


rbeezer:
I believe every `div.image-box`, and every `sidebyside` panel-div (class name is in flux, `sbspanel` right now) containing an image, has inside it an `img` element with

```
style="width: 100%; height: auto;"
```
which is how we get an image to fit the width of the **containing** div (box) while preserving aspect ratio and using only as much height as necessary.  One exception is when the sidebyside panel is a PTX `figure`, and then there is an intermediate HTML `figure` element.  So two cases
```
div.sbspanel
  img/style="width: 100%; height: auto;"

div.sbspanel
  figure
    img/style="width: 100%; height: auto;"
```
Once we rework `sidebyside` and class names settle down, could CSS provide this style information?

Follow-on from #1158

davidfarmer:
Yes, we can put that style on the img with CSS and not inline.

We are close to needing an upgrade of the Javascript (because of
work I am doing on reading questions), so it may make sense to
change the CSS at the same time.

Let's do a creful audit of any other ways that images might
appear.

rbeezer:
I'm 99% confident that a PTX `image` is bare, inside mainline `figure`, a panel of a `sidebyside`, or inside a `figure` inside a `sidebyside`.  But of course, I will check.  This is the beauty of keeping the schema up-to-date.  ;-)

I'll also see if there is any other CSS that might look like it needs backward-incompatible changes.  I'd love to retain `div.sidebyside` as part of obsoleting `div.sbsrow`, and would like to rename `div.sbspanel` to be `div.panel`, reasoning that this might make a screen-reader experience better if those class names were read aloud?   @Alex-Jordan Is this the case?

rbeezer:
Searched on `img` in HTML conversion.  Uses above.  Plus "posters" on videos (still images until they play), will perhaps do similar with `audio`.  MathJax logo, "brand logo" in masthead, some code for Statcounter that we don't choose to show.

rbeezer:
Beta at: https://pretextbook.org/beta/2019-08-04-image-style

Removed all `@style="width: 100%; height: auto;"` extant on HTML `img` elements.

davidfarmer:
I added CSS to replace the inline sizing on img tags (width: 100%, but I left off the
height because that does not seem to matter.  Let me know if I am missing something.)

All the images in the sample article look the same as originally, so the code can
be merged and this issue closed.

Presumably the related issues (#1182 for example) can be addressed by changes to
[...]

PR #1464, opened 2021-03-28
HTML changes for better CSS

davidfarmer:
Three small changes to handle HTML styling more appropriately:

a) A PTX 'paragraphs' now becomes a 'section' in the HTML.  (Formerly an 'article')

b) For an image, the style = "width: 100%; height: auto" is now class="contained".

c) In an sbspanel, the justify-content is now handled by a CSS class (with attribute
name the same as in the source).

Matching CSS changes have already been made (all that I caught, anyway).

Note that for c) the changes as shown in the pull request on GitHub do not give
an accurate impression of what edits were made:  a block of code which set the style
was moved higher and adjusted to add to the existing class.  The GitHub makes
it look like another block was moved down, and then something else was adjusted.

Issue #1160 can be closed after merging.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -5550,9 +5550,8 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 <xsl:attribute name="src">
                     <xsl:value-of select="@source" />
                 </xsl:attribute>
-                <!-- replace with a CSS class -->
-                <xsl:attribute name="style">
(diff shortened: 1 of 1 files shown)


rbeezer:
Looking real good.  I'll get serious about merging in the AM.   About quitting time here.  Thanks.

(Isn't it annoying when the diff doesn't really reflect what you did?  I've made even bigger messes, so have some practice reconstructing the pieces.)

rbeezer:
Merged, and website being re-built right now.  Great to see the `style` elements being lessened.  Thanks.

Issue #1039, opened 2019-04-10
Migrate to MathJax 3


rbeezer:
This is a meta-issue for notes, or related issues involved with an upgrade to MathJax 3, which at this writing is still in beta.

rbeezer:
Using the https://cdn.rawgit.com/mathjax/mj3-demos/3.0.0-beta.3/mj3-tex2html-beta.dist.js variant in a very quick-and-dirty way, seems to resolve the bold math problem at #1040.

rbeezer:
Knowls in display math seem to act better with same variant as above.

rbeezer:
Seems to be some changes in font sizes, noticed in #930.

rbeezer:
Just upgraded my testing branch to beta 4, the final one.  So using

https://cdn.rawgit.com/mathjax/mathjax3/tex-chtml.js

Nothing startling to report.

Alex-Jordan:
Just a note to look into the color.js extension with the migration to MJ3.

rbeezer:
I have no reason to believe that MJ will do a better job of keeping text after math for not line-breaking.  But worth testing.

#697

rbeezer:
Looks like MJ3 is not really mature just yet.  We'll want to watch this:

http://docs.mathjax.org/en/v3.0-latest/upgrading/v2.html

sean-fitzpatrick:
With MathJax 2.7.5, superscripts on the | (\vert) character don't render correctly when accessibility is turned on. This can be worked around by using braces: |^2 doesn't work right, but {|}^2 does.
Maybe using \vert instead of the keyboard character would fix things, too.
(Also, \right| seems OK, but \Big| is not.)

sean-fitzpatrick:
Testing using the utility on the MathJax site suggests this is fixed in 3.0

Alex-Jordan:
When moving to MathJax 3, or after the move, monitor these two issues we currently have and see if they go away:

1. Set MathJax to SVG rendering. Then load [this page](https://pretextbook.org/examples/sample-article/html/section-7.html#p-216). Open the example knowl and see if the second equation column is aligned well.

1. Set MathJax to SVG rendering. Go to [here](https://pretextbook.org/examples/sample-article/html/section-7.html#p-227) and see if the knowls that are inside the math function as they should. Current behavior is that you click them, they open, and then instantly close.

rbeezer:
[...]

PR #1469, opened 2021-04-24
First attempt at updating to MJ3.

zorkow:
As announced in my email last night: Here is a first attempt at integrating MJ3. 
The main issues are:
1. In some proofs contain elements that are rendered but there is no context menu/a11y available. 
2. Math in knowles used as cross-references is not rendered.

If you could give us a hint how to fix problem 1, that would already be sufficient for the experiments we have in mind.

### DIFF
--- a/script/mjsre/package.json
+++ b/script/mjsre/package.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "mathjax-full": "^3.0.5",
+    "mathjax-full": "^3.1.4",
     "speech-rule-engine": "^3.1.0-beta.4",
     "yargs": "^15.4.1"
   }
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -10985,34 +10985,27 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:text>&#xa;</xsl:text>
         <!-- // contrib directory for accessibility menu, moot after v2.6+ -->
         <!-- MathJax.Ajax.config.path["Contrib"] = "<some-url>";           -->
-        <xsl:text>MathJax.Hub.Config({&#xa;</xsl:text>
-        <xsl:text>    tex2jax: {&#xa;</xsl:text>
-        <xsl:text>        inlineMath: [['\\(','\\)']]&#xa;</xsl:text>
-        <xsl:text>    },&#xa;</xsl:text>
-        <xsl:text>    asciimath2jax: {&#xa;</xsl:text>
-        <xsl:text>        ignoreClass: ".*",&#xa;</xsl:text>
-        <xsl:text>        processClass: "has_am"&#xa;</xsl:text>
-        <xsl:text>    },&#xa;</xsl:text>
-        <xsl:text>    jax: ["input/AsciiMath"],&#xa;</xsl:text>
-        <xsl:text>    extensions: ["asciimath2jax.js"],&#xa;</xsl:text>
-        <xsl:text>    TeX: {&#xa;</xsl:text>
-        <xsl:text>        extensions: ["extpfeil.js", "autobold.js", "https://pretextbook.org/js/lib/mathjaxknowl.js", "AMScd.js", ],&#xa;</xsl:text>
-        <xsl:text>        // scrolling to fragment identifiers is controlled by other Javascript&#xa;</xsl:text>
-        <xsl:text>        positionToHash: false,&#xa;</xsl:text>
-        <xsl:text>        equationNumbers: { autoNumber: "none", useLabelIds: true, },&#xa;</xsl:text>
-        <xsl:text>        TagSide: "right",&#xa;</xsl:text>
-        <xsl:text>        TagIndent: ".8em",&#xa;</xsl:text>
-        <xsl:text>    },&#xa;</xsl:text>
-        <!-- key needs quotes since it is not a valid identifier by itself-->
-        <xsl:text>    // HTML-CSS output Jax to be dropped for MathJax 3.0&#xa;</xsl:text>
-        <xsl:text>    "HTML-CSS": {&#xa;</xsl:text>
-        <xsl:text>        scale: 88,&#xa;</xsl:text>
-        <xsl:text>        mtextFontInherit: true,&#xa;</xsl:text>
-        <xsl:text>    },&#xa;</xsl:text>
-        <xsl:text>    CommonHTML: {&#xa;</xsl:text>
-        <xsl:text>        scale: 88,&#xa;</xsl:text>
-        <xsl:text>        mtextFontInherit: true,&#xa;</xsl:text>
-        <xsl:text>    },&#xa;</xsl:text>
+        <xsl:text>window.MathJax = {&#xa;</xsl:text>
+        <xsl:text>  tex: {&#xa;</xsl:text>
+        <xsl:text>    inlineMath: [['\\(','\\)']],&#xa;</xsl:text>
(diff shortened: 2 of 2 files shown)


zorkow:
Not sure why I can't pick reviewers, so I ping you directly:
@davidfarmer @Alex-Jordan @rbeezer

davidfarmer:
If someone can post an html version with these changes, I can look into the knowls issues.

zorkow:
Here's the version of the Tom's abstract algebra book:
https://zorkow.github.io/mathbook/mj3-samples/aata/aata.html

zorkow:
@Alex-Jordan and I looked at [section 2](https://zorkow.github.io/mathbook/mj3-samples/aata/aata.html) last night. 
Three types of elements:
1. Rendered and accessible (i.e. with context menu)
2. Rendered and inaccessible (all inside the hidden proofs)
3. Not rendered, inside knowles (e.g., the crossreferenced knowle like Theorem 2.10 at the start of the Euclidean Algorithm section).

(3) is not a surprise as those live in separated html docs which do not contain a MathJax instance.
(2) are the ones that I am not sure what to do about.

zorkow:
PS: Yesterday I came across a couple of proof environments (usually at the top of the page) that contained accessible elements. But I can't find those anymore now.

rbeezer:
Thanks, @zorkow.  We'll find this very useful for #1039 very soon.  So once we get the knowl issue sorted, this could become the basis for that issue.

A.  Some content is born hidden in knowls always, like hints to exercises and footnotes.  Others are elective, like examples.  Others are elective but "on" be default, like a proof.  This content is in a `div` on the page, so MJ must be more readily available.

B.  Cross-references access duplicates of content stored in files in the `knowl` directory.  As you mention, MJ would be less readily available.  I think David hits these with MJ when they open, but he'll need to provide correct guidance.  A cross-reference to a theorem will open the theorem in a knowl, and likely it will have a proof that is a knowl.  That proof would be "on the page" in (A), but here it is a component of a duplicate of the proof and so there is another file just for this proof duplicate.

So your three scenarios are (1) "regular" math content, (2) my (A), born hidden, (3) my (B), cross-reference knowls?

I don't believe your PS.  ;-) ;-) ;-)

Rob

davidfarmer:
For most of the proofs in AATA, the content is hidden in the page and then
shown when the proof knowl is clicked.

MathJax should not process that content until it is shown.  So, I changed a
line in the configuration to

    ignoreHtmlClass: "tex2jax_ignore",

(Previously in the quotes was ".*").

However, that made no difference.  I also tried, to no effect,

    ignoreHtmlClass: "hidden-content",

Is there a different way to stop MathJax processing content, so it can be done later?

davidfarmer:
I think MathJax is not ignoring material it was asked to ignore.  I posted to
MathJax support.

davidfarmer:
As Davide explained on MathJax support, there are errors in the HTML.
These can be fixed by:

In xsl/pretext-html.xsl, lines 10982-10984 should be deleted:  that script
tag should not have any attributes.

Line 11024 or 11017 (depending on which version you look at),
which implements "close of MathJax.Hub.Config", should be deleted. That
line contains:  "});"

This one I figured out before asking Davide: line 10998 should replace ".*"
by "tex2jax_ignore".

Making the above changes has things basically working, at which point
we find that the "align*" environment is not recognized.  In line 10995,
add 'ams' to the list.

Now the "notsubset" command is broken.  That is odd, because it is defined
on the page.  Sort-of:  On line 10995 also add 'newcommand' to the list.

With those changes, things I checked in AATA were all working.

I have not checked sage cells in knowls, and surely knowls in display math are
not working, but I don't think there are examples in AATA.  I also did not
check remote knowls (because I just hand-edited one HTML file).
I'll look when a new version is posted.

Whoever made those markup errors in the new configuration can
thank me for accepting the blame on the MathJax support group.

rbeezer:
The `newcommand` package was a bit more work than suggested.  Enough details here:

http://docs.mathjax.org/en/latest/input/tex/extensions/newcommand.html

I'm getting pretty good behavior with this now, with fixes from @davidfarmer.  `\sfrac` is broken, and of course, `\knowl`.

Putting in `mathjaxknowl3.js` breaks everything, so you'll need to create a minimalist version, or edit it out of copies to get things working initially.  URL looks right in "not found" error message in the console, so I think the configuration is correct.

Full beta, Section 7 is a good place to start.  I've not yet tested any known problems we hope v3 would fix, this is just to get knowls code working.

https://pretextbook.org/beta/2021-05-07-mj3/section-7.html

pkra:
@rbeezer https://pretextbook.org/js/lib/mathjaxknowl3.js is wrong. To kick things off, I used the "inline configuration" approach, not an extension. I've updated https://codepen.io/pkra/pen/JjWoZVp and added comments "START HERE // END HERE" - if you place just that code into the js file, it should work. 

How about I create a repo for the new knowl extension? Since client-side extensions require extra build steps, it's better as a separate repository. (And for server-side rendering, you said you might want a different behavior so it's good to have both.)

davidfarmer:
I replaced the contents of mathjaxknowl3.js , but that did not
fix it.

The configuration for slanted fractions is wrong for MathJax3
(There is no MathJax.Hub in MJ3), but that is not the problem
because I commented that out in a local version.

With that commented out, I do not see any error messages in the console.

pkra:
> I replaced the contents of mathjaxknowl3.js , but that did not
> fix it.

Next try: add "knowl" to the array in

    packages: {'[+]': ['base', 'extpfeil', 'ams', 'amscd', 'newcommand']}

davidfarmer:
Success!

Not complete success, because the knowl content looks terrible,
probably because I am putting the knowl in the wrong location
(because I have not yet updated the knowl code to handle that
case).  But I am on reasonably solid ground to take care of that
myself.  I'll do that soon (unlikely today, though) and will report
how it goes.

pkra:
Thanks for the update, David!

rbeezer:
I added `'knowl'` to the package list.  Beta is being refreshed right now, give it 10 minutes to upload.  Same location as above.

Knowls are functional!  (Needs work, but they are **are** functional.)  `\sfrac` is the only obvious thing that is wrong.

@pkra - I'll let @davidfarmer advise on repositories, etc.  He's got a plan for where resources like this live.

Once this settles, I'd like to pursue #1471.

davidfarmer:
Knowls code adjusted so that knowl content from math looks as expected.

Appears after teh display, as before.

I don't think it is possible (certainly not easy), and maybe not even
desirable, for the knowl to appear split apart a multiline display.

For the repository, is there a reason to handle "mathjaxknowl3.js"
differently than "mathjaxknowl.js"?  That is, add it as a file in
https://github.com/PreTeXtBook/JS_lib ?

pkra:
> For the repository, is there a reason to handle "mathjaxknowl3.js"
> differently than "mathjaxknowl.js"?  That is, add it as a file in
> https://github.com/PreTeXtBook/JS_lib ?

That depends largely on how you want to handle server-side rendering. I'm not actually sure but I'm guessing mathjaxknowl3 will work with MathJax's "components" example for server-side use (e.g. https://github.com/mathjax/MathJax-demos-node/tree/master/component)  but it won't work with the "direct" example (e.g. https://github.com/mathjax/MathJax-demos-node/tree/master/direct) which I personally prefer.

rbeezer:
Original code from PR is at pushed 9c91af4e107ea75c8f255624aba2d4742ab6b3fb - thanks for starting this @zorkow!

Changes discussed here are on other commits.

PR #1470, opened 2021-04-25
Stylesheet for PG macros

Alex-Jordan:
This is a new stylesheet to build a .pl file that will contain perl macros to support WeBWorK problems.

The only thing it has right now is a copy of the latex-image-preamble. I can confirm that with the resulting file, if it is loaded as a macro library by a WW problem, then I can make a tikz image in that WW problem and it uses the latex-image-preamble styling. Here is a live example:

https://spot.pcc.edu/math/orcca/ed2/html/section-slope.html#XRM

I have not yet tried to understand why the font size is so small, but the styling (yellow background, "ruby" line, etc) is all from latex-image-preamble packaged from this stylesheet into a perl library.

Eventually pretext/pretext will use this stylesheet and know to name the file something like "ORCCA.pl".

Other things that will go here eventually:

- LaTeX macros.
- PG macros specific to the PTX project. For example ORCCA has some that only exist to support unit conversion exercises in the unit conversion section.

### DIFF
--- a/xsl/pretext-pg-macros.xsl
+++ b/xsl/pretext-pg-macros.xsl
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<!-- ********************************************************************* -->
+<!-- Copyright 2021                                                        -->
+<!-- Robert A. Beezer                                                      -->
+<!--                                                                       -->
+<!-- This file is part of PreTeXt.                                         -->
+<!--                                                                       -->
+<!-- PreTeXt is free software: you can redistribute it and/or modify       -->
+<!-- it under the terms of the GNU General Public License as published by  -->
+<!-- the Free Software Foundation, either version 2 or version 3 of the    -->
+<!-- License (at your option).                                             -->
+<!--                                                                       -->
+<!-- PreTeXt is distributed in the hope that it will be useful,            -->
+<!-- but WITHOUT ANY WARRANTY; without even the implied warranty of        -->
+<!-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         -->
+<!-- GNU General Public License for more details.                          -->
+<!--                                                                       -->
+<!-- You should have received a copy of the GNU General Public License     -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks for getting this started.  Some questions/nits:

*  With assembly (good!) `mathbook` at the top level gets replaced.  So  `select="mathbook|pretext"` can be modernized.
*  Does this **need** to be a separate stylesheet?  Could it be created via `exsl:document()`?  As-is you'll need a predictable output filename, but maybe it'll *always* be controlled by the `pretext/pretext` script?  I guess I don't know just when this is being used, and how?
* If it **must** be a separate stylesheet, I've started putting these into a `xsl/support` directory.  The idea is that eventually `xsl` will just have "user level" stylesheets, the "main" ones for creating a primary output.

Alex-Jordan:
> Does this need to be a separate stylesheet?  Could it be created via exsl:document()?

As part of which other stylesheet? There is `extract-pg.xsl` which builds a python dictionary and is only ever invoked by pretext/pretext during the same run of pretext/pretext that sends problems to the WW server for processing. At that point in time, you need to have already made this .pl file and put it on the server. It needs to be there waiting to be used when the problem files are processed.

In this new stylesheet, it could (should) use exsl:document(). Then the name of the file can be forced, which is good so that in extract-pg.pl, we can load the macro file automatically using the same name.

> I guess I don't know just when this is being used, and how?

1.) You have edited your book and you have changed things with "latex-image-preamble", or changed your LaTeX macros,  or changed your pg-macros.
2.) You run this to make a file like `ORCCA.pl`.
3.) You upload `ORCCA.pl` to the webwork host course, updating any old version that was already there.
4.) You make webwork-representations, with the updated pl file there in the host course, ready for use.

> a xsl/support directory

Sure, I can move it there.

rbeezer:
Thanks for the explanations.  Yes, a consistent naming process will save authors/publishers some grief.

Sounds good.  I'll look for a force-push?  Thanks.

Alex-Jordan:
Closing. Bringing this into my `pg-tikz` branch with changes.

PR #1477, opened 2021-05-19
Miscellaneous side commits during PGtikz work

Alex-Jordan:
While I was doing the last batch of PGtikz edits, I ran into these small things that I wanted to change. They are small things that didn't really fit into the PGtikz pull request. So I separated them, but I still think they are good changes to assess separately. It may help to look at them one commit at a time.

Why `<webwork/>` in place of `<webwork />`?  It made it a little easier for me to search for the real `<webwork ...>` that are not just generators.

The "description fix" fixes something that is actually broken.

One commit is trivial typo correction for WeBWorK camelcase.

The other two (I believe) give helpful debugging info during image and webwork extraction.

### DIFF
--- a/doc/guide/author/overview.xml
+++ b/doc/guide/author/overview.xml
@@ -399,10 +399,10 @@
     </section>
 
     <section xml:id="overview-webwork">
-        <title><webwork /> Exercises</title>
-        <idx><h><webwork /></h><h><webwork /> exercise</h></idx>
+        <title><webwork/> Exercises</title>
+        <idx><h><webwork/></h><h><webwork/> exercise</h></idx>
 
-        <p>It is possible to author <webwork /> automated homework problems<idx><h><webwork /></h><h><webwork /> exercise</h></idx> directly within your source.  A static version will be rendered in your <latex />/PDF/print output, and a <q>live</q> version will be rendered into HTML output (though a student is not able to authenticate against a course).  However, you can also extract <em>all</em> of the <webwork /> questions from a textbook into a single archive suitable for uploading into a traditional <webwork /> server.  This is a big topic, so see the dedicated <xref ref="webwork-author" /> for details.</p>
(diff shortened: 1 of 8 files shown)


Alex-Jordan:
Added one commit with something I had forgotten about. What to do with `$`, `\@`, and `%`? They each have special perl meaning, and two of them have special LaTeX meaning.

In the text() template, I added the right substitutions. And I added an example to the sample chapter. And I expanded the documentation to some discussion of special characters.

It's no problem if I should rearrange things for PRs, now that this one has one "important" thing mixed in among the earlier commits.

rbeezer:
No problem.  I can always split it out myself if I feele it is necessary.  I'll take a closer look very soon.

rbeezer:
Thanks for the clean-up!  I'm glad I will never need to author `\~~$`.  ;-)

Nits.  There are always nits.  ;-)

*  Let's use `_verbose()` for progress indicators for the impatient.  And `_debug()` for blow-by-blow indicators of where trouble lies.  I didn't look closely, but **renaming** a file feels more like the latter.
*  Do you think `$refreshCachedImages` might be controlled by an attribute somewhere?  Which might even see greater utility?  E.g., `webwork/@status="draft|stable"`?

Looking forward to more on #1470!  Website will be updated in a few minutes.

Alex-Jordan:
Thanks. Did you change to `_debug` then? If not I can fold that in to the next round.

> Do you think $refreshCachedImages might be controlled by an attribute somewhere? Which might even see greater utility? E.g., webwork/@status="draft|stable"?

That is worth thinking about. But my head goes to projects that are eternally in draft form. Even then with the project as a whole undergoing evolution, `$refreshCachedImages` should not be set to true unless the image itself is stable. So a global attribute doesn't feel right for the local stability of the image.

Alex-Jordan:
source: `\~~$`

WeBWorK translation of source (`\` to `\\`, `~~` to `\`): `\\\$`

Perl interpolation of WW translated source: (`\\` to `\`, `\$` to `$`): `\$`

And in the end, the tex document has its `\$`.

rbeezer:
Website upload got hung-up.  Retrying now.  Have not yet fixed missing images on other PR.

rbeezer:
> Did you change to _debug then?

Did not - I wanted a clean merge without labeling all six commits with the PR number.  ;-)

Just noticed that the URLs are really long and ugly.  Utility of that as a progress indicator might be debatable.

rbeezer:
> Thanks. Did you change to _debug then? If not I can fold that in to the next round.

In df7d8a6238226192ee531eed2bcd177af39992f7, at least.

PR #1482, opened 2021-05-25
WeBWorK: explicitly ask for solutions and hints from server

Alex-Jordan:
In the URLs that we use to return content from the WW server, this change explicitly asks for solutions and hints to be returned as well. Until recently, they were always returned.


rbeezer:
Switches control if these are in the source, no?  (On fone.) In other words these settings just explicitly allow these to be generated, **if present**?

Alex-Jordan:
Right. This is not about what the reader sees. Only about what `pretext` gathers in the representations file.

Another way to put it: this PR makes things be the way they have always been. It compensates for a change that happened with WW 2.16.

rbeezer:
Got it, thanks.

rbeezer:
All done.  Thanks very much.

PR #1483, opened 2021-05-26
WeBWorK: document simpler vars in tikz

Alex-Jordan:
Not urgent. I realized there is nothing wrong with doing like:
```
\draw (0,0) -- ($a,0);
```
in tikz code in a webwork, instead of having to do:
```
\draw (0,0) -- (<var name="$a"/>,0);
```
Latter is still an option, and it rare cases, may be necessary.

The former is much more pleasant. Such is my experience after an afternoon converting one section of APEX webwork images to tikz.

### DIFF
--- a/doc/guide/author/webwork.xml
+++ b/doc/guide/author/webwork.xml
@@ -391,8 +391,14 @@
         </p>
         <p>
           Your image code can use variables that you defined in the <tag>pg-code</tag> section.
-          Include variables just as you would anywhere else in a problem statement,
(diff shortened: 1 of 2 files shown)


rbeezer:
Blown closing tag on a `c` element.  Can you force-push a better version?

Alex-Jordan:
Sorry about that!  I thought I couldn't mess up something so small. Force pushed.

rbeezer:
> I thought I couldn't mess up something so small

I think that all the time, but still catch myself often with a quick test.  ;-)

PR #1484, opened 2021-05-26
Update README.md

dqnykamp:
I believe the element name for attributes is attr.


rbeezer:
PreTeXt Developer!  Thanks for the fix.  I renamed this shortly after it was first built, and I guess I forgot it was documented here like that.  Appreciate you taking the time for a PR.

I edited your commit message (which I often do), so rebased as well.  For hints on cleaning up, see

Cleaning Up 
https://pretextbook.org/doc/guide/html/section-203.html#paragraphs-94

PR #1489, opened 2021-06-01
JSXGraph JessieCode

dbrianwalton:
Modified xsl/pretext-html.xsl to look for slate within an interactive with @surface="jessiecode" and import the resulting JessieCode script. Supports script included directly in slate as text (requires escaping characters) or as a separately loaded script file.

Modified sample-article.xml to include a relevant section with two examples, one using directly included script and another using a script file inside the code/jsxgraph/ folder.


rbeezer:
Visual first-pass through code looks good.  I'll get to this more seriously in a day or two.  **Thanks** for the contribution!

dbrianwalton:
Rob, I'm not sure how to amend a pull request, but I just added commit 04a31d0 to make this more consistent with other interactives in terms of sizing the iframe content.

rbeezer:
Great contribution - thanks @dbrianwalton.

Website rebuild is broken, so advertising will need to wait.  (Unrelated to this work.)

Providing "preview" screenshots (images) for each example will greatly improve the PDF realization.

Two git nits:
* Unlikely your branch needs any merge commits, and they mostly just get in my way.  Development is not active enough (yet!) to worry too much about staying current with other changes since they are usually disjoint.  You can rebase onto `dev` privately before pushing as the best way to catch up with newest code.
* I think your editor strips trailing spaces.  Which is good (we shouldn't have them).  But I like trivial whitespace changes on their own commits.  And you must have touched `xsl/mathbook-latex.xsl` which made things go really crazy when I rebased.  It almost looked you built your branch on some really old commit?  Anyway, I bluffed my way through it, but I didn't really understand what was happening.

I repackaged everything as two commits (code, examples), all with your name on them.

PR #1490, opened 2021-06-03
WeBWorK: use document-language

Alex-Jordan:
This changes from using `pretext/@xml:lang` to using `$document-language` to address the problem I caused with #1486.


rbeezer:
Thanks!  I should be able to merge tomorrow afternoon.

rbeezer:
Not sure I understand the previous failure, but this seems to be behaving now.  Thanks very much.  Website update after another PR.

Alex-Jordan:
Previous failure: the python file that `extract-pg` builds was trying to use the variable `localization` even when it was never defined. Now in those instances it has:

```
localization = 'en-US'
```

Maybe later I will change the URL-maker to leave it out of the query string when the language is English, as you suggested.

rbeezer:
Got it!  Thanks.  Yes, that change would be fine.

PR #1491, opened 2021-06-03
Additional fr translations

JeanSebastienTurcotte:
I translated the newly added Appendices (plural) to... Appendices and upon researching found out that Listing also translates to Listing.


rbeezer:
Thanks for helping keep this current with the most recent additions!

PR #1494, opened 2021-06-10
Update Czecjh translation

jirilebl:
Translate Appendices, and actually a more Czech less Czenglish translation of Appendix to begin with.  I have seen both in use, but Dodatek is definitely more Czech and less a loan-word.  Plus a Czech person might more think of going to the hospital with an appendix than finding it in a book (though a medical book might have an appendix on the appendix).


jirilebl:
And I misspelled Czech in the title ... the irony of that.  I'll leave it as is and take the humiliation ...

rbeezer:
Well, your bad spelling is more than counteracted by the entertaining description.  ;-)  Merged with an edited commit message.  Thanks for keeping up with the changes here.

PR #1496, opened 2021-06-11
Guide: broaden scope of FAQ#1

Alex-Jordan:
I think we could point to FAQ#1 more often if it were not specific to knowls and sage.


rbeezer:
Perfect!  I **love** the blank, plus it is great to get all that local server stuff de-duplicated into a cross-reference.  Thanks for getting this one done.  On website soon.

PR #1497, opened 2021-06-11
GeoGebra: do the right thing with all whitespace text content

Alex-Jordan:
This addresses an issue that Juan Carlos Bustamante and D Brian Walton unearthed here: https://groups.google.com/g/pretext-support/c/5L_XUwcWzBw/m/EuVniXy0AgAJ

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -6129,7 +6129,16 @@ the xsltproc executable.
                     </interactive>
                 </figure>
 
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks for touching this up (including the sample article).  Two comments, neither is important.

*  From early on, I'd use `normalize-space(.)` in this situation.  Not sure if that is better or worse.  The `slate` should have content that is just **one** text node anyway, so perhaps irrelevant.
*  I was tempted to see if the `if` could have been consolidated with the `when`, perhaps **after** testing all the attributes, but I can see it is more complicated than that (since content can be the whole applet, or just a modification).

PR #1498, opened 2021-06-11
Sage: load special MathJax3 findScript config when sage is used

Alex-Jordan:
Before this, try `pretty_print(html("$a^2$"))` in a Sage cell from `section-sage-cells.html` of the sample article. With MJ3, you will see no output. After this, you will see the output math.

With each test, make sure to clear your cache. The MathJax config from a previous run may stick around.


rbeezer:
I added a follow-on commit with this example.  it could be doctested to alert us to changes, but that's not been part of our routine.

At 7350002a0d780bdd2ade8cb6a01f60bd93f606b8

PR #1499, opened 2021-06-17
pretext Python script: add command line option for config file

jjrsylvestre:
Allows authors with multiple projects to easily have per-project user settings for the pretext script.

### DIFF
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -102,7 +102,7 @@ def string_parameters_as_dict(pairs):
 # Configuration File
 ####################
 
-def get_executables():
+def get_executables(arg_supplied_config_file):
     """Query the configuration file, return dictionary of executables/commands"""
 
     import configparser # ConfigParser()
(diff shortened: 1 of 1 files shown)


rbeezer:
@jjrsylvestre , I guess I've made sort of a mess of this one.  I couldn't get to it before the workshop in Portland, and the CLI project asked me to create a better split between the front-end and back-end of the`pretext/pretext` script, and specifically with the configuration file.  See 68673d29cf1c2c1f5d2e918aa411cbba65d26713

I thought maybe I could split this up by hand (I'm pretty sure it won't rebase easily!) but I'm not confident I would get it right.

Do you think you would be able to put the pieces back together in the right places?  It'd be great to have this flexibility.

Again, I'm sorry I didn't move on this faster, and get it in before the big workshop push.

jjrsylvestre:
No problem, it was easy to redo. Even easier than before with the new reorganization of pretext/pretext vs pretext/pretext.py.

rbeezer:
**THANKS!**  First in line for take-off in the morning.  Not going to let this one rot.

rbeezer:
Live now.  Merged as-is, just with edits on commit message.  I'm glad this was not a lot of trouble!

*  Never thought to use capital letters for switches.  :-)
*  I've been in the practice of converting user-supplied filenames into absolute paths, first chance I get.  If `-C` gets a non-existent file, there is no error due to the list of files, and `-vv` indicates the "problem".  What do you think of calling `verify_input_file()` in `pretext/pretext` to convert to an absolute path (or `None`)?  The error need not be fatal (just report `None` if the file does not exist), but an error message would be more prominent, and a mis-typed filename (say) would be much less mysterious.

Thanks for this.  Its a great feature to have.

PR #1503, opened 2021-06-28
Add a worksheet to showcase article

mitchkeller:
This adds a worksheet from the sample article to the showcase article, including the various image files. It also rather silently moves the baseurl from docinfo to the publisher file.


examples/showcase/publication.xml
```diff
@@ -22,6 +22,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <publication>
 
     <html>
+	<baseurl href="https://pretextbook.org/examples/pretext-showcase/html" />
```
Alex-Jordan, line 25:
```suggestion
        <baseurl href="https://pretextbook.org/examples/pretext-showcase/html" />
```

examples/showcase/worksheets.ptx
```diff
@@ -0,0 +1,138 @@
...
+        </figure>
+        <figure xml:id="figure-triangle-cyclic-vectors">
+          <caption>The sides of a triangle presented as vectors.</caption>
+          <image xml:id="worksheet-triangle-cyclic-vectors">
+            <latex-image>
+              \begin{tikzpicture}[xscale=1.5, yscale=1.5]
+              \draw[->,>=latex, style={black, thick}] (0,0) -- (2.30 , 0) node [below] {$\vec{A}$} --  (3,0);
+              \draw[->,>=latex, style={black, thick}] (3,0) -- (1.47, 1.53) node [above right =1mm] {$\vec{B}$} -- (1,2);
+              \draw[->,>=latex, style={black, thick}] (1,2) -- (0.23 , 0.46) node [above left=1mm] {$\vec{C}$} -- (0,0);
+              % \draw[->,>=latex, style={black,semithick,dashed}] (1,2) -- (7/6, 4/3) node
```
Alex-Jordan, line 42: Do we need all of the commented out lines here? The cleaner the Showcase Article is, the better.

Alex-Jordan:
Is it a problem that adding a worksheet here introduces a substantial schema violation? I checked, and there are currently three. I feel like the showcase article should have zero.

And a nit: can we write in the showcase article like:
```
<p>
    This is a paragraph.
    It line breaks in git diff friendly ways.
</p>
```
?

Instead of like:
```
<p>This is a paragraph. It does not line break in git diff friendly ways and you may need to side scroll to read this sentence all the way to the end.</p>
```

examples/showcase/worksheets.ptx
rbeezer: I prefer not to have "commented out" stuff in source.  It confuses searches for problems, etc.  And...it can be recalled from the git history.  Will look locally in a minute.

rbeezer:
To merge shortly.  I

* consolidated whitespace type fixes into Mitch's main commit, included Alex's lightning commit.
* added my own commit to trash commented-out lines in some `latex-image` code

@Alex-Jordan: I would accept a reformatting commit, and you can negotiate with @mitchkeller as to whose responsibility that is.  ;-)

rbeezer:
Added TikZ `positioning` library into showcase article, so LaTeX version builds.  Comments-excised is on a separate commit.  Thanks for getting this started @mitchkeller - we'll need to be aware of any schema changes.

PR #1505, opened 2021-06-29
Update mj-sre-page.js to work with MathJax 3.2.0

mitchkeller:
Incorporates the fix from @dpvc for MathJax 3.2.0 and updates packages.json to current versions.


rbeezer:
Thanks very much @mitchkeller  and @dpvc!

PR #1506, opened 2021-06-29
Hardcode assets directory destinations rather than using provided relative paths

StevenClontz:
This makes the necessary changes in Python - XSL needs to be updated accordingly.


rbeezer:
Merged, with a follow-on commit (90e811f06051806e7438aeadc21c8748a2d43f5d) to adjust creation via XSL.  Also, experimental warnings removed and now `pretext/pretext` is at Python 3.6.

PR #1507, opened 2021-06-29
Remove Vagrant appendix prior to deprecation

daverosoff:
I couldn't find any references in any other sections.


rbeezer:
Merged.  Thanks, Dave.  Send me a message for `pretext-announce` warning about discontinuation at the end of August, and I will broadcast.

PR #1508, opened 2021-06-30
Add XML processing directive for EPUB/Kindle XHTML files

mitchkeller:


rbeezer:
Thanks, @mitchkeller.  Will push once GH brownout ends.

PR #1509, opened 2021-06-30
Prevent (some) EPUB readers from paging to endnotes and remove duplic…

mitchkeller:
…ated titles from asides


rbeezer:
Great - thanks!

PR #1516, opened 2021-07-01
WeBWorK: Use iframe for live html

drgrice1:
This won't work until we simultaneously submit a companion pull request to JScore.


rbeezer:
Thanks, @drgrice1!  First-time contribution!  ;-)

PR #1528, opened 2021-07-06
WeBWorK: allow Make Randomize button even when there is no answer

Alex-Jordan:
The code I am removing here prevented the Make Randomize button from showing when there was no answer field in the problem. This seemed to make sense at the time. But:

1. There are applet-based problems where you do submit answers even though there is no answer input field.
2. Maybe WW could be used someday to just show random versions of something even though there is nothing to submit.
3. Really "Make Randomize" should only be prevented when there just are no new random versions. And this did not do that.

Because of 1, @drgrice1 and I were seeing a problem without "Make Randomize" that should have had it.


rbeezer:
I removed a pile of documentation right before this, which might have been very confusing after the change (it confused me briefly!).  I **think** I did the right thing.  Please review the commit:
* Good? Close this.
* Bad? Make a new PR with corrections.

Update in a bit, will look at other PR first.

Alex-Jordan:
Good catch.

> Good? Close this.

Closing.

PR #1530, opened 2021-07-08
EPUB/Kindle: Various formatting enhancements and fixes

mitchkeller:


rbeezer:
I'm running with this now.

rbeezer:
Thanks, @mitchkeller .  Really nice work.  Expanded to 8 commits.  Tested sampler as EPUB, seems OK in Calibre.  Built Kindle just fine, looks OK in Calibe, including **downshifted** math.


Notes:

*  `&ASIDE-LIKE` in the `match=` will behave as you expect.  But when prefaced with `$document-root` it will be syntactically correct, but will not behave as expected.  An entity just results in a literal substitution, think about that in this case.  So I expanded with aside, historical, biographical.  Since the `aside` was first, you got the result you were looking for.
*  The three additions of `epub.css` and `kindle.css` could have been done with a named template, for consistency and economy?  I left it as-is.
*  I left the `**stringparams`.  I didn't have the energy to think about conflicts with the publisher file in the list of parameters.  Should likely be done once, carefully, in the front end.  I think.  Could be already?
*  Stray changes in indentation, tabs, newlines are of varying importance, and a fiddle to get cleaned up.  Most important: no tabs, and please indent new code with 4 spaces.  That's house style.

Thanks for advancing this one.  Gonna be great to have quality EPUB and Kindle.

PR #1532, opened 2021-07-10
EPUB/Kindle: Cleanup matters

mitchkeller:
I think I got indentation right (hadn't even noticed the odd spots tabs were sneaking in before). Tried to bundle into nice atomic commits so you can deal with them separately as you see fit. This is ready for review.

### DIFF
--- a/css/README.md
+++ b/css/README.md
@@ -30,7 +30,7 @@ number when that changes.  At a minimum, change the date and
 commit that as a record of a "no-change" update.
 
 CSS version: 0.31
-Date: 2021-07-02
+Date: 2021-07-10
 
 
 Experimental Jupyter Notebook CSS
--- a/css/epub.css
+++ b/css/epub.css
@@ -1,4 +1,22 @@
 
+.pretext-content.epub img {
+    display: block;
+}
+.pretext-content.epub .solutions {
+    margin-top: 1em;
+}
+
+.pretext-content.epub .solutions .solution .type {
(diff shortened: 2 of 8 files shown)


rbeezer:
Looks real good with a visual scan.  Thanks for breaking it up so carefully.  I'll see about testing and merging later today, perhaps.

rbeezer:
(Merged yesterday, but forgot to comment.)

Real good.  Sampler with SVG has no validation errors, except for items internal to `sageplot-sentence-multigraph.svg`, which is a `matplotlib` creation, I think.

Validation of Kindle version complains about PNG images.  Maybe not important.
```
EPUBCheck v4.2.2
Validating using EPUB version 3.2 rules.
FATAL(RSC-016): epub-sampler.epub/EPUB/xhtml/external/complete-graph.png(1,1):
Fatal Error while parsing file: Invalid byte 1 of 1-byte UTF-8 sequence.
ERROR(RSC-005): epub-sampler.epub/EPUB/xhtml/external/complete-graph.png(-1,-1):
Error while parsing file: Invalid byte 1 of 1-byte UTF-8 sequence.
```

*  `test="local-name(.) = 'frontmatter'"`  works better with an axis, `test="self::frontmatter"`.  (I changed it.)
*  Caught a `Cotents` (or similar).
*  Otherwise, I rubbed out a bunch of tabs and I think I prefer lines beyond 80 (or so) characters, when they aren't really long, and don't need a break in an awkward place.  All these on their own commit.

PR #1533, opened 2021-07-12
EPUB: Add @id to span containing MathJax output 

mitchkeller:
With this three-line addition, _Applied Combinatorics_ passes `epubcheck`. My EPUB aggregator distributor would not distribute the book to certain channels unless it passed `epubcheck`. Added something to the sampler for testing purposes as well. In Apple Books and Calibre, the `a` does properly jump back to an earlier page with the `@id` on the `span` containing math.


rbeezer:
Tabs silently edited out.  ;-)

Thanks!  Great to know that the validator is at maximum happieness with complete (**real**) books!

PR #1534, opened 2021-07-12
Guide: Basics Reference update, EPUB/Kindle instructions

mitchkeller:
I'm thinking that once these instructions are published, it would be good to do a bit of a soft launch to see what steps people can't follow or what glitches have been missed because we have only really tested against two significant projects. Then `-announce` could get a real blurb.

My Kindle version of AC is live. I know that @oscarlevin has DMOI live on the Kindle store (and exclusive there for 90 days). I am waiting on my ebook aggregator to get distribution to Kobo, Scribd, Tolino, OverDrive (How cool is it that libraries can get our ebooks now!?), bibliotheca, Baker & Taylor, Vivlio, and Borrow Box. Apple is taking forever to verify my SSN. Google Play wasn't on the aggregator I chose, so that one is pending now as well.


rbeezer:
Excellent!  Thanks.  Some light editing on the EPUB instructions.

* No tabs!
* `url/@visual`, not `url/@visible`.  I can **never** remember which.  It was a poor name choice, apparently.  I strip the protocol from the visual version.
*  `cd/cline` works well for overly long command-lines, including that you can use spaces to show indentation of subsequent lines.

Yes, a soft-launch on `pretext-dev` might be a good idea.  Pretty sure WW server-generated images are not yet playing nice, so warn about that.  And I should perhaps document the directory-management piece (but maybe do not wait for me).

mitchkeller:
Emacs slips in tabs stealthily for attribute continuation, and I have to run an extra step to remove them…I wasn't sure how heavily impacted the Guide was by the house style instructions anyway. Sorry about the ongoing hassle there, but I do swear that I removed way more than I left behind in the XSL that was in the prior PR.

rbeezer:
I guess I wasn't clear, maybe.  That was an exuberant statement that I did not find **any** tabs, not another exhortation.  ;-)  That is why it was the first (most important) bullet.

PR #1535, opened 2021-07-13
Added Catalan localization

saludes:


rbeezer:
Thanks very much, @saludes!  Great to have Catalan!

Some light editing to a couple associated files and then combined all the commits.  So after you pull to update with the new commits, you'll want to delete your branch.  If you need them, details at:

Cleaning Up 
https://pretextbook.org/doc/guide/html/section-215.html#paragraphs-95

PR #1536, opened 2021-07-15
Script: verify existence of command-line-supplied config file

jjrsylvestre:
I included "from the command line" in the file-does-not-exist warning message for `file_type == 'config'` (even though files for both `file_type == 'source'` and `file_type == 'publisher'`  also come from the command line but don't include that extra bit in their logging messages) because the most likely name for a user-supplied config file is `pretext.cfg`. So a warning message that just says "executables configuration file pretext.cfg does not exist" would be ambiguous without the "from the command line" part because there are two other pretext.cfg files it could be talking about: pretext/pretext.cfg and user/pretext.cfg from the mathbook path.

### DIFF
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -34,7 +34,7 @@ import pretext as ptx
 
 def verify_input_file(f, file_type):
     """Verify file exists, or raise error.  Return absolute path"""
-    # file_type is 'source' or 'publisher'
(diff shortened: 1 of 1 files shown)


rbeezer:
Looks beautiful.  Thanks.  Off-duty much of tomorrow, but will get this in soon.

rbeezer:
In.  Thanks very much for the follow-up!

PR #1541, opened 2021-07-18
Exercisegroup: within a worksheet page, handle numbering specially

Alex-Jordan:
There is a special template for the number of an exercisegroup, but it fails right now when the eg is inside a worksheet page. This makes it look one level above the page to get the right structure number.

Two commits: first one adds an offending example to the sample article. Next one fixes the issue.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -11219,6 +11219,44 @@ the xsltproc executable.
                     </sidebyside>
                 </page>
             </worksheet>
+
+            <worksheet>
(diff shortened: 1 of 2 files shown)


rbeezer:
An `exercisegroup` is only allowed inside an `exercises` division.  A `worksheet` is a specialized division.  A `sidebyside` in a `worksheet` can be used to arrange `exercise`, and especially a `sbsgroup` could be used to mimic the multi-column look of `exercisegroup/@cols`.  So I'm going to close this one.

PR #1542, opened 2021-07-19
EPUB: offer a simple cover image for testing

Alex-Jordan:
This puts a usable cover image in the repo for an EPUB build. I think this would be helpful for someone testing out EPUB but not ready to try to make a cover image yet.

The doc file has some tab replacements. The only meaningful diff is at line 20.


rbeezer:
Thanks.  Love the cover image.

What do you think of providing this image **automatically** when an author has not yet provided a `epub/@cover` in their publisher file?

It'd save us providing a better error-recovery handler when it is missing, **and** be more consistent with providing sensible defaults for the best out-of-the-box experience.

Alex-Jordan:
Can do. 

I wrestled with *where* to store this image file in the repo. Any suggestions?

And maybe there should be a greater discussion if the open licensing of the project as a whole applies to the PreTeXt logo.

rbeezer:
> I wrestled with where to store this image file in the repo. Any suggestions?

It is not XSL, but it does **support** an XSLT.  `xsl/support`

It'd be asy to locate in  `pretext/pretex` since the leading part of the path is known there already.  I embedded the "big red play" button image into the XSL itself, but I don't think that is necessary or advisable.  It could also go in `pretext`?  Either way, we will want the CLI to pick it up, since the `epub()` routine will be used by them.

> And maybe there should be a greater discussion if the open licensing of the project as a whole applies to the PreTeXt logo.

Oh, I really don't have the bandwidth for that right now.........  :-(

rbeezer:
I'm likely to merge this as-is, but we may want to relocate, as discussed in Friday's Drop-In.  I put the "big red play button" into source, so authors did not need to manage some extraneous file as part of an upload to a web server for HTML.  For EPUB, authors **must** use the managed directories scheme, so we can use Python to locate/copy this stock cover.  So I don't see any upside to making an encoded version to house in a stylesheet.

Some other commit/PR can make this the default cover image when the publisher file has no `epub/@cover`?

Alex-Jordan:
I am planning to do things along the lines of Friday's discussion. Just might not get to it until next week. Was planning to force push over this, but I could also close?

I confirmed that svg is fine for an EPUB cover image, and I think I would like to try generating the PTX logo in SVG form.

rbeezer:
Sounds good.  I'm going to close this, just because I'm up to my eyeballs tracking the various PRs in various states.  Your plan sounds good.  Just link back to this discussion here as a continuation when you open a new PR?  Thanks.

mitchkeller:
@Alex-Jordan before putting in an SVG cover, please test that Kindle Previewer accepts it. I'm not finding any documentation that says that SVG covers are not allowed, and the Kindle Publishing Guidelines link to the EPUB spec when talking about covers, but we know that Amazon isn't super keen on SVGs in Kindle books.

PR #1546, opened 2021-07-21
EPUB: Bug fixes for outcomes and conclusion 

mitchkeller:
* Add `outcomes` and `conclusion` to the things that have default titles, which means we don't have to check for `title` before writing the `head/title` into the EPUB XHTML.
* Refine the process of identifying `outcomes` that need to be processed separately for the manifest.
* Add things to the sampler so that we can monitor for this. There's probably a couple more configurations that need testing, but I now can successfully see all the content in the EPUB that is generated.
* Sampler had to be run through `xmllint` because the `latex-image` was horribly confusing my editor's indentation system. This results in a changelog that looks bigger than it is. (Part of this was needing to add a `section` to a `chapter` in order to do more testing.)

### DIFF
--- a/examples/epub/epub-sampler.xml
+++ b/examples/epub/epub-sampler.xml
@@ -43,30 +43,27 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
             <title>One</title>
 
             <p>The quick brown fox jumps over the lazy dog.</p>
-
             <p>Some test cross-references:<ul>
-                <li>To the next chapter: <xref ref="nonsense-chapter"/></li>
-                <li>To a figure much later: <xref ref="figure-cubic-polynomial"/></li>
-            </ul></p>
-
+	        <li>To the next chapter: <xref ref="nonsense-chapter"/></li>
+		<li>To a figure much later: <xref ref="figure-cubic-polynomial"/></li>
+	    </ul></p>
         </chapter>
-
         <chapter xml:id="nonsense-chapter">
             <title>Two</title>
-            <introduction><p>This is the introduction to the chapter.</p></introduction>
+            <introduction>
+                <p>This is the introduction to the chapter.</p>
+            </introduction>
             <section>
                 <title>A silly section.</title>
                 <p>This section just exists so that we can add <tag>introduction</tag> and <tag>conclusion</tag> tags to this chapter.
                 </p>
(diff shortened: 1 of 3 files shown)


kcrisman:
Nice.  Will `outcomes[preceding-sibling::section]` catch the case where the preceding sibling is a conclusion?  (I assume so since that's what you added, just confused by the syntax, perhaps a conclusion is a section in that context.)

mitchkeller:
preceding-sibling:: is an axis. It doesn’t mean *immediately* preceding sibling. It considers all the children of the current node’s parent earlier in document order. 

-- 
Mitch Keller

kcrisman:
>
> preceding-sibling:: is an axis. It doesn’t mean *immediately* preceding
> sibling. It considers all the children of the current node’s parent earlier
> in document order.


Thanks!  (Useful for future reference too.)

rbeezer:
Repackaged.

* Bug fix.  That was a bad cut-and-paste job!  Thanks for the catch.

* Massive whitespace commits really mess with the history, so I'd prefer to avoid them.  So I split it out from real content changes, and then added 67 tabs that needed replacement.

* Giving `outcomes` and `conclusion` the "default title" property would have changed everybody's output, for all conversions, to have change from no-title (if authored that way) to having titles like `Conclusion`.  I took that out and made a more targeted change that will at least satisfy the validator.  Who sees `head/title`?  Not the reader?

Thanks for continuing to refine this.  Making progress!

mitchkeller:
> * Bug fix.  That was a bad cut-and-paste job!  Thanks for the catch.

I’ve made my share of them while working on this stuff. There’s a silly number of things that just need to be done a couple of times.

> * Massive whitespace commits really mess with the history, so I'd prefer to avoid them.  So I split it out from real content changes, and then added 67 tabs that needed replacement.

Yeah, I wasn’t sure the best way to do that one, and I swear that I untabified before committing, but maybe not. Hopefully we are nearing the end of any major white space changes on the sampler, as we are approaching having most of the structures in there we need to robustly test things.

> * Giving `outcomes` and `conclusion` the "default title" property would have changed everybody's output, for all conversions, to have change from no-title (if authored that way) to having titles like `Conclusion`.  I took that out and made a more targeted change that will at least satisfy the validator.  Who sees `head/title`?  Not the reader?

As far as I am aware, `head/title` is not revealed by any reading system. See my comments elsewhere about writing something more generic into `head/title`, as a number of the things we allow inside a PTX `title` are forbidden in `head/title` and so authors might need to have `shorttitle` in a lot of places just for EPUB validation (possibly with adverse consequences for other conversions).

PR #1547, opened 2021-07-22
pass through ol label from slides to revealjs html

StevenClontz:
This is an untested hotfix to preserve list labeling in RevealJS builds - I suspect it's replicating logic from the rest of pretext that converts PTX `@label` to HTML `@type`, but I wasn't sure if Andrew wrote things this way intentionally when implementing lists.

### DIFF
--- a/xsl/pretext-revealjs.xsl
+++ b/xsl/pretext-revealjs.xsl
@@ -342,6 +342,9 @@ dfn {
 
 <xsl:template match="ol">
   <ol>
+    <xsl:if test="@label='a'">
+      <xsl:attribute name="type">a</xsl:attribute>
+    </xsl:if>
     <xsl:apply-templates/>
   </ol>
 </xsl:template>


StevenClontz:
Yeah this seems like not a great strategy: it'd also be good to preserve `@cols` as well:

image

image

rbeezer:
There's a lot more specialized stuff going on in `pretext-html.xsl` for lists.  A more obscure example is a numbered list that starts counting at zero.  Search `pretext-html` for `<xsl:template match="ol|ul">` to get started seeing how much is happening.

I think the first draft of this conversion just brute-forced some constructions, rather than consulting more elaborate "real" templates.  (And this has made the LaTeX conversion un-maintainable.)

What happens if you just let the `-html` version bleed through?  In other words, **remove** the `<xsl:template match="ol">` template?  Entirely possible that the CSS will be conflicting, missing, or insufficient but that should be fixable.

StevenClontz:
> What happens if you just let the `-html` version bleed through? In other words, **remove** the `<xsl:template match="ol">` template? Entirely possible that the CSS will be conflicting, missing, or insufficient but that should be fixable.

Yeah, this seemed plausible as well. Let me investigate...

PR #1549, opened 2021-07-22
EPUB: Resolve svg/mathml property in manifest once and for all, pick up missing frontmatter bits

mitchkeller:
As a bit of explanation, when I started trying to deal with @kcrisman's `list-of` to pass validation, I quickly realized that this was going to rapidly spiral out of control for the manifest, especially with `list-of/@empty='no'` set in PTX source. So I changed the order that the templates are applied in the entry template and passed in the temp directory as a parameter. This allows the template that builds the manifest to actually inspect the files to see if they contain an `svg` or `math` element and thus need the appropriate property set. The one place that this is not done is for the table of contents file, which gets the property set based on whether or not there is a `title/m` in a `chapter` or `appendix`.

This PR also ensures that everything that can be specified in `frontmatter` gets picked up and sets the right `@media-type` for the cover image. (Previously had assumed PNG.)

### DIFF
--- a/examples/epub/epub-sampler.xml
+++ b/examples/epub/epub-sampler.xml
@@ -37,6 +37,33 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                     <personname>Rob Beezer</personname>
                 </author>
             </titlepage>
+	    <colophon>
+		<website>
+		    <name>PreTeXt</name>
+		    <address>https://www.pretextbook.org</address>
+		</website>
+		<copyright>
+		    <year>2016<ndash />2021</year>
+		    <holder>Robert A. Beezer</holder>
+		</copyright>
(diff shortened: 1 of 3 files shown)


mitchkeller:
Sorry this wound up with merge conflicts after my other PR got merged. I had two branches going for those things and didn't handle combining them for testing quite optimally. Looks like the code from my branch is what is preferred in all the conflict spots…except there might be some tabs in there still :-(

rbeezer:
Merge conflicts resolved, so also busted out code v. test document.  Thanks, I like the approach to determining the `mathml` v. `svg` properties - hadn't thought about doing that.

About to quit for dinner, and forgot to annotate with the PR number.  Sorry.  Four commits, ending at bbd38c32d39802bfa075308ec3e0229f91825679

PR #1551, opened 2021-07-22
Shorten YouTube URLs

sean-fitzpatrick:
I am somewhat expecting this to be closed without merging. (More of an FYI pull request? Anyway, something I promised to do back in September.)
I wanted shorter YouTube URLs so I did it on a branch.

Note that there is no equivalent shortened URL for playlists, so there will still be an issue with bad boxes for these (which tend to be longer as well -- you're not helping, playlists).

I removed the `YouTube:` prefix since I think this is implied by the URL and horizontal space is already at a premium.


sean-fitzpatrick:
OK, here is the part that has me stumped: I can't find my way through the XSL tree, I guess:

This is the code that generates the static URL under the QR code:
```
<xsl:text>\begin{tcolorbox}[captionstyle]%&#xa;</xsl:text>
        <xsl:text>\small </xsl:text>
        <xsl:value-of select="$the-caption" />
<xsl:text>\end{tcolorbox}%&#xa;</xsl:text>
```
The output is
```
\begin{tcolorbox}[captionstyle]%
  \small \mono{https://youtube.com/watch\_videos?video\_ids=pdu3nRE8x1I,zA\_u1Hsd0Jg}
\end{tcolorbox}
```

I'd like to be able to break that `\mono` across two lines in some places.
But I might be alone in this, and I have only a few playlists, so I can edit the LaTeX by hand after the fact.

PR #1552, opened 2021-07-22
Add links in HTML for full-size Asymptote

sean-fitzpatrick:
Sometimes an Asymptote image gets clipped a bit by its bounding box when rotating it, and it would be nice to see the full size image in a new window.
This adds a link below each Asymptote image to enable this.

Note: does not distinguish between 2D and 3D Asymptote, and probably not needed for the former.
Can be turned off using the same publisher variable used to enable links for PDF.

(Q: in that case, should `pdf/b-asymptote-links` variable become `common/b-asymptote-links`?


sean-fitzpatrick:
Sorry, wrong branch!

PR #1554, opened 2021-07-23
pass through revealjs list management to pretext-html, add css to match in revealjs

StevenClontz:
Replaces #1547, and this one actually was shown to work for my particular usecase:

image

I'm not super happy with the copypasta CSS from https://github.com/PreTeXtBook/CSS_core/blob/main/pretext_add_on.css but it's a start to get list markers working as expected. I also added `.pretext-content` as a class alongside `.reveal` anticipating that maybe we can just use core CSS directly down the road.

It's been a minute since I thought about inline lists, but `float:left;padding-right:2em` seemed to be the slickest solution. I didn't attempt to preserve choices of columns and just made any column-ed list be an inline list. Maybe RevealJs has a proper grid layout system that should be used instead.

### DIFF
--- a/examples/sample-slideshow/sample-slideshow.xml
+++ b/examples/sample-slideshow/sample-slideshow.xml
@@ -73,8 +73,24 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 
 
             <slide>
-                <title>Ordered Lists, With Pauses</title>
+                <title>Ordered Lists, With Features</title>
 
+                <p>These are enumerated with capital letters.</p>
+
+                <p><ol label="A">
+                    <li>Two conversions: print-on-demand, electronic <init>PDF</init></li>
+                    <li>Extensive use of the <c>tcolorbox</c> package (CSS-like)</li>
+                    <li>Evolving styling/themes (Andrew Rechnitzer, David Farmer)</li>
+                </ol></p>
+
+                <p>These are inline</p>
+
+                <p><ol label="i" cols="2">
+                    <li>print-on-demand</li>
+                    <li>electronic <init>PDF</init></li>
+                </ol></p>
+
+                <p>And these have pauses.</p>
                 <p><ol pause="yes">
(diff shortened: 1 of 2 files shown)


StevenClontz:
(Maybe https://css-tricks.com/snippets/css/a-guide-to-flexbox/#flex-wrap is the better solution for inline lists.)

rbeezer:
Looks good on a visual scan.  I think of an "inline" list as (a) stuff without a break, (b) all on one line, (c) which may wrap around like a sentence, (d) and so on.  You have a "two-column" list, no?  There've been requests for "inline" lists, so I'd rather not look like we really have them.  OK if I make your two-column list have another row or two?

Yes, the "copypasta" has to stop.  Incurring significant debt.  But not yet, this is fine.  We've got CSS figured out for EPUB, and once that settles, we can migrate that scheme over here.

StevenClontz:
> OK if I make your two-column list have another row or two?

Yeah, I'd prefer if `.col2` actually was two columns and `.col3` was three, but my quick hack to get something comparable (not knowing much about PTX CSS) was to let `.colN` be an inline list.

rbeezer:
Minor edits for the history.  Great!  I'm really glad we can just pass these through.  Thanks for getting that worked out.

Not happy about the technical debt.  Never happy about technical debt!  ;-)  Let's get CSS sorted rather than continuing to cut corners in that department.

Tagging @davidfarmer since maybe we can mimic the CSS setup for EPUB and preserve the one-off **improvements** here in a file like `revealjs.css`?

davidfarmer:
I created (in the usual place) the file
   revealjs.css
with the css currently in pretext-revealjs.xsl

So, you can switch to maintaining the css there instead of
in the xsl, if you wish.

rbeezer:
Excellent!  Thanks.  Today is a mess, but I'll try to get to this real soon.  I think I want to do a network-version and a standalone version, als EPUB, so it'll take a little bit of work.

@StevenClontz we should experiment and see if "standard" CSS will "do the right thing."

PR #1555, opened 2021-07-23
EPUB: Improvements to IDs on math for xref

mitchkeller:
You pointed me in all the right directions, and I have checked both validation reports and behavior when following links. Everything seems to check out here.

### DIFF
--- a/examples/epub/epub-sampler.xml
+++ b/examples/epub/epub-sampler.xml
@@ -45,7 +45,8 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
             <p>The quick brown fox jumps over the lazy dog.</p>
             <p>Some test cross-references:<ul>
             <li>To the next chapter: <xref ref="nonsense-chapter"/></li>
-        <li>To a figure much later: <xref ref="figure-cubic-polynomial"/></li>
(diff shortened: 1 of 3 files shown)


rbeezer:
Looks real good.  Thanks.  I like the match on just `mrow`, which will limit any side-effects.

I might fine-tune two things:

* `<xsl:if test="self::men|self::mrow">` is always true.
* `$context = 'me'` is never true, since you can't have an `me` as a target of an `xref` since it does not have a number.

mitchkeller:
* I wasn't sure how many liberties I should take in deviating from the template in -html. 
* That's what I was thinking, but your email on -dev said "the math in EPUB output should have an "html-id" for me|men|md|mdn", so I went along with that. 

Can definitely clean these up and push another commit if you'd like.

rbeezer:
OK, all done (for now!).  Good work.  I put my changes (as discussed) into new commits so I could take the blame.  ;-) See 41d959103919724ae90a2ba258e00ec20adf8c95 and 086eb1e5f946faa16fcd882dcf3c01ca9657da69.

Split out the changes to the sampler, so I could do some careful testing of your work, and of my changes (which looked real good.)

PR #1556, opened 2021-07-23
Slide image scaling

StevenClontz:
Extends #1554, sorry.

- Centers `.figure-like`
- allows `@href` for images (since I think I recall full URLs are intended to be in `@href` not `@source`)
- Caps image height to 50% of the vertical viewport (I think this in the spirit of capping horizontal width of images in HTML pretext builds, but eventually should be configurable)
- Links images to image URL for a quick "zoom" feature

### DIFF
--- a/examples/sample-slideshow/sample-slideshow.xml
+++ b/examples/sample-slideshow/sample-slideshow.xml
@@ -73,8 +73,24 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 
 
             <slide>
-                <title>Ordered Lists, With Pauses</title>
+                <title>Ordered Lists, With Features</title>
 
+                <p>These are enumerated with capital letters.</p>
+
+                <p><ol label="A">
+                    <li>Two conversions: print-on-demand, electronic <init>PDF</init></li>
+                    <li>Extensive use of the <c>tcolorbox</c> package (CSS-like)</li>
+                    <li>Evolving styling/themes (Andrew Rechnitzer, David Farmer)</li>
+                </ol></p>
+
+                <p>These are inline</p>
+
+                <p><ol label="i" cols="2">
+                    <li>print-on-demand</li>
+                    <li>electronic <init>PDF</init></li>
+                </ol></p>
+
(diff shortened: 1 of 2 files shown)


kcrisman:
Just a comment: I used revealjs quite heavily last spring, and some images were (nearly) the entire slide, so I'm wondering whether we can revisit that later if testing shows there are situations where it might not work well.

(That is to say, don't let this comment stop you from doing anything to improve revealjs, obviously, but I'd hope you might entertain issues or PRs related to it later if it turns out it messes up some slide decks.)

StevenClontz:
> Just a comment: I used revealjs quite heavily last spring, and some images were (nearly) the entire slide, so I'm wondering whether we can revisit that later if testing shows there are situations where it might not work well.

Yeah it occurred to me that maybe this isn't ideal for anyone doing full-slide images, except that even now if you tried to (say) put a 1600x1200 image in a slide and view it on a 1024x800 screen, it'd look bad without scaling the browser window. I can be convinced that maybe that 50% should be increased, but the limit should be less than 100% I think. Ultimately this should be configurable.

kcrisman:
>
>  Ultimately this should be configurable.
>
> I suppose by hacking the source, it is :smiley: but at any rate, probably
a variety of image types should be tested.  I agree that something needs to
be done to avoid really bad sources.  Unfortunately a lot of source images
may be non-vector graphics.

StevenClontz:
Slides need to support different screen sizes so this can never be solved
by requiring authors to provide source images of the "right" size.

rbeezer:
Not ignoring this one.  Just plowing through numerous PRs, which all require lots of time for various reasons.

`image/@href` would be the right thing to do.  But that will be new, we don't have that anywhere else.  How's it going to go in the Beamer conversion?  And suppose we make LaTeX templates for `slide`?  I could list pros and cons.

Would you open a discussion on `pretext-dev` and I'll hop in there?

StevenClontz:
<https://groups.google.com/g/pretext-dev/c/Ra8fa2h1N24>

PR #1557, opened 2021-07-23
Add "click to enlarge" links for asymptote figures in HTML

sean-fitzpatrick:
Here's another crack at this.

This time I also made a separate publisher variable so html links are controlled separately from pdf.

### DIFF
--- a/doc/guide/publisher/publisher-file.xml
+++ b/doc/guide/publisher/publisher-file.xml
@@ -111,20 +111,13 @@
 
         <subsection xml:id="latex-asymptote-link-options">
             <title><latex/> Asymptote Links</title>
-            <idx><h>Asymptote links</h><h>LaTeX, PDF, HTML</h></idx>
+            <idx><h>Asymptote links</h><h>LaTeX, PDF</h></idx>
             <idx><h>LaTeX</h><h>Asymptote links</h></idx>
             <idx><h>PDF</h><h>Asymptote links</h></idx>
-            <idx><h>HTML</h><h>Asymptote links</h></idx>
 
-            <p>When the <c>html</c> output format is selected for Asymptote graphics, each diagram will generate an <init>HTML</init> file as part of the online version of your book. In some cases you may wish to link to these files.</p>
-            <p>The conversion to <init>HTML</init> can provide a <q>Click to enlarge</q> link below each graphic that allows the reader to view a full-size version of your diagram. This is especially useful for 3D graphics, since they are interactive.</p>
-            <p>The conversion to <latex/> can also provide a link to an <init>HTML</init> version of each Asymptote graphic, allowing readers of the <init>PDF</init> version of your book to access the interactive version of the diagram.</p>
-            <p>The<cd>
-                <cline>/publication/html/asymptote/@links</cline>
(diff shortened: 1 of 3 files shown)


rbeezer:
Very nice!  I've got this all re-organized and ready to go.  But it needs one thing, so might as well need two things.

* Documentation: no publisher variable ever gets in without Grade-A documentation!  Add a vague description of the general capability to the chapter on HTML conversion.  Link to the precise, careful description in the publisher reference, and do this exactly and compactly.  Study other examples and follow house style for consistency and uniformity.  Link back to general description.

* `@style` should be avoided, so we can accomodate styles fully.  Ask @davidfarmer for a class name that suits your purpose, and he'll get it centered and maybe mess with the font.  I can make a "beta" for him to experiment with, or we just have text flush-left for a little while.

1.  Post the class name here and I will add it in.  Do not make any more commits on this branch, I've messed with it already locally.
2.  Make a new PR with the documentation.  I'll merge both PRs at the same time.

sean-fitzpatrick:
Ah, forgot the documentation! Of course!
Might be a bit slow on it though: we have company over the weekend and I've got a kid with me during the day next week.
Admittedly I didn't know how to centre a div in PreTeXt, so I searched `pretext-html.xsl` for examples, and found one with `style="center"` :-)

rbeezer:
Do as I say, not as I do!  As a parent, you know that!  ;-)

Use `git blame` to see how old that is.  You can find any amount of bad code....

sean-fitzpatrick:
Houseguests slept in so I snuck in 10 minutes to update the publisher file documentation. I can update that `div` once @davidfarmer educates me on the correct attribute to use.

PS: while searching for appearances of Asymptote in the documentation, I noticed the mention of using `\newcommand{\foo}{\operatorname{foo}}` instead of `\DeclareMathOperator{\foo}`. I *think* this was because the latter requires loading the `amsmath` package in each Asymptote figure, which we now do anyway. But I recall some recent conversation suggesting that the latter should be preferred anyway, so that all LaTeX macros have the same format.

rbeezer:
Courteous of your house guests!  ;-)

> Make a new PR with the documentation

Please?  I can handle this, but...

A feature of HTML should no be buried in the LaTeX section where it canot be found.  Pleasse make complete documentation in the place where the HTML conversion and options are described.  You can cross-reference the general lateX to the general HTML, and vice-versa.

sean-fitzpatrick:
I've reverted the documentation commit, and have that coming (in the correct place this time) from another branch on a separate pull request.
Also, I noticed that the autocomplete suggestions in Atom included "Asympote" which uncovered two typos. I've committed that change here, since it effects `pretext-html.xsl`, which is already modified by this PR.

sean-fitzpatrick:
I changed the text of the link. Realized "click to view full size" is probably not accessible HTML. Now says "link to full size".

rbeezer:
>  I've got this all re-organized and ready to go

>  Do not make any more commits on this branch, I've messed with it already locally.

Which means substantial work has been done on your first two commits here, and to incorporate anything more, I'd need to start over.  (Which you can guess will not happen.)  That's why I requested a new PR -- additions, reversions, (and any force-pushes) can't be easily assimilated.

I'll see if I can incorporate smaller changes manually on #1560.

PR #1558, opened 2021-07-23
Common: utility template to print warning messages one line at a time

Alex-Jordan:
Before this, when a warning message is issued and it has line break characters, you get like this:

```
    *  ********************************************************************************
    *  EPUB conversion is experimental and not supported.  In particular,
creating an EPUB requires the pretext/pretext script.
    *  ********************************************************************************
```

With this in place, you get:

```
    *  ********************************************************************************
    *  EPUB conversion is experimental and not supported.  In particular,
    *  creating an EPUB requires the pretext/pretext script.
    *  ********************************************************************************
```

I only applied it to this EPUB message. Not sure if warnings with line breaks happen in other places too.


rbeezer:
The leading single dot is coming from the `pretext\pretext` script as it reports messages from the `lxml` library employing a stylesheet.  Unclear why the embedded newlines are not respected?  Anyway, thanks!

PR #1559, opened 2021-07-23
EPUB: external and generated

Alex-Jordan:
This makes it OK for EPUB production to have the publisher file's `@external` and `@generated` be something custom. This is now the case with the epub-sampler, and you may run `~/mathbook/pretext/pretext -c all -f epub-svg -p publication.xml -d . epub-sampler.xml` to see the results.

Also, I am able to make HTML for the epub sampler using `~/mathbook/pretext/pretext -c all -f html -p publication.xml -d . epub-sampler.xml`and it seems to work.

I am _not_ able to build pdf for the epub sampler using `~/mathbook/pretext/pretext -c all -f pdf -p publication.xml -d . epub-sampler.xml`. I believe that the build routines in `pretext/pretext` for PDF (and also generated images and HTML?) need to be examined, following these commits here. In particular, there is this utility `def get_image_directories` that I do not understand, but is perhaps designed assuming that the input image folders match the output image folders, which is not the case. EPUB production does not use this utility.  But image, HTML, and pdf production do.


rbeezer:
PDF fails for `epub-sampler.xml` because there is no PDF image available for one of the "extensionless" `image/@source`.  If the right file gets added, I'm sure it'll build.  That example is not built to necessarily be used for any other format, but that is a good test.  HTML looks fine to me.  But I'll look closer.

rbeezer:
`get_image_directories` is necessary to get a "complete" path to the three source directories.  It returns six items, three are no longer necessary since we are hard-coding destinations.  So there is a bit of easy code-cleanup to do after all this.   I think this is ready to go after a bit more examination.

rbeezer:
Some comments:

1.  `$b-managed-directories` is not a boolean as constructed.  Because `<xsl:value-of select="true()"/>` is text/string.  You can only make a boolean variable with the `select` on a `xsl:variable`.  So that big variable should be renamed, and then the boolean variable can be forned based on its value being `"true"` or `"false"`.  This one has bit me many times.  I think `$b-managed-directories` would be true aleways, outside of the fatal error situation (small test says yes).

2. `$external-directory` will never be used outside of managed directories, so when set to the variable for the source, it'll never get used.  (`$generated-directory` needs care because we already had a scheme of sorts for this.  I'm not going to error-check it (trailing slash), since it never was checked, and it is for the non-managed situation, so I don't want to change behavior.)

3.  I'm going to change the parameterization on `$read` to not rely on the default.  For an internal use like this, I'd rather not rely on a default - just clearer code, I think.

I'll make these changes, but may sleep on it.  Thanks!

rbeezer:
Thanks, Alex, for getting this going in the right direction!   As you can see, I split this out into four commits and kept your name on them.  I made some interspersed changes for both effect and style, under my name, so I can take the blame.  Then comprehensive testing, followed by 5 pages of documentation.  Which is why I've been sitting on my branch for this.

I suspect WW sample-chapter as PDF is broken.  I saw it scroll by as I rebuilt the website.  Catching up on a few other things and will get to it soon.  Don't think about it.

PR #1560, opened 2021-07-26
Add documentation for HTML Asymptote link option

sean-fitzpatrick:
This time with documentation in the right place, and in two places: Conversion to online HTML and Publisher file reference.


rbeezer:
Continuing from #1557.

rbeezer:
Thanks, Sean, for sticking with this one.  No malice in "suggestions" for better workflow.  ;-)

Works very nice once you remember to elect it in your publisher file.  ;-)  I've made it available in the sample article so maybe folks will see it.  Code pushed, sample article will go live shortly.

Rob

sean-fitzpatrick:
Thanks Rob. Sorry for missing the bit about not adding. Need to do fewer things at the same time. Did we hear from @davidfarmer on the styling for the div?

rbeezer:
Up next.

rbeezer:
Lost an edit.  Pushed, but not live in sample article.

PR #1561, opened 2021-07-26
EPUB: Simplify what is written into the XHTML head/title

mitchkeller:
This deals with the issue of authors putting things into a PTX title that are fine inside an XHTML `hn` but not allowed inside an XHTML `head/title`. Since the `head/title` is not reveled by any reading systems, we simply put in a PTX-generated identifying string that we know won't contain any other XHTML tags. (For example, this now renders something like "Section 3.2.1" or "Conclusion".)


rbeezer:
Very good.  Thanks.  Almost caught up with PRs....

PR #1565, opened 2021-07-30
Add experimental stringparam `debug.skip-knowls` to skip building knowls

StevenClontz:
Intended to speed up quick preview builds in HTML.

See discussion in https://groups.google.com/g/pretext-dev/c/JhcdoGU3nWk/m/8oeJDRMGCQAJ


rbeezer:
Thanks!  Ignore my comments about testing empty strings, I was looking at the wrong place.  I'm pretty frazzled by the end of Drop-In.

Still making a few knowls, but that is not the point.  ;-)

PR #1566, opened 2021-08-05
Add CLI documentation to guide

oscarlevin:
This is a first draft, with much work still to be done.  I realized though that some feedback about whether I'm heading in the right direction would be useful before I get too far.  Happy to add some more commits to this in the next couple of days.


rbeezer:
Looking good!  Certainly headed in the right direction. 

Some nits discovered along the way:

*  Capitalize XML, more than once.
*  Typos: "HMLT", "thier", "Instillation"
*  <idx>console</idx>: put it after the period (long story, LaTeX puts period on newline).
*  "Any text editor.": link to appendix on these?
*  "which will create a new folder at your current location": idented in LateX as a new paragraph.  Maybe say:  "This will...", OR use "cd" **within** a paragraph.
*  "unless you change the code for that image": did I miss what an image is here?  Maybe "target"?
*  "Let’s back up create":  add "to"?
*  "For a quick introduction to what is available...": sentence needs a period.
*  Ideas on where you think the legacy example will slot in?
*  A few blank lines filled with 8 or so space characters.

oscarlevin:
Thanks for the careful reading and feedback.  I'll fix those as I add more stuff.

What about putting all the xsltproc info in the last section of chapter 5 (Processing etc)?  That currently needs info about how to get pretext (from github) and maybe some other stuff from the quick minimal example chapter.

rbeezer:
The concole/period comment references the index entry, now with backticks: `<idx>console</idx>`

Yes, Chapter 5 makes sense, as I think maybe you have a forward reference to it already?

I'd like to transition to `xsltproc` being a developer/power-user tool.  I should write something for the developer section explaining its role.

Very glad to see the getting-started documentation getting a serious facelift!

oscarlevin:
Okay, I'm done!  I put documentation about xsltproc in the developer's guide, and also moved the pretext script chapter there.  Most everywhere else, instructions for using xsltproc have been replaced with their pretext-cli counterparts.  

One area the will still need work is WeBWorK.  I want to think more about this, and chat with @Alex-Jordan before ripping up those sections.  In particular, some of the documentation is no longer needed.  Perhaps it makes sense to move some basics to the processing chapter.

Anyway, I'm happy to correct anything you think needs it, but I don't plan on adding any new content in the next week or so.

rbeezer:
Thanks!  Giving a workshop on Wednesday, and still have not decided what tools I plan to use....  That's a job for Monday.

oscarlevin:
I'm biased obviously, but for anyone new, the CLI is really the way to go.  CoCalc comes with 0.4.1 installed, which has everything except the custom XSL piece (and I'm sure they could update if asked).  Since you would need people to use CoCalc or have python installed anyway to use `pretext/pretext`, I don't know what you gain by not using the CLI.

rbeezer:
Thanks, Oscar!  I split off a couple of unrelated items (with your name on them).  Cleaned some stray whitespace from the rest of it.  Otherwise, mostly untouched.  ;-)

If you (with @StevenClontz?) want to fashion a blurb about the new documentation (to encourage collateral uptake of the CLI), send it to me (with a subject line:  [NEW] ....) and I'll get it posted.

StevenClontz:
We should avoid suggesting `pip install pretextbook` and encourage `python -m pip install pretextbook`. The reason is because even if `python --version` returns `3.8.x`, the `pip` command may not use that same Python. Using `python -m pip` ensures that we're using the `pip` that matches the version of python on PATH, and also ensures that `python -m pretext` matches the version of `pretextbook` installed by `pip` (although hopefully the user's system has PATH set up correctly and they can just use `pretext`).

StevenClontz:
Rather than moving `new-pretext-project` to create a second project, perhaps we can show off how `pretext new hello` fails, and then suggest `pretext new hello -d hello-world`.

StevenClontz:
Does it make sense to use `pretext view html --watch` in [Section 2.3: Extending the Minimal Example](https://pretextbook.org/doc/guide/html/quickstart-extend-minimal-example.html#quickstart-extend-minimal-example)? With the caveat that this feature is only feasible for building small [chunks of a] document?

StevenClontz:
BTW thanks for doing this Oscar -- it's looking great, and it's awesome to be in the official guide!

PR #1567, opened 2021-08-07
HTML: merge task headers

Alex-Jordan:
Work on hN revealed what I think is a mistake, fixed here.

See 
Exercise 4.10: A very structured exercise 
https://pretextbook.org/examples/sample-article/html/interesting-corollary.html#exercise-structured

Inspect the HTML and note that task (a) has two h6 elements. One for "(a)" and one for the title. This PR merges them into one.

The same issue is present in the solutions at
https://pretextbook.org/examples/sample-article/html/solutions-backmatter.html

So there is an addition here that adds the title to the right h6. And then two removals, removing the other title where born and when part of a solutions.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -2187,6 +2187,12 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <xsl:apply-templates select="." mode="list-number" />
             <xsl:text>)</xsl:text>
         </span>
+        <xsl:if test="title">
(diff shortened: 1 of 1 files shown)


davidfarmer:
Good.  I have been wanting this.

rbeezer:
Very good - thanks.  If I'd known this was a problem, I'd lost track of it.  Sorry.  (Is there an issue to close?)

Website versions will go live in the next couple of hours, after another PR or two.

Issue #1570, opened 2021-08-08
exercisegroup with no title has no hN


Alex-Jordan:
When an "exercisegroup" does not have a "title", then there is no heading for the exercisegroup. So the introduction to the exercisegroup falls under the header of whatever came before. For example

```
exercise
exercisegroup
   introduction
   exercise
   exercise
```

has like:

h3 (exercise 1)
h3 (exercise 2)
h3 (exercise 3)

Someone using the heading map to get to exercise 2 is not aware that there was an important introduction.

One possible solution is to use a header for the exercisegroup even if it does not have a title. I'm not sure what to put in it though. One thing that comes to mind though is like "Exercise 2–5". Then it would become a question if this should precede the title when present, for consistency.

Alex-Jordan:
It's easy to make the exercisegroup open with a header that says like "2–5."  And easy to make it open with a header that says "Exercise Group 2–5."  But making it open with "Exercises 2–5." would take significant work, because of what would be needed for localization.

Planning to open a PR with just "2–5." to see what you think. It's a small change to the code.

Alex-Jordan:
#1574 addressed this.

PR #1572, opened 2021-08-08
Give all exercisegroup a heading

Alex-Jordan:
In HTML, EPUB, and PDF, this PR makes it so that an exercisegroup will always start with a header like:

2–5. Optional Title

If the "exercisegroup" has no "title" and also has no "introduction", this may look strange. But it already looks strange in that situation for the problems to indent with no explanation. (Perhaps we should consider making the schema require at least one of "title" and "introduction".

This was motivated by #1570. If there is an introduction but no title, the introduction can go unnoticed by a user relying on the heading tree.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -2252,7 +2252,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 </xsl:template>
 
 <!-- Title only -->
-<!-- ASIDE-LIKE, exercisegroup, dl/li -->
+<!-- ASIDE-LIKE, dl/li                -->
 <!-- proof, when optionally titled    -->
 <!-- Subsidiary to paragraphs,        -->
 <!-- and divisions of "exercises"     -->
@@ -3510,7 +3510,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- When born use this heading         -->
 <!-- Never hidden, never gets a heading -->
 <xsl:template match="exercisegroup" mode="heading-birth">
-    <xsl:apply-templates select="." mode="heading-title" />
+    <xsl:apply-templates select="." mode="heading-divisional-exercise-serial" />
 </xsl:template>
 
 <!-- Heading for interior of xref-knowl content -->
--- a/xsl/pretext-latex.xsl
+++ b/xsl/pretext-latex.xsl
@@ -5349,11 +5349,14 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     </xsl:variable>
     <!-- build it -->
     <xsl:text>\par\medskip\noindent%&#xa;</xsl:text>
+    <xsl:text>\textbf{</xsl:text>
+    <xsl:apply-templates select="." mode="serial-number" />
+    <xsl:text>.</xsl:text>
     <xsl:if test="title">
-        <xsl:text>\textbf{</xsl:text>
+        <xsl:text> </xsl:text>
         <xsl:apply-templates select="." mode="title-full" />
-        <xsl:text>}\space\space</xsl:text>
     </xsl:if>
+    <xsl:text>}\space\space</xsl:text>
     <xsl:if test="@xml:id">
         <xsl:apply-templates select="." mode="label"/>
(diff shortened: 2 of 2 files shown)


rbeezer:
I think I would prefer a **default** title that is simply "Exercise Group." which we do have localizations for.  The numbers involved should be made obvious by indentation and other styling, and they are right there.  For an `xref` by-title (discouraged anyway) an author can provide a better title.  But perhaps only if an `introduction` is absent?

I'm going to put the larger question of both `title` and `introduction` onto `-dev` since I think it would be worthwhile to get some broader input.

Alex-Jordan:
A local instructor commenting on a local project feels that the indentation on exercisegroup exercises in print is too subtle for indicating which exercises the instructions refer to. He sent me these pics of Stewart Calculus and Sullivan College Algebra. The former is like what I went with here in this PR.

CE6DA8AB-67BC-4C6C-A1A5-AC94F1435285

45E81439-AA0F-40EF-B23D-4958D9B8D231


He prefers the latter, but that can't be automated. And besides, that runs contrary to the notion of recycling an individual exercise form an exercisegroup like we do with webwork, gluing its introduction to the statement.

rbeezer:
Does an "xref" to an `exercisegroup` with the right `@text` produce "45-48" and "7-22"?  That'd be a way to get semi-automation.  It might even be a "dead" link in a title in PDF.

If something is not obvious enough, then we need to work on styling.

Alex-Jordan:
> Does an "xref" to an exercisegroup with the right @text produce "45-48" and "7-22"?

It produces something like "Exercise Group 45–48".

> That'd be a way to get semi-automation.

Not sure what you mean. This PR automates printing "45–48." as a heading. It could instead do "Exercise Group 45–48." as a heading, but at first that seemed like using up too much space.

Here are some sample EPUB, HTML, and PDF output under this branch that I ran for that local project.


Screen Shot 2021-08-07 at 11 43 44 PM
Screen Shot 2021-08-07 at 11 42 21 PM
Screen Shot 2021-08-07 at 11 39 29 PM

davidfarmer:
"45–48" when it appears.

"Exercise Group 45–48" where you refer to it?

Alex-Jordan:
> "Exercise Group 45–48" where you refer to it?

Right, almost.  That is almost how things work already, like [here](https://pretextbook.org/examples/sample-article/html/section-cross-referencing.html) if you search for "Exercise Group". The difference is that with global numbers, it would be like "Exercise Group 2.7.45–48".

> "45–48" when it appears.

Except with a trailing period. I have a biased preference for this. At first I thought I would dislike leading a paragraph with numbers and a dash like that. But once I saw it in action, I warmed up to it. Then when a colleague sent me that picture of Stewart's book doing this, it confirmed my nascent bias.

But I'd be happy with anything there. This is my present hangup with getting hN sorted out in HTML. I believe it is the last issue for me with the sample article, before I more on to examining book output.

rbeezer:
> Not sure what you mean.

OK, I've now done the experiment myself carefully.  Using `xref/@text="local"` **will** produce something **exactly** of the form "7-22".  So an author has the flexibility to put it into a title or introduction in whatever manner they like.

Alex-Jordan:
OK, I think I understand that you are suggesting that an author, if they choose, can currently open each of their "exercisegroup"s like this:

```
<exercisegroup xml:id="exercisegroup-find-the-derivative">
    <title><xref ref="exercisegroup-find-the-derivative" text="local"/></title>
    <introduction>
        ....
```

And do that for each exercisegroup, including coming up with an xml:id even when the only purpose is to self-reference it in its own title.

But are you opposed to letting people just write:

```
<exercisegroup>
    <introduction>
        ....
```

and automatically get the "7–22."?

rbeezer:
Right.  I really don't like "7-22." as a default title.  "I think "Exercise Group." is fine and we can show authors how to make other constructions if they like.

Alex-Jordan:
So going back to something you said early on, would you like this to be something where there is an effective default title "Exercise Group". But if you declare a title, like `<title>My Title</title>` then you just see "My Title" in output. That is, you don't see something like "Excercise Group: My Title".

rbeezer:
Closed in favor of #1574.

PR #1577, opened 2021-08-13
Adds speech processing for svg enhanced mode.

zorkow:
SVG enhanced option that enriches the SVG output with SRE semantic elements.
As per live session 8/13.


rbeezer:
Thanks, @zorkow!  I'll work on integrating this early next week.

rbeezer:
Well, you know what "early next week" can lead to.  I've merged this, with some notes.

I've had trouble extracting the one enrichment (speech) we need from all the others, plus removing all the ones we don't want.  So I'm going to try to get speech from a separate run, and integrate that.  Once that cruder method it is settled, the output can be used as a check for the real work of integrating this properly.

Thanks very much for putting this together!  And apologies for the delay...

PR #1579, opened 2021-08-16
add symlink info to processing

kcrisman:
For conversion from older projects, mostly


rbeezer:
Excellent!  Thanks, @kcrisman!

PR #1583, opened 2021-08-18
Exercisegroup

JeanSebastienTurcotte:
Added french translations to <exercisegroup>.

### DIFF
--- a/xsl/localizations/fr-CA.xml
+++ b/xsl/localizations/fr-CA.xml
@@ -138,6 +138,23 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <!-- Literate programming, a chunk of computer code -->
     <!-- <localization string-id='fragment'>Fragment</localization> -->
     <!-- Parts of an exercise and its solution -->
+    <!-- An "exercise", at any level, within an "exercises" division is a          -->
+    <!-- "divisional" exercise and the string employed is 'divisionalexercise'.    -->
+    <!-- An "exercise" whose parent is a division (chapter, section, etc) we       -->
+    <!-- call an "inline exercise" and the string employed is 'inlineexercise'.    -->
+    <!-- And an "exercise" in a "worksheet" is a 'worksheetexercise'.              -->
+    <!-- And an "exercise" in a "reading-questions" is a 'readingquestion'.         -->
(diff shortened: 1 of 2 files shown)


xsl/localizations/fr-CA.xml
```diff
@@ -148,6 +165,8 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <localization string-id='hint'>Indice</localization>  <!-- Knowingly differs from fr-FR, which uses "Indication" instead -->
     <localization string-id='answer'>Réponse</localization>
     <localization string-id='solution'>Solution</localization>
+    <!-- A group of divisional exercises (with introduction and conclusion) -->
+    <localization string-id='exercisegroup'>Groupe d'exercices</localization>
```
Alex-Jordan, line 169: @rbeezer should weigh in, but maybe a bare apostrophe could cause trouble. Should it use `<rsq/>`?

rbeezer:
Thanks, @JeanSebastienTurcotte!  That was fast.

I'm trying to use the `en-US` file for documentation, so any edits don't mean multiple edits.  So I added a commit where there is just a reference in **every** file to the explanation of the (confusing) situation with exercises.

JeanSebastienTurcotte:
Sure. I was trying to have the line numbers as close as possible to the US version, hence the copy.

xsl/localizations/fr-CA.xml
rbeezer: Not fast enough.  ;-)  I think a bare apostrophe is fine, authors should be doin' that all the time.  Unclear the exact utility of <rsq/>.  ;-)

rbeezer:
>  line numbers as close as possible

That's a noble aim!  Maybe long explnations in the English file can be labeled (numbers?) so that we can reference from the others.

xsl/localizations/fr-CA.xml
Alex-Jordan: It would be for future proofing, since `'` can be a special character. Just a hair more likely to evade the escaping that we do, since this is an automatically generated string, not author input.

JeanSebastienTurcotte: Better safe than sorry they say

PR #1585, opened 2021-08-20
Pretext classes

davidfarmer:
Change "mathbook-" to "pretext-" in HTML classes, template names, and comments where appropriate.

HTML still compiled as expected.


rbeezer:
Thanks, @davidfarmer!  Website versions going live now.

PR #1586, opened 2021-08-23
port minimal example to CLI

StevenClontz:
This is the first contribution wiring up our examples to be built with the CLI. There's a bit of stale stuff here; I didn't fix every mention of MathBookXML to say PreTeXt for example.


rbeezer:
Merged and pushed.  Thanks!  Keep 'em coming.  But please don't mess with the sample article until last, it'll cause me a lot of pain with the build script.  ;-)

PR #1588, opened 2021-08-24
EPUB: Make WeBWorK exercises always render static

Alex-Jordan:
This makes it so that when making EPUB, WeBWorK exercises ignore the `webwork.....static` stringparam values and just make it all static.


rbeezer:
Thanks.  If we are looking forward, the publisher variable stylesheet will eventually make boolean variables for these (I'd think).  How do you feel about making variables like `b-webwork-inline-static`?  No need for error-checking (that'll happen in publisher variables stylesheet) - just preserve whetehr yes/no makes true/false.

Alex-Jordan:
Got it, I think. I made them in `-html` and `-epub`. Should I have made them in `publisher-variables`?  Force pushed.

PR #1589, opened 2021-08-25
HTML: make inline math wrap span around adjacent non-whitespace

Alex-Jordan:
For discussion. Here is one page where this is in effect: http://spot.pcc.edu/~ajordan/temp/exercises.html

Open the solution to "42a" to see math with adjacent text. For me in Firefox at default Zoom, there is an `x,` at the right edge, and if I narrow the browser they stick together (the comma never moves by itself to the next line). Also there is an example of a right quote from a `q` that does move to the next line. Inspect the HTML to see that `span.mathword` is only used where needed.


Alex-Jordan:
Note in -html, using both `disable-output-escaping="yes"` and `CDATA` so that it can print out `<span>` without a closing tag (within the same template). Is there a more straightforward way?

rbeezer:
Both `disable-output-escaping="yes"` and `CDATA` are monster red flags.  If the former is necessary, you are possibly skirting some work that is more XSL-thonic.

I will actually look at the code ;-) when I get a chance later today.  So take the previous paragraph as a heads-up, not an informed critique.

PR #1594, opened 2021-08-27
LaTeX: stretchy workspace

Alex-Jordan:
Workspace amounts are absolute lengths. Before this commit, LaTeX worksheets come out using exactly those absolute lengths, no more. But more is desirable if there is excess vertical space on the page. This commit uses `\vfill` to evenly spread out any excess vertical space on the page among the various workspaces.

This does not account for workspaces in tasks. A `\vfill` does nothing there, because the tasks and workspace are contained inside a tcolorbox. But the `\vfill` implemented here land immediately after a tcolorbox closes. It's not clear if anything can be done to incorporate spreading out excess vertical space into task workspace.


rbeezer:
Excellent.  Thanks.

Do we need an issue to address the case of tasks with `@workspace`?

PR #1595, opened 2021-08-27
LaTeX: par break before workspace

Alex-Jordan:
Replacing #1593.

In one project, I have a terminal task with `@workspace` and the task ends with a `sidebyside`. This causes a `\\` following the sidebyside to give a "no line to end" error. So this PR replaces the `\\` with a paragraph break.

I think this section (workspace on tasks) is "rough" as already noted in the comments. But changing to `\par` seems good for now.

### DIFF
--- a/xsl/pretext-latex.xsl
+++ b/xsl/pretext-latex.xsl
@@ -6131,7 +6131,8 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             </xsl:apply-templates>
             <!-- this is a bit rough -->
             <xsl:if test="not(task) and @workspace and ancestor::worksheet">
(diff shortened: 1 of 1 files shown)


rbeezer:
Studying this one provokes #1600

rbeezer:
Two places where this affects the sample article, and it looks OK.  Testing with a `\noindent` (as discussed in Drop-In) was totally ineffective it seems.  Perhaps some `tcolorbox` stuff for exercises or tasks is already doing that.

*  `-stringparam latex.draft yes` is very useful for testing this stuff.
*  Limited opportunity to test, so keep an eye on it if you have lots more.
*  I'll try to address revealed bug soon.

Thanks!

PR #1596, opened 2021-08-27
Mathword

Alex-Jordan:
This replaces #1589.

Summary: it scraps the punctuation pull-in for inline math. Replacing it with something that pulls in all adjacent text (both in front and behind). It does not put what it's pulled into the math delimiters. In `-common`, it just prints it before and after the math. But in `-html`, there is  an overriding template that uses a `span.mathword` styled for no line breaking.

No diff in the sample article LaTeX.

The diff with the sample article in HTML is good, replacing things like `\(3+4\text{.}\)` with `<span class="mathword">\(3+4\).</span>`. And things like `(\(\int\))` with `<span class="mathword">(\(\int\))</span>`.

There is a separate commit with an addition to the sample article to take a closer look. See the solution to 42a here:
http://spot.pcc.edu/~ajordan/temp/exercises.html
There is an `x,` and you should not be able to get the comma to move to the next line, even though the comma is not inside the math.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -11845,6 +11845,7 @@ the xsltproc executable.
                 </statement>
                 <solution xml:id="solution-antiderivative">
                     <p>An antiderivative of <m>x^2</m> is <m>F(x)=x^3/3</m>, so by the FTC,<me>\definiteintegral{2}{4}{x^2}{x}=F(4)-F(2)=\frac{1}{3}\left(4^3-2^3\right)=\frac{56}{3}</me>!?!      This is indeed an exciting result, but we are mostly interested in seeing that the sentence-ending punctuation is absorbed properly into the displayed equation.</p>
+                    <p>On a related note, we are interested in seeing that math-adjacent non-whitespace characters (not counting things like quote characters introduced by a <tag>q</tag> element) do not line break awkwardly, so we present some math <m>x</m>, some irrational numbers (<m>\pi</m>, <m>e</m>), and remind you that <m>\lim_{x\to0}\frac{\sin x}{x}</m> has <q><m>0/0</m></q>-indeterminate form. Also, <q>Beware the <m>x</m>-axis!</q></p>
                 </solution>
             </exercise>
 
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -1391,66 +1391,65 @@ Book (with parts), "section" at level 3
 
 <!-- Inline Mathematics ("m") -->
 <!--                                                     -->
-<!-- This is fairly simple.  Differences are             -->
-<!--   (1) Some conversions require different delimiters -->
-<!--   (2) We adjust punctuation for HTML, but not Latex -->
+<!-- Inline math in source may have adjacent text        -->
+<!-- (no spaces in between), and we aim to prevent line  -->
+<!-- breaking between the math and that text.            -->
+
+<!-- So in source like `beforetext<m>math</m>aftertext`  -->
+<!-- the m template reaches out and grabs beforetext and -->
+<!-- aftertext. There is opportunity for output like:    -->
+<!-- [beforetext{math}aftertext]                         -->
+<!-- where [,],{,} are some sort of delimiters (possibly -->
+<!-- empty).                                             -->
+
+<!-- Elsewhere, the text template removes beforetext and -->
+<!-- aftertext so that it is not repeated inoutput.      -->
+
+<!-- Abstract Templates to possibly override             -->
 <!--                                                     -->
-<!-- Abstract Templates                                  -->
-<!--                                                     -->
-<!-- (1) named "inline-math-wrapper"                     -->
+<!-- (1) named "math-adjacent-text-wrapper"              -->
+<!--       Wraps around math, including adjacent text    -->
+<!-- (2) named "inline-math-wrapper"                     -->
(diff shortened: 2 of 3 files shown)


rbeezer:
Catching up on this one.  Improvements here will be welcome.  Thoughts:

* LaTeX itself does not ordinarly need help with inline math.
* We do think LaTeX needs help with display math: author a period logically, then "absorb" it into the end of the display.  A great way to help authors get teh right output.
* If a knowl is simply display math (due to a numbered equation) then the punctuation should not be absorbed, so it is not visible.  It is part of the sentence, not part of teh display/knowl.
* For braille, the punctuation should not migratre in between the Nemeth indicators.
* For EPUB, do we want the punctuation inside a resulting inline/display SVG/PNG?
* I understand inline math to be strictly an HTML/MJ deficiency.

Which is all a long way of saying there needs to be clear and obvious ways to control this manipulation.  For example, braille does math text and literary text separately (and I'm not sure I'm wild about the current state of controlling that, meaning turning it on and off).  This is part of my lukewarm reception towards moving text around with no discernible effect in some situations.  It should be easy to just not have it happen.  Switches or overrides would be teh two approaches to consider?

Alex-Jordan:
> LaTeX itself does not ordinarly need help with inline math.

It's less common, but there are instances like:
```
\documentclass{article}
\usepackage{amssymb}

\begin{document}

The set of coordinates in $\mathbb{R}^3$ where the $z$-coordinate is $0$ is called the $xy$-plane, and it is not equal to $\mathbb{R}^2$.

The set of coordinates in $\mathbb{R}^3$ where the $z$-coordinate is $0$ is called the \mbox{$xy$-plane,} and it is not equal to $\mathbb{R}^2$.

\end{document}
```

> For EPUB, do we want the punctuation inside a resulting inline/display SVG/PNG?

I think yes, or it's the same issue as with HTML. Unless it happens to be the case that in EPUB, text/punctuation that is adjacent to an `img` is "glued" to it. I wouldn't think so, but it warrants testing.

> I understand inline math to be strictly an HTML/MJ deficiency.

I suppose it depends if you view the rare LaTeX cases as a problem too. It's not unreasonable to see the line break in that example as acceptable. But I lean toward the version with the `mbox`.

> braille does math text and literary text separately

I guess I'm not as familiar with the braille as I should be to think about this, but we can end the practice of writing things like `\text{.}` into the math, and be in a place where the only instances of `\text{...}` are things the author wrote. I assume that would simplify the braille end, but maybe not if author-supplied `\tex{...}` still needs separate handling.

More discussion Friday? Or are you back to the day job this Friday?

Alex-Jordan:
Revised outline, ready for an inspection.

First commit makes this happen, as given in code commentary:
```
<!-- Inline Mathematics ("m") -->
<!--                                                     -->
<!-- Inline math in source may have adjacent text        -->
<!-- (no spaces in between), and we aim to prevent line  -->
<!-- breaking between the math and that text.            -->

<!-- So in source like `beforetext<m>math</m>aftertext`  -->
<!-- the m template reaches out and grabs beforetext and -->
<!-- aftertext. There is opportunity for output like:    -->
<!-- [beforetext{math}aftertext]                         -->
<!-- where [,],{,} are some sort of delimiters (possibly -->
<!-- empty).                                             -->

<!-- Elsewhere, the text template removes beforetext and -->
<!-- aftertext so that it is not repeated inoutput.      -->

<!-- Abstract Templates to possibly override             -->
<!--                                                     -->
<!-- (1) named "math-adjacent-text-wrapper"              -->
<!--       Wraps around math, including adjacent text    -->
<!-- (2) named "inline-math-wrapper"                     -->
<!--       Provides the delimiters for inline math       -->
<!--       Stub warnings follow below                    -->
```

The before-after diff at this point will show that output that used to look like `\(x^2\text{,}\)` will now look like `\(x^2\).`. That is, the punctuation (for inline math) is no longer pulled inside the math.

The second commit is small. It just activates the math-adjacent text wrapper for HTML. Now the above will come out `<span  style="white-space: nowrap;">\(x^2\).</span>`. Somehow the inline styling does feel appropriate for this, but it could also be a class.

The third commit adds some examples with math-adjacent text to the sample article. Viewable in the solution to 42a here:
https://spot.pcc.edu/~ajordan/temp/mathword/exercises.html
Note that the quotation marks still line break at a bad place, but that is a different beast. The complications with the `q` tag being involved may be too much without post-processing.

Alex-Jordan:
I was reviewing one of the recent sample pages, and I noticed a "(" at the end of a line, followed by "∫)" on the next line (the integral sign being in math mode. It reminded me of this. Are we still interested in this approach?

rbeezer:
Yes!

rbeezer:
Sounds like Davide Cervone has a very simple fix for this, which should be released soon.  So let's monitor `mathjax-users` for that.

rbeezer:
So I can find it again: relevant thread on `pretext-dev` is titled `Hyphenation in mixed math-text word`

Alex-Jordan:
I think the structure here *might* be useful in the abstract. The LaTeX MWE in this thread demonstrates that.

But the practical impetus for this was hanging punctuation adjacent to math, and @dpvc indicates that MathJax 4 will handle that directly.

So I'm closing this. The code will always be findable if there is some future use for it.

rbeezer:
Thanks, Alex.

>  MathJax 4 will handle that directly.

We'll need to give that a big hard test before we bail out here.  I've just not had an opportunity yet.

PR #1598, opened 2021-08-28
HTML: use hN for duplicate headings

Alex-Jordan:
This returns to using hN (at the appropriate level) for headings in `list-of` and `solutions`. A chunk level 0 example is here:

http://spot.pcc.edu/~ajordan/temp/derivatives.html
(Don't click to follow links to appendices the nav column. They incorrectly point to backmatter.html which should not exist with chunk level 0.)

If you use Firefox and haven't already, I like the [headingsMap extension](https://addons.mozilla.org/en-US/firefox/addon/headingsmap/) to view the headings and their depths in a side column.

There are `solutions` in 4.2.7, and in appendices B–E.  Appendices F and G have `list-of`.

This is the same as where PR #1582 wound up, but there were too many course corrections there and the thread was confusing.

This undoes something that #1580 did, but leaves in place the infrastructure that makes it easy to switch back and forth about this and make future decisions about stacked headings (re our discussion this past Friday). 

It also corrects the value of heading-level that is passed to list-of. Since list-of doesn't have a heading, this subtracts 1 from that value before passing it down.

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -7896,7 +7896,9 @@ Neither: A structural node that is simply a (visual) subdivision of a chunk
     <xsl:call-template name="list-of-begin" />
     <!-- recursive procedure, starting from indicated scope -->
     <xsl:apply-templates select="$subroot" mode="list-of-content">
-        <xsl:with-param name="heading-level" select="$heading-level"/>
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks!

PR #1602, opened 2021-09-02
Doc: clean up incorrect use of 'attr'

Alex-Jordan:
I think I keep being led to write like `<attr>key=value</attr>` because I look up other instances and see it there, and think it is an OK thing to do. So this commit removes those instances.


rbeezer:
Thanks!  Sorry I hadn't caught those sooner and kept you from bad habits!  ;-)

PR #1603, opened 2021-09-02
HTML: collapse stacked headings in solutions

Alex-Jordan:
This PR is perhaps just for discussion.

This commit collapses stacked headings in "solutions". For example, look at the current:

https://pretextbook.org/examples/sample-article/html/solutions-backmatter.html

As I inspect that page now, none of the:
```
4 An Interesting Corollary
4.2 A Pedagogical Note about Subsection 4.1
4.2.1 Symbolic and Numerical Integrals
```
use hN. But I think the sample article has not been rebuilt following the merge of #1598. And with that merged, this is currently coming out as:
```
(h3) 4 An Interesting Corollary
(h4) 4.2 A Pedagogical Note about Subsection 4.1
(h5) 4.2.1 Symbolic and Numerical Integrals
```

This commit collapses that into:
```
(h3) 4.2.1 Symbolic and Numerical Integrals
```

And the whole page of headings becomes:
```
(h3) 4.2.1 Symbolic and Numerical Integrals
(h3) 4.2.3 Advice
(h3) 4.2.4 Exercises
(h3) 4.6.2 Test Two [appears because of answers added in this commit]
(h3) 11.3 More Exercises
(h3) 21 Program Listings
...
```

So, all styled the same and all at h3, even though there are different logical depths. The benefit is that heading tree has been simplified.

One reason that this is just for discussion is because maybe instead of:
```
(h3) 4.2.1 Symbolic and Numerical Integrals
```
people would prefer something like:
```
(h3) 4 An Interesting Corollary
     4.2 A Pedagogical Note about Subsection 4.1
     4.2.1 Symbolic and Numerical Integrals (/h3)
```
Personally I feel that is too busy, but I could understand wanting to show all the titles that lead up to the "leaf" division.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -1448,12 +1448,40 @@ the xsltproc executable.
                     <title>Test Two</title>
 
                     <p>An intervening subsection section which contains an <tag>exercises</tag> division which must be at the level of a Subsubsubsection.</p>
+                    <exercise>
+                        <statement>
+                            <p>An inline exercise to examine any clash with divisional exercises below.</p>
+                        </statement>
+                        <answer>
+                            <p>An answer so there is something to appear in a <tag>solutions</tag>.</p>
+                        </answer>
+                    </exercise>
+
+                    <reading-questions>
+                        <title>What Did You Learn?</title>
+                        <exercise>
+                            <statement>
+                                <p>A mock exercise to appease validation.</p>
+                            </statement>
+                            <answer>
+                                <p>An answer so there is something to appear in a <tag>solutions</tag>.</p>
+                            </answer>
+                        </exercise>
+                        <exercise>
+                            <statement>
+                                <p>And a second to help with formatting the division heading.</p>
+                            </statement>
+                        </exercise>
+                    </reading-questions>
 
                     <exercises>
                         <exercise>
                             <statement>
                                 <p>A mock exercise to appease validation.</p>
                             </statement>
+                            <answer>
+                                <p>An answer so there is something to appear in a <tag>solutions</tag>.</p>
+                            </answer>
                         </exercise>
                         <exercise>
                             <statement>
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -3369,6 +3369,19 @@ Book (with parts), "section" at level 3
     <xsl:value-of select="false()" />
 </xsl:template>
 
+<!-- Specialized divisions (exercises, worksheet, reading-questions) sometimes   -->
+<!-- get special handling when the parent division is unstructured.              -->
+<xsl:template match="*" mode="is-specialized-division-in-unstructured">
+    <xsl:value-of select="false()"/>
+</xsl:template>
+
+<xsl:template match="exercises|worksheet|reading-questions" mode="is-specialized-division-in-unstructured">
+    <xsl:variable name="parent-is-structured">
+        <xsl:apply-templates select="parent::*" mode="is-structured-division"/>
+    </xsl:variable>
+    <xsl:value-of select="not($parent-is-structured = 'true')"/>
+</xsl:template>
+
 <!-- Structural Leaves -->
 <!-- Some structural elements of the document tree     -->
 <!-- are the leaves of that tree, meaning they do      -->
@@ -7369,6 +7382,9 @@ Neither: A structural node that is simply a (visual) subdivision of a chunk
     </xsl:apply-templates>
 </xsl:template>
 
+<!-- If not explicitly addressed in another template, dry-run returns empty -->
+<xsl:template match="*" mode="dry-run"/>
+
 <!-- This template is called for items in a worksheet that can have a  -->
 <!-- workspace specified.  It is important that this sometimes returns -->
 <!-- an empty string, since that is a signal to not construct some     -->
@@ -7492,6 +7508,7 @@ Neither: A structural node that is simply a (visual) subdivision of a chunk
                 <xsl:with-param name="purpose" select="$purpose" />
                 <xsl:with-param name="admit"   select="$admit"/>
                 <xsl:with-param name="heading-level" select="$heading-level"/>
+                <xsl:with-param name="scope" select="$scope"/>
                 <xsl:with-param name="b-inline-statement"     select="contains(@inline,     'statement')" />
                 <xsl:with-param name="b-inline-hint"          select="contains(@inline,     'hint')"      />
                 <xsl:with-param name="b-inline-answer"        select="contains(@inline,     'answer')"    />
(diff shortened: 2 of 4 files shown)


Alex-Jordan:
Here is a page built from my branch (not yet pushed):

https://spot.pcc.edu/~ajordan/temp/solutions-backmatter.html

It groups stacked headings into a single hN. Then each heading is inside a `p.assistive`. Our thinking is that the `p` will give screen readers a pause between reading each heading. I put the `assistive` class on it in case that can help with styling. I think all space between these p's can be removed.

@davidfarmer Can I ask you to give feedback on whether or not `.assistive` is needed, and/or if something else would be better? And then do some styling during wine time.

If styling is worked out and this output looks good to @rbeezer, I'll push the branch for inspection.

rbeezer:
Something's amiss:  `Secton 4` shows up 5 times, and it should only be once.

Note for anybody looking at this:  `Subsection 4.2 A Pedagogical Note about Subsection 4.1` is a heading, and it wraps funny so that there seems to be a `Subsection 4.1` that is out-of-place (perhaps).

Alex-Jordan:
> Something's amiss: Section 4 shows up 5 times, and it should only be once.

Well, it was intentional. Sort of. Each "leaf" division shows all of its ancestors up to the scope of the solutions. But I can adjust to leave out ancestors when appropriate.

Styling and .assistive questions can still be addressed in the meantime.

davidfarmer:
> @davidfarmer Can I ask you to give feedback on whether or not .assistive is needed, and/or if something else would
> be better? And then do some styling during wine time.

Reload and let me know what you think of the styling on the
.assistive lines in the .heading.

Alex-Jordan:
@davidfarmer It looks good!

The second block that ends with "Subsubsection 4.2.3 Advice" is going to become a single line that only has "Subsubsection 4.2.3 Advice". I'm guessing that will mess this up unless I leave some empty `p.assistive` preceding it. Which feels like a hack.

I'll post again when the next iteration is ready to look at.

rbeezer:
All the headings within the scope are meant to be displayed once and only once (subject to the @empty switch).

Alex-Jordan:
OK, next iteration is ready to look at:

https://spot.pcc.edu/~ajordan/temp/solutions-backmatter.html

I think the progressive indentation and decrease in size on progressive `p.assistive` has to go. But if something like that is important to keep, I could work on encoding a depth into a class name on each `p.assistive`. I'm OK with adjacent `p.assistive` all styled the same, just close together.

All the division headings within the "solutions" are `h3` except that 4.6.2 is unstructured, with one "reading-questions" (that I gave a title to) and one "exercises" (that I did not). And I let those be `h4`.

If no flags raised, I'll clean up code, and align it with latex and the solutions making stylesheet.

davidfarmer:
I deleted the progressive indentation.

Alex-Jordan:
Question for Rob. Are "exercises", "worksheet", "reading-questions" considered "structural" when they are in an unstructured division?

Part of the edits on a branch changes:

```
<xsl:template match="&STRUCTURAL;" mode="is-structural">
    <xsl:value-of select="true()" />
</xsl:template>
```
to
```
<xsl:template match="&STRUCTURAL;" mode="is-structural">
    <xsl:choose>
        <xsl:when test="(self::exercises|self::worksheet|self::reading-questions) and not(parent::*/child::section|parent::*/child::subsection|parent::*/child::subsubsection)">
            <xsl:value-of select="false()" />
        </xsl:when>
        <xsl:otherwise>
            <xsl:value-of select="true()" />
        </xsl:otherwise>
    </xsl:choose>
</xsl:template>
```

For me with this branch, this is just a convenient way to handle those kinds of `exercises|worksheets|reading-questions` separately. I could distinguish them using different means if needed. But I do wonder if changing the template like this (or the meaning of `&STRUCTURAL;`) is the "right" thing to do.

rbeezer:
That template might be poorly named, but it should simply indicate if an element is in that list.  It is critical that it not change in any way.  (Sorry!)

I bet all kinds of things will break if you were to do somne testing with your proposed changes.

rbeezer:
> OK, next iteration is ready to look at:

Yes, that's looking more like the status quo.  I suspect we'll want @davidfarmer to tighten-up some of the whitespace preceding each stack of headings following when a solution is displayed.

Alex-Jordan:
> That template might be poorly named, but it should simply indicate if an element is in that list. It is critical that it not change in any way. (Sorry!)

No problem. I just made a new modal template `is-specialized-division-in-unstructured`. Name too long?

Alex-Jordan:
> If no flags raised, I'll clean up code, and align it with latex and the solutions making stylesheet.

It looks like I never completed this. If it's not too late, hold up with reviewing it until I give a green light.

rbeezer:
This was first for takeoff this morning, but we had a power outage.  `is-specialized-division-in-unstructured` is no too long.  Inside code (and maybe everywhere!) clarity trumps verbosity.

Holler when this is ready.

Alex-Jordan:
Force pushed and ready for inspection.

Testing report:
* There are two commits here. The second one has sample article changes, so you can do a cleaner before/after comparison without that commit.
* I found no unwanted changes with sample article and sample book, neither with latex nor html.
* Note the solutions manual is not affected at all by this PR. The way that one builds "solutions" is different enough that it is just not affected. There is no diff at all with the sample-book solutions manual.

Prior to this, the `division-in-solutions` template is not part of -common. By the time I had made changes to -html and -latex, this template was nearly identical in each, so I moved it to -common. (Part of the dream, no?) It relegates the finer point of actually printing the headings to each output-specific stylesheet.

The big picture is that the solutions generator passes a "heading stack" parameter down recursively. It's a node set containing all divisions from the scope of the solutions down to the current division. But excluding any divisions where the heading has already been printed. And ultimately, excluding the scope form being printed too, if it is still part of the heading stack. But it helps to include the scope within the heading stack.

HTML ends up with stacked headings inside a single hN, with each one wrapped in a `p.assistive`. LaTeX groups the headings in a similar block.

Suppose `solutions` is called within a division whose heading level is n. Then these blocks will all be level n+1. Even when these blocks may represent divisions with logically distinct heading depth. For example, one block may show section content, and the next may show subsection content (from a different section). Both headings will be level n+1. 

Then there can be `exercises|reading-questions|worksheet` within an unstructured division. That division may have had content to show, so its heading is printed at level n+1. But then the `exercises|reading-questions|worksheet` has its heading printed at level n+2.

For now, any `subexercises` have their heading at `h6`. This is just how they are already handled, no changes there. They have not been treated as a "division" up to now. In future (but soon) work, I'll change that. But there's enough in this PR to leave that alone for now.

Future work might also remove the `division-in-solutions` template from the solutions manual stylesheet to make it a truly abstract template in -common across stylesheets. But again it feels like a move that can be isolated form this PR.

Alex-Jordan:
Here are some builds, for convenience to reference.

Sample Article. Places to take note are section 4.2.7, and appendices B–E.
- PDF: http://spot.pcc.edu/~ajordan/temp/stacked-headings/sample-article/derivatives.pdf
- HTML (knowl files omitted): http://spot.pcc.edu/~ajordan/temp/stacked-headings/sample-article

Sample Book. Places to take note are appendices B–D.
- PDF: http://spot.pcc.edu/~ajordan/temp/stacked-headings/sample-book/book.pdf
- HTML (knowl files omitted): http://spot.pcc.edu/~ajordan/temp/stacked-headings/sample-book

rbeezer:
As discussed at Drop-In, I added `references` and `glossary` to the template for determining structured division parents.  With more documentation.

Also, -common defines `$b-is-book` and `$b-is-article`, so I employed those where the LaTeX conversion initializes the heading level.

Diffs look good, output looks good, and it seems to be a big improvement for modularity.

Thanks for sticking with this one, I know it has been a slog.  And I know better why I never started pushing hN down below divisions!  ;-)  This (and subsequent work) will make us look so much better during accessibility testing.

Website rebuild in-progress.

PR #1605, opened 2021-09-07
Slideshow: update Reveal.js version

Alex-Jordan:
Updates to load the most recent reveal.js on the cloudflare cdn (4.1.2).

Recent versions in the cdn do not have `css` and `js` folders. You just go straight to the css and js files you are after.

Also [the actual repo](https://github.com/hakimel/reveal.js/tree/master/dist) has a `dist` folder that has the necessary js and css files instead of using css and js subfolders for these. So if you provide locally, you would drop these files all together in your reveal-root folder instead of using the intermediate folders.

Since revealjs slideshows are provisionally supported, not sure what other things I should do at the same time. The current section about revealjs in the doc doesn't appear to get into the details of having a `css` and `js` subfolder. 

BTW, I was writing a revealjs slideshow through PTX, trying to use a newer feature of revealjs, and it wasn't working. That is what led me to make this update. Now it's working: [live background page on a slide](https://spot.pcc.edu/~ajordan/MHECWWdemo/slideshow.html#/1/1).

### DIFF
--- a/xsl/pretext-revealjs.xsl
+++ b/xsl/pretext-revealjs.xsl
@@ -88,16 +88,16 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
             <!-- load reveal.js resources             -->
             <xsl:choose>
                 <xsl:when test="$b-reveal-minified">
-                    <link href="{$reveal-root}/css/reset.min.css" rel="stylesheet"></link>
-                    <link href="{$reveal-root}/css/reveal.min.css" rel="stylesheet"></link>
-                    <link href="{$reveal-root}/css/theme/{$reveal-theme}.min.css" rel="stylesheet"></link>
-                    <script src="{$reveal-root}/js/reveal.min.js"></script>
+                    <link href="{$reveal-root}/reset.min.css" rel="stylesheet"></link>
+                    <link href="{$reveal-root}/reveal.min.css" rel="stylesheet"></link>
+                    <link href="{$reveal-root}/theme/{$reveal-theme}.min.css" rel="stylesheet"></link>
+                    <script src="{$reveal-root}/reveal.min.js"></script>
                 </xsl:when>
                 <xsl:otherwise>
-                    <link href="{$reveal-root}/css/reset.css" rel="stylesheet"></link>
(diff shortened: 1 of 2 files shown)


rbeezer:
Basic testing **reveals** that this is not as simple as it appears.  Some notes for when there is time to do this completely:

*  I'd suggest documentation says: for a local version to copy `dist` directory and the `plugin` directory to directory with HTML output/file.
*  Documentation should say that math will not work out-of-the-box for a local version.
*  Then local paths should use the `dist` prefix.
*  Looks like all CSS and JS have been minified.  Local and CDN.  So that publisher switch should be made obsolete, with warnings, and removed from documentation.
*  Appears the setup for math has changed radically.  I wonder if this means we can switch to MJ 3 for our stuff?  Though the plugin file (JS) seems to have a 2.7 hardcoded.  See https://revealjs.com/math/ for configuration changes.  Perhaps best to get MJ 2 working, and then see what we do relative to 3 that could be tested.  Being stuck on 2 should be documented (even if temporarily).

rbeezer:
I've got the hard parts above either implemented or debugged, so don't do a force push or anything.

Alex-Jordan:
No problem. Working on a monster headings PR for you to review (#1603). :)

rbeezer:
Turns out to be more convenient to not use the `dist` directory for local use.  But a bit more fiddle for presenters.  Hard parts are done.  And I think I'm done for today.

rbeezer:
Four commits later, and I think this has all the changes needed beyond a new version number.  Plus documentation on running a local slideshow.

Alex-Jordan:
Nice. I'll move slide backgrounds up one notch now on my list.

On Tue, Sep 21, 2021 at 4:03 PM Rob Beezer ***@***.***> wrote:
>
> Four commits later, and I think this has all the changes needed beyond a new version number. Plus documentation on running a local slideshow.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android.



-- 
Alex Jordan
Mathematics Instructor
Portland Community College

PR #1607, opened 2021-09-10
WeBWorK: copy problems that use tasks

Alex-Jordan:
The `@copy` mechanism was not upgraded to handle WW exercises with tasks. This commit does that.


rbeezer:
The warning message (about copying a problem with `@source`, etc) needed a similar change.  I built an example for the sample chapter to verify that the change was needed.

Extra code melded into your commit, sample chapter addition on a new commit.

Will be live very shortly.

Alex-Jordan:
Thanks, and sorry for not being more thorough.

I need to more fully retire `stage`. For the rare cases of an old WW server that returns `stage`, it can safely become `task` during assembly.

rbeezer:
> Thanks, and sorry for not being more thorough.

Well, I only stumbled on the necessity semi-accidentally while trying to locate the relevant code for context.  But it illustrates the value of adding a bug-provoking example for every bug.  ;-)

PR #1609, opened 2021-09-13
dl example expanded to show new 'narrow' behavior

davidfarmer:
The `width="narrow"` attribute on a `dl` now adjusts to allow very long terms.
The example in the Sample Article was expanded to illustrate this behavior.

No XSL changes were needed:  just choose the `narrow` option on the `dl`.


rbeezer:
Thanks!  The new bits are almost lyrical.

Merged, and rebuilding website right now.  Give it about 20 minutes before checking.

PR #1606, opened 2021-09-09
Guide: update WeBWorK info

Alex-Jordan:

PR #1610, opened 2021-09-14
Ww tikz intro

Alex-Jordan:
Second commit here adds a `PGtikz` image to an exercisegroup's introduction. 

First commit lays down the infrastructure to make it happen.

They are separated to help you do a before/after diff. It is expected that you will see changes. In particular, the names for the perl variables storing each image get a new numbering scheme, local to their problem.

When #1606 is resolved, I may have related documentation to add. I started that here, but soon realized it would conflict.

### DIFF
--- a/examples/webwork/Makefile
+++ b/examples/webwork/Makefile
@@ -70,6 +70,7 @@ DTD = $(MB)/schema/dtd
 
 # XML to apply templates to
 SMPCHP = $(WWEXAMPLE)/sample-chapter/sample-chapter.xml
+SMPCHPPUB = $(WWEXAMPLE)/sample-chapter/publication.xml
 MINI = $(WWEXAMPLE)/minimal/mini.xml
 
 # These paths are subdirectories of
@@ -79,6 +80,7 @@ HTMLOUT    = $(SCRATCH)/html
 PDFOUT     = $(SCRATCH)/pdf
 EPUBOUT    = $(SCRATCH)/epub
 WWOUT      = $(SCRATCH)/webwork-representations
+IMGOUT     = $(SCRATCH)/images
 
 # Some aspects of producing these examples require a WeBWorK server.
 # Either specify only the protocol and domain (like https://webwork.yourschool.edu)
@@ -100,6 +102,15 @@ SERVER = https://webwork-ptx.aimath.org
 #  webwork-representations.xml which holds multiple versions of each problem.
 #  Also locally store images from the WeBWorK server.
 
+#  BUILD TREE FOR SAMPLE CHAPTER
+#  make sample-chapter-check
+#  make sample-chapter-representations
+#      make sample-chapter-images
+#          make sample-chapter-html
+#          make samepl-chapter-epub (planned)
+#      make sample-chapter-pdf
+#      make sample-chapter-problem-sets
+
 #  For HTML and problem set output, we turn off dates in file headers,
 #  so when we use these for testing we can get more accurate diffs.
 #  Dates are useful for single file output (or at least less of an
@@ -109,7 +120,7 @@ SERVER = https://webwork-ptx.aimath.org
 
 sample-chapter-representations:
 	install -d $(WWOUT)
-	-rm $(WWOUT)/webwork-representations.ptx
+	-rm $(WWOUT)/*.*
 	$(MB)/pretext/pretext -v -c webwork -d $(WWOUT) -s $(SERVER) $(SMPCHP)
 
 mini-representations:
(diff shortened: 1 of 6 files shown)


rbeezer:
Do you expect the graph to randomize when used in a WW problem that is built as an archive of problems extracted from a text?  I guess not, since it would not vary across the problems within the exercise group when present within the text.  So your example is an accurate use case?  One graph, two sets of values for the question.

Why not one question, and a copy with a different seed?

Why not two questions, with the graph repeated, and authored once in a separate file that gets included twice?

Alex-Jordan:
> Do you expect the graph to randomize when used in a WW problem that is built as an archive of problems extracted from a text?

No, for at least two reasons. An `exercisegroup/introduction` is common instructions for a collection of exercises. It's inconsistent if it would be possible for the "common instructions" to change as you move from one exercise to the next. But also, practically that would mean the `exercisegroup/introduction` would need its own `pg-code` section, which would be too big of a new thing to make happen. 

In case you are wondering, so why use `syntax='PGtikz'`? Because these have to be in the form of the guts of a `tikzpicture`, without the environment wrapper, and regular `latex-image` would either include the `tikzpicture` environment wrapper or maybe even not use tikz.

> Why not one question, and a copy with a different seed?

The way it is now there is no need for randomization code, and no `pg-code` in those two exercises. It's just hard coded to use two different Pythagorean triples.

> Why not two questions, with the graph repeated, and authored once in a separate file that gets included twice?

Because then in print you have to see the graph twice. But it's the same graph, part of common instructions for the exercisegroup.

rbeezer:
> Because then in print you have to see the graph twice. But it's the same graph, part of common instructions for the exercisegroup.

OK, two different graphs then with the values from the two problems hard-coded.  Now there is not a common graph and no need for an exercise group.

Alex-Jordan:
I rebased from dev and force pushed. That's the only change with the force push. I am doing some ORCCA work and I needed what's on dev and this branch together. I did the rebase locally and there was a conflict, so it seemed right to force push after resolving that.

Alex-Jordan:
> OK, two different graphs then with the values from the two problems hard-coded. Now there is not a common graph and no need for an exercise group.

I'm not following you. This PR supports when you want to place a graph in an exercisegroup introduction, and the exercises from that exercisegroup are WW exercises.

rbeezer:
It moves WW-specific design decisions (like providing the begin/end environment automatically, something we abandoned) outside of the `webwork` element, which leads to a whole lot of exceptional code in lots of places.  How do you propose the schema account for this?

I understand the use case entirely, but I am very hesitant about WW exceptions that are not contained within the `webwork` element.  The use case might be desirable, but I do not see it as necessary.

Alex-Jordan:
It's a bit of a rock and a hard place. The following is schema-compliant right now:

```
exercisegroup
    introduction
        image
            latex-image
    exercise
        webwork
```

But it will fail if you try to build it. The webwork pulls the exercisegroup/introduction into its PG code.

Using `syntax='PGtikz'` is supposed to isolate this.

Since we pull exercisegroup/introduction into PG code, perhaps there should be a special version of exercisegroup in the schema for WW exercises. This would enforce not mixing WW and non-WW within an exercisegroup, something that has always seemed like the right thing to do.

Alex-Jordan:
On the WW end, perhaps I can add a flag to the tikz tool that would suppress the automatic `begin/end{tikzpicture}`. If I can do that there, maybe the PTX side can have cleaner more isolated code. I'll give this a try.

rbeezer:
> perhaps there should be a special version of exercisegroup in the schema for WW exercises

That's the only real alternative I could come up with overnight.  A lot of tests could become a lot cleaner.  Plus the schema (wr already have a `WWImage`, iirc).

> perhaps I can add a flag to the tikz tool that would suppress the automatic begin/end{tikzpicture}

I think that would help a lot, and maybe the LaTeX conversion (say) would not need as many changes.

Alex-Jordan:
Subject to review and approval by WW maintainers, I have a new avenue in WW for generating LaTeX images that is more like PreTeXt: not tikz-centric. If it is approved, I can take things here in that direction.

A complication: the new avenue will not be "out there" until WW 2.17 is released. 

I think I will leave this PR open for now, at least until the WW PR is resolved. It helps me remember what things are actively cooking.

rbeezer:
Is 2.17 far enough along to resume this one?

Alex-Jordan:
Yes, it's on my list now to return to this.

Alex-Jordan:
#1787 replaces this.

PR #1611, opened 2021-09-14
WeBWorK: use the problem id as problemUUID

Alex-Jordan:
When sending a request to the WW server, there is parameter `&problemUUID` that we have not used up to now. It did not exist when we first started WW-PTX, and I've been slow to come back to using it. This PR uses it, using the PTX visible-id of the problem.

I recently discovered that this is really needed. Take a look at:
https://pretextbook.org/examples/webwork/sample-chapter/html/section-9.html

Open checkpoint 9.1 and you will see the wrong graph. It's a graph of a cubic function that belongs in checkpoint 9.3. The explanation is that not using `problemUUID`, image files are overwriting each other on the server. Caching of image files with the same name compounds the issue, so even if you make that problem live, you still may see the wrong graph. But with `problemUUID` in use, `make sample-chapter-representations` is returning the correct image files.

Note: does your build script copy image files from the representations folder to where the public sample chapter HTML can get them? If not, then after this is merged, you will need to copy those image files over once manually. (But writing that into the build script would be good, in case other work brings in more image files.)

Note: now I will move to use `problemUUID`in the PTX-WW javascript too. Without that, I think that the live versions will still show wrong graphs in some cases.

Note: @sean-fitzpatrick the previous paragraph probably addresses the thing where if seeds are the same, we get images replacing each other.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -678,7 +678,8 @@ def webwork_to_xml(xml_source, pub_file, stringparams, abort_early, server_param
                          ('course_password',course_password),
                          ('outputformat','ptx'),
                          server_params_source,
-                         ('problemSeed',seed[problem]))
+                         ('problemSeed',seed[problem]),
+                         ('problemUUID',problem))
 
         msg = "sending {} to server to save in {}: origin is '{}'"
         _verbose(msg.format(problem, dest_dir, origin[problem]))


Alex-Jordan:
> Note: now I will move to use problemUUIDin the PTX-WW javascript too. Without that, I think that the live versions will still show wrong graphs in some cases.

I've made this change. So now when you load a live problem, you should see the right graphs. In some cases, it may feel slow, if the graph is not cached yet.

sean-fitzpatrick:
Nice! I'll check this out. Sounds like no need to rebuild, just reload?

Alex-Jordan:
@sean-fitzpatrick Yes, unless I am still misdiagnosing that issue, you would only need to reload with a fresh version of the javascript. Hard to do before/after testing though. If it's working, I don't know if you can get an example up now where you will experience the original issue.

rbeezer:
Merged.  Yes, build script regenerates as much as possible each time.  Which is why it takes so long.

I've started a rebuild, should be done in about 20 minutes.  Wanna verify that all is well?  Better, even?

Alex-Jordan:
Looks good. And better, even :)

rbeezer:
Great.  Thanks for checking.  Working through the backlog...

PR #1614, opened 2021-09-16
More dl examples to check spacing when the width is narrow

davidfarmer:
I added several examples to the sample article narrow description list,
so that we can determine whether the switch from inline terms, to floating
above terms, comes at the correct width.

I think it is okay, but I am happy to change the CSS if needed.

Also corrected two typos.


rbeezer:
Thanks, @davidfarmer.  Timely - I'm getting close on the LaTeX version.  Gotta fix up the labels for cross-references, and then should have it.

Error with the Riemman zeta function?  In the git history where it can never be revoked?  ;-)

Should be live on website in a few minutes.

PR #1617, opened 2021-09-17
Updates to Guide and Glossary

davidfarmer:
In the Guide:
Add xref to a section introduction, update the glossary, and omit references to MathJax 2.

In pretext-common:
Omit period at end of glossary terms


rbeezer:
The cross-reference to the `introduction` may work, but it throws a warning at the console (the same one that started the "p" discussion).

There is then a cascade of changes and schema edits necessary.  So this won't happen as fast as suggested.

PR #1618, opened 2021-09-24
HTML: scope for table column headers

Alex-Jordan:


rbeezer:
Should this not also be the case for `@header = 'vertical'`?  In which case, maybe it makes sense to conditon on `$header-row-elt`?

Alex-Jordan:
Maybe I shouldn't write code during the meetings.

rbeezer:
"Can you multitask?"  ;-)

Alex-Jordan:
Is it better now?

Alex-Jordan:
> "Can you multitask?" ;-)

You didn't specify performance :)

rbeezer:
Much better.  ;-)  And if there needs to be another conditional change sometime later, perhaps a boolean variable will be best for readability.  Website update perhaps sometime later today.

PR #1619, opened 2021-09-26
Solution Manual: shift to -commons division-in-solutions template

Alex-Jordan:
Unclear if this is desirable.

This makes the solutions manual stylesheet use the new -common `divisions-in-solutions` template and be more like the -html and -latex stylesheets in this regard.

Since the solutions manual stylesheet is so different (opening chapters/sections with numbers hard coded into titles, adding TOC entries) the move doesn't save lines of code. In fact there is a new growth of 2 lines.

Most of the changes in the diff (I think 92 out of 110) are just indentation of code within a new `for-each` (which is looping through the new heading stack).


rbeezer:
Likely desirable.

I'll test against AATA's (nontrivial) solutions manual to understand it a bit better.  Thanks.

Alex-Jordan:
One thing I realized while dozing off last night. As it stands, I use a `for-each` to go through the `$heading-stack` node set. An alternative would be to use a modal template, selecting the `$heading-stack`. I'm honestly not sure why I defaulted to `for-each`. Should I change it before you get into testing?

rbeezer:
AATA solution manual is in a private repo, so not available.  But I built LaTeX solution manual, which I believe never has stacked headings.

So no change in produced TeX.  I guess that is the desired result, no?  Let me know and I'll merge (or double-check my testing!)

Yep, a `for-each` can often be replaced.  But they are often tempting, and the alternative seems like more work that it is worth?  I suspect it is perfectly fine here.  A purist would suggest the only real use is to change context (which could mean iterating over just **one** node).

PR #1622, opened 2021-10-03
LaTeX: recover WW exercises lost with ca03a3f

Alex-Jordan:
This fixes something that was broken by ca03a3f. That commit explicitly nullified dry-run content for elements that don't otherwise have an explicit dry-run modal template. The downwind effect of that was that in -latex, WW exercises were not printing where born. They were conditioned on dry-run not being empty. Maybe at some point that was needed because of building solutions, but currently, the "solutions" modal template for "exercise" handles this by executing dry-run before calling exercise-components.

I ran diffs on the sample chapter tex file comparing right before ca03a3f, dev at its present tip, and following this commit. I only see diffs that are good.

And did the same with the sample article, same results.

### DIFF
--- a/xsl/pretext-latex.xsl
+++ b/xsl/pretext-latex.xsl
@@ -5925,7 +5925,6 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:param name="b-has-answer"  />
     <xsl:param name="b-has-solution"  />
 
-
     <!-- The introduction (and conclusion) remain in the "exercise" as part -->
     <!-- of the enhanced source, and so lie outside of the "webwork-reps"   -->
     <xsl:if test="$b-has-statement">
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
As noted in a [forum thread](https://groups.google.com/g/pretext-dev/c/HpARdVYHvBw/m/i4-c_pBzBAAJ), while looking into this I think I discovered an issue with tasks in solutions in latex. I'd like to address that separately. It was there before ca03a3f if I understand correctly, and just went unnoticed.

Alex-Jordan:
The diff that I see here in GitHub is more confusing than what it ought to be. All that happened is that where a dry-run variable was created, that is deleted. And then the `xsl:if` on the dry-run is removed, and the content inside loses a level of indentation. Two places: WW exercise with the old "stage" and all other WW exercises.

PR #1625, opened 2021-10-08
HTML: straighten out heading level for block elements

Alex-Jordan:
This puts the appropriate heading level on all of the hN for block objects in HTML. It's all controlled by the modal `hN` template which essentially counts ancestors that would appear on the page with a heading already.

Things I've tested:

- A diff on the before/after HTML output only has good changes. That diff is a monster, so if you want to look at it too, I pre-processed the files before making the diff. I changed `h\d` to `hN` so it wouldn't show lines where only the N changed.
- As noted in a thread, in the knowl files, an svg namespace is either moved or dropped altogether. Seems harmless.
- I inspected the headings all the way through the sample article HTML output, both at chunk level 0 and chunk level 1. The heading tree tool shows no invalid heading depth jumps. And I manually looked for other inappropiate things like jumping down in depth by 1 when that shouldn't happen. Or jumping up in depth when that shouldn't happen. Seems to be all good.
- I ran a validator on EPUB output for the epub sampler and it is has 40 validation errors! But presently, those are there with the dev branch too, and I don't think this PR makes it worse. So there is something to look into and fix, but it is unrelated to this. And perhaps epub validation checking needs to become more routine.

There is a list of heading issues in HTML yet to address. I'll list them here. I think that they are separate projects, but if you are looking at the output here, you might notice some of these "definciencies".

- In list-of output, we still have "stacked headings".
- In the index, I think that we should make it more navigable by making each letter be in an hN.
- In a solutions, exercises with tasks are still a source of stacked headings. Like Exercise 47, task (a), subtask (i) may have an answer to print. But task (a) does not need to be have its own heading.
- Related, tasks in solutions come out (with this PR's commit) labeled like "4.10.c.i". Maybe they will look better just as "(i)" or something like that. But it's something to tackle at the same time as the previous item.

### DIFF
--- a/xsl/entities.ent
+++ b/xsl/entities.ent
@@ -71,6 +71,8 @@
 <!-- statement + proof -->
 <!ENTITY THEOREM-LIKE "theorem|corollary|lemma|algorithm|proposition|claim|fact|identity">
 <!ENTITY THEOREM-FILTER "self::theorem or self::corollary or self::lemma or self::algorithm or self::proposition or self::claim or self::fact or self::identity">
+<!-- catch a proof that is inside THEOREM-LIKE -->
+<!ENTITY INNER-PROOF-FILTER "self::proof and (parent::theorem or parent::corollary or parent::lemma or parent::algorithm or parent::proposition or parent::claim or parent::fact or parent::identity)">
 
 <!-- a mathematical statement, but never with a proof -->
 <!ENTITY AXIOM-LIKE "axiom|conjecture|principle|heuristic|hypothesis|assumption">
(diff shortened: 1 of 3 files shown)


rbeezer:
Almost merged one during Drop-In.....

Alex-Jordan:
> Accessibility video next?

Soon :)  I have 60 midterm exams to grade :(

> Mind if I sing your praises on pretext-announce?

This is worth an -announce since it could encourage people to rebuild their HTML. I don't need any praises sung though. Thanks for the work testing it out.

> Website update in-progress. Very close to single digits...

Hmm, maybe I will have to change my evening plans then ;)

PR #1626, opened 2021-10-11
HTML: (empty) href attribute on knowl links

Alex-Jordan:
Local accessibility staff were having screen reader issues with knowl links that did not identify as something the user can click. That led to this.

This PR adds `href=""` to knowl anchor elements. Without an `@href`, an anchor element does not default to having role "link", which is what the screen reader wants to identify the anchor as clickable (either that, or role "button"). Note we could put `role="link"` on this instead, but should not do both.

Things about this may be deceptive. If I visit a (current) page with Firefox and use the inspector tool, I see `href=""`. But it's something the browser is adding while building the DOM, it's not currenlty in the PTX HTML. And apparently, in some browser/screen-reader combinations, that is not good enough for the screen reader to understand the role is "link".

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -2576,6 +2576,11 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:param name="placement"/>
 
     <xsl:element name="a">
+        <!-- empty, but presence needed for accessibility -->
+        <!-- An HTML "a" without an href attribute does   -->
+        <!-- not default to role "link" and does not read -->
+        <!-- as clickable by a screen reader.             -->
+        <xsl:attribute name="href"/>
         <!-- empty, indicates content *not* in a file -->
         <xsl:attribute name="data-knowl" />
         <!-- id-ref class: content is in div referenced by id       -->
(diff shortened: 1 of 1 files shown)


davidfarmer:
I thought we had already added the empty href, for the same reason.

I am torn between the simplicity of adding the empty href,
and the fact we are using a side-effect to accomplish the intent.

It won't break knowls, if that is the question.  Or if it does
(I have not tested) I should be able to fix it.

Alex-Jordan:
I didn't see any badness with knowl behavior after building the sample article with this.

> I thought we had already added the empty href, for the same reason.

I thought so as well. I do wonder if I was misled by the browser inspector inserting an `href=""` into the DOM even though it is not in the source.

That's the same reason why I lean toward `href=""` rather than `role="link"`. Because if we end up having both (we added `role="link"` and the browser put `href=""` in the DOM) then that runs contrary to an ARIA role principle: never redundantly put a role equal to the default role.  Officially, `a[href]` has default role "link". While `a[not(href)]` does not*. It's all a bit more confusing when the source says one thing and the DOM says something else, and it's unclear at which stage the screen reader and the accessibility tree come into play.

*Also, what is the purpose of `a[not(href)]`? Googling tells me that this is for indicating that something *would* be a link under different conditions. Like this is a link to see some content, but only after you log in to the site. So it is still styled to look like a link, but it's not a link. It sounds like another reason why we really should have the `href=""` in there.

Alex-Jordan:
Force pushed. 

Unsure if all the repeat commentary is desired.

rbeezer:
More commentary is better than less.  ;-)

Rebuilding website now, give it 15 minutes.  Testing on Section 4 of the sample article does not reveal any harm to output.  But please double-check!

PR #1629, opened 2021-10-20
html: Add versioning which helps avoid cached js and css files

davidfarmer:
Small change which helps developers avoid aggressive caching of Javascript files.

Changing this line in pretext-html.xsl
`<script>miniversion=0.6</script>`
will force reloading of various dynamically added files.

Changing this line
`<script src="{$html.js.server}/js/{$html.js.version}/pretext_add_on.js?x=1"></script>`
will force reloading the main `pretext_add_on.js` file.

No doubt there are better ways, but this seems to be adequate for now.


rbeezer:
I'll merge this soon.

Brad puts a hash on the end of a URL, which serves a similar purpose.  Maybe peek at where we load RS JS, or ask Brad (if interested).

davidfarmer:
I think I misunderstood.  Bryan and I are exchanging email.

I thnk it means I will do it better in the future, but this can
be merged when convenient.

PR #1634, opened 2021-11-04
Guide: edits to table description

Alex-Jordan:
I read the new table stuff in the guide. It's good. I found a few things to fix. Of the four things here, the latter three I *think* are things to actually fix. The first thing is a little different.

### DIFF
--- a/doc/guide/author/topics.xml
+++ b/doc/guide/author/topics.xml
@@ -2802,7 +2802,7 @@
         <introduction>
             <p>A <tag>table</tag> is a container that houses a <tag>tabular</tag>, which is the actual rows and columns of table entries.</p>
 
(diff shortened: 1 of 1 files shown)


rbeezer:
Excellent!  Thanks.  Will merge tomorrow.

Didn't know about Thing One, but it doesn't surprise me.

Alex-Jordan:
I came upon thing 1 when someone used it for a local project and then later I was looking over their source. Imagine every cell having `bottom="minor"` instead of just the tabular itself. That kind of thing.

rbeezer:
Thanks again.

>  That kind of thing.

Yes, as I can imagine.  Another data point for auto-generated code.

PR #1635, opened 2021-11-07
Clean up uses of 'header' and 'heading'

Alex-Jordan:
Since PR count dropped to single digits, it is a good time for this. I tried to straighten out where "header" and "heading" are used in the repo. Roughly, "header" is appropriate for:
* Something at the start of a file, often a comment or file type identifier
* Things inside an HTML `head`
* Things inside the LaTeX preamble
* Something that is visible in the HTML banner
* Something that appears at the top of printed pages
* In a table, the cells that would be `th` in HTML
* Extend the previous to sidebysides with titles on panels
* WeBWorK  "set header" files

Maybe I missed something. Meanwhile "heading" is appropriate for:
* the visible title of a division or block
* headings in the index

Recently when I was working on hN, I found it helpful to not intermix the two terms. There was at least one instance of a "something-header" modal template that wrapped around "something-heading-content", which bugged me. This PR is either obsessive compulsive cleanup, or could make it more smooth for a future contributor working on head(er|ing)s. 

Three kinds of change here:

- commentary and docuemntation text have "header"->"heading" or vice versa
- one instance of PreTeXt source code where there is an ineffective table row `heading` attribute that is corrected to become `header`
- modes of modal templates change. This one might make you nervous. For each such change, I made sure to do a global search. For example if `mode="section-header"` needed to become `mode="section-heading"`, I did a global search for `"section-head(er|ing)"` to (a) make sure all instances of `"section-header"` were found and changed, and (b) make sure that `"section-heading"` was not already in use.

Have you ever stared at or repeated a word so many times that it starts to feel "wrong"? Like it's spelled wrong or o longer has meaning? I was starting to feel that way about "header" and "heading" at the end of this.


rbeezer:
> Since PR count dropped to single digits, it is a good time for this.

And maybe some of the others!

Yes, that was a mess.

> Have you ever stared at or repeated a word so many times that it starts to feel "wrong"?

Yes, too.  **Thanks** for the much needed clean-up.  Now I have to adapt to the correct names.  ;-)

For the record: no unexpected changes in testing HTML, LaTeX, and braille outputs on testing samples.

PR #1637, opened 2021-11-17
Guide: multiline math, paragraphs, long titles

davidfarmer:
A few items added to documentation:

-the `paragraphs` element
-more details about multiline math
-how to break long titles into lines


rbeezer:
Pushed, and will update website later.  **Thanks!**

PR #1639, opened 2021-11-19
add publisher switch to allow for continuous formatting of worksheets

oscarlevin:
The publisher switch `publication/latex/worksheet/@formatted` with value `no` allows a publisher to have worksheets formatted continuously with the surrounding sections (no new page, no difference in margins, no workspace).  

Added documentation to the guide and sample-article publication file.

### DIFF
--- a/doc/guide/publisher/publisher-file.xml
+++ b/doc/guide/publisher/publisher-file.xml
@@ -131,6 +131,15 @@
             </ul></p>
         </subsection>
 
+        <subsection xml:id="latex-worksheet-options">
+            <title><latex /> Worksheet Options</title>
+            <p>
+                By default, worksheets are formatted differently than other pages, including customizable margins, workspace between exercises, and on their own pages.  This separate formatting can be ignored, causing worksheets to be treated like any other division, using the<cd>
+                <cline>/publication/latex/worksheet/@formatted</cline>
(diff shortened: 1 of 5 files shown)


rbeezer:
Thanks, @oscarlevin!  This is a nice option to have.

I added a pile-on commit, and took responsibility for the changes.  ;-)

*  Warned if `@formatted` is not `yes` or `no`, and then set to default.
*  Produced a boolean variable in the publisher file, marginally simplifying the employment in the LaTeX conversion.

This is exposing some goofieness in the LaTeX output, so paragraphs following an exercise are starting in the margin.  Holler if this is urgent for you, and otherwise, I'm going to say worksheets are not official yet.  ;-)

rbeezer:
Forgot to say, for anybody using this.  LaTeX may take three passes to get all the tcolorbox to settle down.  So don't panic.

rbeezer:
@oscarlevin noticed that an `exercise` inside a `worksheet` (at the end of a `sidebyside` only?) did not finish cleanly, and subsequent text (e.g. a new paragraph) started in the right margin.  Not a good look.

2fec967deff480cf6e67ed54dc3e5435741a7553 adds the necessary `\par` for this situation (and many many more situations).  Seems to fix the observed problen, with no harm elsewhere.

@oscarlevin - Thanks for reporting this.  Can you keep an eye on it as you work through your project, especially any ill-effects elsewhere?  Thanks.

PR #1641, opened 2021-12-02
Bibtex-style bibliography entries

davidfarmer:
Within `<biblio type="bibtex">` there are new tags available, sufficient for
bibliography entries in a typical math paper.  It is expected that all content is
tagged, with no text elements.  Output preserves the order of the tags in the
source.

Examples added to the sample article.  Tested in HTML and LaTeX.

Update to the schema will be submitted in an issue.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -4173,8 +4173,27 @@ the xsltproc executable.
 
                 <biblio type="raw" xml:id="biblio-strang-article">Gilbert Strang, <title>The Fundamental Theorem of Linear Algebra</title>, <journal>The American Mathematical Monthly</journal> November 1993, <volume>100</volume> <number>9</number>, 848<ndash />855.</biblio>
 
+                <biblio type="bibtex" xml:id="CF">
+                <author>J. B. Conrey and D. W. Farmer</author>
+                <title>Mean values of <m>L</m>-functions and symmetry</title>
+                <journal>Internat. Math. Res. Notices</journal>
+                <number>17</number>
+                <year>2000</year>
+                <pages start="883" end="908"/>
+                </biblio>
+
                 <biblio type="raw" xml:id="biblio-beezer-fcla">Robert A. Beezer, <title>A First Course in Linear Algebra</title>, 3rd Edition, Congruent Press, 2012.
                 <note><p>An online, open-source offering.</p></note>
+                </biblio>
+
+                <biblio type="bibtex" xml:id="Dav">
+                <author>H. Davenport</author>
+                <title>Multiplicative Number Theory</title>
+                <series>GTM</series>
+                <volume>74</volume>
+                <publisher>Springer-Verlag New York, NY</publisher>
+                <year>2000</year>
+                <pages>xiv+177</pages>
                 </biblio>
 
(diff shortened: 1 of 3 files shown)


rbeezer:
Visual scan looks good.  One item to discuss.

Most templates use
```xsl
<apply-templates/>
```
 in the "middle".  Good, since we need to manage extraneous whitespace, and this will happen.  Bad, because any stray markup will get processed.  Say a "q" element (which would not be a disaster).  This is as permissive as you can be.

However, the schema proposal, at #1642, has items like:
```
BibEditor = element editor {text}
```
which says that the `editor` element has **no** markup allowed in it.  Which is about as restrictive as you can be (`empty` is moreso!).

Now, I'm guilty of lots of this from the old days, but with new elements, we can start fresh and avoid the "but it looks good" syndrome.  So the question here is your intent: what the code says or what the schema says?

Keeping the schema could mean changing the code to:
```xsl
<apply-templates select="text()"/>
```

davidfarmer:
Eventually BibEditor and BibAuthor will need markup in them,
because each contains the names of people.  To add that to the
schema now would require a way to describe such content
(something like the `TextLong` used in BibTitle).

Since we are not close to properly implementing editor and author,
for now the source should only have text for those fields.
Not sure it is good for the schema to allow something which will not work.

If `<apply-templates />` is wrong for editor, isn't also wrong
in the existing code for "number" because the number is just text:
   
```
<xsl:template match="biblio[@type='raw']/number">
    <xsl:text>no. </xsl:text>
    <xsl:apply-templates />
</xsl:template>
```
I don't object to making incremental improvements, but the code which would
be improved would still be on the chopping block when the bibliography
is done properly.

rbeezer:
> Eventually BibEditor and BibAuthor will need markup in them,
because each contains the names of people.

I thought about that, but forgot that we need things like given name, family name, suffix, etc.  You don't mean accented characters, or punctuation?  Those are all under control.  **When** we decide to do names carefully, then we can adjust the schema and code in concert.

> Not sure it is good for the schema to allow something which will not work.

Definitely not good (but I'm sure one can find instances quickly).

> the existing code for "number"

Yes, that is what I meant about code from the old days.  `type="raw"` is very, very old.

> I don't object to making incremental improvements, but the code which would
be improved would still be on the chopping block when the bibliography
is done properly.

Agreed, which is why this is the only thing I thought might be changed.  I'm trying to avoid us digging a hole for authors, where they don't ever validate, get good-looking results, and then are bummed when we catch up to them.

rbeezer:
Merged, as-is.  Thanks for this improvement.

I added a follow-on commit, c99b455f41a885a5a323f3ad22b8a3773b57cf9b, which limits the markup in each item.  To basically nothing.  But I got caught out by your (excellent) examples in the sample article.  Need to process `m` for math!  So I turned that on for `title` and `journal`.  Holler if your complete (current) writing project implicates other exceptions.

I'm going to go consider #1642 (assuming you don't want to do it!).

How about you announce this on -announce (`[NEW]`), as you see fit?  Get the balance between short-term necessity and long-term commitment right?

davidfarmer:
Thanks!  It was just random that I had math in a title.

I wasn't thinking this was -announce worthy, but I will think about it.

rbeezer:
> I wasn't thinking this was -announce worthy, but I will think about it.

Your call.  Schema is coming along, but it will also be of a short-term quality.  ;-)  Maybe complete before the day is out.

PR #1644, opened 2021-12-13
WeBWorK: Update WW examples to use pretext/pretext and managed images

Alex-Jordan:
This accomplishes two basic things.

* A project using WeBWorK and managed directories now will have the representations file and downloaded image files as part of the managed directories scheme.
* The Makefile for the sample chapter is updated to use `pretext/pretext` and managed directories. The only remaining direct use of `xsltproc` is for making the problem set archives. Additionally, there are versions of the make recipes at the bottom that do not use managed directories.

I'm sorry I was not able to keep these things clean and separate. However I did thorough testing with file diffs between the `dev` branch output, this branch using managed images, and this branch not using managed images. With .tex and .html output. With the minimal example and the sample chapter.

And I examined PDF and HTML output in all cases to verify WW exercises function, and all image files are where they need to be.

One update was the change from `$(MB)` to `$(PTX)`, so after you checkout this branch, you will need to add a definition of `$(PTX)` to `Makefile.paths`. No harm in leaving `$(MB)` there too.


rbeezer:
Real nice - a lot of documentation has disappeared and the Makefile is much simpler.  And the cost is not a lot of **additional** code (just lots of **changes**).  Great to have the representations fully managed now.

Notes:
*  Did usual testing drill with sample chapter, in large part to understand the new process (which it was good for).
*  Lots of namespace pollution, which I do not fully understand/appreciate.  Did you try adding in `pi` to the `pretext-html.xsl` stylesheet in the `stylesheet/@exclude-result-prefixes` attribute?  That might squelch many of the `xml:ns` attributes, or it might totally break everything.  ;-)
*  I added a commit with some `-vv` in the Makefile, with a comment.  I do A/B testing on `*.tex` and so I needed the name of the temporary directory where the new version of that landed (and also it is not removed).  I'll need old and new once we move forward with this PR.  Not strictly necessary for HTML, etc, but no real harm I think.  Holler if that needs adjustment.
*  Cutover my (messy) website build script to align with changes to managed directories for the sample chapter.  Harder than the review itself.  :-(
*  I'm still building the Guide and Showcase the old way (not enough patience today to cut over everything for those too, there are other PRs to look at!).
*  I edited the message about no managed directories in the publisher file to be a verbose (`-v`) message, and to say it was **WW** output going someplace.
*  Do you want to put onto -announce, with a pointer to edited instructions in the Guide?  (Updating right about now, double-check first.)  Does the `(MB) -> (PTX)` change need to be announced?  Do we think many people have their own `Makefile.paths`?

Thanks - this was a lot of work and a big improvement.

Alex-Jordan:
> Lots of namespace pollution, which I do not fully understand/appreciate. Did you try adding in pi to the pretext-html.xsl stylesheet in the stylesheet/@exclude-result-prefixes attribute? That might squelch many of the xml:ns attributes, or it might totally break everything. ;-)

I did not, but I'll put it on my list of things to do try when I have a random 15 minutes or so to try things. My file diff review workflow involves identifying things like this and sedding them away first, so they don't clutter the diff.

> I added a commit with some -vv in the Makefile, with a comment. I do A/B testing on *.tex and so I needed the name of the temporary directory where the new version of that landed (and also it is not removed). I'll need old and new once we move forward with this PR. Not strictly necessary for HTML, etc, but no real harm I think. Holler if that needs adjustment.

My version of the A/B testing was to change the `-f pdf` to `-f latex` in the Makefile. Is the "once we move forward" part now moot?

> I edited the message about no managed directories in the publisher file to be a verbose (-v) message, and to say it was WW output going someplace.

In this comment you stressed *WW* output. Should I be seeing that in 7de7218? I only see the change to verbose.

I'll make an -announce post. Might be tomorrow though.

rbeezer:
> My version of the A/B testing was to change the -f pdf to -f latex in the Makefile. Is the "once we move forward" part now moot?

Well, OK, that is a good way to go.  Don't even remember what my own code can do.  ;-)

But I'd still just as soon get all of `-vv` every time.  Every fiddle to make testing work makes me inclined to not always bother.  :-(

> In this comment you stressed WW output. Should I be seeing that in 7de7218? I only see the change to verbose.

Lost an edit.  See next commit?

> I'll make an -announce post. Might be tomorrow though.

Seen it - thanks.

rbeezer:
>>  Lots of namespace pollution, which I do not fully understand/appreciate. Did you try adding in pi to the pretext-html.xsl stylesheet in the stylesheet/@exclude-result-prefixes attribute? That might squelch many of the xml:ns attributes, or it might totally break everything. ;-)

> I did not, but I'll put it on my list of things to do try when I have a random 15 minutes or so to try things. My file diff review workflow involves identifying things like this and sedding them away first, so they don't clutter the diff.

Needed this badly for fixing up the Jupyter conversion without ill effects.  Seems to be the right thing to do (and it is done lots of other places).  I wish I understood this well enough to totally get it 100% under control.  Let's be aggressive and squash these extraneous declarations until it breaks something and is really needed.

rbeezer:
See #1652 for (beneficial) changes to HTML and EPUB namespaces.

Issue #1664, opened 2022-02-14
Refactor stacked headings


rbeezer:
For @alex-jordan.  Based on Friday's discussion, so brief.

Experiment:

1.  In `-common` at current line 2980.  Uncomment `BUG` message in `mode="is-structured-division"`.
2.  Process sample article to HTML.
3.  Figure out how stacked headings should change in order that message goes away.

Goals:

1.  Move `mode="is-specialized-division-in-unstructured"` in its own commit closer to similar templates around line 2980.  No changes, just a move.
2.  In `-common` at current line 2980.  Replace return value of `false` by `BUG` message.
3.  Rename `mode="is-specialized-division-in-unstructured"` in a positive fashion (remove 'un")
4.  Have main part of `mode="is-specialized-division-in-unstructured"` match on only specialized divisions, return true/false, BUG when called on wrong type of divisions.

I guess `mode="is-specialized-own-number"` is not exactly what is needed.  But it could be a good model for what is needed.  Thanks for looking into this one.

rbeezer:
Closed via #1667.  Thanks.

PR #1667, opened 2022-02-18
Unstructured

Alex-Jordan:
This is not to be merged. I am checking in to ask for an inspection that so far, so good. See #1664.

### DIFF
--- a/xsl/entities.ent
+++ b/xsl/entities.ent
@@ -33,8 +33,16 @@
 <!-- for match/select -->
 <!ENTITY STRUCTURAL "book|article|slideshow|letter|memo|frontmatter|part|chapter|appendix|index[index-list]|index-part|preface|acknowledgement|biography|foreword|dedication|colophon|section|subsection|subsubsection|slide|exercises|worksheet|reading-questions|solutions|references|glossary|backmatter">
 
+<!ENTITY TRADITIONAL-DIVISION "book|article|part|chapter|section|subsection|subsubsection|appendix">
+
+<!ENTITY SPECIALIZED-DIVISION "exercises|worksheet|reading-questions|references|glossary|solutions">
+
 <!-- for filtering nodes, eg *[&STRUCTURAL-FILTER;], *[not(&STRUCTURAL-FILTER;)] -->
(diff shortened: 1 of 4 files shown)


rbeezer:
Looking good.  Do you get any changes (perhaps just for `worksheet` since I've made an exception for a division full of them)?

Filters will be useful.  I think "appendix" belongs in "traditional" and "solutions" belongs in "specialized".  But I don't think `match="*[<filter>]"` is best, instead it should just be a list of the elements - so perhaps there should be lists as entities, too.  I know there would be other places I might use them if they existed.

Alex-Jordan:
No changes in output with what is here so far.

I tried a list of the elements first, but couldn't get the match priority to behave. At one point I had `match="*"` with the error, then `match="&TRADITIONAL-DIVISION;"` and then the other one, but `match="*"` was winning the precedence every time. But I could try harder to make that work.

Do you have a preference between three different templates, versus one that uses a `xsl:choose`?

Alex-Jordan:
At the moment I have it as:

```
<xsl:template match="*" mode="is-traditional-division">
    <xsl:choose>
        <xsl:when test="self::*[&TRADITIONAL-DIVISION-FILTER;]">
            <xsl:value-of select="true()"/>
        </xsl:when>
        <xsl:when test="self::*[&SPECIALIZED-DIVISION-FILTER;]">
            <xsl:value-of select="false()"/>
        </xsl:when>
        <xsl:otherwise>
            <xsl:message>PTX:BUG: asking if a non-division (<xsl:value-of select="local-name(.)"/>) is a traditional division</xsl:message>
        </xsl:otherwise>
    </xsl:choose>
</xsl:template>
```

Any red flags there? Or opportunities to simplify?

rbeezer:
I think self::*[ ] is redundant.  Just set the test= to the entity?

Alex-Jordan:
OK, ready for review. I cannot come up with a good way to remove the "un". So I've just replaced all those instances with a logical combination of two more basic templates. So like:
```
<xsl:when test="$is-specialized-division-in-unstructured = 'true'">
```
becomes

```
<xsl:when test="$is-specialized-division = 'true' and $is-child-of-structured = 'false'">
```

rbeezer:
Very good!  Thanks for chasing through this one.  Yes, much more verbose, but it is complicated and I think this makes it clearer when the code needs maintenance.

Exploded one commit into three, with you as author, fixing one typo along the way.  Used a merge commit to label the branch of four commits with the GH PR number.  So basically zero change to the code.

Thanks again.  I'll be using parts of this in the next few minutes.  ;-)

PR #1668, opened 2022-02-19
HTML: replace p with br for stacked headings

Alex-Jordan:
In a stacked heading, we were wrapping each line with a `p`. The intent was to give a screen readers a clear signal to pause between lines. But that makes invalid HTML and invalid EPUB. This removes the `p` and drops a `br` where needed instead.


rbeezer:
Beautiful.  Output looks great.

Does @davidfarmer  need to know that `p.assistive` (not `a.assistive`) is now obsolete?

rbeezer:
Forgot to say: website update may be a bit later, got a bug to fix.

Alex-Jordan:
Yes @davidfarmer, `.assistive` reverts now to only affecting the skip to main content link. Scrub it from CSS if you like.

davidfarmer:
I removed the .para.assistive from the CSS

PR #1671, opened 2022-02-24
EPUB: put authors on their own line on LaTeX-based cover image

Alex-Jordan:
This does two things.

1. Put authors on their own line on LaTeX-based cover image. This will make it much less likely that names run off into the margin.
2. Remove the page number "1" that was previously appearing on this cover page image.


rbeezer:
Done.  Thanks for the incremental improvement.

PR #1675, opened 2022-03-04
Correct some error messages

Alex-Jordan:
I think that these error messages need this edit.


rbeezer:
Thanks very much!  (That was quick!)

PR #1679, opened 2022-03-09
HTML: classes signifying special content in knowls

Alex-Jordan:
This stems from a discussion at last week's drop-in. The background is that I find people (students and faculty) are unaware that many APEX solutions have @sean-fitzpatrick 's videos in them. For example, in the solution here:

Example 7.3.5: Finding volume using the Shell Method 
https://opentext.uleth.ca/apex-video/sec_shell_method.html#ex_shell1

They are within born-hidden knowls, and some people are just skimming through. People who would otherwise stop to watch a video pass right over it. 

This puts classes like `has-video` on the clickable for a born-hidden knowl. With these classes, I can experiment with ways to let users know (discreetly, but not too discreetly) that there is something "special" inside. The first thing I will try is to have an icon follow the knowl clickable, like a little movie reel if there is a video, a little landscape if there is an image, etc. All with CSS.


davidfarmer:
There is a little movie camera after the section title here:

https://sl2x.aimath.org/book/aafmt/section2-2.html

In case you want a suggestion for a camera icon.

rbeezer:
I like it.  I love sniffing around like this.

Will merge when I get a chance.

Opinion: I think this will work well for activities, such as video and interactive.  I'm not yet sold on the value for image and tabular.

rbeezer:
Can you put the additions to the sample article on a first commit, and the new XSL on a second one?  I can't test when they are all together and it takes too long to unravel **and** continue to give you credit.

Alex-Jordan:
For some reason I have stopped getting alerts on PRs. Maybe since the move to PreTeXtBook? Just seeing this (and other posts) now.

Alex-Jordan:
>  I'm not yet sold on the value for image and tabular.

One possibility I was thinking is that a publisher specifies which things should pop out. The default could be videos and interactives (and "computation" things like Sage?). But the machinery could be there for, say, an accounting book that wanted to make it clear where the tabulars are hidden.

Alex-Jordan:
> Can you put the additions to the sample article on a first commit, and the new XSL on a second one? 

It's on my radar now that I found this trove of unread GitHub alerts. Here I thought you'd just been too busy to look at my PRs over the last few months.

rbeezer:
> Here I thought you'd just been too busy to look at my PRs over the last few months.

And I thought it was **you** who was too busy!

Alex-Jordan:
I just force pushed this branch where the previous single commit is now two commits: additions to the sample article, then changes to the XSL.

rbeezer:
Thanks for splitting this up, that made testing easier.

I had it backwards - put sample article changes on the **last** commit.  That way I can test the sample article without source changes, by just switching to the prior commit.  Learned something here - I did a **temporary** `git revert` to recover from my mistake.  ;-)

I'll likely update the website version sometime later today.

PR #1681, opened 2022-03-11
Update it-IT.xml

valelav:
Translations of proof-like elements and of 'Fragment'


rbeezer:
Thanks very much, Valerio!  (Watch the -dev discussion, there may be more!  :-( )

PR #1683, opened 2022-03-11
Update URLs in guide to use @visual

sean-fitzpatrick:
While adding an appendix to the guide on transitioning an existing project to the CLI, I noticed (while testing) that most of the URLs in the guide were deprecated because they did not have a `@visual` attribute.

This pull request adds that attribute throughout the guide.

Some notes:

- unfortunately, there are also some unintended whitespace changes added by my editor. Harder to avoid when working across many files
- where there was a hard-coded visual URL added as a `fn`, I've removed the footnote, since it's now redundant.
- one place where I was a bit unsure how to proceed was in the instructions for adding URLs themselves. There are a few places that talk about adding an `example.com` URL. Early in the document, the now deprecated syntax is used. Later in the same section the use of `@visual` is explained. In the first instance, I added a much shorter explanation of the `@visual` attribute.
- the `href` to the Berne Convention (in the section on licensing) was empty. I added what I think is a suitable URL.


rbeezer:
Thanks!  I'll likely get to it (and the other one!) tomorrow.

rbeezer:
Thanks!  Whitespace changes are beneficial, it seems.  Better to have them on their own commit, but not worth the trouble here.  Should not have been there in the first place.

Appreciate the assist with this.  Every deprecation causes me a long long list of new chores.

Consolidated commits and edited commit message.  Otherwise, as-is.

PR #1689, opened 2022-03-22
Simplify extraction-to-file-to-Python pipeline

StevenClontz:


rbeezer:
Works well, tested all three on the sample article.  Except the YouTube (thumbnail) for Led Zeppelin is gone AWOL.  Damn.

We need more graeceful failure for that...  #1690

**THANKS** for seeing that optimization.  Sorta naive on my part!

StevenClontz:
Something to consider at some point: if this extraction xsl is only ever used by the Python scripts, perhaps the script could query the data needed directly with lxml. (Or at least, do the transform in memory to avoid writing to disk and reading back the file.)

rbeezer:
> the script could query the data needed directly with lxml

**Yes**!  See one subtlety on #1691

PR #1693, opened 2022-03-24
Showcase: establish CLI setup

Alex-Jordan:
This sets up the Showcase Article to run with the CLI. Should work with `pretext build html` and `pretext build pdf` with no fussing over images. All image formats are committed in `generated`. Note however that the CLI is not yet capable of making all of these formats, so I used `pretext/pretext` to make most of the archival asymptote formats.

Another thing the CLI is not yet doing is copying the archival formats to the appropriate folder in `output/html`. So the little links to the archival formats do not work, unless you manually copy those image files.

Also this cleans up a few schema violations and runtime warnings.

And this comments out the pstricks image from the images section. That only works with TeXLive 2019 (not earlier, not later).

For this one move of image files to a new folder, I let the image files be rebuilt. So most (all?) have nontrivial diffs. But in general we won't clutter things up with such diffs, and I put a note about that in the README for contributors.

@rbeezer You will have to change your big build script. Either move to the CLI for the Showcase, or adjust what you have now to use the new locations of the various files.


Alex-Jordan:
Force pushed to add eps versions of the asymptote images. I hadn't realized that was an output option before.

rbeezer:
Thanks, Alex.  It'll require changes to the build script, as you say, so I'll have a go at this when I'm prepared for all that.  ;-)

Alex-Jordan:
Note: I think don't switch the CLI for this. I'm keeping in mind how the CLI lags a little behind this repo.

rbeezer:
Thanks very much!  I was tearing my hair out to get the ww representations file right in my build script.  Modernizing and not thinking about it was the solution.  ;-)  Just need to fix two simple WW in the Guide now (later).

* Will rebuild website later today.
* Using `pretext/pretext`.
* Added WW server spec into publication file.
* Watermarks temporarily gone, they don't do well from my build script to the pretext script.

Unclear why GH doesn't see your commit as a merge?

Issue #1687, opened 2022-03-18
"all" option for asymptote


Alex-Jordan:
When generating latex-image image files, there is an `all` option for the `-f` switch, so you can make all formats (tex, pdf, svg, png, eps) in one go.

For asymptote there is not currently such an option. It would be convenient to have this, and make all that we currently make in one go (asy, html, svg, png, pdf).

sean-fitzpatrick:
I'm not sure there is a built in option to do this in Asymptote. You might need to program this as multiple runs; one for each file format. (For APEX this would probably take a couple of hours.)

sean-fitzpatrick:
I've posted on the Asymptote forum to see what I can find out. As an experiment, I tried running with `-f all` and it turns out there is a file format called "all" that it supports!!

PR #1694, opened 2022-03-24
Script: 'all' format option for generated Asymptote images

Alex-Jordan:
This addresses #1687. Perhaps inefficiently, but it works.


rbeezer:
> Perhaps inefficiently

It looks very much like what I would have done.  ;-)  Thanks for the improvement!

PR #1696, opened 2022-03-29
WeBWorK: document using pg-code for statement

Alex-Jordan:
This documents a feature we have always had*, but I only realized it recently. Formerly we thought of the `pg-code` element as purely for setup code. But there is no reason it cannot include the statement, hint, solution with these things coded using PG directly.

Why would we want this? There is a barrier to using WeBWorK in more books. People can't simply copy-paste more of their PG (or PG taken from online references) into the PTX source. Having to translate some existing PG over to PTX statement, hint, solution can be a burden in some cases. (Even while in other cases, coding it in PTX is more pleasant.)

I acknowledge that this will allow authors to slip features into WW exercises for which we do not have a PTX counterpart. But no more so than with `.pg` problem files that live on the server, which we mostly support. And then we examine the static PTX that comes back looking for bad XML/PTX. Also the WW server drops comments that such and such a feature (eg horizontal rule) does not even have a PTX counterpart. 

*Well, it wasn't really there until with this commit, that adds `webwork[pg-code]` to some template matches.

### DIFF
--- a/doc/guide/author/webwork.xml
+++ b/doc/guide/author/webwork.xml
@@ -65,7 +65,7 @@
     </introduction>
 
     <subsection>
-      <title>Using an Existing <webwork/> Problem</title>
+      <title>Using a <webwork/> Problem from the Host Course</title>
       <p>
         If a problem already exists and is accessible from the hosting course's <c>templates/</c> folder,
         then you can simply include it as a <attr>source</attr> attribute.
@@ -162,7 +162,7 @@
         requires using a <tag>pg-code</tag> tag.
         Having at least a little familiarity with coding problems in <webwork/> is necessary,
         although for simpler problems you could get away with mimicking the sample article in <c>mathbook/examples/webwork/</c>.
-        A <tag>statement</tag>, <em>optional</em> <tag>hint</tag>, and <em>optional</em> <tag>solution</tag> follow.
+        An <em>optional</em> <tag>statement</tag>, <em>optional</em> <tag>hint</tag>, and <em>optional</em> <tag>solution</tag> follow.
       </p>
       <pre>
         <![CDATA[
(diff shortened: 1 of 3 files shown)


rbeezer:
I've been giving this one a lot of thought.  My goal with incorporating WW problems was to have as much as possible of the problem formulation in PreTeXt.  Ideally broad classes of "simple" problems would not need any PG.  And would therefore be translate-able to other systems as well.  I think some of this is going to happen as we incorporate more Runestone-flavored exercises.

And generally, I have never been sympathetic to "this markup is too much trouble."  So I do not want to encourage dropping in an entire problem as raw PG.  We should be encouraging proper use of markup (as a philosophical underpinning of PreTeXt), not encouraging avoidance.

So while the code changes here are minimal (this is not purely a documentation commit), I view the nature of the matches (`statement|task|stage|@source` and `statement|task|stage`) as being critical as-is.

Would some sort of online converter be a service that would remove the barrier?  Perhaps it would be a better place to warn of "bad XML/PTX"?  We would have most of the tools for this.  Could even be a routine in `pretext/pretext`?

Alex-Jordan:
I forget if we chatted about this during a drop-in session once upon a time. A converter would be a big project and after some thought it would be beyond my skills to get it to work for two large classes of OPL problems: those written without PGML, and those written with PGML that use the `[@...@]` construct. I haven't surveyed PGML thinking about this, but I wouldn't be surprised if there are other things too.

If this change is not made, as an author I can still accomplish the same thing by putting the entire problem statement in the `pg-code` and then have an empty (or trivial) PTX `statement` follow. This change was in part to bring that potential back alley practice out into the light. 

> And generally, I have never been sympathetic to "this markup is too much trouble."

Yes, but are there nuances to that? It is one thing when you are writing the markup from scratch, where I completely agree. Another flavor is when you have a huge library not using that markup, and it is a labor-intensive manual process to convert. (Hence your suggestion for a converter, but I can't see it happening in a way that covers a reasonable proportion of PG in the wild.)

> Ideally broad classes of "simple" problems would not need any PG.

We do have that for when there is no randomization:
Subsection 7.1.2: Perl-free Problems 
https://pretextbook.org/doc/guide/html/webwork-source.html#subsection-104
And then as soon as you and I can continue your multiple choice work over to WW output.

If I can't sway you, no problem. In places ORCCA edition 3 may end up with that "back alley" practice. Maybe by edition 4 everything will be manually converted.

Alex-Jordan:
Closing this. But if you are more open to it now, holler at me. I do foresee that I'll be using the back alley with upcoming ORCCA revisions. All at my own risk of course.

PR #1698, opened 2022-03-31
Comment on the importance of <initialism> for Tikz in the WeBWorK sample-chapter.xml

enriqueacosta:
Adding a comment about the importance of the <initialism> command for Tikz figure generation.


rbeezer:
Thanks, @enriqueacosta, that's a great addition.  I edited the commit message (as I do), so you'll need to delete your copies of the original branch.

PR #1702, opened 2022-04-16
remove old alert-warning classes

bnmnetp:
This removes old class names from the runestone component top level container.

Points the "web" build at a test build that has updated javascript and css for a new look that does not style components using the old alert-warning class for component backgrounds.

### DIFF
--- a/xsl/pretext-runestone.xsl
+++ b/xsl/pretext-runestone.xsl
@@ -542,7 +542,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:variable name="the-id">
         <xsl:apply-templates select="parent::exercise" mode="html-id"/>
     </xsl:variable>
-    <div class="runestone alert alert-warning">
+    <div class="runestone ">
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks!  I rebased, re-titled, and tagged the commits (as is my usual practice).  But content/changes is unchanged.  Works great on my end.  And looks great.

Website sample is going live right now, give it about 15 minutes for my script to finish everything.

Question:  are the spaces in `@class` necessary?  I left them, but could delete them next time I'm in the neighborhood.

rbeezer:
Forgot to say: if you want more eyes on this or more testing of some sort, put a plea onto `pretext-dev`.  Pinging @davidfarmer here, since he should know at a minimum.

davidfarmer:
There is no reason to have the space at the end of `@class`, but it does not hurt and can
just be cleaned up along with some other pull request.

bnmnetp:
The space was just the result of me doing a global search and replace to get rid of "alert alert-warning"

PR #1704, opened 2022-04-23
Guide: missing slash in closing tag of idx in section 22.4

valelav:
Not a big issue, but while reading section 22.4 of the guide about index entries, I noticed that there are two instances of idx without the slash in the closing tag


rbeezer:
Thanks, @valelav!  Merged, and will  appear online next time I do a website update (probably by the end of Monday).

PR #1709, opened 2022-04-30
WeBWorK: remove p whose only child is a fillin when first writing sta…

Alex-Jordan:
…tic representation

As discussed in drop-in today. Right before writing WW representations to disk, eliminate any p whose only child is a fillin. Except don't do that when the p is inside an li. Except do do that if there is a preceding sibling.

And then we can remove the expensive templates in -latex and -html.

Only diff in representations file is removal of those p, as expected.

Sample article goes from 28905 ms down to 13949 ms for building .tex!

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -1044,8 +1044,8 @@ def webwork_to_xml(xml_source, pub_file, stringparams, abort_early, server_param
                 os.remove(os.path.join(ww_images_dir, ptx_image_filename))
 
         # Start appending XML children
-        # Use "webwork-reps" as parent tag for the various representations of a problem
         response_root = ET.fromstring(response_text)
+        # Use "webwork-reps" as parent tag for the various representations of a problem
         webwork_reps = ET.SubElement(webwork_representations,'webwork-reps')
         webwork_reps.set('version',ww_reps_version)
         webwork_reps.set("{%s}id" % (XML),'extracted-' + problem)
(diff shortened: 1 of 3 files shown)


rbeezer:
>  Sample article goes from 28905 ms down to 13949 ms for building .tex!

Wow!  (Let's see who notices?)

rbeezer:
AATA:

Old, LaTeX: 395 sec
New, LaTeX: 343 sec

Old, HTML: 186 sec
New, HTML: 106 sec

rbeezer:
Thanks!  I feel much better.  I predict Sean F will be the first to notice.  Hopefully not a self-fulfilling prophecy.

sean-fitzpatrick:
Are you suggesting I should do an APEX build before and after pulling to update my PreTeXt to this commit?

(Confession: it has been a month or so since my last pull, since I've been playing with the CLI.)

One more tough week to get through and then I can start playing around with this stuff again

rbeezer:
> Are you suggesting I should do an APEX build before and after pulling to update 
> my PreTeXt to this commit?

You might see up to about a 50% improvement in run time.  No need to do it now, unless you are curious.  But pay attention next time and be delighted.  ;-)

> (Confession: it has been a month or so since my last pull, since I've been 
> playing with the CLI.)

That does sound pretty much like an appeal for an absolution of sins.  ;-)

Alex-Jordan:
This may be clear, but just in case: the speed improvement won't be there yet with the CLI. Only if you use `pretext/pretext/pretext` (or, I guess, `xsltproc`).

rbeezer:
Thanks, I keep forgetting that!

PR #1712, opened 2022-05-03
LaTeX: make fillin only contract up to 20%

Alex-Jordan:
Following from discussion at #1710, this is a smaller change such that fillin can now only compress down to 80% of intended width.

Later #1710 will be replaced with something that is a more significant overhaul that I have on a local branch.


rbeezer:
Thanks!  Should get to it very soon.

rbeezer:
Thanks for getting the original shortcoming spruced up.  Lets continue on #1710.

PR #1714, opened 2022-05-04
WeBWorK: declare server and credentials using publication file

Alex-Jordan:
Makes it so that the WeBWorK server and corresponding credentials can be (should be) handled using the publication file.

If the old way (`-s`) is still used along with a publication file, there is a warning that the `-s` argument will be ignored.

If the old way is used without a publication file, there is a suggestion to start using a publication file.

This is accomplished by `extract-pg.xsl`, which has access to the publisher variables, writing out the server parameters into a python dictionary. Later in the python routine, these are used (when present) to identify the WW server. When absent, it relies on the `-s` argument.

Documentation in the publisher guide (webwork section) is updated.

### DIFF
--- a/doc/guide/publisher/webwork.xml
+++ b/doc/guide/publisher/webwork.xml
@@ -213,7 +213,7 @@
       </p>
       <console>
         <prompt>$ </prompt>
-        <input>pretext -c webwork -p &lt;publisher&gt; -s &lt;server&gt; aota.ptx</input>
+        <input>pretext -c webwork -p &lt;publisher&gt; aota.ptx</input>
       </console>
       <p>
         <c>aota.ptx</c> in the example is the root file for your <pretext/> project.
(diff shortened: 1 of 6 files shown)


Alex-Jordan:
The "run black" commit appears to be responsible for this having conflicts. Is it easy for you to resolve?

This PR would be nice to have merged and all the way through to the CLI by Ann Arbor.

rbeezer:
**responsible** is a strong word.  ;-)

Sorry, I'd lost track of this one owing to travel.  I'd have merged it before requesting the black run.

Looks great and testing on `mini` example went smoothly.

Each new publisher variable needs a **very terse** specification in the _Publication File Reference_, with cross-references back-and-forth to more expository explanation (which is here already).  No exceptions, we're not incurring that sort of technical debt again.  ;-)  Should be clear what to do by looking at what's there.

I've manually fixed-up this PR.  So please put the rest of the documentation on a new PR, it won't do well here.  I'll then move fast on my end, in hopes that this will be in the CLI for Ann Arbor.

Alex-Jordan:
OK, I made #1723. It doesn't say much , but the references are there.

I wonder if we will be asking @StevenClontz and @oscarlevin to frequently release a new CLI version during Ann Arbor and how much of a pain/distraction that might be for them.

StevenClontz:
I think we need a serious discussion on versioning and deployment in general. My vision for CLI once we hit 1.0 is to adhere to [semantic versioning](https://semver.org/) `Major.Minor.Bugfix`: always update for Bugfixes, consider Minor updates, and know that a Major update might require some updating of your source. But that would require more coordination with core development.

I'll also feel more comfortable with frequent releases if we have automated tests to ensure nothing backwards-incompatible was introduced and no major bugs have popped up.

Alex-Jordan:
During Ann Arbor, we may develop things in PreTeXt core and need to see them in action with the CLI quickly to make sure things are working for Runestone, etc. So yes to what you bring up, but also what do you think about quasi-releases that week that people can manually install? How much of your time does it take to make something like a new unnumbered version?

rbeezer:
Shouldn't this discussion be somewhere else?  ;-)

rbeezer:
Thanks for finishing this up on #1723.  I took the opportunity to break into 3 topical commits.   Live, but will need to wait on website update for a few hours.

rbeezer:
Forgot to say: does this need a `pretext-announce` post?  Doesn't need to be long, and can point to the documentation for details.  Do mention if anything is urgent or backward-incompatible.

Updating website right now - give it about 15 minutes.

Alex-Jordan:
I don't feel this warrants a post to -announce. The only situation that might be wanted is if someone was previously using `-s` *and* they had their publication file set up to use the mechanism that wasn't enabled until this PR. But in that case, we issue a warning at run time.

Alex-Jordan:
Er, maybe you mean a post to advertise this new way of doing things. Hmm, well we still alert people with run time suggestions that they move to using the publication file. Is that enough?

rbeezer:
The price of rolling releases is announcements of surprises.  Any new warning when nothing has been changed on the authors end deserves an announcement, I think.

Alex-Jordan:
OK, I just sent something to -announce.

rbeezer:
Perfect!  Thanks.

Issue #1700, opened 2022-04-11
Localizations in multi-lingual documents


rbeezer:
Requested by  @mdoob  at Drop-In on April 1, 2022.  I hope it was not April Fools, since I thought an issue was being created and I can't find it.

Interpret an interior `@xml:lang`  to request a different set of localizations for the contained subtree (until some other subtree intervenes).

rbeezer:
Good progress on this - lots of under-the-hood work at 2f5e167e86c2035e24e644e55405d9c2cf81581b and six following commits that should make the dependence on an intermediate language fairly easy to implement.  As "no change" refactoring, I'm going to let this sit for a bit to make sure nothing unexpected happens.  And the LaTeX conversion needs some tedious editing.

mdoob:
Thank you Rob. It's really quite à propos for my current work. 

I am working a clone of pretext/master on my home machine, 
and the update from today included the files like `fr-CA.xml`.
Using
`<section xml:id="x1980-E" xml:lang="fr-CA">`
had no effect. Should I be pulling a different branch? Like 2fe167?

Another thing: I tried to use an xml:id like
`<section xml:id="x1980-français" xml:lang="fr-CA">` 
and the "c with a cidilla" character threw an error. This may be a consequence of the description above.

rbeezer:
> had no effect.

> should make the dependence on an intermediate language fairly easy to implement.

As in, it is not implemented yet.  Just check that nothing is broken.

> character threw an error.

Correct behavior.

mdoob:
Nothing is broken, so it's a good start. Will stand by.

rbeezer:
See #1446

rbeezer:
Closing as implemented. Localizations now follow the closest interior `@xml:lang`: the localization language is taken from the nearest `@xml:lang` indication ("Get languages for localizations from closest xml:lang indication", `33aed4a8`; interactives' pages adjusted in `38a643e1`), so a multi-lingual document gets the right localized strings per subtree. (Unrelated to the Czech-locale update that GitHub had cross-linked here.)

*Claude Opus 4.8 (1M context), acting as a review assistant for Rob Beezer*

PR #1715, opened 2022-05-05
A couple of new translations

jirilebl:
Translate the new messages


rbeezer:
Thanks, @jirilebl!  Appreciate you keeping this up-to-date.

Use of these will soon become more flexible, see #1700.

PR #1718, opened 2022-05-17
Guide: Correct publication file element name for project component visibility

mitchkeller:
The section of the Guide on using the publication file to control exercise component visibility has a bit of a contradictory statement regarding using `project` or `exercise-project` as the element name in the publication file. Just using `project` as the tabular calls for at present results in `exploration` not hiding all the things that are set to `no`, even though `project` appears to do what is expected. By using `exercise-project` in the publication file, this behavior is unified across `project` and `exploration`. It's possible that this PR is in error and something should be changed in the processing of the publication file, but this seems to get things into the state that is described before the tabular.


rbeezer:
Thanks, @mitchkeller !  You got it right.  It'll go live on next website rebuild, perhaps sometime tomorrow.

PR #1720, opened 2022-05-19
New: Run transform in a thread

bnmnetp:
This allows the code to consume error messages as they are created rather
than waiting until the transform is complete.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -2183,7 +2183,7 @@ def pdf(xml, pub_file, stringparams, extra_xsl, out_file, dest_dir, method):
 #################
 
 # Pythonic replacement for xsltproc executable
-def xsltproc(xsl, xml, result, output_dir=None, stringparams={}):
+def xsltproc(xsl, xml, result, output_dir=None, stringparams={}, outputfn=print):
     """
     Apply an XSL stylesheet to an XML source, with control over location of results.
 
@@ -2195,6 +2195,8 @@ def xsltproc(xsl, xml, result, output_dir=None, stringparams={}):
     output_dir   - a directory for exsl:document() templates to write to
     stringparams - a dictionary of option/value string:string pairs to
                    pass to  xsl:param  elements of the stylesheet
+    outputfn     - a function for routing output of error messages. Any
+                   such function should process its parameters like print
 
     N.B. The value of a "publisher" string parameter passed in the
     "stringparams" argument must be a complete path, since a relative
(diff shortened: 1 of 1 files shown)


rbeezer:
Finally, some peace, so I could look at this.

Looking good.  Two suggested changes (which I can meld in):

1) `error_log` is a sequence of `LogEntry` (or similar) and we just grab the `message` part.  So the line with all the action becomes

```python
outputfn('    * ', xslt.error_log[line].message)
```
so we don't get prefixed clutter like:

```
<string>:0:0:ERROR:XSLT:ERR_OK: PTX:DEPRECATE: (2022-01-05) a "url" ...
```

2) Much discussion, long ago, ended with the decision to simply import modules, and access their functions by prefixing the function call.  In other words, we would go

```python
import threading
parse_t = threading.Thread(...)
```
I hope that it is not religion.  Just house style.

bnmnetp:
OK, done.

I'm curious about the import decision.  That is definitely not Python's (PEP8) recommended style of doing imports. Further,  I would advocate strongly against having an import inside a loop even if that is the first use, or even inside a frequently called function.

rbeezer:
> I'm curious about the import decision.

I didn't mean to suggest they be adjacent, or burired, or repeated.  Just the conventions on naming.

I believe it all started here, with @davidfarmer :

https://groups.google.com/g/pretext-dev/c/QXi1IK9T7Vg/m/EFvCavqcBQAJ

rbeezer:
I've already got the two changes on a follow-up commit.

And all that clean-up is great.  But your changes and my changes are going to clash.  I can fool git, but don't really want to.

The clean-up is automated, no?  (Black?)  Can we do that **later** on a new PR?  Discuss this afternoon?

davidfarmer:
I have skimmed through the code and I agree that it seems odd to have all those imports
inside every function.  So, consider putting those imports at a higher level.

bnmnetp:
I removed the black formatting, but then had to do a force push, so you may not like that.

You seem to have your own process for picking commits from a PR anyway so maybe it doesn't matter.

rbeezer:
> I removed the black formatting, but then had to do a force push, so you may not like that.

Love it.

rbeezer:
OK, really done now.  All in one commit with your name on it.  I restored a preliminary line that explained (and bracketed) the messages that were spewing out.  Thanks very much - I've got little experience with threads in Python.

Can you do a massive (whitespace) PR using `black`?  Yes, I could figure out how to run `black` myself.  But there's lots to fix right now...

StevenClontz:
I have a theory that a recent issue I've noticed recently is related to this enhancement. Let's say a build is interrupted for some reason. In the CLI, we have to change the working directory during a build to the desired output folder because XSL just spits out build files in the current working directory. But if we interupt the build, and our working directory is cleaned up to be wherever it was originally, what if this thread isn't killed correctly? Then that'd explain what I've noticed recently: interrupted builds are being leaked into the project's root, rather than the intended build directory.

bnmnetp:
Hmmm, I am not sure about that.  the thread is part of the same process and has the same working directory and file handles as the parent...  Are you doing any signal processing in the cli to catch a ctrl-c and then doing something?

rbeezer:
See #1747 for a fix.

PR #1722, opened 2022-05-21
Run black on all python files

bnmnetp:


rbeezer:
Thanks!  I'll need to get organized and do that myself from now on...

bnmnetp:
Sure,

If you are into hooks I have a script that will run it locally as a pre-commit hook, or I set it up in my editor to run on save.

But for you it is really as easy as `cd pretext` followed by `black pretext *.py`

PR #1725, opened 2022-05-26
WWSC: Makefile .tex recipe

Alex-Jordan:
This gives

`make mini-latex`

`make sample-chapter-latex`

which will drop .tex files in the scratch PDF folder.


rbeezer:
Thanks!  That'll be a big help.

rbeezer:
Dropping LaTeX into the `pdf` directory means it gets removed if you also do a PDF run.  For a before/after test, I run all the targets, change scratch directory name, adjust code, then rerun all the targets.  Then I can do a bunch of diffs.

Makefile adjusted at 71748e3f0adebe316824039601ca2fe2da9978fd

PR #1727, opened 2022-05-30
Fillin overhaul

Alex-Jordan:
This removes `latex.fillin.style` (so there will need to be an -announce message) just for that.

It puts `publication/common/fillin/@textstyle` and `@mathstyle` in place in publication files. Each may be `underline` (default for text), `box`, or `shade` (default for math).

With shade becoming default for math (an improvement, I think), that is another reason for an -announce message.

The text fillin can compress in LaTeX up to 20% to help with line fitting. The text fillin in HTML puts classes in place, but a separate CSS PR is needed to actually make things happen there.

The math fillin templates are all in -common. In the end I found a way to make that work either way.

Text fillin uses `@characters` with default 10. Math fillin can use `@characters` or like `@fill="\frac{x+1}{2}"` with default `@fill="XXX"`. fill trumps characters. 

This does not improve accessibility for fillins. It's already OK for fillins in text. And things are not good for fillins in math and it stays that way with these commits. But I have a lead on @pkra's aria-label MathJax extension that might work out separately, later. That extension is not playing will with `\phantom` right now.

Lots of doc cutting and editing for this. I tried to pay lots of attention, but a second set of eyes on that would be good.

Not sure how you want to handle schema here. Now two types of fillin, one in math and one not?

### DIFF
--- a/doc/guide/author/topics.xml
+++ b/doc/guide/author/topics.xml
@@ -760,7 +760,7 @@
 
             <paragraphs>
                 <title><tage>fillin</tage>, <fillin characters="5" />, fill-in blank</title>
-                <p>A <q>fill in the blank</q> blank.  While meant for use with normal text, it may also be used within mathematics contexts.  The <attr>characters</attr> attribute may be used to hint at how long the line will be, should the default value of 10 be too long or too short.  Here we have set <attr>characters</attr> to the value <c>5</c>.</p>
+                <p>A <q>fill in the blank</q> blank.  May be used in normal text or within mathematics contexts.  The <attr>characters</attr> attribute may be used to hint at how long the line will be.  Here we have set <attr>characters</attr> to the value <c>5</c>. If used inside math, a <attr>fill</attr> attribute (in lieu of <attr>characters</attr>) may be some string of math that will be used to determine width, height, and depth of the blank. In text, the default value of <c>10</c> is used for <attr>characters</attr>. In math, the default value of <c>XXX</c> is used for <attr>fill</attr>.</p>
             </paragraphs>
 
             <paragraphs>
@@ -1514,7 +1514,7 @@ displayed line, and there are no <c>\\</c>s.  Use <c>\amp</c> to mark the alignm
         <subsection>
             <title>Fill-In Blanks in Mathematics</title>
 
-            <p>The other mix-in <init>XML</init> element is <tag>fillin</tag> with an optional <attr>characters</attr> attribute that takes an integer value.  You will get a thin horizontal line, on the baseline, which can suggest to a reader that they should supply something within the surrounding mathematics.  The attribute suggests the length of the line<mdash />experiment a bit, since it is not super-precise.</p>
+            <p>The other mix-in <init>XML</init> element is <tag>fillin</tag>. It may use a <attr>fill</attr> attribute with some math string to determine width, height, and depth of the blank. Or it may use a <attr>characters</attr> attribute that takes an integer value to hint at the width.</p>
         </subsection>
 
         <subsection>
--- a/doc/guide/publisher/pdf-print.xml
+++ b/doc/guide/publisher/pdf-print.xml
@@ -126,12 +126,6 @@
         <p>This is meant to help you create a professional electronic <init>PDF</init>.  A print-on-demand service (<xref ref="print-on-demand"/>) will likely want a standalone image (possibly with the front and back, plus a spine, all rolled into one <term>wrap</term> image).  So build your <em>real</em> cover images (<xref ref="cover-design"/>), and then modify them for this use.</p>
     </section>
 
-    <section>
-        <title>Fill-Ins</title>
-
-        <p>The <tag>fillin</tag> element can be rendered as a line near the baseline of the current line of text, or as a complete rectangle (<q>box</q>).  Set <c>latex.fillin.style</c> to the values <c>underline</c> or <c>box</c>.</p>
-    </section>
-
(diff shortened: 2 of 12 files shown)


Alex-Jordan:
This replaces #1710 .

rbeezer:
A lot of work here!  ;-)

Comments based on visual review.  Feel free to pile on commits, or force-push, neither is a problem.

* `DOCUMENT-FILTER` is a nice thing, but I don't see it being used anywhere?
* I'm on a campaign to move out as many warnings and errors as possible when the schema (or validation-plus) can determine the problem.  For example, to address your shema question, I expect to add a `FillinText` and `FillinMath` pattern, and these will keep the roles of `@characters` and `@fill` straight.  There might even be something like an `xs:number` type that will keep `@characters` under control.  Can you review messages with an eye to not duplicating the role of schema validation?
* Publisher variables should do its best to respect options selected with old string parameter.  In other words, we should do our best to not abandon selections that we have replacements for.  Should be many examples, test for this situation first.
* When you respect the string parameter, you may need/want to unwind the `otherwise/if` construction on the end of the two publisher variables.  I **think** I like a final `otherwise` that guarantees the default value, even if it may not logically be necessary.

Alex-Jordan:
> DOCUMENT-FILTER

Oh! I had that for something that was cut once I found a way to do `shade` in math that worked for both LaTeX and MathJax. Can cut with the next force push.

The new publisher variables here apply to all output forms. But the old stringparam latex.fillin.style is only for LaTeX. So I don't see a way to make the new publisher variables respect the old stringparam.

The only way I see to continue to respect the old stringparam is to do things with it inside the LaTeX stylesheet. I will do that if you'd like. The change here would affect anyone who uses `latex.fillin.style="box"`. Maybe just my opinion, but moving to the text default of `underline` and math default of `shade` feels like improvements in both places. I was thnking it's like an opportunity to really kill an old stringparam without anyone complaining.

In the next force push I will modularize the latex macro definition so that the beamer template just calls it instead of repeating that code.

rbeezer:
> The new publisher variables here apply to all output forms.

OK, I see.  Make that distinction clear then for `-announce`.  "LaTeX output will revert to the default value which is 'underline'. An election for "box" needs to use the new publisher variable."  Or something along those lines, I'm not checking that what I just said is correct.

And maybe review the deprecation message.  We want to reduce panic.  ;-)

Alex-Jordan:
Will the schema be able to say that a `fillin` inside math can have `@characters`, or it can have `@fill`, but it cannot have both?

rbeezer:
Yes.  Easy-peasy.  (I think.)

Alex-Jordan:
OK, Take 2 is ready.

Even if everything here looks good, still need:

- schema updates
- CSS for `.fillin.underline`, `.fillin.box`, `.fillin.shade`
- message to -announce

Should I try the schema updates or is it trivial for you?

rbeezer:
@rbeezer can do the schema.

We don't need to wait on CSS, right?  That'll end up with @davidfarmer post-facto?

-announce is for @alex-jordan.

Alex-Jordan:
Got it.

@davidfarmer At Subsection 4.7: Some Paragraph-Level Markup 
https://spot.pcc.edu/~ajordan/temp/interesting-corollary.html#subsection-paragraph-markup
I repeated a paragraph with fillins three times.

The first paragraph has `.fillin.underline`. Then `.fillin.box`. Then `.fillin.shade`.

We need plain `.fillin` to continue looking like it looks now. And `.fillin.underline` should look the same way.

`.fillin.box` should put a border around the span and give it an appropriate height.

`.fillin.shade` should give it an appropriate height, but no border. And make the background color be gray. Specifically `#e6e6e6`.

I would add these except I know CSS is in so much transition right now.

davidfarmer:
I can do the CSS once an example is posted.

Alex-Jordan:
My most recent message here has an example posted. A "real" example will only have one of the fillin classes so I doctored that one to have all three.

rbeezer:
Very good!  That's all a big improvement.  I'll rebuild website shortly.

Removed a mention in the sample article of `@characters` being deprecated.  And exploded out into severtal topical commits keeping all references to the work being yours.

Alex-Jordan:
Thanks! Is it ready for an announce message? Maybe after @davidfarmer styles the HTML blanks?

Alex-Jordan:
Math fillin in HTML is covered. CSS is needed for HTML text fillin. But the default is `underline` and that is already what you get even without new CSS. So I'll do the -announce now. Now rush on `box` and `shade` CSS for text fillin.

Alex-Jordan:
My most recent message here has an example posted. A "real" example will
only have one of the fillin classes so I doctored that one to have all
three.

davidfarmer:
I added CSS for:
   span.fillin.underline
   span.fillin.box
   span.fillin.shade

Formerly there was only span.fillin , which is the same as what now is
span.fillin.underline .

Let me know if something does not look right when those are actually
in the sample article.

Alex-Jordan:
> Let me know if something does not look right when those are actually
> in the sample article.

It is a publisher switch to use one of the options globally, so it will never be the case that you can see them all on the same page with the untouched sample article. But using the inspector to change a class should be OK to get a glimpse of the other two styles.

@rbeezer had some suggestions at last week's drop-in about where the baseline of these things should be. Rob, if you look [here](https://spot.pcc.edu/~ajordan/temp/interesting-corollary.html#subsection-paragraph-markup), what would you change? I can't recall the differences (if any) for the baseline between underline, box, and shade.

PR #1728, opened 2022-05-30
update runestone-services file

bnmnetp:
This has the code for the new graph based grading in parsons.
It also has:
* new css for the runestone components
* Horizontal parsons
* First released version of the conversion code rst to PreTeXt (no impact on pretext)

### DIFF
--- a/xsl/support/runestone-services.xml
+++ b/xsl/support/runestone-services.xml
@@ -1,14 +1,14 @@
 <?xml version="1.0" ?>
 <all>
-    <js type="list">
-        <item type="str">runtime.45581e16620a950c.bundle.js</item>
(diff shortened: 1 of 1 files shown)


rbeezer:
Me or you?
```
Uncaught (in promise) TypeError: document.getElementById(...) is null
    21294 bookfuncs.js:211
    21294 bookfuncs.js:254
    jQuery 13
bookfuncs.js:211:17
```
Interactive exercises behave before this change, now are not interactive when this change is applied.

bnmnetp:
Just discovered the problem.... Passes all my unit tests but breaks html-for-all books in pretext.  Time for some updates to unit tests.

rbeezer:
OK, thanks.  I'm trying to get to moving the PTX sample material to the sample book.  Should be 100% independent of this.  But it'll give you something additional to teest with once we iron out switch.

Issue #929, opened 2018-07-27
Problem with \fillin when at start of line


mitchkeller:
Preview Activity 1.2.1 in Active Calculus is demonstrating a small issue with @Alex-Jordan's LaTeX code for `\fillin`. With the way the sentence containing `<fillin>` is currently phrased, the blank winds up at the start of a line. However, LaTeX fails to actually produce the line in this case. If I insert some extra words before the `\fillin`, then the line is drawn in the output. If I manually insert a `\newline` into the LaTeX before `\fillin`, the line still isn't rendered (and no horizontal space is consumed—the period and close quote appear at the start of the line).

Resolving this is well beyond my level of TeXpertise, unfortunately. I'm going to hack something that gets the blank to appear in the 2018 edition of Active Calculus, and then this can get fixed later.

Alex-Jordan:
Hmm, sorry about this. I suggest for now that you add something to `latex.preamble.late` that renews `\fillin`. The code right now looks so complicated so that it can support the box style, and also so that it can be very stretchy/compressible. If it didn't have to support the box style, it could be a lot simpler. The XSLT could change to make the definition a lot simpler in the underscored case. That would make a bigger if/then in the XSLT, but a cleaner .tex when using underlined fillins.

Or maybe there is a similar problem if using the box style. 

Here is another thing you could try in the short term. Note how the code allows the fillin to compress all the way to 0 width. Edit so that something like 1em is minimal and see what effect that has. Of course, it might make answer blanks move on to new lines where previously there was a tiny answer blank on the previous line.

mitchkeller:
The definition of `\fillin` that I'm getting now is

```
\newcommand{\fillin}[1]{\leavevmode\leaders\vrule height -1.2pt depth 1.5pt \hskip #1em minus #1em \null}
```

I'm really bad at rule drawing commands in (La)TeX, so I really can't make heads or tails out of this. I'm not sure how to change the minimum width in this command.

Alex-Jordan:
There is a place where you see `\hskip`. It takes the argument `#1` times as many em, but adds rubber space of negative that many em. So it starts wanting to go `#1em` forward, but is willing to shrink back down to 0 if that helps fit something nicely on a line. Not sure if you can get away with `\hskip #1em minus (#1-1)em`. Maybe you need to use `\dimexpr` like 
```
\hskip #1em minus \dimexpr#1em-1em\relax`

Sorry I can'ttest right now, but that is the basic logic.
```

[...]

PR #1730, opened 2022-06-02
LaTeX: prevent an underline fillin at the start of a line from collap…

Alex-Jordan:
…sing

This addresses issue #929, which was still a problem even after the recent refactor. But at least now I understand how to fix it.

I tested the macro in a .tex file I wrote from scratch. I doctored it so that a fillin happened right after a natural line break, and it was collapsing as in #929. But then this edit fixes that. And I also tested that it doesn't harm fillins in other places.

The same change is not needed for box or shade.


rbeezer:
Thanks, I'll close the issue.  Maybe a website update later today.

PR #1734, opened 2022-06-08
LaTeX: make \fillin behave at start of paragraph

Alex-Jordan:
I'm making dinner too :)

This is only tested on the sample book. But works for all three fillin text styles as far as my eyes could catch.

Haven't tested against Mitch's issue when the fillin is at the start of a new line. I'll need to fiddle up an example to test that.


Alex-Jordan:
Second commit adds two fillin examples to the sample article. One lands a fillin at the natural start of a second line in a paragraph. The other is a paragraph containing only a fillin.

Tested the above with underline, box, and shade style and all is well. Same with sample book.

rbeezer:
Thanks, I'll merge in the AM.

rbeezer:
Thanks again for the rush job.  One more small PR and I'll update website examples.

PR #1735, opened 2022-06-08
Remove double footnotes in the guide

valelav:
In a few circumstances there are double footnotes with the same content in the guide: see for instance

https://pretextbook.org/doc/guide/html/atom.html

One footnote is generated by the url tag, while the other one is explicitly authored with a fn tag

I removed the fn elements and (if not already present) added a visual attribute to the url element

I made a search in the source for this string `</url><fn>` and there shouldn't be any other occurrences


rbeezer:
Thanks, Valerio!  I make changes to PreTeXt and then do not always get around to fixing up the various documents I manage.  Thanks very much for the assist.

I merged these all into one commit.  Website is updating just now, give it 15 minutes or so.  Thanks again.

Issue #1713, opened 2022-05-04
Live WW exercises not working on repo docs


Alex-Jordan:
The Guide has live WW exercises at:
https://pretextbook.org/doc/guide/html/basics-s-webwork.html
But they are not working.

Neither are those at the Showcase Article:
https://pretextbook.org/examples/showcase/html/webwork.html

My suspicion is that this has to do with Rob's build script, not something with the PreTeXt repo itself.

rbeezer:
Known problem, been unable to investigate.

Alex-Jordan:
Just a ping about this. It may be that I should help update the publication file for each document.

rbeezer:
Showcase Article is fixed via #1693.  Plus new way to specify WW server in publication file.

I'm **guessing** some `*.ptx` v. `*.xml` confusion on my part is to blame for the outage.

rbeezer:
Recent changes mean missing representations files are raising processing errors.  So I removed the two examples in the Guide at 8e2c030c7e5288c8eb2072af90d510561e7dca8d.  That change can be easily reverted.

rbeezer:
Content in the Guide is back.  Still not building, but there is no longer a processing failure.  These WW were not to blame, I'd introduced a bug, which is corrected at fb3fc82cfc28a80a156e3d00ddf286656a595d99

rbeezer:
PR #1740 fixes this for me locally, but I may still need to adjust my build script.  Closing now and will get it straight shortly.

PR #1740, opened 2022-06-11
Documentation (guide) updates

bnmnetp:
I am not very good at reading documentation from start to finish.  I am more apt to look at the TOC or do a search for something close to what I want to do.  So, I'm attemping to do both read through the documentation, but also think about how someone might arrive at a page who has a converted Runestone book and has a particular task in mind. 

I'm not sure if you are ready for me (or anyone) to update this guide for all of the interactive exercises?

I also added a project.ptx file so that one can `pretext build html -d -w` to build the guide for themselves.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,12 @@ doc/guide/webwork-representations.ptx
 script/mjsre/node_modules/
 script/mjsre/package-lock.json
 pretext/__pycache__/**
+.error_schema.log
+*~
 .vscode
 **/_build
 **/output
+**/generated
+**/rsbuild
+**/build
+
--- a/doc/guide/README.md
+++ b/doc/guide/README.md
@@ -1,20 +1,20 @@
-The PreTeXt Guide
-=================
+# The PreTeXt Guide
 
 PDF and HTML versions of this guide are available at the [PreTeXt](https://pretextbook.org) site in the Documentation area.
 
 If you wish to build from source, possibly as part of contributing improvements, follow these steps:
 
 1.  To build LaTeX for input to `pdflatex`:
-        cd /path/to/mathbook
-        xsltproc -xinclude -o guide.tex xsl/pretext-latex.xsl doc/guide/guide.xml
+    cd /path/to/guide
+    pretext build latex -d -w
 1.  And for HTML output:
-        cd /path/to/mathbook
-        xsltproc -xinclude xsl/pretext-html.xsl doc/guide/guide.xml
+    cd /path/to/guide
+    pretext build html -d -w
+1.  The preceeding two steps will attempt to build all of the webwork representations and diagrams needed for your book. You may be missing some of the prerequisites, such as Sage that will need to be installed before a full build can be completed.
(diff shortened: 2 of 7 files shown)


rbeezer:
Thank-you, thank-you, thank-you.  Documentation definitely needs fresh eyes.  I started a project to really overhaul the Guide, during a recent long road trip, but it got derailed by some major improvements.

Yahoo! Webwork builds now since you switched to managed directories.  See #1713.

Rebuilding website now.  May need to be redone, I think there is some repo nonsense related to the `webwork-representations` file.  I never ends.

I think this week will be my time to document all the Runestone progress.  Perhaps I'll make an issue to report progress and invite reviews.

Some nits.  All **minor**.  Thanks again.

* I have to wait long enough for my build script - I'm not re-generating `/generated` over and over.  ;-)  I consider them fixed for the examples - build, save, commit.  So I took that out of `.gitignore`.
* Removed another `xsltproc` mention in the README.
* I added a bit more to the overview material, with a cross-reference to `topic-specialized-divisions`.
* "user" `->` "reader"  ;-)  We don't have any users.  Readers, authors, publisher, developers.
* I changed "html" to `<init>HTML</init>`.  House style.
* Setting `generated="images"` might not be best practice.  But "it works" and why not incure some technical debt here and there?
* Idly curious.  Was it your editor that put elements with lots of long attributes onto a single line?  Or some other reason?

bnmnetp:
* generated -> Hmmm I had to build them all?
* "user" -> "reader"  I predict that will not be the last time I forget that distinction
* I set external="images" which I thought was a good practice...  but correct me if I'm wrong because some of this diagram and image stuff still feels like magic to me.  Maybe a good section of the docs for me to read.
* Maybe I should turn off autoformat for all `.ptx` and `.xml` documents.  at least when I'm editing docs I didn't start.

rbeezer:
> generated -> Hmmm I had to build them all?

There was nothing to build automatically in the Guide.  I had a small problem when the directory didn't exist (I should set a trap for that).  But I did a follow-on commit putting the WW problems in there, that was a (personal) fix waiting to happen.

> I set external="images"

Puzzling over breakfast:

```
generated="images"
<image source="images/foo.png"/>
```
should make a build expect
```
images/images/foo.png
```
In other words, you should have had to edit source to remove a lot of `images/` from `@source` attributes (a PITA I have done many times).  And then, "aha", I bet a `.join()` of two paths is collapsing the adjacent `images\`!  Hopefully there is a switch to correct that behavior.

https://pretextbook.org/doc/guide/html/processing-directory-management.html

...has some speculation that has actually been accomplished.

> Maybe I should turn off autoformat

Yes!  ;-)  I can't tell what is automatic and what is what you think is best practice.  ;-)

bnmnetp:
OK, we will talk about generated sometime this week.... 

I would not claim to know best practice for xml format...  We use black for python  and prettier for javascript to delegate that responsibility and make sure everyone does it the same.  It really helps with readability for everyone to follow the style and it makes everyone more productive to just mash in code and then hit the save button to have it formatted nicely.

PR #1742, opened 2022-06-17
Require runestone package (as is done with requests and lxml).

StevenClontz:
This PR no longer requires us to maintain separate copies of Runestone components. I've added a requirements.txt file to keep track of external dependencies (right now, `requests`, `lxml`, and as of this PR `runestone`).


rbeezer:
This further complicates my development workflow.

When I added this, I would have preferred to just use `tracer.runestone.academy` to get Python traces, like we are doing now for Java, C, and C++.  This would be moot, and the core routines would be simpler.

StevenClontz:
let's discuss synchronously in drop-in (I'm there now if you're able to join a bit early)

StevenClontz:
Looked a bit closer at this code. Setting aside issues viz a viz [this relevant XKCD](https://xkcd.com/1172/), I think it makes a lot of sense to run tracing on Python code the same way it's done with other languages (e.g. via an online Runestone service instead of locally). This removes both the local runestone dependency (whether by pip installation or redistributing code).

StevenClontz:
While this is still the *right* way (cough 🙃) I'm closing this for now and finding a way to redistirbute scripts that works whether you're importing pretext.py as a script or the CLI pretext module.

rbeezer:
See #1746.  **Thanks!**

PR #1743, opened 2022-06-17
Cli examples

StevenClontz:
This PR will CLI-ify a few of the example projects. The sample article publication files get a few tweaks for managed directories.

Note that as requested I *didn't* reorganize source in the sample article to fit the convention we're generating with `pretext new`.


rbeezer:
Looks good!  Thanks.

rbeezer:
Thanks!

PR #1745, opened 2022-06-17
eliminate need to call pdfcropmargins as an external executable

StevenClontz:


rbeezer:
Very nice!  Can we have more of this, please?  ;-)

Edited deprecation message seems unreachable, after I tried to provoke it.  The key `pdfcrop` is out-of-service, so the `get_executable_cmd()` never gets called with that key.  I removed it in a follow-on commit.

I added a line to the `pretext.cfg` file which is the only way folks know about edits they might make to their personal non-repo copies.  Added that into commit posted here.

Great progress, thanks.

rbeezer:
Parameters (such as a threshold for "darkness" of what is cropped or not) got lost from the `pretext.cfg` file.  A fix is at 3ce881b7c7bf9fe29b3419804911cbb81f522ef7, plus a bit of tidying up in two subsequent commits.

@StevenClontz, @oscarlevin  - clear for CLI 0.7 from where I sit.

oscarlevin:
I've updated the CLI to your most recent commit and restored our `project.ptx` templates to once again include `pdfcrop`.

StevenClontz:
> I've updated the CLI to your most recent commit and restored our `project.ptx` templates to once again include `pdfcrop`.

Thanks for doing this. I cut 0.7.0 before I thought about this revert!

rbeezer:
Threshold is now hard-coded in the Python to be byte-value 254, meaning white is only value 255 and even the faintest gray will be retained when cropped.  At 6e0cafb87384f7e45cbf5c59a60a1203ef49988f

Lots of cruft blown away, too, at c12ac037445a3b27e656d015a425d6118f67fe45

PR #1746, opened 2022-06-17
attempt to solve import issues based on how pretext.py is called

StevenClontz:
I think this might work with the core pretext cli script, with the bonus that it could be replaced with a `pip install runestone` later and deleting the runestone submodule we're distributing.


rbeezer:
Reverted problematic edit on an initial commit.

Edited import and use to house style.  Should not have consequences, but perhaps should be tested.

Not ready for CLI 0.7, still need to get "coding exercises" whipped into shape.

StevenClontz:
> coding exercises

To clarify, this won't involve changing the core API/imports (and I'll just need to update the core commit when it's ready), right?

rbeezer:
Right - just XSL changes.  And got rougher than I expected.  Will tackle again later tonight perhaps.  (I'm hoping to be off-duty tomorrow.)

StevenClontz:
Cool, my plan is to update core commit and publish 0.7 Sunday afternoon - if this isn't working by then, we'll pick it up next week for the first of what I expect will be daily releases. :)

rbeezer:
```xml
<chuckle/>
```
The 0.7.x series will always remind us of the U of Michigan.  Go Blue!

Can you do `0.7.tuesday`?  Don't answer that.

StevenClontz:
Too bad, I'm answering. While https://peps.python.org/pep-0440/ doesn't
allow "Tuesday" as a version identifier, I'm hoping it works out that we
release 0.7.0 on Sunday, 0.7.1 on Monday...

StevenClontz:
This was fun while it lasted, but with our update to using poetry (like Runestone) for packaging, this hack has stopped working.

Rob do you use `pretext/pretext.py` directly ever, or only via your `pretext/pretext` script? If its the latter, I think have an idea for a PR that fixes these import issues that I think won't mess with your workflow.

PR #1749, opened 2022-06-19
add info on deploying to GhPages using CLI

StevenClontz:
Let folks know they can deploy any PreTeXt project to GitHub Pages with minimal(?) fuss if they use the CLI.


rbeezer:
Got it!  Thanks.  Updating website right now.

PR #1752, opened 2022-06-21
add the cli.log, built by cli, to gitignore

oscarlevin:
After we wired up the CLI to work in the sample-article, if you build it in that folder, you get a cli.log there.  This shouldn't be tracked by git.


rbeezer:
Done.  Thanks.

PR #1754, opened 2022-06-22
Require image descriptions to be plain text

Alex-Jordan:
This changes the schema for an image description to be plain text only.

Will require an -announce message? Suddenly schema-compliant source will become noncompliant.

Also since this goes into an alt attribute, we should do something to the text inside to escape double quotes. But I would think that should that be a separate commit/PR.


rbeezer:
Looks good.  I don't have my schema browser tool on mhy laptop, so I'll pick this up later.  Thanks!

rbeezer:
Pausing this to wait on #1754.

rbeezer:
Can the "circle-I" info icon go on the left of an image, perhaps jutting into the left margin?  Well, maybe not if the images are in a `sidebyside`.  It'd need to not fight with the permalink icon.  Perhaps centered vertically?  I like this a lot - just musing on locations that work well for readers and maybe don't also add extra whitespace in any one direction.

Alex-Jordan:
> Can the "circle-I" info icon go on the left of an image, perhaps jutting into the left margin? Well, maybe not if the images are in a sidebyside.

This is wrt a mockup I made to do #1754. I'll continue over there.

Alex-Jordan:
Er, I mean #1762.

Alex-Jordan:
In the time since this was opened, we've had some more discussions and realizations about description/short-description/long-description, etc. But I can never remember the details weeks later when I might sit down to hammer through implementation. Maybe an upcoming drop-in, we can go over it again and I'll write it all down somewhere.

Alex-Jordan:
Closing in favor of #2054.

PR #1755, opened 2022-06-22
Package core PreTeXt python scripts as importable library

StevenClontz:
Went ahead and gave [this](https://github.com/PreTeXtBook/pretext/pull/1746#issuecomment-1162588811) a shot. All `.py` files are collected into a `lib` directory that should give everyone less headaches when importing.

My hope is that you're only running the `pretext/pretext` script, which I believe is updated to import things correctly. Since `pretext/pretext.py` is now `pretext/lib/__init__.py` anything hardwired to `pretext/pretext.py` would be broken. I'm unsure if there's a contract with users that `pretext/pretext.py` won't be moved.


rbeezer:
As discussed, we will soon get a Python tracer on a Runestone server, and we'll just get rid of these additional files.  So this won't be necessary.

rbeezer:
Following up, we now have a Python endpoint on the Runestone `tracer` server.  Thanks @bnmnetp!  Main bit of code at d88127f0c246d6588300fbc8bd332f2aac042c72, with other surrounding commits.

So all this `pylogger` code is now removed and the situation is back to being a little simpler.

PR #1757, opened 2022-06-22
WeBWorK: make valid inside PROJECT-LIKE

Alex-Jordan:
This makes it so you can have `(introduction?|webwork|conclusion?)` in a PROJECT-LIKE. PROJECT-LIKE is similar enough to inline exercise that this did not take much work.

### DIFF
--- a/doc/guide/author/webwork.xml
+++ b/doc/guide/author/webwork.xml
@@ -40,7 +40,7 @@
         before you will be able to see any <webwork/> content in your output.  For simple cases, you might be able to run <c>pretext build html -w</c> to process <webwork /> exercises.</alert>
       </p>
       <p>
-        A <tag>webwork</tag> tag must be inside an <tag>exercise</tag>,
+        A <tag>webwork</tag> tag must be inside an <tag>exercise</tag> or a PROJECT-LIKE block,
         <em>optionally</em> preceded by an <tag>introduction</tag>,
         and <em>optionally</em> followed by a <tag>conclusion</tag>.
       </p>
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -308,6 +308,18 @@
                 </conclusion>
             </exercise>
 
+            <project>
+                <title>Inside a <tag>project</tag></title>
+                <introduction>
+                    <p>
+                        If you like, you can have a <webwork/> inside a PROJECT-LIKE block.
+                        Just like with an <tag>exercise</tag>, it can be preceded with an optional <tag>introduction</tag>
+                        and followed by an optional <tag>conclusion</tag>.
+                    </p>
+                </introduction>
+                <webwork copy="webwork-add-numbers" seed="511"/>
+            </project>
+
         </section>
 
         <section>
--- a/schema/pretext.xml
+++ b/schema/pretext.xml
@@ -1258,7 +1258,8 @@
                        Prelude?,
                        (
                           (Statement, Hint*, Answer*, Solution*) |
-                          (IntroductionStatement?, Task+, ConclusionStatement?)
+                          (IntroductionStatement?, Task+, ConclusionStatement?) |
+                          (IntroductionText?, WebWork, ConclusionText?)
                        ),
                        Postlude?
                     )
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -287,11 +287,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <xsl:param name="webwork.divisional.static" select="'yes'" />
 <xsl:param name="webwork.reading.static" select="'yes'" />
 <xsl:param name="webwork.worksheet.static" select="'yes'" />
+<xsl:param name="webwork.project.static" select="'no'" />
 <!-- We make variables instead of using the params directly, so that in EPUB we can overrule -->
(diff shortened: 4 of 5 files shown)


Alex-Jordan:
Motivated by @mitchkeller 's request to turn the Active Calculus preview activities into WeBWorK exercises.

Alex-Jordan:
>     Examples in the sample chapter?

Done on a second commit.

> The final line of current commit looks like our recent time sinks?

For posterity: we chatted offline and this should not be so much of a time sink. In -html, it is within a variable selection. In -latex, within a match but for a modal template.

rbeezer:
I'm not seeing the new example come through to the problem archive?

rbeezer:
@Alex-Jordan notes that the (serial) number of a PROJECT-LIKE and the number of am "inline exercise" can be identical, which is a bit of a train wreck in the problem set formulation.  Projects can be "on their own counter" or not, so sometimes a problem, sometimes not.  At any rate, this PR is paused.

rbeezer:
Thanks, Alex.

*  Made an issue for the serial number limitation of WW:

https://github.com/PreTeXtBook/pretext/issues/1859

*  Should we announce on `pretext-announce`?  [NEW]?  Briefly mention numbers in problem sets? (You or me, your call.)

*  Using `@copy` in the tests is great.  But to have our **only** example that way makes me nervous.  Such as, the change in the problem sets is negligible.  (Doing a cut/paste copy of some other problem would be better.)

*  The `match="*[...]` construction we discussed is not so bad.  (A) runtime seems unaffected.  (B) it is a **modal** template so only gets fed a limited number of elements anyway.  It is wildcard, filter, all elements that is dangerous.  I missed that yesterday.  But I silently changed it anyway, and added a `PROJECT-LIKE` nearby.

*  String parameter `webwork.project.static`, and its cousins, will be deprecated very soon.  Weeks, not days, not months.  Your call if you want to update documentation now or wait.

rbeezer:
> String parameter webwork.project.static, and its cousins, will be deprecated very soon. Weeks, not days, not months.

Not true!  Doing it now.

Alex-Jordan:
Midterm exams and Thanksgiving-related stuff had my attention when this was in the final stages.

My guess is that with this merged, it's OK to have a `webwork` in a PROJECT-LIKE, but the archiving process ignores such `webwork`. So for now, no standalone `.pg` file, and no set definition file for all the PROJECT-LIKE. Does that sound right? If so I might try to get archiving working, then do the announcement.

rbeezer:
Is this a guess, or did you try?  My careful reading of the code just now is that they won't be missed.  Can you do a check?  Will want such a thing in the sample chapter in any event, if it is not there already.

But if you want three set definitions (in `exercises`, inline `exercise`, project-like) then some adjustments are needed.

Alex-Jordan:
It's a guess, I haven't tried anything yet. But in `pretext-ww-problem-sets.pl`, line 81, it has `<xsl:apply-templates select="$document-root//exercise/webwork-reps" />`. That makes me thing that `webwork` inside a project-like will be missed for the .pg file creation.

Meanwhile it looks to me that there will be two kinds of set definition files built. One for each `exercises` and one with all the rest, including those in a project-like. (Is that what you were seeing?) But it's no good if the .pg file doesn't exist. And besides, I think there should be one set def for `exercises`, one for each `worksheet`, one for all of the section's inline exercises, and one for all of the section's project-like. And I guess one for all of the section's reading-exercises. So it's also missing that aspect and lumping all the latter into one def file.

On my radar. Just 1.5 more weeks of teaching to get through...

rbeezer:
Well, yes, I glossed over that one.  Since **we** place the `webwork-reps` we know just where they are, and thus, the `exercise` step can just be **scrubbed**.  Rather than a complicated version with all the PROJECT-LIKE.

Alex-Jordan:
There's just a bit more though to this. Currently, if there is no title to a `webwork`, the pg file comes out like this: `4_2_4.pg`. That's inline exercise number 4 from section 4.2. So if section 4.2 also has project-like number 4, there would be a clash. So something should be done. I'm not sure if we want to make files like `Project-4_2_4.pg` or put things in a folder like `Project/4_2_4.pg` or something else entirely.

rbeezer:
There is a PreTeXt-wide notion of a "default title".  That could be impelmented for WW with no author-provided title.

rbeezer:
> There is a PreTeXt-wide notion of a "default title".

On second, thought, no.  It'd bleed through into too many places.  Let's keep anything like this in the dedicated stylesheet.  A utility template could go into `pretext-common.xsl` since we just need text for a filename.

rbeezer:
Do we think Active Calculus thinks they can put a WW in a `task` in a `project`?

mitchkeller:
I think we could live without that. WW has enough structuring options that just using the WW as a child of `project` should be enough without wanting a WW inside `project/task`. (I think the only part of Active Calculus currently using `task` is the vector calculus chapter, but long term, I do want to migrate toward task and away from lists.)

rbeezer:
> Using @copy in the tests is great. But to have our only example that way makes me nervous. Such as, the change in the problem sets is negligible. (Doing a cut/paste copy of some other problem would be better.)

Now using the sample chapter to test completing this addition.  We need another example that does not use `@copy`, and as such it might as well be something other than `project`.

rbeezer:
To a Runestone manifest at: bb90dc05550f3c75c40142a203dd13e8a9a0b0da

PR #1759, opened 2022-06-22
Sample article typo

jjrsylvestre:
Fixed a reference to a "second" item when it really should have been "third."


rbeezer:
Much better!  Thanks.

PR #1760, opened 2022-06-22
update verbose and debug messages to use logging package

oscarlevin:
Removed reliance on `__verbose` and `__debug` in favor of logging library's `log.info` and `log.debug`.  Messages now start with `PTX:INFO:` or `PTX:DEBUG:`, but some messages are `log.warning`, and `log.error` and `log.critical` are also now possible.

In addition to this being a cleaner way to manage messages to the console, this will allow the CLI to log all messages, both from the CLI and pretxt\pretext.py  in a log file.

### DIFF
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -24,6 +24,21 @@
 
 import pretext as ptx
 
+# Set up logging (replacing _verbose/_debug).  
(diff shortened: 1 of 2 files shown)


rbeezer:
Love it!  Thanks.

(Now I have to remember to keep using the new style!)

PR #1761, opened 2022-06-23
Logging improvements

oscarlevin:
Improved formatting of log messages (uniform spacing).

print() commands that were clearly log messages (starting with PTX:ERROR or PTX:WARNING) converted to log.error and log.warning.  There are still a few print() commands, but I think these are something else.


rbeezer:
Looking good. Thanks!

PR #1767, opened 2022-06-24
WeBWorK: support for a 2.17+ server

Alex-Jordan:


Alex-Jordan:
I've already changed things to `pretext-webwork.js`. More things will change but they should only affect cosmetics, not functionality.

Make sure to disable caching if you try it out.

But yes, I don't expect you would have a problem if you use this branch now with your 2.17 server. If you do, please let me know. It's working with the 2.17 I have in the AIM dev server.

And of course this won't be in the CLI.

sean-fitzpatrick:
Thanks! I'm trying to remember: was there a way to disable caching globally, or do we have to specify this in every ww problem with an image?

Alex-Jordan:
What I do is with Firefox, open the developer tools. Go to the Network tab, and there is a checkbox to disable caching. Then as long as the developer tools are open, when you load a page it will never rely on cached things.

sean-fitzpatrick:
Ah, OK. I thought you were referring to the cache on the webwork server: without cleaning it, a new build of webwork representations ends up using whatever images are in there instead of new ones.

sean-fitzpatrick:
Looks like things are working as expected.
Still needs some CSS attention though:
image

Only other thing I noticed: clicking the make interactive button gives you a different question than the static version.
That isn't the case with the versions of the book using 2.16.

Alex-Jordan:
Yes, styling will change in significant ways.

I suspect you built with platform=runestone? The version changing is
intentional, but only with rs builds.

sean-fitzpatrick:
I didn't set it to use Runestone.

sean-fitzpatrick:
Unless Runestone is a default now (I don't think it is) I've never put anything in the publication file for Runestone.

I've got another book that I'm busy putting Runestone exercises into, but even that I'm not building for Runestone, since I don't have a Runestone server and it's a work in progress

rbeezer:
> I should be able to fix that rerandinization issue

@Alex-Jordan: should I merge, or wait?  If the latter, please then send me a clear green light.

Alex-Jordan:
You can merge this, assuming you didn't find a problem with it that I missed. The above was in reference to `pretext-webwork.js`, that I can edit any time.

Alex-Jordan:
> Only other thing I noticed: clicking the make interactive button gives you a different question than the static version.

Should be fixed now.

For the record, it turns out building for Runestone is not the only thing that triggers PTX putting Runestone javascript into the page. That was what was leading to the randomization when we didn't want it.

sean-fitzpatrick:
Excellent. Time to rebuild all the things.

Alex-Jordan:
Oh, the thing I just fixed doesn't require a rebuild. I edited an AIM-hosted javascript file. So you should get no re-randomization if you do what you did before.

sean-fitzpatrick:
Yeah, but all of my official HTML builds are temporarily using my production WeBWorK server instead of the one intended for PreTeXt.
Now that 2.17 is working, I'll rebuild to move everything back to the right server

rbeezer:
Pushed from Hyatt Place lobby.  Limited testing, no website update yet.

Forgot to annotate with PR #.  Sorry.  :-(  At e5bb62a7000cd521fc33401085e45c50b18e242a

PR #1770, opened 2022-06-27
Update sre 4

zorkow:
Updates the MathJax integration script to work with latest MathJax (v3.2.2) and SRE (v4.X.X). In particular we take care of:
* Importing the new `Sre` module from MathJax.
* Setting the correct path to SRE's `MathMaps`.
* Waiting for all required locales to be loaded. 

Here we front-load all the locales, before actually rendering any expressions. This has the advantage that the rest of the code can be kept clean of any asynchronous operations. Note, that this is only possible since we are assuming the code is used as a batch script. In the event that the code should ever be integrated dynamically, this needs to be changed, as resetting a locale will not necessarily wait for that locale to be loaded.

### DIFF
--- a/script/mjsre/mj-sre-page.js
+++ b/script/mjsre/mj-sre-page.js
@@ -279,8 +279,17 @@ if (!(argv.svg || argv.svgenhanced)) {
   //  Wait for SRE, if needed
   //
   if (needsSRE) {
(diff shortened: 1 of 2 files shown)


rbeezer:
Working nicely - thanks for the maintenance on this piece!

PR #1776, opened 2022-06-29
Reveal-slides = when defn source-number not supplied do not print empty brackets

arechnitzer:
Very minor change to pretext->reveal translation. 
* Previously the defn template would attempt to print source-number even if not there resulting in empty brackets in the title line of the defn in the slide.
* Now this tests to see if source-number is supplied and prints accordingly.
* Also updated the copyright line.

Also some minor tweaks to center side-by-side and so that images work with decktape.


arechnitzer:
Sigh - sorry. This is not ready yet. I will make some more careful updates of things. Closing.

PR #1779, opened 2022-07-06
make opening files more robust to avoid permission error on windows

oscarlevin:
The previous line 116 left the file open, causing a permission error on windows.  Switched to the `with open(..) as file:` pattern.  Followed a similar pattern for the two "inplace" substitutions later (so they now match what substitutions were done in the kindle version (lines 85-87).  

Tested before and after on the epub-sampler and only build times changed.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -82,7 +82,7 @@ def mathjax_latex(xml_source, pub_file, out_file, dest_dir, math_format):
     # Trying to correct baseline for inline math in Kindle, so we
     # insert a \mathstrut into all the inline math before feeding to MathJax
     if math_format == "kindle":
-        with fileinput.FileInput(mjinput, inplace=True, backup=".bak") as file:
+        with fileinput.FileInput(mjinput, inplace=True) as file:
             for line in file:
                 print(line.replace(r"\(", r"\(\mathstrut "), end="")
 
@@ -133,7 +133,7 @@ def mathjax_latex(xml_source, pub_file, out_file, dest_dir, math_format):
     owd = os.getcwd()
     os.chdir(tmp_dir)
     html_file = mjoutput
-    with fileinput.FileInput(html_file, inplace=True, backup=".bak") as file:
+    with fileinput.FileInput(html_file, inplace=True) as file:
         for line in file:
             print(xhtml_elt.sub(repl, line), end="")
     os.chdir(owd)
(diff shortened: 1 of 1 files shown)


rbeezer:
Mea culpa.  Sorry.

Why are we saving a backup?  These are intermediate files that are are being massaged to satisfy EPUB validation.  I didn't actually test (just **reading** the code, always my first step), but if there are backup files, I don't think we want them in the EPUB.

oscarlevin:
I don't know why or whether we create backups; I just copied the pattern from the kindle replacement step.  I suspect the .bak file is destroyed when the temp directory is cleaned up anyway.

mitchkeller:
@rbeezer The backup files aren't going into the EPUB file. The `inplace=True` replacement mechanism fails on Windows (at least when working on files in the temp directory) without creating a backup. Per @oscarlevin, this is something that has come up on Stackexchange as a problem for others when dealing with Windows. I think that's why in one area we (I…last summer…when testing with Oscar) were using `bak` files for a stage of the Kindle process. This simply uses the same pattern during all of the file fix-ups and then still only brings over the files that we need.

rbeezer:
Thanks, @mitchkeller - I wanted to make sure I understood **why** we needed this - I might drop a comment in alongside.

rbeezer:
Thanks, @oscarlevin.  Yes, a test would be great.  (I'd be trouble for me to get this going on my Windows VM!)  Then we can let @mitchkeller experiment if he likes, next time his head is into Kindle on Windows.

oscarlevin:
I've removed the `backup=".bak"` argument in the three spots it was/I put it.  Still works fine on both windows and linux (wsl).

Reading about how fileinput works, no matter what, it creates a backup file.  Passing `backup=".bak"` both specifies that the backup should have the `.bak` suffix, and also that the backup files are not deleted after the inplace substitutions are made.

I believe that using the `with fileinput.FileInput()...` instead of just `for line in fileinput.input(...)` seems better, only because the files would still be closed even if there is an error (with no error, the `fileinput.input()` would close the file, apparently).

oscarlevin:
Testing a few output formats with the `turtle` line removed works fine on windows.  And vscode agrees that it isn't being used anywhere.  I don't know when that was added; I don't think it was related to any of the logging stuff, even though it is called in that block of code.

rbeezer:
Thanks for this, Oscar, and for sticking wih it.  Much better...

PR #1780, opened 2022-07-06
WeBWorK: localize terms used by pretext-webwork.js

Alex-Jordan:
This puts localized versions of strings like "Correct" into a data attribute for an embedded WW exercise. The javascript will grab that and use it for feedback messages. I think I'll wait to edit the javascript until this PR is resolved, unless there is a reason to do that first.

Also, I understand that you are refactoring localization. With the current tools, I felt like I needed to use the "type-name" template. I understand "type-name" to refer to the name of an element type, like "Theorem" for a `<theorem>`. That's not what this localization is doing; it's just looking for localized versions of certain words. So did I do this the wrong way? Or will your refactor be more general than just names of element types?

### DIFF
--- a/xsl/localizations/af-ZA.xml
+++ b/xsl/localizations/af-ZA.xml
@@ -227,4 +227,10 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <localization string-id='evaluate'>Evalueer</localization>
     <localization string-id='evaluate'>Evalueer</localization>
     <!-- <localization string-id='code'>Kode</localization> -->
+    <!-- Interactive exercise feedback messages                          -->
+    <!-- These might be part of a feedback message for submitting        -->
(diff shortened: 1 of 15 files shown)


rbeezer:
Thanks!  The "type-name" template looks right.  The "lang" parameter will eventually be controlled by the context of a `match` on a change to a not-named template.  So it can vary for parts of a document with different languages.  I won't rant again.  ;-)

> The javascript will grab that and use it for feedback messages.

I was wondering how you were going to push this through the pipeline.

> Document concisely in the US-en file, do not repeat documentation in every other file.

I'm trying to make the English file the documentation, and then not **repeat** that in the all the other files, except in extreme cases.  So I've removed the thirteen instances of repeated explanations.

No website update yet, but I did check against the HTML sample chapter.

PR #1783, opened 2022-07-11
Wwlateximage

Alex-Jordan:
This cleans up the situation with `latex-image` in a `webwork`. No more `synatx='PGtikz'`. Just write `latex-image` code like you would anywhere else. (Except author must take special care with dollar signs.)

So `synatx='PGtikz'` is removed from schema.

There is a deprecation warning.

If you still use `synatx='PGtikz'`, the repair phase takes care of that and makes your source ready for a 2.17 WW server.

If it turns out you are actually contacting a 2.16 WW server during the representations collecting, then the python script adjusts at the last moment, reverting to 2.16 structure. 

Once this is resolved, I will revisit #1610.

### DIFF
--- a/doc/guide/author/webwork.xml
+++ b/doc/guide/author/webwork.xml
@@ -374,22 +374,17 @@
       <subsubsection>
         <title>Using <c>TikZ</c></title>
         <p>
-          In a problem statement (or hint, or solution), you may use an <tag>image</tag> with a <tag>latex-image</tag> child
-          that has <attr>syntax</attr> set to <c>PGtikz</c>. (This requires the <webwork/> server to be version 2.16 or later.)
+          In a problem statement (or hint, or solution), you may use an <tag>image</tag> with a <tag>latex-image</tag> child,
+          as long as the host <webwork/> server is version 2.16 or later.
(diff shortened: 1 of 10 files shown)


rbeezer:
Nice (comprehensive) work.  Deprecations can be no fun.  But the pre-processor's "repair" phase is a dream, no?

I guess we need to `sanitize-latex` during "repair" since adding the `tikzpicture` environment is going in with no idea about existing indentation.  No matter - if repairs are a bit crazy, that's tough, authors will need to heed warnings sooner rather than later.

One spelling error corrected:  `invironment` in the deprecation warning.  That was sure to set off **somebody** if they saw it repeatedly.  ;-)

Had to split out the schema changes (no small feat) so I could add to them, and I like them on their own commit.  So while I was at it anyway...I like to have the actual act of deprecation (warning, fix) on a commit of its own.  And then "Guide" material.  And then the meat is about all that is left.

Thanks - nice to have this all consolidated!

rbeezer:
Big push yesterday, waiting on major review by you-know-who.  Good timing!  You never know what shiny object will get my attention.

PR #1784, opened 2022-07-11
WeBWorK: warning in the guide about incompatible PG problem files

Alex-Jordan:
This was a good suggestion from the forums recently.


rbeezer:
Would you like to suggest "coding" the problem "from scratch" in PreTeXt  if it is really appealing?  (Maybe that is already somewhere else, I didn't hunt around.)

Alex-Jordan:
Thanks, I added a paragraph about what to do when you really want that problem anyway. Force pushed.

PR #1786, opened 2022-07-11
WeBWorK: remove PGtikz instance from validation-plus

Alex-Jordan:
One thing got through that should have been in #1783. Removing this (now unnecessary) thing from validation-plus.


rbeezer:
OK, now back to grant-panic, so be patient!

PR #1787, opened 2022-07-13
Lateximageinwwegintro

Alex-Jordan:
This makes it so when you have a `latex-image` in an `exercisegroup/introduction`, and the `exercisegroup` has WW exercises, each archived .pg file will have the code for the `latex-image`.


rbeezer:
Rebased, just to make sure this was up-to-date.  Otherwise, no edits whatsoever.

Ran this through testing, but can't say I was horribly thorough.  ;-)

PR #1788, opened 2022-07-14
Use pubnote for publication note in bibliography

davidfarmer:
A "note" is for a note about a reference.

This adds "pubnote" for a note about publication information.


rbeezer:
Do you explicity want the content of `pubnote` to have no markup?  No `m`, no `em`, no `pubtitle`, etc.?

Matching on `text()` means any element like the above will be **dropped**.

rbeezer:
I added an example to the sample article so this would get tested regularly.  At b60fed9f91488e91597d81c6b7b09801c7b4fa93

PR #1799, opened 2022-07-21
ensure posix path is passed to xsl via string params

oscarlevin:
This fixes a problem with epub (and I assume braille) under windows where math is not inserted into the output, because the windows style path is being used inside a stringparam.  Should have no effect on unix systems.


rbeezer:
Nice work.  Thanks!

PR #1804, opened 2022-08-01
Update CLI quick start for image generation

kcrisman:
This image generation instructions currently lead to errors.  This should be accurate for the current state of the CLI, but edits are welcome since the precise verbiage, use of remark, etc. may not be as desired by the maintainers.

### DIFF
--- a/doc/guide/introduction/quickstart-example.xml
+++ b/doc/guide/introduction/quickstart-example.xml
@@ -130,37 +130,47 @@
         </console>
 
         <p>
-          Now we build an HTML version with the following command:
+          Now we build a version for the web (HTML) with the following command:
         </p>
 
         <console>
-          <input>pretext build html -d</input>
+          <input>pretext build web</input>
         </console>
 
-        <p>
-          The <c>-d</c> flag tells <pretext /> to generate any images described in the source.  After doing this once, you can omit the <c>-d</c>, unless you modify the code for one of the images described in the source.
-        </p>
+        <remark>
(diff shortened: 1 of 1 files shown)


kcrisman:
Comments: I don't know how to change the name of the branch using Github alone here, and I am leaving this in draft mode so nobody hurts me, but this draft is ready for merging as far as my own thoughts.  Primarily in draft to see whether @oscarlevin or others think the styling is right, or whether they have some more comprehensive stuff already on the way.

rbeezer:
Looks fine to me, but @oscarlevin can decide the ultimate disposition.  At your service.

StevenClontz (approved):

kcrisman:
I'll set it ready for review based on @StevenClontz approval (not sure what exactly that means on GH?).

StevenClontz:
It means nothing other than I "approve" it, though it's possible on GitHub to require N approving reviews before merging a PR.

oscarlevin:
I'm in the middle of fixing this section as well as the more comprehensive cli page, so some of this will be rewriten.  But I like your language better than mine, so I would recommend merging now.

rbeezer:
Thanks, everyone, especially KDC.

*  Had to fix an unbalanced `c` element after a `p`, which should have gone after an `xi:include` example.  Just merged that into current commit.
*  Pushed, so @oscarlevin can proceed, but have not yet rebuilt website version (maybe tonight?)

PR #1805, opened 2022-08-02
Fix errors in beamer template

sean-fitzpatrick:
I noticed (when `xsltproc` choked on it) that a couple of errors were introduced in the last update to this template.

These changes should ensure that the `pretext-beamer.xsl` template runs successfully. (I make no claim to have improved the output in any way.)

Two problems resolved:

1. On (what was) line 442, `<xsl:call-template name="fillin-text">` should be self-closing.
2. The use of `&PROOF-LIKE;` requires the `entities.ent` file to be loaded. I added some lines at the top to ensure this happens.

### DIFF
--- a/xsl/pretext-beamer.xsl
+++ b/xsl/pretext-beamer.xsl
@@ -19,6 +19,11 @@ You should have received a copy of the GNU General Public License
 along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 *********************************************************************-->
 
+<!DOCTYPE xsl:stylesheet [
+    <!ENTITY % entities SYSTEM "./entities.ent">
+    %entities;
+]>
+
 <xsl:stylesheet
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Sorry for the not-closed tag, that was me. If we don't have a test doc for
beamer, I'll make one next time I touch that file.

rbeezer:
Mea culpa, I didn't test a previous PR against this conversion, when I should have.

Sample slideshow to test against:

https://github.com/PreTeXtBook/pretext/blob/master/examples/sample-slideshow/sample-slideshow.xml

XSL succeeds now, but the produced LaTeX does not compile.  This conversion needs some serious work, and hence needs a maintainer.  There will have to be some coordinated work to de-monolithicasize the template of the stock LaTeX conversion, so **only** relevant pieces can be overridden.

Thanks for the fixes!

PR #1807, opened 2022-08-04
add CLI quickref

oscarlevin:
This is the quick reference sheet for using the CLI, to be released once the 1.0 version of pretext goes live.  Please don't merge until then.


rbeezer:
Looking good (as in, I built the PDF).  I might change the one `pdftex` driver so that this can be built for website via `xelatex`.

Give me some advance warning and I can have it all ready-to-go for 1.0.  I'll be offline for part of this month.

Thanks for this, it should really help folks getting started.

rbeezer:
Added, and integrated into website documentation area.  Rebuilding site now, fingers crossed.

Switched driver on color package to `xetex` and switched date over to a more-international standard format.

Thanks for this - should be really, really useful.

Will do main documentation soon, then maybe a "pretex-announce" is in order?

PR #1808, opened 2022-08-05
untested attempt for interactive preview logic, assuming a single local HTML file with all interactives

StevenClontz:
For @rbeezer consideration.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -1689,26 +1689,30 @@ def preview_images(xml_source, pub_file, stringparams, xmlid_root, dest_dir):
 
     # Interior asynchronous routine to manage the Chromium
     # headless browser and snapshot the desired iframe
-    async def snapshot(input_page, fragment, out_file):
+    async def snapshot(input_page, interactives):
 
-        # input_page: the "standalone" page of the interactive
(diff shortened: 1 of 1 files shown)


rbeezer:
Yes, that seems workable.  There will need to be a core `xsltproc()` call to apply the stylesheet I'm working on to the source, with output into a temporary directory full of iframe content files and support files (JS scripts, etc).  We can give the HTML file of interactives (fed into snapshotting) a predictable name, say `interactives.html`.

I suspect that the Python routine can use a small amout of `lxml` to simply identify the`iframe/@id` for the things to snapshot.  In other words, the present `extract-interactives.xsl` stylesheet can be retired and its output not consulted.

StevenClontz:
Agreed on retiring `extract-interactives.xsl` (and I suspect most `extract-` sheets could be similarly retired at some point.)

rbeezer:
@dbrianwalton is well down the road on this one, with preliminary work at #1894, #1895, and more to come.

So I'm going to close this.  Thanks, @StevenClontz, for promoting this approach.

StevenClontz:
Thanks @dbrianwalton !

PR #1810, opened 2022-08-10
Update quickrefs

bnmnetp:
I just fetch and merge out of habit!  I've got to figure out how to like rebase.

But I know you can split this however you want....

If not let me know and I'll figure out how to undo and submit this as two separate PRs

### DIFF
--- a/doc/quickref/quickref-pretext.tex
+++ b/doc/quickref/quickref-pretext.tex
@@ -302,6 +302,7 @@ An \texttt{<exercise>} has the following structure.
   <statement>
     <p>The <c>statement</c> is mandatory.</p>
   </statement>
+  <marker[s] />
   <hint>
     <p>Optional.</p>
   </hint>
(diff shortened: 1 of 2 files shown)


rbeezer:
*  Only two commits left after rebase.  Cleaned up real nice.  `2463318`  and  `fc01bf2`  are identical?  `git log --graph` showed some crazieness I did not understand.  ;-)
*  First newline-character needed a trailing semi-colon to be legal XML.
*  Promise:  I don't see multi-line math in a Parsons block rendering right.  That is the problem this is meant to solve?  Added a commit to sample book which will test this.
*  Exercises:  Where you have `<marker[s]/>` in the pseudo-XML is what I call elsewhere a "signal" (thinking like an XSL programmer).  Is that what you meant?  `blocks` is a signal/marker, and I don't know what to call `block` in this context.  Only signal not like this is for True/False, where it is an attribute on `statement`.

bnmnetp:
I meant to say signal but apparently my fingers did not get the message.

This is not meant to solve the display math problem in Parsons. -- Sean made a nice issue with some things for the two of us to look at together to help get to the bottom of that.

This is a more reliable solution to the problem where math was no always rendering correctly in a multiple choice question.

rbeezer:
> the problem where math was no always rendering correctly in a multiple choice question

Right!  Got that confused with Parsons.  In any event, MathJax still seems happy.  I could not provoke a failure before with success after, but that is not enough to stop this going in.

Som heavy edits on interactive exercises in the QuickRef, so I took the blame on a new commit.  Might rebuild website in a bit later today.

Thanks for the fix and for the documentation.

PR #1818, opened 2022-08-24
Fix: remove youtube player_api script tag

bnmnetp:
I have been playing around a bit.  Using `data-video-width="100%'` and omitting `data-video-height` does a passable job.  

the video fills the width, but uses letterboxing to maintain the aspect ratio of the video rather than increasing the height. So you end up with wide videos with lots of black on the side.  See below:
Screen Shot 2022-08-24 at 2 30 59 PM

I think it is kind of OK, especially since you can expand it to full screen if you want details.  But it would look nicer if we could somehow give a height hint.  For example I tried `data-video-width="100%"` with `data-video-height="350px"` and that looked very viewable. See below:

Screen Shot 2022-08-24 at 2 29 15 PM

PS - if you are looking closely you will notice that the margins are not asymmetrical, that is **me** not youtube.

### DIFF
--- a/xsl/pretext-runestone.xsl
+++ b/xsl/pretext-runestone.xsl
@@ -160,9 +160,8 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
             <!-- When hosted, we embed YouTube videos in a slightly different -->
             <!-- manner, and with the next script it is possible to monitor   -->
             <!-- reader events associated with the use of the videos          -->
-            <!-- NB: placed here just for initial testing w/ diffs, -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged, with some edits.

* This was on my (long) list from Tuesday Drop-In.
* I didn't realize it had a deadline.
* Thanks for the sizing experiments.  I figured these were going to be part of the changes, which was going to mean a lot more experimenting, so I was saving it for when I had time for that.

bnmnetp:
Thanks,  at least this will make the videos available, even though not perfect.

PR #1820, opened 2022-08-27
WeBWorK: don't mark webwork tasks as fillin-basic

Alex-Jordan:
This addresses the issue reported here:
https://groups.google.com/g/pretext-dev/c/taENGzE4wIw/m/hn4WGkGbCAAJ

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -1189,7 +1189,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
             <xsl:when test="statement and areas">
                 <xsl:text>clickablearea</xsl:text>
             </xsl:when>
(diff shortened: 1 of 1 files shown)


rbeezer:
That looks right - sorry for the omission.  Testing looks good - only mildly confused by the `.gitignore` (which are fine, just forgot about the ones more local to WW).  More praise on -dev.  Rebuilding website now.

Alex-Jordan:
I was going to check in this morning about that selector. I know you've learned how to be more efficient with those, and wondered if you see a better way.

You lost me with `.gitignore` though.

rbeezer:
Selector looked/looks good.

Was looking for diffs in testing in the wrong places.  git-aware editor showed representations file as untracked.  `.gitignore` in **root** didn't show it ignored.  Ahh, three `.gitignore` down in the webwork examples.  ;-)

PR #1827, opened 2022-08-31
new search page template

bnmnetp:
I just dumped in the search.html I had been working on CSS and JS included.  I know both of those should be factored out into  separate files.  But for this early stage its nice for fast prototyping.


rbeezer:
Thanks!  I'll get a couple of betas up in a couple hours.

I was getting a leading slash in the links, so I edited the one line to become:
```javascript
link.href = `${res.url}`;
```
which seems to behave better since all teh files are in the same directory.

bnmnetp:
Hmmm, I wonder why you were getting different behavior?  I tested this locally and had no problem bringing up pages.

rbeezer:
I'm not sure either.  Putting onto a server now, so we'll see how it behaves.

PR #1828, opened 2022-08-31
Update fr-CA.xml

JeanSebastienTurcotte:
Added interactive exercises feedback in french

### DIFF
--- a/xsl/localizations/fr-CA.xml
+++ b/xsl/localizations/fr-CA.xml
@@ -222,13 +222,13 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <localization string-id='evaluate'>Évaluer</localization>
     <localization string-id='code'>Code</localization>
     <!-- Interactive exercise feedback messages -->
-    <!-- <localization string-id='correct'>Correct</localization> -->
-    <!-- <localization string-id='incorrect'>Incorrect</localization> -->
-    <!-- <localization string-id='blank'>Blank</localization> -->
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Many of these terms will appear as the text on a button. As such, it's ideal if that text can be short. Can you imagine something short for `Générer une nouvelle version aléatoire`?  If not, it is what it is, and the button will probably break to a new line.

sean-fitzpatrick:
It might not be perfect, but `Randomiser` would be shorter. Is it somehow misleading in French?

Alex-Jordan:
Maybe we should look into nice icons for the buttons, and these text strings can just be their titles. So it won't matter if they are long. If `Randomiser` or something short works, great. If not, please don't let it hold this up. The long text will work, and maybe not even matter if icons come into play.

JeanSebastienTurcotte:
I based them on what we already have in our server. You can change them if
you want. See the attached file.

Le mer. 31 août 2022, 15 h 23, Alex Jordan ***@***.***> a
écrit :

JeanSebastienTurcotte:
Picture does not seem to go through github. I'll post it on -support at https://groups.google.com/g/pretext-support/c/pOTzg-y5X5U.

I'm fine with "Randomiser" if the longer name causes problem

Alex-Jordan:
@rbeezer Please go ahead and merge.

rbeezer:
Thanks, everybody.  Merged just now.  Of course, it may be a while until the CLI has it.

PR #1836, opened 2022-09-27
Search box

davidfarmer:
Inserts a search box in the banner which calls the recently implemented lunr search function,
when `debug.search.page` is set to `yes`.

Will submit issues to clean up the old search code.


rbeezer:
Thanks!  Merged with one whitespace change (an indent).

I've got about six follow-on things to do, but need to get talk ready for tomorrow.  Might be able to announce this later today.

PR #1847, opened 2022-10-23
Fix: Adjust menu items for logged in/out user

bnmnetp:
Should not show Login/Register when user is logged in.


rbeezer:
Had to remove a couple of tabs.  Possible they were there from long ago.  Thanks - that looks more rational.  ;-)

PR #1850, opened 2022-10-29
WeBWorK: support --restrict argument

Alex-Jordan:
This adds support for `-r` for building WeBWorK representations restricted to a subtree.

To test, temporarily edit `examples/webwork/Makefile` to have these recipes:

```
sample-chapter-representations:
	$(PTX)/pretext/pretext -v -r section-the-quadratic-formula -c webwork -p $(SMPCPUB) $(SMPC)
```

```
sample-chapter-html:
	-rm -r $(WWEXAMPLE)/sample-chapter/generated/latex-image
	$(PTX)/pretext/pretext -c latex-image -r section-the-quadratic-formula -f svg -p $(SMPCPUB) $(SMPC)
	-rm -r $(HTMLOUT)
	install -d $(HTMLOUT)
	$(PTX)/pretext/pretext -vv -c all -r section-the-quadratic-formula -f html -p $(SMPCPUB) -x webwork.divisional.static no debug.datedfiles no -d $(HTMLOUT) $(SMPC); \
```

All that is changed is the `-r` option.

Then run `make sample-chapter-representations` and you will see that only three WW are processed.

And then run `make sample-chapter-html` and open the resulting HTML in the scratch folder. All three WW exercise function.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -305,7 +305,7 @@
 
         </section>
 
-        <section>
+        <section xml:id="section-the-quadratic-formula">
             <title>The Quadratic Formula</title>
 
             <p>In the previous section, we saw relatively simple <webwork /> questions. This section demonstrates how even very complicated <webwork /> problems can still behave well.</p>
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -638,6 +638,7 @@ def main():
             xml_source,
             publication_file,
             stringparams,
+            args.xmlid,
             args.abort,
             args.server,
             dest_dir,
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -886,7 +886,7 @@ def tracer(xml_source, pub_file, stringparams, xmlid_root, dest_dir):
 
 
 def webwork_to_xml(
-    xml_source, pub_file, stringparams, abort_early, server_params, dest_dir
+    xml_source, pub_file, stringparams, xmlid_root, abort_early, server_params, dest_dir
 ):
     import urllib.parse  # urlparse()
     import base64  # b64encode()
@@ -901,9 +901,11 @@ def webwork_to_xml(
         global __module_warning
         raise ImportError(__module_warning.format("requests"))
 
-    # support publisher file, but not subtree argument
+    # support publisher file, subtree argument
     if pub_file:
         stringparams["publisher"] = pub_file
+    if xmlid_root:
+        stringparams["subtree"] = xmlid_root
     log.info(
         "string parameters passed to extraction stylesheet: {}".format(stringparams)
     )
--- a/xsl/extract-pg.xsl
+++ b/xsl/extract-pg.xsl
@@ -62,6 +62,11 @@
 <xsl:import href="./pretext-assembly.xsl"/>
 <xsl:import href="./pretext-common.xsl"/>
 
+<!-- Get a "subtree" xml:id value   -->
+<!-- Then walk the XML source tree  -->
+<!-- applying specializations below -->
+<xsl:import href="./extract-identity.xsl" />
+
 <!-- Override the corresponding param in pretext-assembly so that webwork  -->
 <!-- copies can be made.                                                   -->
 <xsl:variable name="b-extracting-pg" select="true()"/>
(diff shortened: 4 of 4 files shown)


rbeezer:
Some suggestions based on a visual review.

Took me a little bit to figure out what makes this would work.  Basically the "extraction" template matches any element, but will only be applied once, then a massive "select" is used to round up all the webwork elements.

Whereas the "extract-identity" template is meant to provide careful management of the subtree selection **and** provide the entry template.  Then items of interest are rounded up by traversing the entire tree.  This is meant to keep from providing this infrastructure over and over.

But I see the problem: there is a bunch of one-time setup that needs to come first.

Suppose `extract-identity.xsl` had a new stub/trivial template that wrapped all the extracted items allowing for one-time material before and after all the items?  "wrap-extraction"?

You could put the one-time setup in a redefinition of the stub, and then "dictionaries" gets renamed as "extraction".  Done.

I see that I've confronted the same problem in `extract-interactive.xsl`, by redefining the entry template and totally neutering the whole subtree feature.  :-(  Maybe some other places as well.  :-( :-(

So I'd suggest the stub suggested above is then a necessary part of `extract-identity.xsl`.

What if I enable it, test with interactives, and provide guidance here on how to use with WW?

Rob

Alex-Jordan:
> What if I enable it, test with interactives, and provide guidance here on how to use with WW?

That sounds OK, but I could also move forward with starting the other way round. I could get it working here with WW and then you could apply it to those other things. I can go either way but I'm a little eager to get this feature (restrictions on WW) all the way through to the CLI because soon-ish, my colleagues (including new-to-PTX colleagues) will be working on ORCCA. If I can train them with the CLI and nothing more, that will be a big deal. And restricting to a subtree for WW will also be essential.

> But I see the problem: there is a bunch of one-time setup that needs to come first.

Exactly. It took me a little bit to understand it all the same as you've outlined. It's a nice mechanism you built, well done.

So to clear it up for myself, are you suggesting that the subtree (which can default to the entire tree) always gets wrapped in a new root level element named "wrap-extraction"?

rbeezer:
A template, not an element.

```xsl
<xsl:template match/name? mode="extraction-wrapper">
  <xsl:param name="the-subtree-root">
    <!-- do some set-up -->
   <xsl:apply-templates select="$the-subtree-root" mode="extraction"/>
  <!-- do some wrap-up -->
</xsl:template>
```
would be the stub version.  Or something like that.  Importing stylesheets would override this by fleshing out the comments (but preserving the application of the `"extraction"`).

* Maybe the subtree root is passed in via a select?
* Maybe this can be organized so the use is just an "apply-imports"?

I'll see if I can get to this by Tuesday's Drop-In.  Can you attend that?  I'm not in favor of development being rushed by a release schedule.  Maybe you can alert the CLI to watch this PR?

Alex-Jordan:
I can't make Tuesdays (or Thursdays). I can be there Friday though (but closer to 1:00).

December 2 is when the committee overseeing ORCCA next meets. I'm optimistic that this can be in play ahead of that so I can demo some things at that meeting. Of course, it is not an imperative. Sometimes these quasi-necessities are the mother to my getting some coding done is all.

rbeezer:
OK, `extract-identity.xsl` is in much better shape now

4edcab6352611c6593ecfce82558d0bf202cec10: changes

090b130d9145ec14b7b2814434225aab3430edec: much better documented

0a7aa2025cfb47111d87eab8c3614513523f6b48: simple example in `extract-interactive.xsl`

*  Migrate your initializations in the entry template to a "extraction-wrapper" template, like model inside `extract-identity.xsl`
*  Change `mode="dictionaries"` to `mode="extraction"`

A force-push might be best.  Comments on documentation welcome.

I may test this out further with `extract-math.xsl`.  Very non-trivial.

PS you will need to alert CLI that signature of the Python WW function is changing to include subtree root.

rbeezer:
> I may test this out further with extract-math.xsl

At 3ea1b8d49f4fcc2e9b9887a51058c69b459b58b5

Alex-Jordan:
I made those changes and force pushed. Same testing routine works for me.

rbeezer:
Excellent.  Doesn't that feel better?  ;-)  Cleaner, anyway.  Rebuilding website right now (along with some other stuff).

rbeezer:
Did an extra test.  Restricted representations, but no restriction for HTML build.  Should we drop a text warning in HTML when there is no representation?  (We do that for static/LaTeX, no?  Forgot to check that one also.)

Alex-Jordan:
Yes, and it's easier for someone else to understand now. You know, after you and I each get hit by that bus...

PR #1854, opened 2022-11-09
WeBWorK: fail gracefully when iamge formats are missing

Alex-Jordan:
Occasionally when WW processes a latex-image, there may be a failure on the WW server end to produce one of the image formats. (This happened to me recently with the png format on one image. It is ultimately a flaw with how WW uses dvipng to make pngs.)

Prior to this PR, that leads to an error and everything halts. This PR allows processing to proceed and logs an alert that some image format was missing from the tarball file.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -1391,10 +1391,9 @@ def webwork_to_xml(
                 msg = "PTX:ERROR: there was a problem downloading an image file,\n URL: {}\n"
                 raise ValueError(msg.format(image_url) + root_cause)
             # and save the image itself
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks for the increase in robustness!

Merged as-is, but I had to correct the spelling in the commit message.

I saw the opportunity for tightening up some repetitions, so went ahead and did that.  I really should have made a pull request for you to review.  Could you have a look anyway?  At dc775487d0907abe85005d2851d6fd2bb9c503b6.  Testing indicates no harm.

PR #1863, opened 2022-11-22
Make paragraph column error nonfatal

Alex-Jordan:


rbeezer:
Very good.  I like it.  Thanks for hopping on this one.

PR #1866, opened 2022-12-01
Guide: commas as thousands place separators

Alex-Jordan:
Here's a small thing for math best practices.


rbeezer:
Thanks!  (That was fast.)

I'm going to run it by @zorkow  in the morning and make sure we aren't messing up heuristics for braille and speech.  ;-)

rbeezer:
@zorkow blessed this.  And I added a follow-on commit with a bit extra at the end, specifically an example of **exactly** how to author this.  And then I couldn't help myself and showed the **bad** example with the commas giving space.

Thanks for the improvements to this one.  It'll go public on next refresh.

PR #1867, opened 2022-12-12
Fix: Use an import from the CLI to produce relative pathnames 

bjones1:
Specifically, relative path names in the HTML mapping from source paths (relative to the project path) to XML IDs.

The history: the original add in b4395ea3, which performed the mapping for all HTML builds, was made optional in ea4acc57 due to security concerns -- the mapping file contains potentially confidential information (username / local filesystem paths) to the mapping file, which might be published to the web. This update now make the mapping default behavior, but only stores a relative path to address the security concerns.

### DIFF
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -24,15 +24,15 @@
 
 import pretext as ptx
 
-# Set up logging (replacing _verbose/_debug).  
-# Use log.info(<string>) or log.debug(<string>).  
-# Also available now are log.critical(), log.error(), and log.warning(). 
+# Set up logging (replacing _verbose/_debug).
+# Use log.info(<string>) or log.debug(<string>).
+# Also available now are log.critical(), log.error(), and log.warning().
 # 0-level verbosity will be set to only show log.critical() messages.
 import logging
 #ptxlogger will be compatible with the CLI logger.
 log = logging.getLogger('ptxlogger')
 log.setLevel(logging.DEBUG)
-# Create stream handler for console output.  
+# Create stream handler for console output.
 # Later we will set the level with log.setLevel(logging.[level])
 console_log = logging.StreamHandler()
 log_format = logging.Formatter('PTX:%(levelname)-8s: %(message)s')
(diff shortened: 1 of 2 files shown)


bjones1:
@oscarlevin and @StevenClontz, this would make using CodeChat in CodeSpaces a bit easier. Feedback is welcome!

After this is merged, I'll need to update CodeChat so that it expects relative instead of absolute paths.

bjones1:
Good point. This code is currently in the core `pretext.py` since it's (as I see it) a part of the processing pipeline used to transform PreTeXt source to HTML. That makes the most sense to me from a structural perspective. However, the code needs to know the project directory, which only seems available from the CLI. So, given the `import`s needed, I suggest we put it in the CLI. Putting such a small amount of code in an entirely separate package seems like overkill to me.

StevenClontz:
I view the CLI as focused on "authoring interface and project management", so this seems like a reasonable port unless Rob disagrees.

bjones1:
OK, I'll wait to hear from @rbeezer. Assuming he agrees, I'm guessing I'd put this in the CLI's `build.py`?

StevenClontz:
I think `codechat.py` would be fine.

rbeezer:
Sorry for the delay.  Just to make sure I understand correctly.  The routine `map_path_to_xml_id()` would leave `pretext.py` (along with associated calls, etc.).  Then development work with CodeChat features would take place with the CLI going forward.

If I have that right, then it sounds fine to me.  Thanks for letting me check-off.  I'll be at Tuesday Drop-In if this needs more discussion.  @davidfarmer may need an alert to the change.

bjones1:
Rob, correct re: move from `pretext.py` to somewhere in the CLI. I'll be at today's drop-in for any discussion. David, any thoughts/preferences?

bjones1:
I'm closing this in favor of https://github.com/PreTeXtBook/pretext-cli/pull/370. I'll submit a new PR to remove the mapping code from this repo.

rbeezer:
> I'll submit a new PR to remove the mapping code from this repo.

Thanks, that'll work well.  Let me know on that PR if there are considerations for timing with the CLI PR.

bjones1:
Will do.

bjones1:
See #1868, which removes the mapping from this repo.

PR #1874, opened 2022-12-18
WeBWorK: move Activate button

Alex-Jordan:
This PR is for discussion, until some details can be worked out. I'll start a discussion thread now and reference this PR.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -10410,6 +10410,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:attribute name="aria-live">
             <xsl:value-of select="'polite'"/>
         </xsl:attribute>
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Force pushed this following a rebase over master and resolving a conflict.

rbeezer:
Is this ready, then?  Do you want it now, or still need a pause?

Alex-Jordan:
No, this is not good for at least two reasons.
* Having the button off to the side doesn't work well with narrow situations like two-column or phone.
* When for example Runestone renders an exercise independently, outside of the actual book, it does not get the pretext_add_on CSS applied and then this button is really bad.

So more experimentation/research is needed.

bnmnetp:
Alex,

I can include the required CSS **if** I know about it.  The reason I hadn't included it recently was because I've not had to add CSS from PreTeXt or WebWork on the assignment page in the past.   If it makes sense to to do that I'm perfectly happy to do so.

PR #1875, opened 2022-12-18
Add @include attribute for <program>.

jdeisenberg:
Runestone allows a `.. include` in activecode; this allows you to have several activecode examples that build on one another without having to duplicate the code.  PreTeXt does not have this capability; this PR adds a `include` attribute to the `<program>` element; its value is a whitespace-separated list of `id`s for the activecode segments you want included.


rbeezer:
Thanks, @jdeisenberg, for starting this.  Getting through PreTeXt `@xml:id` and '@label` was a bit of a dance.  And I split off the one typo correction - thanks for that.  Pushing up new examples right now at the end of

https://pretextbook.org/examples/sample-book/noparts/html/activecode.html

Give it about another 30 minutes to finish.

Issue #468, opened 2016-12-17
shorten WeBWorK tabular halign output


Alex-Jordan:
Presently putting something like 'halign="center"' in a webwork `tabular` will put halign specification into every single cell. This can quickly make even a modest table bump up against the cap on url character length.

It doesn't have to be this way. I was just lazy when initially writing the tabular template. It is possible to have the tabular's halign attribute feed into the DataTable's top-level arguments instead of into every single cell argument. I believe there is a clash with MBX's row/column priorities compared to niceTable.pl's row/column priorities, and I wanted to work around that by putting alignment into every single cell after it was calculated the MBX way. But relieving the url character cap would be worth figuring that puzzle out.

PR #1876, opened 2022-12-20
WeBWorK: update table routines

Alex-Jordan:
This one needs some discussion because of the dance between a rolling PTX release and annual WeBWorK releases.

The effort here revises the PG table-making routine to be better. A "big" problem is https://github.com/PreTeXtBook/pretext/issues/468. Recent conversion of Active Calculus preview activities to WeBWorK butts up against he URL character length cap because of this. So one thing this does is remove a lot of redundancy that was present from the previous brute force approach.

Also this change captures almost all tabular features that we have. For example it captures a `right="major"` whereas previously, that essentially got morphed into `right="minor"`. I believe the only things I'm not trying to capture now are `top` on a `col`, and `top`,`bottom` on a cell.

OK, but then the problem gets to the server and a file called `niceTables.pl` processes it. For PTX output (for static renderings) that file has a lot of hacks that rely on the former "brute force" handling. So even though better code is being sent to the server, now it comes back missing more in the static. Notably, vertical rules are gone.

In parallel, I have rebuilt `niceTables.pl` from scratch with more modularity in mind, and with the new `niceTables.pl` (destined for WW 2.18) we get back good PTX static tabulars, nearly totally faithful to the author's source.

For diff testing (if you want to):

1. Make the sample chapter representations file before this commit.
2. Make it after this commit (so it will still be using the old `niceTables.pl`.
3. Make it again, but in the publisher file, change the course to `active-calculus-single-mbx` where I have set up the new `niceTables.,pl`.

If you diff 1 versus 2, and then 1 versus 3, you should see what I have described.

Now I do not know what should be done. Options:

1. Nothing, until WeBWorK 2.18 comes out and we have it running on the AIM server. Then merge this and anyone relying on a 2.17 server will lose some cosmetic table features.
2. Merge this now. Everyone loses some cosmetic table features but we are forward thinking.
3. Merge this now, and I update `niceTables.pl` in the AIM host course. Everyone using the AIM server gets all the benefits right away. I could post to -announce with details for how other people could upgrade `niceTables.pl` too.

Each option above also has the possibility that I do a lot of work to support 2.17 servers too. I'm not keen on this because (a) it seems like a lot of work for not much benefit, and (b) I think it is impossible to do that and still address the too-long-URL problem.

### DIFF
--- a/xsl/extract-pg.xsl
+++ b/xsl/extract-pg.xsl
@@ -2524,8 +2524,6 @@
 </xsl:template>
 
 <xsl:template match="tabular" mode="components">
-    <!-- PTX tabular attributes top, bottom, left, right, halign are essentially passed -->
-    <!-- down to cells, rather than used at the tabular level.                          -->
     <xsl:param name="b-human-readable" />
 
     <xsl:text>[@DataTable(</xsl:text>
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I completely forgot to mention something. In addition to the purposes already mentioned, this lays the ground work so that a "webwork" can have "sidybyside" and "sbsgroup". They will translate to the `LayoutTable()` in the new `niceTables.pl`.

Alex-Jordan:
I just force pushed this following a rebase over master and condensing to one commit.

Alex-Jordan:
We decided to merge this (pending a review) in about two weeks (mid January). I posted to -announce. At the same time this is merged, I need to tweak a few small things on the AIM WW server.

rbeezer:
Put this through some paces, so I'd be ready when the time comes.  Looking good.  Thanks for the careful instructions for the diffs.

One versus Three: a stray "hi" is being inserted.  Before a "statement" of a "task"?  Debugging device?  I don't see it in One versus Two, and I don't see it in the diff for the XSL itself, making me think it is in `niceTables.pl` for AC?

That post on `-announce` was **perfect**.  I've made myself a January 16 reminder to merge this.

Alex-Jordan:
Thanks for testing this. Funny timing, yes, I'm debugging something over on the AC course. But "hi" is coming from my edited`PGML.pl` in that course.

What am I debugging? Well it turns out `f''(x)` works fine inside a table. So does something like `f'(x)+g'(x)`. But if the table has an odd number of single quotes, it just disappears. No error message on the WW side. And not fixed by my wrapping in that extended ASCII character (which is still good for a different reason). I'm slowly tracking down this PGML bug. I should have done that somewhere else, I wasn't thinking you or anyone would have reason to use that course.

rbeezer:
> I wasn't thinking you or anyone would have reason to use that course.

I'm quantum.  Everywhere at once.

rbeezer:
Rebased on `master` locally and building PDF without errors, so pretty much ready to go on my end.  Holler once ready.

Alex-Jordan:
Thanks. The rest of the day was a slog and I'm retiring early tonight. Plannning to hit this tomorrow morning.

Alex-Jordan:
OK, the AIM production WW server has the new macros in place. Ready to merge this.

I will do the same now to the AIM dev WW server.

rbeezer:
Will not update website sample chapter until I hear from you.  Here, or on "update" thread, or ...

Alex-Jordan:
It should be good, but I haven't tested with a build. I can do that, but it would need to wait until this evening. I'm not in a rush if you aren't.

rbeezer:
At some point I will have a need to rebuild.  But I'm not seeing anything that'll happen this weekend that would prompt that.  Eventually it will just happen.

Alex-Jordan:
I built the sample article, HTML and PDF. Looks good. But for a moment there I was scratching my head. It turned out a file local to the anonymous host course was overwriting the new table making macro file. Deleted it, and then things are good.

#1882 would make certain tables look better in static form, like the one in Checkpoint 8.1.

PR #1878, opened 2022-12-21
Add "datafile" attribute to <program>.

jdeisenberg:
The runestone `:datafile:` option specifies where to find a file if an ActiveCode script needs to open and read a file. This PR adds that attribute to `<program>`

### DIFF
--- a/xsl/pretext-runestone.xsl
+++ b/xsl/pretext-runestone.xsl
@@ -1734,6 +1734,22 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                                     </xsl:otherwise>
                                 </xsl:choose>
                             </xsl:attribute>
+                            
+                            <!-- allow @datafile attribute on <program> -->
+                            <xsl:if test="@datafile">
+                                <xsl:attribute name="data-datafile">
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, been busy with a different Runeestone `datafile` which was a big job.

Notes for me:  Brad says just one file per program 9as implemented here already).  Rob says a (new) `dataurl` element might be a good static representation.

rbeezer:
Have you been able to use this successfully?  I can't get various combinations of `@datafile` attribute, `open()` commands, using different relative paths and locations, to open a file.

This includes the dead-simple version of putting the file in the same directory as the page and using no paths at all.  Do you have an example I can test with?

bnmnetp:
@jdeisenberg The `@datafile` option is only useful for running programs on the JOBE server.  It says find the data from the `datafile` directive and upload it to the server before running this program.

rbeezer:
@jdeisenberg - I slimmed down your commit, but left your name on it.  Added an example to the "sample book" (going live ritht now) and will return to this and add support for multiple files (if this one passes muster).

Thanks for the contribution!

PR #1879, opened 2022-12-23
Use HTTPS protocol for CDNFonts.com

StevenClontz:


StevenClontz:
This is necessary for the import to work on HTTPS-served documents.

image

rbeezer:
Good catch!  Thanks.

PR #1880, opened 2022-12-26
Schema support for xincluded fragments

oscarlevin:
This modifies the schema by:
1. Puts the `pretext` element into its own definition, separate from `start = ...`
2. Defines `start = ...` to include a choice of (almost) all divisions and blocks.

This will allow for real time schema validation in VSCode and other editors with rng support when fragments of a PreTeXt document are included using `<xi:include>`.

### DIFF
--- a/schema/pretext.xml
+++ b/schema/pretext.xml
@@ -63,14 +63,27 @@
     </introduction>
 
     <section>
+        <title>Start Elements</title>
+
+        <p>To support modular source files, we specify which elements can naturally be the root of a <em>fragment</em> file in a <pretext /> document.  These include the <c>pretext</c> element itself, as well as most divisions and blocks.  All of these are defined as elements below.</p>
+
+        <fragment xml:id="start-elements">
+            <title>Start elements</title>
+            <code>
+            start = Pretext | DocInfo | Part | Chapter | Section | Subsection | Subsubsection | Paragraphs | ReadingQuestions | Exercises | Subexercises |  Solutions | ArticleAppendix | BookAppendix | IndexDivision | References | Glossary | BlockText | BlockStatementNoCaption | BlockStatement | BlockSolution | BlockDivision
+            </code>
+        </fragment>
+    </section>
+
+    <section>
         <title>Gross Structure</title>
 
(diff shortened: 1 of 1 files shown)


rbeezer:
Sorry, been meaning to comment on original thread.  I'm not a fan of most projects including small pieces of source, except perhaps for TikZ code that is used in more than one place.  And nobody is prevented from doing that, it just looks ugly in VSCode.  Once folks are including blocks, maybe they know well enough to not be thrown off in VSCode?

But I'm all in favor of divisions!  And where to draw the line?  I could argue two ways, as usual, for `paragraphs`.

Would you be amenable to starting with divisions?  We could revist smaller chunks as necessary.

oscarlevin:
I think that divisions is very reasonable for drawing the line.  I assume that would include exercises?  Glossary?  References?  Not that I would include the latter two.  

I could see a good argument for figures, but until we have someone reasonably complain about the warnings, I think divisions is a good middle ground.

I'm happy to add a commit to fix this, but it might be easier for you to do so.  Just let me know.

rbeezer:
Thanks for the reply.

>  I assume that would include exercises? Glossary? References? 

Yes!  And `worksheet` and `reading-questions`.  And I meant to say, `frontmatter` and `backmatter` make good sense, along with (the shrinking) `docinfo`.  `solutions` is a "generator" so a bit odd - but maybe for completeness?  So think about those "big" units that also make sense.  Portions of the front matter?  `preface`, `acknowledgement`, etc?  Your call, and it will be easy to adjust if we miss something.

I'd love it if you can adjust the commit (pile on and I will squash, or force-push; your call).  I have a script for all the derived versions, and we'll see how the schema browser reacts.

Thanks for figuring this one out and acting on it.

oscarlevin:
Done.  Thanks for considering this.  It will make the editing experience in VS Code a lot nicer.

rbeezer:
A little hard for you to debug without all the tools.  ;-)  Fragment needed to be incorporated in a grand collection at the bottom.  And one missing pipe/or.  Both added silently.

Minimal impact on `pretext.xsd` which means minimal impact on the schema browser it seems.  `pretext` leaves a "top-level elements" section, but `interlude` remains - not sure what that is all about.  ;-)

Anyway, great work and I hope it works well in the VSCode setting/application.  Feel free to advertise how and where you feel necessary.  Updating website and schema browser right now.

We can update/improve as necessary.

PR #1882, opened 2022-12-28
WeBWorK: when a static tabular cell contains only a fillin, cut the f…

Alex-Jordan:
…illin

In the context of a WW exercise, there may be a tabular with some cells that only have answer blanks. As with `p` that only contain a `fillin`, it is better in the static version if these `fillin` were just gone.

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -607,6 +607,10 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <!-- preceding siblings"                                           -->
 <xsl:template match="webwork-reps/static//p[not(normalize-space(text()))][count(fillin)=1 and count(*)=1][not(parent::li) or (parent::li and preceding-sibling::*)]" mode="webwork"/>
 
+<!-- and likewise when a tabular cell contains only a fillin,      -->
+<!-- keep the cell but not the fillin                              -->
+<xsl:template match="webwork-reps/static//cell[not(normalize-space(text())) and count(fillin)=1 and count(*)=1]/fillin" mode="webwork"/>
+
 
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
Do you have thoughts on this one?

Here is a place where it would have an effect (if Oscar rebuilds):
[Exercise 2](https://runestone.academy/ns/books/published/MATH228-Spring23/sec_propositional.html#exercise-62) 
https://runestone.academy/ns/books/published/MATH228-Spring23/sec_propositional.html#exercise-62

The little underlines would go away from the static preview. (Also if Oscar rebuilds, the table will have all its vertical and horizontal lines that separate grid cells, now that the tables overhaul is merged.)

rbeezer:
*  Can the two filters be combined into one?  Use two `and` with three booleans in one filter?  There's no context change between the first and second filter.
* Do we have some tests in the sample chapter?  Not off in Oscar's stuff?  One fillin, two fillin, fillin plus other element, some text or not.  Situations that will expose anybody tinkering with this later.

Alex-Jordan:
Hmm, we do have an example here:

https://pretextbook.org/examples/webwork/sample-chapter/html/section-8.html

But when I build it on this branch, with the `fillin` removed, those cells are so tiny. Without the width from the `fillin`, they look like this: 

Screen Shot 2023-04-02 at 9 01 13 PM

It's not good that the cell no longer displays with the author's intended width.

What if, upon the match from this PR, the `fillin` gets some sort of `pi` attribute and that leads to an "empty" fillin (in both HTML and LaTeX)? No underline, no shade, no box. Just the author's horizontal space.

rbeezer:
> some sort of pi attribute

Sure, I think that could work cleanly.

PR #1883, opened 2022-12-30
Webwork array

Alex-Jordan:
This makes it so `fillin` can have `@rows` and `@cols`.

Current implementation: if either is present and greater than 1, then the answer blank is followed by ` (m x n array)`. The word "array" is localized. The "x" is the times-character. The spaces in `m x n` are nbsp.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -1685,6 +1685,37 @@
 
             </exercise>
 
+            <exercise>
+                <title>Answer Arrays</title>
+                <webwork>
+                  <pg-code>
+                    Context("Point");
+                    $p = Point(0,0);
+                    Context("Vector");
+                    $v = Vector(1,2);
+                    $c = ColumnVector(3,4);
+                    Context("Matrix");
+                    $m = Matrix([[1,2],[3,4]]);
+                  </pg-code>
+                  <statement>
+                    <p>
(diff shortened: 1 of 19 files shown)


rbeezer:
How about one big template in `-common`?  You are just producing text, like ` (2 x2 array)` and since you are using the character templates (very good!) you can just make one base/abstract template?  Maybe a mode like "webwork-fillin-array"?  I've looked closely for a difference, but I'm not seeing any.

Alex-Jordan:
Do you want this to be webwork-specific? As in `@rows` and `@cols` are only something that comes back from a WW server, and never something an author writes into a `fillin`? I was assuming not, and added the attributes to the schema.

Alex-Jordan:
Force pushed with a common template now. I thought to use an `xsl:if` to reduce calls to the new template, but am unsure if that will have any real impact on profile.

rbeezer:
Looking good, visually.  May merge tomorrow morning.  Where the time goes is not always what you expect.  And an `xsl:if` hear and there isn't really noticeable I think.

rbeezer:
OK, all buttoned-up.  Thanks, this worked well.  Missed your comment about authors yesterday.  Yes!  So I did include the schema change.

But that means...can you add one example to the sample article for more regular testing, and especially add it to the documentation.  Only needs a sentence, I think.  Someplace in Chapter 4?  Thanks.  (Noted as #1886.)

PR #1884, opened 2022-12-30
Showcase article: update WeBWorK representations

Alex-Jordan:
I noticed that the second Showcase WW exercise was not working:

https://pretextbook.org/examples/showcase/html/webwork.html

It turns out this is because we are committing the representations file instead of regenerating it for each build, and it is committed with a version back when we were on 2.16. So this commit updates the representations file. There were apparently changes to the .tex for the image here as well, so I committed new versions of the image. And I added the sample chapter's macro file (that contains latex-image-preamble for WW to use) to the repo so it need not be regenerated in the future, barring changes to latex-image-preamble.


rbeezer:
Thanks!  Likely update website later today.  PDF builds, perhaps have a peek at HTML?

Next time you are in this source, there are some `stage` that should become `task`.

PR #1885, opened 2022-12-31
Splitting BRF into chapters update

alexeikolesnikov:
Tested the code with "chunking" = 1 on the sample book.


rbeezer:
PR looks great procedurally - I have it locally right now.  I'll likely dig-in in the morning.

rbeezer:
Very good!  Thanks.

I smashed the commits together and retitled the one commit.

Some whitespace needed to be removed (blank lines with a few tabs, some trailing stuff at at the end of lines).  Gotta eradicte that sort of stuff because it becomes a git problem very quickly.

I turned orphaned triply-quoted strings into lines of comments.

And a couple of very minor edits, but none of it touched code.

Built one massive file for AATA, then rebuilt chapter-by-chapter.  The did a diff with Chapter 1 chunk, against Chapter 1 edited out of massive file by hand.  Expected no change, got **everything** changing.  But if I get both up in my editor and switch tabs, I see no changes.  Maybe the types of spaces I use for math?  Still - very odd and I can't explain it.  But not enough to stop adding this!

alexeikolesnikov:
Thanks!

On the test book, I pulled out the first chapter in a text editor and did
the diff with
--strip-trailing-cr
This killed all the differences. Tried adding '\n' at the end of each line
(bad idea) and '\r' (not immediately bad, but did not resolve the
differences).

With the current version, produced Chapter 1 chunk of test book and grabbed
Chapter 1 from the large file. Embossed both versions on Romeo, they are
indistinguishable. So the current version works.

One note from embossing: the newpage character at the start of the chapter
produces a blank page. If needed, I can remove it, but this could be
counted as a feature: Al likes having a blank page at the start of the
volume: it works as a cover when the volume is bound preventing damage to
the first page.

--Alexei

rbeezer:
> On the test book, I pulled out the first chapter in a text editor and did
the diff with
--strip-trailing-cr
This killed all the differences. Tried adding '\n' at the end of each line
(bad idea) and '\r' (not immediately bad, but did not resolve the
differences).

Aah-ha!  Thanks.  And I like `git diff` as my primary tool where `--ignore-space-at-eol` works very well.

> Embossed both versions on Romeo, they are indistinguishable. So the current version works.

Very good.  Thanks for thinking to check that.

I think this describes the symptom, though I wouldn't dare try any of those (global) git changes:

https://stackoverflow.com/questions/1889559/make-git-diff-ignore-m

It'd be nice to get to the bottom of this, i.e. where Python is changing line endings.  We might get caught out on other embossers, where a full BRF and a chapter-busted BRF act different?  Perhaps unlikely, but still...?

Do you think iterating over the list of lines and using `write()` (rather than `writelines(self.body)` one-time) would (a) do the right thing, or (b) allow more control over adding a line ending?

rbeezer:
> One note from embossing: the newpage character at the start of the chapter
produces a blank page. If needed, I can remove it, but this could be
counted as a feature: Al likes having a blank page at the start of the
volume: it works as a cover when the volume is bound preventing damage to
the first page.

Without any impact on page-numbering, I think that is fine.  And sighted folks might want to scribble on that first page to keep track of the volumes.

mitchkeller:
>  One note from embossing: the newpage character at the start of the chapter produces a blank page. If needed, I can remove it, but this could be counted as a feature: Al likes having a blank page at the start of the volume: it works as a cover when the volume is bound preventing damage to the first page.
> […](#)

Can this be clearly documented in a README of some sort that is produced when the .brf files are generated? I sent a .brf of one chapter of AATA to our disability services office for embossing so I could use it as a demo during a talk, and he was very confused and thought that something was wrong with the file.

rbeezer:
I think we would have the content, at the right time, to do something like
```
Volume Cover Page
Chapter 5 Groups
```
in braille, on that blank page.  Yes, it will get trashed (unless we cover it with a blank page!) but it might help the **blind** reader to organize multiple volumes.

alexeikolesnikov:
Yes, I can work on extracting the name of the chapter when TOC is parsed
and then prepend an unnumbered page at the beginning.

@Mitch, which chapter did you pick? Are there images? If yes, and the
disability services have an embosser that can handle PDFs, do you also want
to add the images?

mitchkeller:
@alexeikolesnikov This was back in September. It was Chapter 13 and then I grabbed a few image files separately. The embosser here is not  one that can handle PDFs. Michael Cantino helped me work through some things to get the guy some stuff, but the pages were not great. I can try to describe in person if you'll be at JMM. Alas, the embossed pages are in my office, and I don't plan to get back there before heading to the airport.

PR #1891, opened 2023-01-03
Dutch localization

chosia:
Created a Dutch localization file (nl-NL).


rbeezer:
Thanks!

* I added your name to the overall README.
* Removed the "commented-out" US language overall element.
* **One** stray ">" was missing.  Took me a while to track that down!  ;-)

Since I added code, combined commits and added the PR number, you will need to get the new commits and trash your old branch.  Read about "Cleaning Up" in The Guide.  And if anything seems suspect, please see current `pretext-dev` thread about this and chime in!

chosia:
Ah, sorry about that. I should know better than committing untested code. I thought I'd quickly add a few missing translations :-) Thanks for fixing it.

PR #1892, opened 2023-01-11
Latex multiline headers

Alex-Jordan:
First commit here adds tables to the sample article with header cells that have multiple line. LaTeX output does not boldface such content. (Discovered while working on a project where a multiline column header would be good to use.)

Second commit changes -latex to use \bfseries around such cells. Sample article output looks good.

There is a comment in the current -latex:
```
<!-- adding \textbf, \bfseries not 100% effective here -->
```
I'm unsure how to provoke how it might not be effective. But it may be that this comment is wrt trying to use a boldface command within the `\tablecelllines` argument. But the approach I've used puts the boldface command around the whole `\tablecelllines` macro.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -5710,6 +5710,42 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             </table>
 
             <table>
+                <title>One Row Header, Multiline</title>
+                <tabular>
+                    <row header="yes">
+                        <cell>State</cell>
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I doubt you'd already looked at this, but just in case, a heads up that I just force pushed. I remembered it's not great to nest a `choose` inside a `choose`, so I moved an inner `choose` to a new modal template.

rbeezer:
Thanks!  That's great.  Rebuilding website now...

PR #1894, opened 2023-01-13
Playwright replace pyppeteer for interactive previews

dbrianwalton:
Modified pretext/pretext.py to use the playwright module in place of pyppeteer. Pyppeteer is seriously lagging and using a 2018 version of Chromium, so that the features of javascript are not current. Playwright is being actively developed and sponsored by Microsoft and will be a more reliable remote tool for accessing the DOM.

Note that playwright supports three different browsers, while we only access the chromium tool. The default install of playwright will download all three.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -1791,10 +1791,10 @@ def preview_images(xml_source, pub_file, stringparams, xmlid_root, dest_dir):
     # imported here, used only in interior
     # routine to launch browser
     try:
-        import pyppeteer  # launch()
+        import playwright.async_api  # launch()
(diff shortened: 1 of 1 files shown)


oscarlevin:
We can update the CLI requirements to match this when it gets approved.  Mostly posting here so I get updates.

rbeezer:
Really nice work - love the split into two topical commits.  All merged in.

I edited the commit messages slightly, as I do, so rebased and thus you will need to trash your branch after getting this one.

A small bit of weirdness on my end.  I'll post on the lists in a bit.

PR #1895, opened 2023-01-14
Fix call of template syntax-highlight in pretext-interactives.xsl

dbrianwalton:
Fix call of template syntax-highlight in pretext-interactives.xsl which originally had "-header".


rbeezer:
Thanks!  Merged with minor commit message edit.

Sorry for the fiddle, I shoulda just fixed this yesterday when you mentioned it.

PR #1896, opened 2023-01-14
Implement local server for generating interactive previews

dbrianwalton:
This code builds off the addition of playwright and adds in the creation of local html representations for all of the interactive elements as individual pages (using pretext-interactives.xsl), saved to the temporary directory in a folder named "html-interactives". A subprocess is created calling python's http.server to serve this directory to localhost with a default port 8888, and if the server halts prematurely (typically port is already used), then it attempts to use different ports. Once the server is verified as still active, an asynchronous loop is called using playwright to open the pages showing the interactives and generate the resulting preview images. Once the loop is complete, the server is terminated.

Comment: The pretext-interactives.xsl generates a single file containing all of the interactives. I'm not sure that this page was necessary unless it prompted the creation of the other individual pages for single interactives (possibly using knowl generation templates to do so?).

Comment: extract-interactive.xsl creates a list of fragments and puts the baseurl for the project in the first spot. Since we are no longer referencing the baseurl, should we remove that first entry and only create the list of interactive fragment ids?

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -1796,44 +1796,41 @@ def preview_images(xml_source, pub_file, stringparams, xmlid_root, dest_dir):
         global __module_warning
         raise ImportError(__module_warning.format("playwright"))
 
-    # Interior asynchronous routine to manage the Chromium
-    # headless browser and snapshot the desired iframe
-    async def snapshot(page, input_page, fragment, out_file):
-
-        # page:       browser page object loading the URL
-        # input_page: the "standalone" page of the interactive
-        #             hosted at the base URL
-        # fragment:  the hash/fragment identifier of the iframe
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, @dbrianwalton!  Merged as-is, pushed, and rebuilding website right now.  Nice work.

I added on a small commit to trash the base URL business.  Tried a 15 second delay, no improvement to numerical integration Sage interact.  Update previews for the sample article, so that fills a few holes, I think.

A couple of nits:

*  Running with sample article, I see one 404, but could not figure just what was missing.  Any ideas.
*  The stylesheet building the page cannot seem to get a localized version of the button text for Sage interacts.  (I think.)  Probably on me to sort that out.

PR #1905, opened 2023-01-16
Resolve sample article interactives jumping focus by making perspecti…

dbrianwalton:
See Issue #1086.

Using an attribute when creating the Geogebra applet avoids the setting of focus in the script. Propose that we add the attribute to the Geogebra slate and take it out of the script.


rbeezer:
Yahoo!  No more jumping to GeoGebra!

The placement of the "otherwise" confused me for a bit, so I added a code comment.

Thanks - glad to have this resolved.

PR #1906, opened 2023-01-17
French quotation marks html and Latex

JeanSebastienTurcotte:
Added code so that <lsq/>,<rsq/> and <q/> are french when fr-CA or fr-FR are selected as document language.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -8773,12 +8773,26 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
 <!-- Left (Double) Quote -->
 <xsl:template name="lq-character">
-    <xsl:text>&#x201c;</xsl:text>
+    <xsl:choose>
+        <xsl:when test="$document-language = 'fr-CA' or $document-language = 'fr-FR'"> <!-- If french, double quotation marks are << rather than " -->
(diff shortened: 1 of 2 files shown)


rbeezer:
@JeanSebastienTurcotte,  thanks for taking the initiative on this one.  We just moved away from the "document-language" being the only language in play.  Motivated by Doob's French-English Canadian Olympiad project.

So something more general needs to be put in place.  Like a "local" language being a parameter to the templates for the characters (or the context itself).  I haven't yet come to a conclusion on the best approach.

Related question: how does French look for a mix of quotes, like you might see in a novel?  Single quotes inside of double quotes.  Some very old LaTeX code tries to make sure 8 backticks in a row resolve to the right mix of single and double.  Can't illustrate this in markdown!!!

JeanSebastienTurcotte:
@rbeezer A quote inside a quote is depicted with standard english double quote. So you'd write << Rob asked me: "How does French look for a mix of quotes" in Github >>.

I was aware that double language was now a thing. Maybe this can help when it's decided how to implemented. While looking at how spaces behave in French (We need thin spaces before some punctuation, which english does not have, I'll have a post on -dev later).

I guess the easy way would be to create a <flq/> and <rlq/> for french left quote and french right quote and treat them as something more available. But I understand not wanting more element.

PR #1908, opened 2023-01-19
Some french localization

JeanSebastienTurcotte:
Added some newly created translation. There were also missing lines in fr-FR that I added from fr-CA.


rbeezer:
Thanks!  Sorry for the missing lines, I try to keep them in sync.  Consolidated into one commit.

PR #1912, opened 2023-01-21
Close http.server on interrupt of interactive preview generation

dbrianwalton:
Put the asynchronous loop inside a try block with the clean-up of the server in a finally block. This causes the server cleanup to occur on interruptions of the file generation (e.g., keyboard interrupt). Also now including the process in the debug output.


rbeezer:
Beautiful!  Works great.  Thanks.

PR #1913, opened 2023-01-21
Sample article: comment about Geogebra perspective

dbrianwalton:
Made a note in sample article that Geogebra interactives written from scratch should use the perspective attribute for setup rather than a scripted setPerspective command.


rbeezer:
Yep!  Thanks.  Commit is unharmed (since the message was perfect), so you don't need to delete the commit, likely just delete the branch pointer/name on your end.  Thanks.

PR #1915, opened 2023-01-22
Proof of concept for extended dev schema

oscarlevin:
As discussed and requested on the [pretext-dev](https://groups.google.com/g/pretext-dev/c/crogb8-meWs/m/WJukDjmpAgAJ), here is a proposal for how to set up a secondary development schema that can be validated against instead of (or in addition to) the current stable schema.  A very rough attempt at implementing `<interactive>` is provided (but not knowing its current status, I made guesses based on the sample article).  Results in fewer schema violations, but would need much improvement to completely capture what interactives are supposed to be.

Basic setup is this:

* New features under development get their own section in a reasonable place in `schema/pretext.xml` along with the rest of the schema.  This includes providing their definitions just like any other element, but will also mean giving additional definitions to the elements that are meant to include them (using the `|=` notation for `combine="choice"`).  
* The `<fragment>` gets an `@xml:id` just like any other, and while the feature is still under development, it is then referenced in a second collection of fragments at the end of the file (with `filename="pretext-dev.rnc"`; see line 2659).  
* Once the feature is ready for the stable schema, this reference can be moved to the main collection of fragments.  
* Additionally, the additional definitions of elements that can include the feature can be cleaned up: merge them into the correct place in the original definition of the including element.
* Processing the `xml` file with `xsl/pretext-litprog.xsl` creates both files.  Then use trang to convert both to `.rng`.  (No ideas/opinions about whether this should be included as part of the other versions of the schema.)

If this all seems reasonable, then I'd be happy to make a first pass at adding other experimental features to capture whatever is in the sample article.

### DIFF
--- a/schema/build.sh
+++ b/schema/build.sh
@@ -69,6 +69,7 @@ xsltproc ${MBXSL}/pretext-litprog.xsl pretext.xml
 
 # System trang conversion to RELAX-NG XML schema
 trang -I rnc -O rng pretext.rnc pretext.rng
+trang -I rnc -O rng pretext-dev.rnc pretext-dev.rng
 
 # System trang conversion to W3C XSD schema
 # "abstract groups" make schema browser too obtuse
--- a/schema/pretext-dev.rnc
+++ b/schema/pretext-dev.rnc
@@ -1,10 +1,10 @@
 
-            grammar {
+                    grammar {
+
+                    include "pretext.rnc"
                 
-            include "pretext.rnc"
-            
-                    Interactive =
-                    element interactive {
+            Interactive =
+                element interactive {
                     UniqueID?,
                     PermanentID?,
                     attribute aspect { text }?,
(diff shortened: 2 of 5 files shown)


rbeezer:
Putting out fires, so finally got a good look at this.  You are basically leveraging the RELAX-NG `include` and our `fragment` support, yes?

*  I like the idea of refining a semi-crappy piece of the schema into something that can migrare easily into the "real" schema.  More likely to see some gradual evolution than neglect once a feature becomes official.

*  Looks like you captured `interactive` pretty well.

*  I'd say, yes, please see what else you can add.  `docinfo` might see some more removal/changes, so don't work too hard on that part.  Global search of teh XSL for `$docinfo` could find bits that are not apparent from the sample article.

Thanks!

oscarlevin:
Finally found some time for this.  I think I have a pretty good idea how to add things, and I should be able to make some good progress soon.  To start, I'm just adding stuff to fix the validation errors in the sample book and sample article.

But before I get too carried away, are there some changes to the stable schema that should be made?  For example, I just spent some time figuring out how to add a footnote to the shortlicense.   You can look at the most recent commit to see what I did, but this requires copying over the entire defined element that contains shortlicense and modifying that version (in this case, `ColophonFront`).  Now the schema has two slightly different versions of `ColophonFront`, which need to be kept in sync.  The alternative is giving a definition for `ShortLicense` that can have two versions, but this is a bigger lift.  Both methods are definitely doable, but perhaps it is better to just add Footnote to shortlicense in the stable version.

So how would you like to proceed?  We could start a discussion on -dev about what is ready for the stable schema and what should be experimental?  Or we could go through the validation errors from the sample- documents and label each as stable, dev, or do not add to either.

rbeezer:
Yes, a big duplication for a small fix is not going to be very practical.

How about:
* A long-running thread on `-dev` about what is missing versus what is woring versus what is in-progress?  (or a drop-in breakout?)
* Discrete (pun intended) pull requests for topical/related changes.

I have a script that starts with `pretext.xml` and makes the RNG, RNC, XSD versions, **plus** runs the commercial tool that creates the schema browser.  I can do that easily with each commit and merge it into your work on the source.

oscarlevin:
Okay, I think I've made enough progress on the low-hanging fruit that this should be merged.  No changes to the standard schema (I hope); everything goes into the -dev schema.  

Next step is to carefully consider what the experimental elements *should* allow, and perhaps fix some obvious schema elements in the main version.

rbeezer:
Merged and pushed, with a bit of reorganization.

Really nice.  I love the combination of integration **AND** modularization/separation.  Took me a minute to see how you accomplished that.

I've done no review of the actual patterns added.  As suggested, we could start a slow-burn thread on `-dev` to sort out fiction from fact.

Would you like to have the honor of announcing the existence on `-dev` for the cognescenti?

Thanks!

PR #1916, opened 2023-01-27
Showcase article: convert WeBWorK stages to tasks

Alex-Jordan:


rbeezer:
Very good!  Thanks.  I had to rebase because of a gitastrophe on my end, all my fault.

I have some ideas about timing the "repair" phase of the pre-processor and this might just be the ticket when that time comes for testing.

Slowly digging out.

PR #1917, opened 2023-01-27
WeBWorK: delimit table cell content better

Alex-Jordan:
#1876 changed how table cell contents are delimited, to use &#x00a6;. This PR changes it to use the existing `mode="delimit"` template instead, which I had forgotten existed.

At the same time, now that I know more about perl's `q` function, a bug is fixed with that template. A big that was probably never encountered by a real user, but a bug nonetheless.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -2166,6 +2166,26 @@
                     </statement>
                 </webwork>
             </exercise>
+
+            <exercise>
+                <introduction>
+                    <p>
+                        This exercise has a single quote in it.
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I rebased and force pushed this one. At the same time, fixed a typo, improved some commentary, and put the allowed delimiter characters in a sensible order. Also I noticed that `#` was missing but there is no reason to leave it out. And I notices that `\` was present and actually that one won't work as a perl `q`-delimiter.

rbeezer:
OK, seems to function as intended.

I had to break this into two for testing, and left it that way.

Thanks for keeping this one humming.

PR #1921, opened 2023-02-03
Sanatize code element in titles for html title element

oscarlevin:
Currently, if a user uses `<c>` elements in a title of a chunked page, a bunch of junk gets put into the html title element, making the title bar of the browser ugly.

Just like what is done for math, the `<c>` tag should now not be a problem.


rbeezer:
Pushed, and will rebuild website example sometime soon.  Got me thinking - see #1922.

PR #1924, opened 2023-02-04
WeBWorK: PG archiving via pretext.py

Alex-Jordan:
This removes the last direct use of xsltproc from the sample chapter Makefile. In its place, it uses `pretext/pretext`, with a new function for making a WeBWorK archive.

Also I added the recipe for making the sample chapter's PG macro library file to the Makefile. You won't need that because the host course already has the macro library up. But now the Makefile has a complete model.

As we discussed today, once this is merged and in the nightly CLI build, I will ask @oscarlevin to build CLI tools that execute `pretext -c pg-macros` and `pretext -c ww-archive`. Possibly nothing new is needed? Just a target in your project file with `<format>pg-macros</format>` or `<format>ww-aarchive</format>`?

And then I will update documentation in the guide.

### DIFF
--- a/examples/webwork/Makefile
+++ b/examples/webwork/Makefile
@@ -100,6 +100,18 @@ PGOUT      = $(SCRATCH)/pg
 # testing purposes.This should not be taken as necessary or best
 # practice for a real PreTeXt project using WeBWorK problems.
 
+
+#########################################################################
+#  Construct macros file. This must be uploaded into the macros folder
+#  of the server's host course before building representations below.
+#  There are no macros for the minimal example, so omitted
+
+sample-chapter-macros:
+	install -d $(PGOUT)/macros
+	cd $(PGOUT)/macros; \
+	$(PTX)/pretext/pretext -v -c pg-macros -p $(SMPCPUB) $(SMPC)
+
+
 #########################################################################
 #  Extract webwork problems into a single XML file.
 #  Location is within generated folder (named in the publisher file),
@@ -177,13 +189,13 @@ sample-chapter-problem-sets:
 	install -d $(PGOUT)
 	cd $(PGOUT); \
 	rm -r Integrating_WeBWorK_into_Textbooks; \
-	xsltproc -stringparam publisher $(SMPCPUB) -stringparam chunk.level 1 -stringparam debug.datedfiles no $(PTXXSL)/pretext-ww-problem-sets.xsl $(SMPC)
+	$(PTX)/pretext/pretext -v -c all -f webwork-sets -z -x chunk.level 1 debug.datedfiles no -p $(SMPCPUB) $(SMPC)
 
 mini-problem-sets:
 	install -d $(PGOUT)
 	cd $(PGOUT); \
 	rm -r WeBWorK_Minimal_Example; \
-	xsltproc -stringparam publisher $(MINIPUB) -stringparam chunk.level 1 -stringparam debug.datedfiles no $(PTXXSL)/pretext-ww-problem-sets.xsl $(MINI)
+	$(PTX)/pretext/pretext -v -c all -f webwork-sets -z -x chunk.level 1 debug.datedfiles no -p $(MINIPUB) $(MINI)
 
 
 #########################################################################
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -376,6 +376,7 @@ def get_cli_arguments():
         ),
         ("sagenb", "Sage worksheet conversion (removed)"),
         ("all", "All available output formats"),
+        ("webwork-sets", "Folder tree of PG files, set defintions, and set headers"),
     ]
     format_help = "Output formats are:\n" + "\n".join(
         ["  {} - {}".format(info[0], info[1]) for info in format_info]
@@ -466,6 +467,13 @@ def get_cli_arguments():
         help="abort script upon recoverable errors",
         action="store_true",
         dest="abort",
+    )
+    parser.add_argument(
+        "-z",
+        "--tgz",
+        help="output a compressed tar archive",
+        action="store_true",
+        dest="tgz",
     )
     # Needs help-file, once public
     # Will default to False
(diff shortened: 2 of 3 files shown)


Alex-Jordan:
Should I change it from `pretext -c ww-archive` to `pretext -c webwork -f archive`? Or maybe `pretext -c webwork -f source` since `source` is already an option for `-f`?

Along the same lines, is `pretext -c pg-maacros` best?

rbeezer:
For the archive, I could argue both ways.  A component (`-c ...`) or an output format (`-c all -f ...`).  I lean to output format, since it uses titles and divisions to form the organization of the archive, and it just feels more like a "product".

Macro-library feels more like a small piece, so `-c`?

Do we want the option to get the archive all zipped up?  One file?  `-f` could be `ww-archive` and `ww-archive-zip` which would dictate an argument with two different options, provided to the new routine.  First uses destination directory, second uses output file.

Alex-Jordan:
I agree that output format feels better here for the archive. Are you sure it should be `-c all -f ...` and not `-c webwork -f ...`? If I make a new `-f archive` option, then (a) the `-f` option need not be webwork-centric and (b) it could already apply to a few other things, like `-c latex-image -f archive` makes a zip of all the images.

Speaking of zip, it is of course doable, but will take some thinking. The python function does not know the name of the folder that is being created to put all the .pg and .def files. (Like `Integrating_WeBWorK_into_Textbooks` for the sample chapter.)

Macro file feels more like an extraction (extracting latex-image-preamble, LaTeX macros, etc) than it does about creating output from source. And in my head, the extractions are `-c` like `-c latex-image`. So I think I will leave that one alone. (Good, because it has been that way for a while now and I'd have to deprecate and announce if that one changed.)

rbeezer:
I think of "-c" as things we need to rip out and process in ways XSL cannot.  Such as `latex-image` in the early days.  The essential nature of the script in the first place.

Then `-c all -f ...` is newer and are things you can do with `xsltproc` (like `latex`), or maybe with a complicated series of `xsltproc` invocations (like `braille-emboss`).

I think I only want to zip things that are expected to zipped.  EPUB.  And a WW server can ingest a zip,  no?

Ooops, I forgot the macro file already had a routine!  Yes, don't change it.

Alex-Jordan:
I wonder if zipping will become an OS specific headache. WeBWorK can automatically handle a .tgz. Does Windows make .tgz easily? Cursory Googling says somehting like `tar` or `gzip` doesn't come with Windows and people download special applications if they need to make a .tgz.

So are you suggesting it should be `-c all -f ww-archive`?

rbeezer:
> So are you suggesting it should be -c all -f ww-archive?

Yes.  But how about `ww-problem-sets` (or similar) and `ww-problem-sets-archive` or `ww-problem-sets-tgz` for the `*.tgz` version?

rbeezer:
> I wonder if zipping will become an OS specific headache.

It is not for zip files (so far).  We use the zipfile module.

Have you looked at the `tarfile` module?

https://docs.python.org/3/library/tarfile.html

rbeezer:
I think the naming discussion is done.  Any progress on using `tarfile`?  I think Windows is irrelevant.  The `*.tgz` is for a WW server to ingest, not for a Windows user to unpack.

Alex-Jordan:
OK, this is ready. To recap:

* No documentation yet. We will wait until the CLI has this and @oscarlevin  adds something like a `-g webwork-sets` option and then write documentation in the guide only for the CLI approach.
* Now you can do `pretext/pretext/pretext -c all -f webwork-sets` or `pretext/pretext/pretext -c all -f webwork-sets -z` to make a folder tree with all the WW exercises, set def files, set header files, and the PG macros file for a project. The `-z` option gives you a `.tgz` file instead of the actual folder tree.
* I added a general utility `targz()` which may be useful elsewhere. For example we could apply it to all of the image file output.

oscarlevin:
I can try to work on the CLI function later this week.  Right now I think making a new target in the `project.ptx` manifest is the right way to go.  Note that even before that is ready, once this is in `pretext/pretext` it will be available to the CLI via the devscript command.

Incidentally, no need to worry about archive formats on windows.  I regularly zip and unzip tar files (using 7zip, I think).

rbeezer:
Very nice!  Two spelling fixes, and a bit of very light editing on the `pretext/pretext` help messages.  Otherwise no code was harmed.

Wasn't that considerate of me to leave in `chunk.level` as a string parameter?

Alex-Jordan:
Continuing a discussion here with @oscarlevin. Currently, we can use something like:
```
pretext devscript -p publication/publication.xml -c all -f webwork-sets src/main.ptx 
```
to produce a folder tree with .pg files, set definition files, set header files, and the project-specific macro file. Also we can add the `-z` option to produce this folder tree as a `.tgz` file.

Note that this should be done after you have generated webwork representations, perhaps with `pretext generate webwork`.

What should we do so that the CLI can create that folder tree without calling on the devscript? It sounds like since this is ultimately a `-f` option, that a new target could go into the project file like:
```
    <target name="pg">
      <format>webwork-sets</format>
      <source>src/main.ptx</source>
      <publication>publication/publication.xml</publication>
      <output-dir>output/pg</output-dir>
    </target>
```
But currently, `webwork-sets` is not a recognized CLI `format`. Is that easy to add?

Also I'm not sure about how the CLI works internally, but it would be good if there was a check for the existence of the representations file before trying to make this folder tree. And if the right flag is used on this target telling it to generate all required assets first, that it generate the webwork representations.

When we have this sorted out, I need to update the PTX guide, which is still telling people to use `xsltproc` to make the PG files.

oscarlevin:
This sounds like a great and relatively easy CLI feature to add.  The checking for the representations file is already on my list of things we need to fix for regular builds (https://github.com/PreTeXtBook/pretext-cli/issues/393)

oscarlevin:
This will be available in the next nightly build.

Alex-Jordan:
Thanks Oscar. Can you say when the next regular release might be?  I will have (at least) until then to update documentation.

oscarlevin:
I plan to do a regular release June 1st, but of course, there is nothing that this new feature breaks from the current documentation.  So no rush.

PR #1925, opened 2023-02-07
Fix for temporary directories not being deleted

oscarlevin:
A few spots, the directory is changed to the temp directory, but not changed back.  This prevents windows from deleting the temporary directory.  

It would probably be good to clean up this code so that we are not constantly switching in and out, but using a context of some sort.  But for now, this should solve the problem.


rbeezer:
Great catch!  Thanks.  I just cherry-picked the tip of this branch, it looks like it was built off of part of a previous PR.

rbeezer:
And a "to do":  #1926

PR #1930, opened 2023-02-09
Guide: Revise chapter 1

oscarlevin:
Various editing improvements to chapter 1, making it slightly more user-friendly and more focused on using the new workflow.

@StevenClontz is similarly revising chapter 2 in a separate PR.


StevenClontz:
If this is merged first, I'll update my PR to match with fixed xml:ids.

rbeezer:
Thanks for keeping this current!  I added a follow-on commit to remove some stray whitespace that I couldn't figure out how distribute among the (appreciated) topical commits.

PR #1937, opened 2023-03-01
Corollarium -> Gevolg in nl-NL.xml

pjbruin:
The usual translation of "corollary" in Dutch is "gevolg"; I have never seen "corollarium" and it sounds antiquated to me. @chosia do you agree with this change?

This commit also fixes a typo in "berekening".


rbeezer:
Thanks, @pjbruin!  As is my practice, I've edited the commit message so the commit hash has changed.  We can change again if there is any more input from others.  Thanks for improving this.

So you'll want to delete (or ignore) your `patch-1` branch.

PR #1939, opened 2023-03-03
Add cr

mdoob:
An unterminated \end{document} line causes LaTeX to throw a warning message. No danger to the LaTeX but might confuse the user.


rbeezer:
Thanks, Michael.  We had a few of these in place, but not all.

I think you started your branch on `dev`, which is long dead.  You submitted four commits, when only one should be necessary.  And only rarely should a merge commit be necessary to submit.  

So I had to do some extra work to get this in.  Best to always `pull` master and then build a new branch from there.

mdoob:
Hi Rob --

Here is my login to the computer in my office:

You have mail.
Last login: Mon Mar  6 11:00:59 2023 from 10.176.6.197
[2000] net276:~ $ cd pretext
[2001] net276:pretext $ git branch
* addCR
  master
[2002] net276:pretext $

I pushed this to my githup account and did the pull request.
I saw the extra commits and was mystified by it. Do you think that
I need to create a new fork on github?

Cheers,
Michael

rbeezer:
> I saw the extra commits and was mystified by it.

If you knew they were there, then it would have been better not to have left the PR in that state.  A fresh clone/fork is cheap, and a new PR can be formed while noting that the old one is not to be considered.

`pretext-dev` would be an appropriate place to seek help deciphering exactly how you began by including the stub commit on `dev`.

PR #1945, opened 2023-03-07
GUIDE: Add subsection on installing pretext (cli)

### LINKED ISSUE #1943: self-reference for installing CLI
https://github.com/PreTeXtBook/pretext/blob/master/doc/guide/author/processing.xml#L55 says that `Instructions for installing this command line interface can be found in <xref ref="processing-CLI"/>.` But that is the very section in which this quote is found.

oscarlevin:
Also fixed an old reference to `pretextbook` and added a list of the assets you can generate with `pretext generate`

Will close #1943


kcrisman:
Thanks for bringing this back!  This will be a big help to the kind of people I was presenting to last week, all of whom are quite familiar with command-line LaTeX and weren't that interested in working in the cloud.

I do note that it says something about no newline at the end of the file.

rbeezer:
Excellent!  This will be a huge help with support questions, and it looks great.

Combined the two commits and removed some stray whitespace - otherwise as-is.  Thanks for contributing to the documentation.

PR #1946, opened 2023-03-07
React update

siefkenj:
The React frontend is switching to a new build system that requires being imported via a `type="module"` script tag. Additionally, the name of the built source file has been fixed as `main.js`, so `bundle.js` is no longer relevant.


siefkenj:
Good question. I have no idea how I ended up with https://github.com/PreTeXtBook/pretext/commit/8b373f775dbbc44861000f22810eaf5768ecb8ad . My fork was still called `mathbook`. I just renamed it yesterday and pulled...Hopefully it won't happen again.

PR #1948, opened 2023-03-09
Worksheets: button to print page

Alex-Jordan:
This adds a button to print the HTML page. Demonstrated [here](https://spot.pcc.edu/~ajordan/temp/worksheet-1-letter.html).

Only available on standalone worksheet pages, but could easily be extended.

The styling is uncommitted to navbar_default.css. Not sure if I should be messing with that file, so I'll await input from @davidfarmer. As well as input on the particular styling I went with (just copying the Calc button).

### DIFF
--- a/xsl/localizations/af-ZA.xml
+++ b/xsl/localizations/af-ZA.xml
@@ -262,4 +262,5 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <!-- <localization string-id='activate'>Activate</localization> -->
     <!-- <localization string-id='reset'>Reset</localization> -->
     <!-- <localization string-id='array'>array</localization> -->
+    <!-- <localization string-id='print'>Print</localization> -->
 </locale>
--- a/xsl/localizations/bg-BG.xml
+++ b/xsl/localizations/bg-BG.xml
@@ -286,4 +286,5 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <!-- <localization string-id='activate'>Activate</localization> -->
     <!-- <localization string-id='reset'>Reset</localization> -->
     <!-- <localization string-id='array'>array</localization> -->
+    <!-- <localization string-id='print'>Print</localization> -->
 </locale>
--- a/xsl/localizations/ca-ES.xml
+++ b/xsl/localizations/ca-ES.xml
@@ -265,4 +265,5 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <!-- <localization string-id='activate'>Activate</localization> -->
     <!-- <localization string-id='reset'>Reset</localization> -->
     <!-- <localization string-id='array'>array</localization> -->
+    <!-- <localization string-id='print'>Print</localization> -->
 </locale>
--- a/xsl/localizations/cs-CZ.xml
+++ b/xsl/localizations/cs-CZ.xml
@@ -260,4 +260,5 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <!-- <localization string-id='activate'>Activate</localization> -->
     <!-- <localization string-id='reset'>Reset</localization> -->
     <!-- <localization string-id='array'>array</localization> -->
+    <!-- <localization string-id='print'>Print</localization> -->
 </locale>
--- a/xsl/localizations/de-DE.xml
+++ b/xsl/localizations/de-DE.xml
@@ -271,4 +271,5 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <!-- <localization string-id='activate'>Activate</localization> -->
     <!-- <localization string-id='reset'>Reset</localization> -->
     <!-- <localization string-id='array'>array</localization> -->
+    <!-- <localization string-id='print'>Print</localization> -->
 </locale>
--- a/xsl/localizations/en-US.xml
+++ b/xsl/localizations/en-US.xml
@@ -304,5 +304,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <localization string-id='reset'>Reset</localization>
     <!-- Describe an array of fill-in-the-blanks, say for a matrix -->
     <localization string-id='array'>array</localization>
+    <!-- Print page button text -->
+    <localization string-id='print'>Print</localization>
 
 </locale>
--- a/xsl/localizations/es-ES.xml
+++ b/xsl/localizations/es-ES.xml
@@ -266,4 +266,5 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <!-- <localization string-id='activate'>Activate</localization> -->
     <!-- <localization string-id='reset'>Reset</localization> -->
     <!-- <localization string-id='array'>array</localization> -->
+    <!-- <localization string-id='print'>Print</localization> -->
 </locale>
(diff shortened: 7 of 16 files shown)


davidfarmer:
The print button has an unknown character it in, when I see the page.
(I assume it is a little printer emoji.)

But do we really need a print button, when the reader can just
do control-p ?

Alex-Jordan:
The reader can do that if they know about that. Or if they know about
printing from a browser's menu. But I doubt most users think about a web
browser as capable of printing.

And even when they do, if they are like me, they assume it would print the
whole page including navigation and were surprised to learn otherwise. So
might not even try.

A button says to the user that the web page designer intends for you to be
able to print this worksheet and therefore set it up to print only the
worksheet.

Alex-Jordan:
I will change the Unicode character to a word. I also need to change
something else because the tab order on the button is wrong.

kcrisman:
> The reader can do that if they know about that. Or if they know about
> printing from a browser's menu. But I doubt most users think about a web
> browser as capable of printing.
>

Agreed.

davidfarmer:
I recall a similar argument(?) in another setting for the issue of
whether a link should open a new browser window.

Someone argued against that, with "You are forcing me to open a new
window.  If I want to do that, I'll just right-click."

Someone else had an eloquent argument against that logic:
"What is right-click?"

I'll leave you to speculate which side I was on and which side won
the argument.

------

Such examples disprove the claim that kids these days understand 
technology.

davidfarmer:
I am okay with the print button on the individual worksheet pages
and I will work on the page fitting so that the 80% trick is not needed.

rbeezer:
Nice.  XSL looks good.

Sounds like you are going to replace the printer character?  Will we need a beta to test styling?  Or wait on changes in another repository?

Holler when this is ready.  Might catch a few PRs Friday AM, but there is no rush.

Alex-Jordan:
I'm wresting with some CSS for this, and potentially the need to change the HTML. The issue starts because I feel like the print button should be over on the right (but left of the tree buttons). So let me check in. Does anyone want to convince me to put the print button on the left? Like next to the Calc button, or Index button?

Placing the button where I want to place it raises issues of floating things to the right. To accomplish that in [this demo](https://spot.pcc.edu/~ajordan/temp/worksheet-1-letter.html) I had to move the tree buttons earlier in the DOM. They do not appear to be placed differently on that page, but the tab order is not what you would expect. It goes through the tree buttons first, then back to index, calc, print.

Is that a good change? Get the tab-navigating user to the tree buttons first? I think it's considered bad practice to have the tab order misaligned with the visual order.

To do this "right", the buttons navbar should use a flex-box scheme. That would be a significant change to the CSS and make it awkward to support existing pages that float the tree buttons right, while also supporting a new flex-box scheme.

Thoughts?

davidfarmer:
The layout in the default and CRC styles are intentionally done
differently:  default is mostly vanilla CSS; CRC uses a mixture,
including flex and grid.  This is part of the main point of overhaul:
we were changing the structure of the HTML and we needed to be sure it
would work with multiple styles.

The navbars may be styled similarly at the moment, but I have misgivings
about requiring a particular method.

There also is an ongoing problem which maybe we can take a step toward
fixing:  the work required to maintain multiple styles.  Most recently
this arose when we changed the layout of the footer content.

If the general layout of the navbar and footer were required to be
basically the same (in terms of position if not style), then that can
be common to all styles.  We would split the navbar_*.css files and
put part in pretext.css.

Then we have to decide of we go to the next CSS version.  Probably yes.

Of just temporarily live with the print button next to the calculator
button, or having the print button appear in a different location than
suggested by the document order.

oscarlevin:
As a user, I think I would expect to see a "print" button or link "inside" the page, and not as a stealthy new button in the navigation.  As a bonus, this wouldn't require any tweaks to the pages "frame"

Alex-Jordan:
Alright, that helps. I put the button inside the page, and CSS changes are minimal additions:

https://spot.pcc.edu/~ajordan/temp/worksheet-1-letter.html

The button itself is not styled. If anyone has suggestions, I'll try them out. Going to force push with updated XSL now.

davidfarmer:
For some reason I pucture the print button as green.

I can style after it is live.

Alex-Jordan:
Now with the updated XSLT.

There are uncommitted edits to pretext.css that only impact .print-button.

In editing the localization files, I noticed two of them had CRLF style line breaks. Must have been submitted by a Windows user. I removed the CR part in a separate commit (along with a few other places in the repository where they showed up. They are also in some MOM static content that is committed to the repo, but I opened #1949 about that.

Alex-Jordan:
Now "LightGreen". Just until David makes it better.

rbeezer:
A project may be chunked in a manner where a `worksheet` is one of several divisions on the page.  One such is 33.4 in sample article (screenshot).  The "A4" and "US" buttons take you to a standalone version.

This suggests to me that a "Print" button maybe belongs only on the standalone page.  Is that where this is going?

Am I just missing some CSS?  See button placement in screenshots.

Screenshot 2023-03-10 at 10-42-48 SA Worksheets
Screenshot 2023-03-10 at 10-43-26 SA Dot products and projection

Alex-Jordan:
You are missing some CSS. Refresh on the standalone page.

But the presence of the button on other pages is my mistake. I'll think that through and update. 

Roger on the CR. IT was uglifying the diff on those localization files for this which is what got me going on it here.

rbeezer:
> Refresh on the standalone page.

That's better!  Sorry, should have thought about that sooner.

rbeezer:
> But the presence of the button on other pages is my mistake. I'll think that through and update.

From the hip.  Isn't there a template that makes the standalone version?  Add the button in there, always?  Without even having a separate template?

Alex-Jordan:
> From the hip. Isn't there a template that makes the standalone version? Add the button in there, always? Without even having a separate template?

One thing:  If we are sure the button is only for printing worksheet pages, then I considered this. I was trying to make it abstract enough that it could apply to other things easily when they come up in the future. Still thinking, but I may give up.

Other thing: for the button to literally be inside the worksheet as it is in the current sample, the styling is simpler (I think?) if it is within the "structural division content". But that led to it showing up where I didn't want it...

Alex-Jordan:
CR removal removed.

Now it builds the button outside of the actual content, but within the file wrapper. For the standalone worksheet, the context node is the "worksheet", not a section, etc. And the modal template for the print-button applied to a "worksheet" builds the button.

I don't think I can place the button on the top right corner of the worksheet this way. So it's in the top left corner, as here:
https://spot.pcc.edu/~ajordan/temp/worksheet-1-letter.html

It's not really "in" the worksheet with the border outline. It is before that, but positioned to overlay.

davidfarmer:
I'll move the button when Alex is done with what he is doing.

Or, Alex can change around line 732 of pretext.css

.print-button {
   position: relative;
   right: 2px;
   top: 66px;
   background-color: LightGreen;
   z-index: 1;
   margin-top: -4em;
   float: right;
}

Alex-Jordan:
David's CSS is applied. 

Not sure yet if Rob is sold on the print-button template being applied inside the file-wrap template. If not, this CSS will change.

Alex-Jordan:
OK, you can take a more serious look at this now.

rbeezer:
Thanks for sticking with this one!  Merged, and might rebuild tonight (check first).

I'll leave publicity (pretext-announce) to you.  I may post on -dev just now, have a look first.

Alex-Jordan:
I don't think this is worth publicizing. But if we are revitalizing worksheets in other ways, then after all that is done, we could make one post to -announce.

PR #1953, opened 2023-03-13
WeBWorK: move activate button to top

Alex-Jordan:
This moves the "Activate" button to the top of WW exercises. (It currently appears at the bottom. This is problematic when the exercise is very long.)

Then the "Submit" buttons, etc. still appear at the bottom.

I think styling is fine, but maybe it could use some fine tuning.


rbeezer:
Merged and pushed, as-is.

Pretty simple, so only limited testing.  And I think styling is good enough for now and can be tweaked as needed.

Sample chapter should be rebuilt sometime before late evening, if you want a live testing example.

Issue #1872, opened 2022-12-16
Ensure managed directories can function with CLI without a publication file


StevenClontz:

oscarlevin:
We have a few possible paths forward with this.  

1. Modify core pretext to default to the same default managed directories structure as the CLI would use with the default publication file.
2. Simply remove the requirement of managed directories for the CLI (not sure what would have to be done or what the results would be).
3. Modify the CLI to create a default publication file for any target that doesn't specify one (and put that path into `project.ptx`, or tell the user to do so).

I'm leaning to 3, but would be interested in hearing @StevenClontz and @rbeezer weigh in.

StevenClontz:
Maybe we could do this variation(?) of 3, since (I think I recall) all our work on a project happens in a temp directory:

- When a publication file is not present, silently copy our default publication file into the default location within the temp directory, and pass that on to core pretext.

oscarlevin:
This is definitely fixed in 2.0.  We can close this.

PR #1954, opened 2023-03-17
attempt to get codespaces working with core pretext repo

StevenClontz:


StevenClontz:
@oscarlevin 

Try opening this commit in a codespace. For whatever reason, PreTeXt-tools isn't finding the targets.

oscarlevin:
That's because your pretext.ptx has two targets without a publication file.  The way I hacked this to work with the CLI is that the CLI has a `-t` option that prints a list of targets, and that gets read by pretext-tools.  But for some reason the CLI is looking for a entry for `<publication>` in the manifest (maybe to make sure it really is a manifest??), and there is an error when it doesn't find it.  Obviously, I should work to fix that in the CLI, to at least make the failure more graceful

StevenClontz:
Feel free to close, I will redo this in a fresh PR.

oscarlevin:
Just FYI, the issue wasn't that there were no publication files, it was that `project.ptx` didn't have a `<publication>` element for at least one target.  See #1872 and https://github.com/PreTeXtBook/pretext-cli/issues/424

PR #1955, opened 2023-03-17
Update tutorial video

StevenClontz:
The tutorial itself needs updating as well but that won't happen before the weekend.


rbeezer:
This needs a couple of things.  I started to fix it up, but gave up after the first one.  ;-)

*  Let's make **one** good representative multi-purpose image for all time, and use the `@preview` attribute to point to it.  I copied the old generated one to a new filename (due to the `xml:id` change).  But then I hit...
* The QR code is tied to the `xml:id`.  A new one needs to be generated.

The Guide gets tested regularly as part of pull requests, etc.  And so should build without the CLI.   Please check that your PR builds both PDF and HTML successfully.

StevenClontz:
I guess I'm confused. Is there some invalid PreTeXt in my PR? I'm happy to make sure my PRs pass whatever tests you'd like, including non-CLI builds, but I'm not sure where these are documented. I checked [here](https://github.com/PreTeXtBook/pretext/blob/master/CONTRIBUTING.md) but found nothing. Is there a makefile or something available?

StevenClontz:
9bb6978 adds the assets needed for a PDF build (I forgot these are committed to source for the guide). I'll work on the preview video request.

rbeezer:
I think you are not confused.  Recent commit is what I'm suggesting.  I do not "generate assets" every time I build the Guide.  They all live in generated, ready to go.  So when I built PDF a few minutes ago, it failed since it did not find images it needs.

By testing, I meant when a contributor makes a PR against the Guide, at a minimum I make sure I can use `pretext/pretext` to build HTML and PDF.  (Nothing fancier than that.)  But I don't want to make generating **anything** as part of that process.

Gotta run now, will get back to this.

rbeezer:
Got it!  Thanks for finishing this one off.  Merged and pushed, and will rebuild website in the AM.

Next month, I think a PR will just need:
* New YouTube identifier
* New month in title
* New QR code image (this will be easy to miss!)

Thanks!

PR #1956, opened 2023-03-18
Remove fixed HTML ids and prefer classes

siefkenj:
As discussed, we should prefer classes to HTML ids, which makes CSS easier and makes managing ids in XSL easier.


rbeezer:
Thanks, @siefkenj.  No objections here.  `extract-math.xsl` just makes an intermediate file for offline MathJax (braille, EPUB), so nobody sees it, so nobody styles it.  And I'll need to review the comment there.  More on `-dev`.

rbeezer:
Closed via #1979.  Thanks, @siefkenj, for getting this going.

PR #1959, opened 2023-03-20
common project.ptx + codespace config for contributions from CLI/codespace users

StevenClontz:
@oscarlevin please review. The idea that is whenever someone wants to make (say) a Guide contribution, they can do so using codespaces and our usual authoring tools, lowering barrier to helping improve documentation.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,46 @@
...
+  // 		"protocol": "https"
+  // 	}
+  // },
+  "onCreateCommand": "pip install pretext",
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "sudo bash ./.devcontainer/postCreateCommand.sh",
+
+  // Configure tool-specific properties.
+  "customizations": {
+    "codespaces": {
```
oscarlevin, line 27: We should remove this here, since there isn't such a file

oscarlevin:
I think this is probably good.  I like the idea of having a single project.ptx for all the different builds.  Probably.  Of course, I don't do that for all of my projects (a single manifest for all books on my computer), but I guess one per repository makes sense.

Just checking: the `.gitignore` is set up to ignore `**/output`, so all the different output folders get ignored, correct?

StevenClontz:
> the `.gitignore` is set up to ignore `**/output`, so all the different output folders get ignored, correct?

yes

StevenClontz:
> I don't do that for all of my projects (a single manifest for all books on my computer), but I guess one per repository makes sense.

Yeah, exactly. And the fact that there's no other way (currently, at least) to use pretext-tools other than with such a model. And I'm not sure there should be.

StevenClontz:
Yes, I set this up with `texlive-full`. https://tex.stackexchange.com/a/312867/

So I don't think 3 GB is unreasonable to avoid all the tricky LaTeX gotchas (basic GitHub users get 15 GB-mo for free). But what we need to do for pretextbook/pretext-codespace is set up https://docs.github.com/en/codespaces/prebuilding-your-codespaces so users get a quick full LaTeX+SageMath+etc machine on demand.

kcrisman:
> ... so users get a quick full LaTeX+SageMath+etc machine on demand.

But ... isn't that what CoCalc also provides?  And at least when I tried setting it up a codespaces in preparation for a workshop, it didn't really seem particularly faster to load than that solution.  I guess at the least, it would be good to check how a normal-load browser (e.g. with multiple open tabs, email, etc.) actually loads that the first couple times.

StevenClontz:
CoCalc doesn't give users internet access or automatically wire projects to
a GitHub repo. The wait to set up a new Codespace is still smaller than the
time+money doing those extra steps.

kcrisman:
> CoCalc doesn't give users internet access or automatically wire projects to
> a GitHub repo.


Yes, that's true.  I guess GH must have really good spam blockers.

rbeezer:
Not sure I got an answer?  Is this ready to go?

StevenClontz:
You didn't - the Codespace took a while to set up and I'm just now running builds on all the projects.

StevenClontz:
I mean it didn't take two days; I just let it provision in the background and didn't get around to using it until just now. (-:

I just successfully built PDFs for the guide, showcase article, sample article, minimal, and humanities examples (the ones we've wired up so far). So I think this is good to go.

rbeezer:
Thanks!  Didn't mean to rush you - hadn't realized there was a lag to get things setup.

Likely merge tomorrow.

bnmnetp:
There are pre-built textlive docker images... If we use one of those it would really reduce the amount of time it takes to set up a codespace.  The installation of texlive with apt is one of the longest parts when I build new docker containers for Runestone and has had me looking for alternative ways of getting that set up.

rbeezer:
I'd like to keep the top-level directory of the repository relatively un-cluttered.  And it would be nice if `doc` (or even `doc/guide`) were each relatively self-contained.

Q: could we have **two** such setups?  One in `doc` and one in `examples`?

StevenClontz:
Closing in leiu of #2110

PR #1960, opened 2023-03-21
use context manager for opening list of qrcodes

oscarlevin:
I believe this will fix the issue reported by Mark F. here: https://groups.google.com/g/pretext-support/c/lOaxRqCbu7g/m/vGQO9kqUEQAJ


rbeezer:
Thanks, @oscarlevin!  I'll seek confirmation from Mark.

PR #1961, opened 2023-03-24
HTML: add space character to fillin spans

Alex-Jordan:
This puts a space character in the span for a text fillin. Without this, the span is empty, which makes it more challenging to style.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -8806,6 +8806,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <xsl:value-of select="5 * $characters div 11" />
             <xsl:text>em;</xsl:text>
         </xsl:attribute>
+        <xsl:text> </xsl:text>
     </span>
     <xsl:if test="@rows or @cols">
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
@davidfarmer In #1727 you recently reported you added CSS for the various fillin styles. I had been trying to do this recently, but I felt the need to add the space character being added here in this PR. I didn't see how to give the span height without giving it some content first. In your opinion, is this PR no longer necessary and can be closed unmerged?

davidfarmer:
It was a bit tricky to get the box and shade versions to work.
I added a CSS height to the span, otherwise you don't see the effect.

It does not seem to need a space inside that span.

I added that CSS to the 0.4 CSS, so you can see it here:

https://spot.pcc.edu/~ajordan/temp/interesting-corollary.html#subsection-paragraph-markup

Alex-Jordan:
OK. When I was trying, II came to the conclusion that adding a CSS height had no effect on an empty span. But I must have been doing something wrong. I will close this PR.

rbeezer:
Ooops, I'd thought I'd merged this.  ;-)  Guess not, and I guess for the best.  Thanks for getting this sorted.

PR #1965, opened 2023-04-07
Description var

Alex-Jordan:
Briefly discussed in drop-in today. This started as just adding `var` to the schema as a legal child of a `webwork//image/description`.

I also noticed an issue and an omission in the sample chapter.

And I rewrote/condensed some of the code in `extract-pg` surrounding images and their descriptions.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -1272,7 +1272,7 @@
                         <image width="50%">
                             <description>
                                 a rectangle whose width is labeled <var name="$b"/> cm
-                                and height is labeled <var name="$b"/> cm
+                                and height is labeled <var name="$a"/> cm
                             </description>
                             <latex-image>
                                 \begin{tikzpicture}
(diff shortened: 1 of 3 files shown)


rbeezer:
Visual scan looks good.  If authors have the results of the "webwork-representations" run comitted in their repository, will they see the names of files for some images changing?  Such as: from `image-4` to `parabola-inverted`?

Alex-Jordan:
Are you asking because of the changes surrounding `mod="pg-name"`? If so, I don't think so. That is the name of image as a perl variable, and does not go into what is used to make an image file name.

Presently, images that are grabbed from the WW server are stored like `webwork-79-image-2.png` or `theWebworksXMLid-image-2.png` . The filename is constructed in `pretext.py` (search for `ptx_image_filename`). It ultimately comes from `@ww-id` in `-assembly`. Which presently either automates the `webwork-79` or uses the `xml:id` on the `webwork` element.

rbeezer:
> Are you asking because of the changes surrounding mod="pg-name"?

Yes.

> If so, I don't think so.

OK, thanks for the explanation.

Q: Advisability of merging this now, or in about three weeks?  In the meantime, there won't be much opportunity for any fixes, but also likely not a CLI release, nor much course-building.

Alex-Jordan:
No rush for this, although I think it is 99.9999% likely fine. I inspected all the output for the sample chapter, including alt text on images. But no rush. Maybe no later than summer 2024? ;)  That is ORCCA edition 3 release date.

PR #1969, opened 2023-04-12
Checkbox

Alex-Jordan:
Now you can have checkbox-style multiple choice questions in a `webwork`. For a long time, something in WeBWorK caused all of the checkbox inputs to have the same HTML id, which led to badness.  That is all fixed now.

At [this page](https://spot.pcc.edu/~ajordan/temp/checkboxes/section-7.html) you can try the last three exercises to see live output.

WeBWorK will always write static PTX for this kind of thing as one of ol, ul, dl. (Which one depends on how checkbox labeling was done, and the three examples show the three styles.) It is almost "legal" PreTeXt, except there is a `@name` attribute on the (ol|ul|dl) that has the name of the corresponding WeBWorK answer. (Something like AnSwEr0001.) We need this in case the checkboxes list is inside a task, so that we know to associate (for example) AnSwEr0003 as an answer to something from within (for example) task 1b.

This works with the AIM server, which is 2.17, plus some small things form 2.18 to make this work. For anyone else with a 2.17 or earlier server, they would need to add `parserCheckboxList.pl` to their macros and make a small edit to `PGbasicmacros.pl`. I haven't thought through yet how/where to document this quirk. But I'm opening the PR for now to discuss it.

### DIFF
--- a/doc/guide/author/webwork.xml
+++ b/doc/guide/author/webwork.xml
@@ -279,7 +279,7 @@
         to cause the input form to be an array of smaller fields instead of one big field.
       </p>
       <p>
-        A <tag>var</tag> can have <attr>form</attr> with value <c>popup</c> or <c>buttons</c> for multiple choice questions.
+        A <tag>var</tag> can have <attr>form</attr> with value <c>popup</c>, <c>buttons</c>, or <c>checkboxes</c> for multiple choice questions.
       </p>
       <p>
         If you are familiar with PG, then in your <tag>pg-code</tag> you might write a custom evaluator
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -1087,40 +1087,114 @@
                 </webwork>
             </exercise>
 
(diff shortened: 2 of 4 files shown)


rbeezer:
Alright!

Alex-Jordan:
I just force pushed a rebase of this to keep it more current.

rbeezer:
Merged and pushed.  Website update will come later on.  Holler if you plan publicity and I can do update sooner.

Schema had a typo.  Missing a quote where addition was made.  Maybe you can setup `trang` and include a run just to check parsing as part of testing?

Noted: in the new three-part problem in "Subject Area Templates", I get checkbox part correct on first try.  Only.  I leave the other two empty - one says incorrect, other is silent.  Expected?  I guess it is the "?" in the drop-down that is the culprit?

Thanks for the improvement!

Alex-Jordan:
Thanks. Yes, the '?' counts as a submitted answer, and is incorrect. Upcoming version of WW has a new `DropDown` macro in lieu of `PopUp` that automates a placeholder for select elements and we can fix this when that is available.

PR #1979, opened 2023-05-17
No ids in css

davidfarmer:
Added classes in all places where previously the CSS was using an id selector.
The new CSS is live.

Did not remove any HTML ids.  Some are not necessary, but some are, so I thought
it best to just leave them as-is.

Built a document and it looked fine.  It is possible that I missed something, but that
can be sorted once people build from the new version.

Note that this change required incrementing the CSS and JS versions
(done as a separate commit).

This replaces pull request #1956


rbeezer:
Beautiful.  That was a bare-minimum low-maintenance PR!  As indicated by the fact that no commit hashes were harmed.  ;-)

Rebuilding website righ now, and will follow-up elsewhere.  Thanks!

Alex-Jordan:
I missed my year.

rbeezer:
I hate it when that happens.

PR #1981, opened 2023-05-19
slight updates to make Google Analytics 4 status clear

kcrisman:
It turns out that nothing really needed to change in terms of code, but the documentation could be slightly clearer given all Google's many warnings, so I took the liberty.  Hope this is helpful.


rbeezer:
Thanks very much for the contribution.  Merged.  Will update website in due course, likely before Monday.

PR #1982, opened 2023-05-19
Update print info for decktape

kcrisman:
Trying decktape out reveals it sometimes causes problems with the grid option, so I added a useful comment.


rbeezer:
Thanks very much for the contribution.  Merged.  Will update website in due course, likely before Monday.

PR #1985, opened 2023-05-21
Update Czech translation

jirilebl:


rbeezer:
Thanks for keeping up with all the additions!

PR #1989, opened 2023-05-29
GUIDE: update tutorial with current VS Code workflow

oscarlevin:
As someone recently pointed out, the directions in chapter 2 are now out of date.  This should fix that issue.


rbeezer:
Merged as-is.  Thanks!

PR #1992, opened 2023-06-01
Conditionally disable uneeded HTML/CSS when React is enabled

siefkenj:
Several parts of the HTML generation should be disabled when React is used, since React handles those things. This PR should catch most of them.


rbeezer:
Merged as-is, just an edited commit message (harming the hash).

Thanks - just what I was after.

PR #1994, opened 2023-06-04
Add XML to JSON utilities to pretext-html

siefkenj:
This PR adds `xml-to-json.xsl`, which allows writing JSON as structured XML. No more manual tracking of commas and indentation.

The XML format follows the format specified by XSLT 3.0, as I understand it. Thus, I have used the same namespace `http://www.w3.org/2005/xpath-function`.

This PR modifies the MathJax and sageCell Javascript initialization to use the new templates. All conditionals should be carried over. Hopefully it looks more readable!

The commit history should be squashed before merging, since this came from a debugging experience.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -49,9 +49,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     xmlns:exsl="http://exslt.org/common"
     xmlns:date="http://exslt.org/dates-and-times"
     xmlns:str="http://exslt.org/strings"
+    xmlns:fn="http://www.w3.org/2005/xpath-functions"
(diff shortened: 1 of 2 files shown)


siefkenj:
This PR adds `xml-to-json.xsl`, which allows writing JSON as structured XML. No more manual tracking of commas and indentation.

The XML format follows the format specified by XSLT 3.0, as I understand it. Thus, I have used the same namespace `http://www.w3.org/2005/xpath-function`.

This PR modifies the MathJax and sageCell Javascript initialization to use the new templates. All conditionals should be carried over. Hopefully it looks more readable!


The commit history should be squashed before merging, since this came from a debugging experience.

siefkenj:
I have a new PR with a better commit history.

PR #1995, opened 2023-06-05
Add xml-to-json.xsl for XML to JSON conversion

siefkenj:
This PR adds `xml-to-json.xsl`, which allows writing JSON as structured XML. No more manual tracking of commas and indentation.

The XML format follows the format specified by XSLT 3.0, as I understand it. Thus, I have used the same namespace `http://www.w3.org/2005/xpath-function`.

This PR modifies the MathJax and sageCell Javascript initialization to use the new templates. All conditionals should be carried over. Hopefully it looks more readable!

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -49,9 +49,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     xmlns:exsl="http://exslt.org/common"
     xmlns:date="http://exslt.org/dates-and-times"
     xmlns:str="http://exslt.org/strings"
+    xmlns:fn="http://www.w3.org/2005/xpath-functions"
     exclude-result-prefixes="svg pi"
     extension-element-prefixes="exsl date str"
 >
+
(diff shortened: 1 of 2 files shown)


siefkenj:
Those changes sound good :-). Go ahead and slip them in.

I left `fn` out of the `exclude-result-prefixes` because, theoretically, unprocessed `<map>`, etc., tags should never show up in the output, so I didn't think it would make a difference, but if there's an impact in another way, might as well exclude it.

rbeezer:
Great, thanks.

There are places, like maybe a `xsl:copy-of`, where all the aliases/namespaces get stated again.  I'd like to get even "cleaner" HTML than we do, and that is one piece of this.  Since it is a **developer** who would output a `map`, not an **author**, I think we should rely on not screwing up.  ;-)  I think.

Will test changes and merge shortly.

rbeezer:
Edits made silently, and broke into two commits.  Thanks!  I'll advertise just now on -dev, since I can think of at least two people who need to know.

PR #1997, opened 2023-06-06
Validation-plus: Allow certain Unicode in math only

kcrisman:
Resolves discussion at [this pretext-support link](https://groups.google.com/g/pretext-support/c/H6a969qCZeM)


rbeezer:
Love the test case!  ;-)  Visual scan looks good `<emdash/>` will merge soon.

kcrisman:
Well, we've all seen students do that.  Thanks!

rbeezer:
Broke this into two commits, otherwise as-is.  Usually there is a reference here, but I'm not seeing it.  Still the commits **do** point here.

Thanks for this one!

PR #2002, opened 2023-06-09
HTML: Change body/@class in interactive iframe generation

mitchkeller:
This one line should make Sage interacts generate usable output again.


rbeezer:
Thanks, Mitch!  More on the lists.  Merged and re-building website right now, give it a bit.

PR #2010, opened 2023-06-20
Guide updates

StevenClontz:
I imagine a fuller update needs to be made with a lot of changes @oscarlevin has worked on recently (e.g. no longer having to generate assets), but I wanted to address https://groups.google.com/g/pretext-support/c/le-2d8gTXS0/m/TVIDgY_vBAAJ and do some general standardization of language around Codespaces vs repositories vs Pages.


rbeezer:
Thanks for the updates!  Merged, and website version should be rebuilt later today.

Removed three stray hunks of whitespace, and reorganized commits into two topical pieces.

Issue #2016, opened 2023-06-30
pretext generate interactive previews fails to launch server


dbrianwalton:
See https://groups.google.com/g/pretext-dev/c/f7MAHC0Tg-s

Python on newer Mac's seem to be executing different Python executable interpreters, causing the creation of a subprocess where the Python http.server can be running to fail.

This should be resolvable by launching the same Python interpreter as the current pretext/pretext.py script is using.

rbeezer:
Closed via #2017.

PR #2017, opened 2023-06-30
pretext/pretext.py: Modified call to open http.server to us sys.execu…

dbrianwalton:
I think this should address #2016


rbeezer:
Thanks!  Tested fine on Linux.  More details on the `-dev` thread.

PR #2018, opened 2023-07-07
Menu css

davidfarmer:
This corrects some problems with the "bust menu" in Runestone.

There was inline CSS which needed some attention.  This pull request deletes
that from the XSL.  It has been replaced by PreTeXt CSS.

Also, the CSS class "ptx-runestone-container" was added to the top navbar,
only when building for Runestone.


rbeezer:
Pushed and merged, with no changes to the code.  Thanks.

Let me know if a website update is urgent, otherwise it might be 24 hours or so.

PR #2019, opened 2023-07-12
Add support for landscape figures and tables.

whaynes:
This pull request which adds support for for rotating wide images and tables in latex output.   It uses the latex rotating package.

When a @landscape attribute is added to figures, tables, lists, or listings, the item and its caption/title will be rotated 90° and placed on its own page, but only if publication/latex/@print = 'yes'.   

A landscape figure containing a sidebyside will rotate properly, but because landscaped items move to their own page, putting a landscaped item inside a sidebyside will produce undesirable results.

The PR also includes the ability to rotate bare images by an arbitrary amount with a @rotate attribute,  but I think this feature is a novelty, and probably unnecessary.

The attached pdf demonstrates the results.

[main.pdf](https://github.com/PreTeXtBook/pretext/files/12033075/main.pdf)

### DIFF
--- a/xsl/pretext-latex.xsl
+++ b/xsl/pretext-latex.xsl
@@ -404,6 +404,8 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:text>\usepackage{ifxetex,ifluatex}&#xa;</xsl:text>
     <xsl:text>%% Raster graphics inclusion&#xa;</xsl:text>
     <xsl:text>\usepackage{graphicx}&#xa;</xsl:text>
+    <xsl:text>%% The rotating package provides sidewaysfigure and sidewaystables environments&#xa;</xsl:text>
+    <xsl:text>\usepackage{rotating}&#xa;</xsl:text>
     <xsl:text>%% Color support, xcolor package&#xa;</xsl:text>
     <xsl:text>%% Always loaded, for: add/delete text, author tools&#xa;</xsl:text>
     <xsl:text>%% Here, since tcolorbox loads tikz, and tikz loads xcolor&#xa;</xsl:text>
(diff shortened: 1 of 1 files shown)


rbeezer:
Looking real nice!  I hadn't thought ahead about using the "print" flag, I like that.  Some suggestions:

*  Let's only load the package in the preamble if it is required.  See variables defining necessity at about line 81 and then you can follow on to see applications.  And not next to all the always-loaded packages (later on in the preamble somewhere sensible?).
*  I was going to say that rotation by an arbitrary angle seems a bit too cute?  But I do like the look of a tilted Reddy, and maybe folks will add a bit of visual appeal by rotating some bare images?  It might make future implementations more complicated (I have ideas for a non-LaTeX print conversion), but we can deal with that later.  And maybe we can respect `@rotate` in HTML?
*  Could you add some of your good test cases (figure, sidebyside, bare image, etc.) to the sample article so we can see this as part of testing?   Great place might be:
Subsection 29.1: <image> 
https://pretextbook.org/examples/sample-article/html/atomic-items.html#atomic-image
Definitely use your schematic, and I like Reddy if there are no copyright issues.
*  Documentation needs a just a line or two making it clear these are both possible, and just for print.
*  Remind me to add attributes to the schema!

A pile-on commit is fine (I'll repackage), or you can force-push, either way works for me.

whaynes:
- The latex rotating package now only loads when needed.  
- I have added support for rotating html images with a css style attribute, however the browser doesn't allocate any more vertical space for the rotated image so it will sometimes overlap with the text above and below.  
- Added some examples and brief documentation.
- Regarding Reddy Kilowatt, the copyright status is unclear to me but he is still a registered trademark, so I removed him to be safe.
- If you decide to include these, don't forget to add them to the schema!

rbeezer:
Shoot.  Forgot the schema.  Coming up.

rbeezer:
Thanks for all the changes and additions.  Love the yellow duck.

*  Lots of tabs in the code.  Which really makes a mess for those with different tab widths and for git.  My editor will convert them to spaces automatically as I write.
*  The image rotation is only affecting those given by `@source` and in one case only if the image name has an extension.  So there is work to do if we want universal applicability. See #2036.
*  I've tried to avoid floats in LaTeX.  This may be a place where it makes sense.  We'll see.  It will definitely mess with "page fidelity" so we may want a publisher option? See #2037.
*  HTML rotation will need work for spacing changes.  See #2038.
*  Some repackaging and light editing, but nothing serious enough to mention.

Thanks again for working this one up!

rbeezer:
Schema additions: 7af51c6808a0f3107cf7f572e895da3c71898f14

PR #2021, opened 2023-07-19
Allow multiple macros elements

jjrsylvestre:
Allowing multiple `<macros>` lets the user specify alternative notation for different versions without having to duplicate the entire set of macros, by using `@component` on macro subsets.

The schema already allows _one-or-more_ `Configuration` elements as child of `<docinfo>` so I don't think any changes are needed there.

The string `docinfo/macros` also shows up in `xsl/pretext-beamer.xsl` --- not sure if that needs tweaking.

Tested against html, latex, and pdf outputs.

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -1512,7 +1512,9 @@ Book (with parts), "section" at level 3
     <xsl:variable name="latex-left-justified">
         <xsl:call-template name="sanitize-text">
             <xsl:with-param name="text">
-                <xsl:value-of select="$docinfo/macros" />
+                <xsl:for-each select="$docinfo/macros">
+                    <xsl:value-of select="." />
+                </xsl:for-each>
             </xsl:with-param>
         </xsl:call-template>
     </xsl:variable>


jjrsylvestre:
Alternative solution to this pull request: introduce a no-op element. Then macro subsets in a single `<macros>` element could be enclosed inside the no-op element for inclusion/exclusion via `@component`.

xsl/pretext-common.xsl
```diff
@@ -1512,7 +1512,9 @@ Book (with parts), "section" at level 3
     <xsl:variable name="latex-left-justified">
         <xsl:call-template name="sanitize-text">
             <xsl:with-param name="text">
-                <xsl:value-of select="$docinfo/macros" />
+                <xsl:for-each select="$docinfo/macros">
+                    <xsl:value-of select="." />
+                </xsl:for-each>
```
Alex-Jordan, line 1517: An alternative to this XSL (since we try to rarely use `for-each`):

```
    <xsl:apply-templates select="$docinfo/macros">
```

That should be enough (but test it if you are inclined). The default template for the "macros" element will just apply templates to its child nodes, which are all text. And the the special (non-deefault) template we have for `text()` should end up just doing `<xsl:value-of select="." />` since there are no math elements inside the "macros" element. So, same thing in the end.

Alex-Jordan:
Forgive my musing, but maybe it helps.

Perhaps your should should validate when you have multiple `macros` in source, but only the first one in the tree should be actually used. I'm assuming the `component` mechanism could leave you with a tree where all but one of the `macros` has been removed. Then this PR is not needed, as `<xsl:value-of select="$docinfo/macros" />` will grab the first (and only) one still standing.

If it could actually concatenate two `macros` together, you just have to monitor things like not using `newcommand` in each `macros` block for the same new macro name.  Slightly easier to police yourself on that when it's the same constraint, but only within one `macros` at a time. 

I suppose you are hoping to avoid having to maintain some identical "common" things across multiple `macros` elements. But you could see if using an `<xi:include parse="text">` works for the "common" bits.

rbeezer:
@Alex-Jordan is right about `for-each`.  But the default template for a text node gets extended in each conversion by a "hook" that is template named something like "text-processing" and in particular will convert some characters to tex-isms, which might be dangerous in macros.

Alex-Jordan:
The following is a "trick" I've found will work if I use `xsltproc` or `pretext/pretext/pretext`, but not if I use the CLI.  Since it works with `pretext/pretext/pretext` I feel like it should be possible to get it to work with the CLI too, but it hasn't risen to the level of investigating for me.

You can do things like
```
<xi:include parse="text" xparser=(id('common-macros'))>
```

Without an `href`, then `href` defaults to `""`, meaning this current file you are in. And the above looks for some element with `@xml-id="common-macros"`. So you don't need to manage the common things in a separate file. But you *do* need to put them in something like a "no-op" element with that id :)

Without "no-op", you might also be able to do like:

```
<whatever-is-here xml:id="its-id"/>

<!--    some
    common
    macros -->

<macros component="A">
    <xi:include parse="text" xparser=(id('its-id')/following-sibling::comment())>
    some macros
</macros>

<macros component="B">
    <xi:include parse="text" xparser=(id('its-id')/following-sibling::comment())>
    some macros
</macros>
```

Note that you can't point to an id of something that is self-or-ancestor of the `xi:include`, and that's why I'm referencing the element that came before it all.

Alex-Jordan:
Sorry, I wrote `xparser` where I meant `xpointer`. I updated the comment in GitHub, but this is for your email.

PR #2022, opened 2023-07-20
Fix: Make regex raw strings, to avoid quoting confusion.

bjones1:
For example, to match an asterisk, the regex is either r"\*" or "\\*". The first option is more readable IMHO.


rbeezer:
Thanks, @bjones1.  These all relate to a somewhat brittle method for busting a braille BRF into a sequence of files, one per chapter.  With the major refactor of producing braille mostly in good shape, it will be way more reliable to make this an option of that process (and no regular expressions!).  So I'm going to let this one go unmerged since all the surrounding code will sonnget removed.

Pingin @alexeikolesnikov just because this was his work.

bjones1:
Sounds good!

PR #2024, opened 2023-07-21
pass bytes to ET.fromstring to avoid issue with XML prolog

Alex-Jordan:
This does not hurt anything when contacting a 2.17 or earlier server. It is needed when contacting a 2.18 server because now the server will include the XML prolog line (`<?xml version="1.0" encoding="UTF-8" ?>`) and we have to convert the response to bytes as a consequence.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -1330,7 +1330,7 @@ def webwork_to_xml(
 
         bad_xml = False
         try:
-            response_root = ET.fromstring(response.text)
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks.  Only tested with whichever server is given in the sample chapter configurations.  I've lost track of which are 2.17, 2.18, or something inbetween.  ;-)

rbeezer:
I was careless about recoring this one and fogot to at least drop the commit here.

Merged as b8ae684a47a916a9cc92d5e1df9dfed8a255695f

PR #2026, opened 2023-07-25
Add .logical to an auto-inserted div.para if it contains a block element

siefkenj:
Currently when
```
<ul>
  <li>
    <p><ul><li>foo</li></ul></p>
  </li>
</ul>
```
is processed, the resulting `p` turns into a `div.para.logical`. However, 
```
<ul>
  <li>
    <ul><li>foo</li></ul>
  </li>
</ul>
```
gets an auto-inserted `p` which turns to a `div.para` even though it contains a block element. This PR makes the behavior of the auto-inserted `p` consistent.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -5923,7 +5923,15 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     </xsl:when>
                     <!-- No good test for unstructured? -->
                     <xsl:otherwise>
-                        <div class="para">
+                        <div>
+                            <xsl:attribute name="class">
+                                <xsl:text>para</xsl:text>
+                                <!-- Just like for explicitly added <p>'s, if we contain a block  -->
+                                <!-- element, an additional "logical" class should be added       -->
+                                <xsl:if test="ol|ul|dl|me|men|md|mdn|cd">
+                                    <xsl:text> logical</xsl:text>
+                                </xsl:if>
+                            </xsl:attribute>
                             <!-- Create a derived id, if original.  Somewhat  -->
                             <!-- contrived so it doesn't collide with another. -->
                             <xsl:if test="$b-original">


rbeezer:
Thanks @siefkenj.  Good catch.  It'll be a few days before I can merge.

@davidfarmer: do you expect CSS ramifications?  Would a beta be useful or necessary?

rbeezer:
OK, put this through testing on the sample article.

* Only changes seem to be when the `li` content is a `cd`.  Which may illustrate a shortage of examples.
* Nothing harmful for CSS to fix, it appears.
* I believe an `ol`,`ul`, `dl` cannot appear as a child if an`li` without an intervening `p`.  Or said differently, the list structure **must** be a child of a `p`.  So could be removed from the `@test`?

Shouldn't these "unstructured" `li` be exploded into several HTML `p`?  I wonder if the pre-processor should just wrap the content of an "unstrcuctured" `li` into a `p` and then it will be handled appropriately in all conversions (especially HTML)?

siefkenj:
I am not sure what `@test` refers to. And since we do not use any `<p>` tags anywhere in HTML, I don't think exploding is necessary.

Yes, moving this step to the preprocessor would be an option (that's what I do in the JS version. I first add a `<p>` around anything that is not allowed as a direct child via the schema; of course you'll want to take special care for the `<md>/<m>` tags, as you pointed out to me)

rbeezer:
> I am not sure what @test refers to

```
 <xsl:if test="ol|ul|dl|me|men|md|mdn|cd">
```

I don't think `ol|ul|dl` are necessary as it never happens this way.

siefkenj:
Are you saying that the case
```
<ul>
  <li>
    <ul><li>foo</li></ul>
  </li>
</ul>
```
is handled elsewhere in the code?

rbeezer:
Your interior `ul` has to live in a `p`, according to the schema.  Who knows how permissive the conversions are.

Some exceptions, such as `GOAL-LIKE`, presumes an interior list not in a `p`.

siefkenj:
The `ul` is current auto-wrapped in a `p` as per line 5919.

PR #2027, opened 2023-07-28
Configurable asymptote server URL

StevenClontz:
Allows library users to configure their remote asymptote server URL

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -246,7 +246,7 @@ def mathjax_latex(xml_source, pub_file, out_file, dest_dir, math_format):
 
 
 def asymptote_conversion(
-    xml_source, pub_file, stringparams, xmlid_root, dest_dir, outformat, method
+    xml_source, pub_file, stringparams, xmlid_root, dest_dir, outformat, method, server_url="http://asymptote.ualberta.ca:10007"
 ):
     """Extract asymptote code for diagrams and convert to graphics formats"""
     # stringparams is a dictionary, best for lxml parsing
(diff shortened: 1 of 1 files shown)


sean-fitzpatrick:
Has anyone else set up another server?

And is it also possible to specify local compilation instead of a server?

StevenClontz:
This is part of a standardization of how CLI 2.0 will handle features that can be run locally or via SaaS, so you'll get your overdue local asymptote support finally.

StevenClontz:
I have no clue if anyone else is running a server but with this PR it will be an option for our users.

sean-fitzpatrick:
Great! To my knowledge there is no other server, but it is now possible for one to exist.

rbeezer:
Core functions are predicated on a publisher file, and will provide different output depending on that input.

We specify the WeBWorK server in the publication file, having moved it away from a command-line argument.

So, to keep life simpler for publishers, I think to specify an alternate Asy server we should mimic the procedure in place for a WW server.  The local v. server dichotomy will be new ground, since we haven't had this option for a WW server.  Probably local/server should be a yes/no attribute with an additional attribute for a server URL (defaulting to Alberta).  Not sure just where it should go in the publisher file, perhaps under `common` since it affects **all** output formats.

bnmnetp:
How do I configure a book to tell it what sageplot server to use?

Both of these are interesting and important to me, because post 1.7.0 many books want to rebuild these plots to get their hash...  I don't have sage installed on the server that does the builds, and I would rather just have it pointed to a server in any case.

rbeezer:
There is no "sageplot" server.  You are assumed to have a local installation of Sage with an executable, sorta like you are assumed to have a local installation of LaTeX.  (Yes, I know that sounds suboptimal.)

The Sage Cell server has at least two APIs which could be leveraged in the core PreTeXt Python, I guess, but I'm not familiar with the details.  I think this is the simpler one:

https://github.com/sagemath/sagecell/blob/master/doc/session.md

Start a new issue on the PreTeXt side?  (You've hijacked this one!)

StevenClontz:
Missed these comments while at MathFest - will review how webwork server stuff is handled and propose a comparable pattern here.

rbeezer:
This has been dormant for over a month.  Reopen if you want to pursue this.

StevenClontz:
Sorry, I got sidetracked with https://github.com/PreTeXtBook/pretext/pull/2044 and didn't return to this.

I think the "right" thing here is to find the right design for allowing users to specify local vs server for webwork (pretty sure local webwork is in the plans), asymptote, and whatever else may have both local and server support in the future. I can contribute to this but maybe a more active user of one or the other should take lead.

Issue #1347, opened 2020-08-12
Programming languages for Sage cells inside knowls


rbeezer:
Sage cells in knowls are not being configured for different (programming) languages, like R, Python, Octave, etc.

Outside of knowls, we have put the language on the Evaluate button.  So, we might have button text "Evaluate (Python)".

Can we pass in a **list** to the knowls code to make multiple configurations, each with its own label text?  Items of this list might need to be pairs (with correct syntax!):
```
[(python, 'Evaluate (Python)'),....]
```
This is a follow-on from #981.

davidfarmer:
Passing a list may involve a major change to the knowls code.
(Or maybe not: I will investigate.)

A global variable, which is not particularly frowned upon
in Javascript (not like in some other languages) could do
what you want.

rbeezer:
I meant setting a variable, like have now.  "Passing" was colloquial, not technical.  (Sorry!)

I think it would mean looping over that list, building several "makeSageCell()" calls.  (Function name may not be correct.)

JeanSebastienTurcotte:
@rbeezer Sorry if you already had my ping in #981 , but it seems something is still lacking there. I didn't get an email when I posted in 981, so I figured notification are maybe off on a closed thread.

PR #2041, opened 2023-08-12
first go at adding autoeval to sage cells

rpruim:
This is a POC implementing `@autoeval='true'` **for python only**.  It would be easy enough (though a bit tedious) to replicate this for the other languages.  But probably best to figure out if we like this approach or want to improve it before implementing it across the other languages.

I've tested it with a book I'm writing, and it seems to work in the main text.  Still need to do some checking regarding knowls.

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -1678,6 +1678,9 @@ Book (with parts), "section" at level 3
         <xsl:with-param name="language-attribute">
             <xsl:value-of select="@language" />
         </xsl:with-param>
+        <xsl:with-param name="autoeval-attribute">
+            <xsl:value-of select="@autoeval" />
+        </xsl:with-param>
         <xsl:with-param name="in" select="''"/>
         <xsl:with-param name="out" select="''" />
     </xsl:apply-templates>
@@ -1731,6 +1734,10 @@ Book (with parts), "section" at level 3
         <!-- OK to send empty string, implementation reacts -->
         <xsl:with-param name="language-attribute">
             <xsl:value-of select="@language" />
+        </xsl:with-param>
+        <xsl:with-param name="autoeval-attribute">
+            <xsl:if test="not(@autoeval)">false</xsl:if>
+            <xsl:value-of select="@autoeval" />
         </xsl:with-param>
         <xsl:with-param name="in">
             <xsl:call-template name="sanitize-text">
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -9591,6 +9591,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <xsl:template match="sage" mode="sage-active-markup">
     <xsl:param name="block-type"/>
     <xsl:param name="language-attribute" />
+    <xsl:param name="autoeval-attribute" />
     <xsl:param name="in" />
     <xsl:param name="out" />
     <xsl:param name="b-original"/>
@@ -9609,7 +9610,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 <xsl:text>sage</xsl:text>
             </xsl:if>
             <xsl:value-of select="$language-attribute" />
+            <xsl:text>-</xsl:text>
+            <xsl:value-of select="$autoeval-attribute" />
         </xsl:attribute>
+        <xsl:attribute name="autoeval">
+            <xsl:value-of select="$autoeval-attribute" />
+        </xsl:attribute>
+
         <xsl:attribute name="id">
             <xsl:apply-templates select="." mode="html-id"/>
         </xsl:attribute>
@@ -10058,7 +10065,6 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <xsl:with-param name="content">
                 <map xmlns="http://www.w3.org/2005/xpath-functions">
                     <string key="inputLocation">.sage-interact</string>
-                    <boolean key="autoeval">true</boolean>
                     <array key="hide">
                         <string>editor</string>
                         <string>evalButton</string>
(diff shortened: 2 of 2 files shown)


rpruim:
I agree that

* It would be nice if we could come up with a way to reduce the code redundancy.  I'm not familiar enough with xsl to know whether there is something that can be done on that side of things.  Seems like a great setting for a macro that could be called with a couple of arguments...  Does the current approach indicate that this sort of thing isn't possible?
* I like the idea of allowing an author to indicate whether autoeval cells should appear.  This could apply to both static and dynamic formats potentially.  It seems like there are two sorts of use cases:
    * A short bit of code -- for example code to load some standard packages (my main use case)
    * A longer bit of code that loads some custom functions, datasets, whatever.
   It's possible that the second use case could be replaced with a short bit of code that loads code located in a github repository or similar.  (See David Austin's Activity 6.1.3 
https://understandinglinearalgebra.org/sec-dot-product.html#activity-69, for example.)  That might be a good practice to encourage. 

    Another thought: perhaps we could allow a non-editable, auto-evaluated cell to be collapsable (in dynamic formats).  But that might go against the advice to keep the code chunks short.

rbeezer:
OK, a couple puzzling hours with the Sage Cell and knowls.  (Some of the puzzlement was my own damn fault.)

Short answer: no new code, and I think I understand things a little better.

Gotcha: a page could have a cross-reference to an `example` (say) far-away on some other page and where that example is born it has a Sage cell in it.

Cross-reference is implemented as a knowl.  The file with the content of the knowl (the example) has a `head` which does the necessary `sagecell.makeSageCell()` (or whatever) given what sort of cells are in teh example.  But if the far-away page has setup/autoeval code (near its beginning), then that is not available to the knowl.  This isn't new with auto-evaluation, we just need to advise something like the `example` to be self-contained.

rbeezer:
Well, you can view XSL as one giant templating language and everything is a macro.  Or marco-ish.  You've used the `xsl:param` way of passing arguments.  Clumsy, but about as good as it gets I think.

I think I will edit your PR here slightly, just for Python, and leave your name on it.  Then I'll expand to multiple languages.

Let's just support the "short code" use case for right now.  No changes for static, no new arguments.  "Long code" brings up hiding in the different formats, which can be an extension once we see how this shakes out.

rpruim:
Thanks.

This sounds like a good plan for the moment.  Where should we document this, including some advice about what happens to code chunks in knowls?

rbeezer:
*  First, nice job on the XSL - this stuff is a bit of a mess to work with.
*  Booleans in XSL are odd.  You gotta use a variable (or param) to build them.  They test normally, but their "value" is a textual `true` or `false`.  You might pick up some tricks looking through the commits.
*  Built a template to make the necessary class name.  With `autoeval` is the signal, otherwise missing (like before this work).  Made testing easier (which was **not** the reason to do it!).
*  Your "internal" `autoeval` attribute is gone, I think the XSL booleans replaced its function.
*  Source attribute is now `auto-evaluate` since we try to minimize abbreviations except for frequently-used elements.  And we consistently use `yes` and `no` for values.
*  In the schema now, which restricts the attribute values, so we only need to do a simple check for `yes`.
*  Double the JS setup calls, couldn't think of a way around it.
*  Two small examples in sample article, Subsection 4.4, iirc.  A day and a half on this, so not much enthusiasm for more documentation.
*  Six commits total, put your name on two of them.

Thanks for your patience while I waded through other things to get here.

rpruim:
Thanks for taking a look and for cleaning things up.  Since I'm new to both xls and pretext, I've still got lots to learn there. 

Too bad about the duplicated code for the JS setup.  I also couldn't figure out if there was a cleaner way to do this, but I thought that might just have been due to my lack of experience with this stuff.

rbeezer:
> Too bad about the duplicated code for the JS setup

It is supporting `@hide` that has me worried.  :-(  Maybe we'll just build a class for hidden cells automatically?  I dunno.

I know the people who designed this, and I wish it had been more flexible...

rpruim:
@rbeezer , as you seemed to fear,  I think I have discovered an issue with the current implementation of `<sage>`.  Something doesn't seem to be working quite right when `<sage>` is in a `<solution>` or `<answer>`.  Here is an example:

Activity 2.3.1: Using Python to find row reduced echelon matrices 
https://ula-ds.netlify.app/sec-python-introduction#activity-13

Things are working well enough that the sage cell knows to evaluate as Python, but the initial code is missing and the cell is not editable.

Source for the <sage> looks like this:

```
		<sage language="python">
			<input>
A = np.array([[-1, -2, 2, -1], [2, 4, -1, 5], [1, 2, 0, 3]])
print(A)
print(sympy.Matrix(A).rref())
			</input>
			<output>
[[-1 -2  2 -1]
[ 2  4 -1  5]
[ 1  2  0  3]]
(Matrix([
[1, 2, 0, 3],
[0, 0, 1, 1],
[0, 0, 0, 0]]), (0, 2))
			</output>
		</sage>
```

kcrisman:
One stab: Is it possible that the issue is that where SOLUTION_LIKE is processed, [beginning here](https://github.com/PreTeXtBook/pretext/blob/master/xsl/pretext-html.xsl#L4743), it doesn't process properly?  After all, the solution/answer is in a knowl, which in the past caused problems with Sage (https://github.com/PreTeXtBook/pretext/issues/190, see also https://github.com/PreTeXtBook/pretext/issues/1347).  Randy, the code looks very similar to the code for `DISCUSSION-LIKE` (also born hidden) immediately after that, do discussions work okay?

rpruim:
I'm not sure I know what discussion-likes are and I'm not locating them at https://pretextbook.org/doc/schema/ or in the guide.  I'm happy to give it a try, but I need a nudge.

kcrisman:
As it turns out, I wasn't sure either!  And not in schema, see https://github.com/PreTeXtBook/pretext/blob/master/xsl/entities.ent  

```
<!-- appendages of OPENPROBLEM-LIKE -->
<!ENTITY DISCUSSION-LIKE "context|discussion|opinion|status|suggestion">
<!ENTITY DISCUSSION-FILTER "self::context or self::discussion or self::opinion or self::status or self::suggestion">
```

Open problems appear [here](https://pretextbook.org/examples/sample-article/annotated/section-open-problems.html) but without this "appendage", so I'm not sure quite how to use it.

rpruim:
`<sage>` in `<discussion>` in `<openquestion>` works for me, but it isn't born hidden:

image

rbeezer:
This discussion should probably happen on `pretext-dev` where more people will see it, rather than on a closed pull request.

(a) Open problems are a work-in-progress and the requesting party could have taken responsibility for letting you know that.

(b) I am skeptical that this is a consequence of the pull request.

@rpruim: Can you checkout a reasonable commit prior to the PR and see if a `sage` in a `solution` behaved correctly at that point?  I'm going to guess not.

In any event, the original diagnosis, with updaed information, could be copied into a new thread on `-dev`.  Thanks.

kcrisman:
> (a) Open problems are a work-in-progress and the requesting party could have taken responsibility for letting you know that.

If the requesting party had realized that, yes! 😃  I was just looking at the code near `SOLUTION_LIKE`.

Yes, at this point -dev seems reasonable.  Initially, I thought maybe it could be quickly figured out here.

rbeezer:
> Something doesn't seem to be working quite right when <sage> is in a <solution> or <answer>.

I did the suggested checkout of a prior commit.  I suspect the problem is the use of a language other than Sage.

I've started a `-dev` thread.  Please make anymore comments there, I'm only posting here so interested parties are aware.

PR #2043, opened 2023-08-12
Avoid unintentional mutation of stringparams

StevenClontz:
Currently core takes given stringparams and modifies them to implement certain functionality. This modification affects the arguments passed, which can be a surprise if the dictionary is reused in the calling script or library.

This PR creates copies of passed stringparams, so they can be modified safely without mutating the passed argument.


rbeezer:
Yep - good idea!  Thanks.

Some whitespace edits.  Limited testing (three, of several, routines).

I guess you are clear to unwind https://github.com/PreTeXtBook/pretext-cli/commit/1196a540dd12461e74476f9d4436010d71d81690

PR #2044, opened 2023-08-14
Webwork extract refactor

StevenClontz:
This PR modifies `extract-pg.xsl` to output an XML file, which Python reads with lxml to obtain the data it requires to create webwork-representations.xml.

I am able to successfully build webwork-representations.xml for both the minimal and sample-chapter webwork examples using this script, and they are identical to the current result of the head of the master branch.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -1068,16 +1068,50 @@ def webwork_to_xml(
 
     # Build dictionaries and localization string into a scratch directory/file
     tmp_dir = get_temporary_directory()
-    ww_filename = os.path.join(tmp_dir, "webwork-dicts.txt")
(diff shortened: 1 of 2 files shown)


rbeezer:
I checked the first part of this yesterday, just the effect of changes to `extract-pg.xsl`.  They are as expected with regard to new XML containers.

With a switch from text output to XML output, the ampersand and less-than are being excaped, as they would need to be.  And I guess that is fine since once consumed by `lxml` they behave properly.

I'll check the whole pipeline here shortly.  I'd also like to have input from @Alex-Jordan, or at least have him say he can't get to this one.

rbeezer:
Reorganized as two logical commits, but no change to the overall code.

Zero appreciable change to the representations file for the sample chapter.

Issue #2033, opened 2023-08-03
Running head oddities in two-sided latex output


jjrsylvestre:
In two-sided latex output, there are blank pages inserted to ensure new parts/chapters open on a recto page. Shouldn't these pages be completely blank, without running head content? They don't really belong to the chapter that has just ended.

In any case, here are a couple of oddities I noticed. The following refer to the [print PDF version of my Discover Linear Algebra book](https://sites.ualberta.ca/~jsylvest/books/pdf/JSylvestre-DiscoverLinearAlgebra1-2023-Print.pdf).

1. Appendix A ends on document page 347 (pdf page 355). Then a blank page (with running header) so the Bibliography can open on an odd page, then the Bibliography (single page) then another blank page so that the index can open on an odd page. But look at that blank page after the Bibliography --- it still has running header referring to Appendix A.
2. The last chapter (chapter 22) ends on document page 337 (pdf page 345). Then the next page is blank so that the appendices can begin on an odd page. The last page of chapter 22 has no running head at all (running head for a recto page should be [Section#] [SectionTitle]). The blank page following the last page does have a running head, but it's only the chapter title (running head for a verso page should be Chapter [Chapter#]. [ChapterTitle]). This is the only chapter ending in the whole book that displays these oddities --- for example, compare with the end of chapter 17 (document pages 257-8, pdf pages 265-6).

I think some of this can be fixed by just inserting `\thispagestyle{empty}` into the definition of `\cleardoublepage`.

Alex-Jordan:
> But look at that blank page after the Bibliography --- it still has running header referring to Appendix A.

This sounds like a familiar issue with latex. You start an appendix with `\appendix` but you have no real tool to declare the appendix has ended. Only the option to start a new appendix.

jjrsylvestre:
Seems like the missing/incomplete running head is a last-page-of-a-part problem, and the manifestation of the issue depends on whether that last page is even or odd:
* Last page of Part I is document page 133 (PDF page 141), and the running head is missing completely, when it should be "10.5 Theory"
* Last page of Part II is document page 306 (PDF page 314), and the running head is "Column, Row, and Null Spaces" when it should be "Chapter 20. Column, Row, and Null Spaces"
* Last page of Part III is document page 337 (PDF page 345), and the running head is missing again, when it should be "22.6 Theory"

jjrsylvestre:
[...]

PR #2045, opened 2023-08-15
LaTeX: empty page style on extra blank pages

jjrsylvestre:
Addresses part of #2033.


rbeezer:
Very good!  Merged unharmed.  Website examples, etc will be updated sometime later today.

Thanks for all the work to clean up these details!

PR #2046, opened 2023-08-15
Guide: use latex generator in the preface

jjrsylvestre:
`<m>\LaTeX</m>` causes a compilation error using TeX Live 2023.66984.


rbeezer:
Oops, shouldn't have let than one slip in as part of dbf48da9ea5910b95647ebdf93d636a8a4e9d41e.  Thanks for the fix.

N.B. A true merge this time.  No commit hashes were harmed. ;-)

PR #2047, opened 2023-08-16
Add extra classes to identify runestone component types earlier

ascholerChemeketa:
Should not affect any existing css - only adds a few new classes to make it easier to identify what kind of item a runestone component is at the point it may need to become wide


ascholerChemeketa:
This is finished AFAIK

rbeezer:
Thanks, all merged in.  I may rebuild the website samples shortly.

Edits to commit messages, but no changes to the code.

PR #2048, opened 2023-08-17
Navbar updates

ascholerChemeketa:
1. Makes sure every navbar control has a span.name and span.icon for maximal flexibility/consistency of display
2. Changes runestone controls to be buttons instead of divs to better semantically match their behavior
3. Adds grouping elements and reorders navbar controls only if HTML navbar style is "wide"

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -10946,7 +10946,16 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     <xsl:call-template name="native-search-results"/>
                 </div>  <!-- banner -->
             </header>  <!-- masthead -->
-            <xsl:apply-templates select="." mode="primary-navigation"/>
+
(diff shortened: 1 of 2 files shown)


ascholerChemeketa:
I am happy to extend part 3 to be universal - but handling the changes in default and CRC themes involves making decisions on content ordering I don't want to make assumptions about (i.e. Does Next/Up/Prev belong on the far right or does Runestone? Where does search button go in relation to both of those?)

It is a little ugly to have a switch in the xsl to generate different versions of the navbar based on chosen HTML theme. But it may be necessary. Reordering items visually using CSS is bad accessibility practice, so the only way to do it gracefully in the final output is to produce different HTML.

ascholerChemeketa:
Question about the implementation of part 3
Would it be better to move that to a separate xsl file that overrides the default?

ascholerChemeketa:
Closing for now - will reopen when ready

PR #2050, opened 2023-08-22
WeBWorK: recognize DropDown macro

Alex-Jordan:
WeBWorK has a new macro `DropDown` for making a "select" style multiple choice. It's just a better version of `PopUp`. The new `DropDown` has an automatic placeholder. This commit lets PTX recognize the macro and automatically load its library file.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -1642,7 +1642,7 @@
 
                 <webwork>
                     <pg-code>
-                        $color1 = PopUp(["?","Red","Blue","Green"], 2);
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
Bah, I had the server set to `webwork.runestone.academy` for recent testing about that.

The new tool here is a 2.18 tool, and I haven't upgraded AIM servers yet. Hold on this until I make that upgrade? Sorry for the wild goose chase.

Alex-Jordan:
Once #2115 is processed, I will close this and open something similar. #2115 gets the basic idea going for how to recognize that a certain feature is only going to work in certain versions of WeBWorK, and then catch that when the server version is established at the start of gathering representations.

PR #2051, opened 2023-08-31
Add a database attr for sql activecode

ascholerChemeketa:
It looks like an option to specify something that turns into the data-dburl attribute needed in HTML output didn't make it into activecode. There is an option to do so in hparsons. This copies the code from there to allow activecode elements of language SQL to specify a database file.

Also adds some minimal documentation to the guide.

### DIFF
--- a/doc/guide/author/topics.xml
+++ b/doc/guide/author/topics.xml
@@ -3197,6 +3197,8 @@ displayed line, and there are no <c>\\</c>s.  Use <c>\amp</c> to mark the alignm
 
             <p>Note also that a data file may be provided independently for consumption by an ActiveCode program.  See <xref ref="topic-datafile"/>.</p>
 
+            <p>ActiveCode elements that are language <c>sql</c> can make use of an SQLite database file. This is different than a <c>datafile</c> (which is expected to be text or an image). To include a database file, use the <attr>database</attr> on the <tag>program</tag> element and specify the file to load as a string relative to the <attr>@external</attr> top-level directory</p>
+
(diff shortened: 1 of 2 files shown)


rbeezer:
I'm trying to add a small example to the sample book and not coming anywhere near close, so it feels like there are bigger problems (in ohter words, I think the code here on the PR looks fine).  Trying

```xml
<program label="activecode-sql-database" interactive="codelens" language="sql" database="sql/test.db">
    <input>SELECT * FROM test</input>
</program>
```

where `sql/test.db` already exists for the horizontal Parsons problems.  Were you successful making an example with this PR applied?  We can talk at Drop-In if that is easier.

ascholerChemeketa:
I've got it deployed as a monkey patch in an add on xsl file on a book. It is mid-conversion to PTX, but here is a page that uses it:
https://author.runestone.academy/ns/books/published/welcomecs2/sql-intro_selecting.html?mode=browsing

Here is the source:
https://github.com/ascholerChemeketa/cs160Reader/blob/0290f1e5776604e20e808cd0d705437054c421ef/pretext/SQLIntro/selecting.ptx#L34

Looks like your sample has `interactive="codelens"` - try `interactive="activecode"`

rbeezer:
>  try `interactive="activecode"`

That was it, I knew it must have been something silly.  Confused myself with trying to figure out where to place it in the sample book.  I'm going to merge this now.

rbeezer:
Merged just now.  Thanks for catching this, I didn't know that was an option for the ActiveCode version of using SQL.

Which is why I like to have one of everything in the samples.  Especially when they illustrate some sort of fix.  ;-)

Would you be willing to add your current example to the sample book via a PR?  It could go at the end of last chapter in the first section on "ActiveCode".  Right before the "nonsense" paragraph.

PR #2053, opened 2023-09-01
Add activecode example that loads an SQLite db

ascholerChemeketa:
As requested in previous PR

### DIFF
--- a/examples/sample-book/rune.xml
+++ b/examples/sample-book/rune.xml
@@ -233,11 +233,12 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <program xml:id="sql-using-db" interactive='activecode' language="sql" label="sql-using-db" database="sql/test.db">
                 <input>
                     SELECT * FROM test
-                    ====
+                </input>
+                <tests>
                     assert 1,1 == world
                     assert 0,1 == hello
                     assert 2,1 == 35
-                </input>
+                </tests>
             </program>
         </exercise>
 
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
@rbeezer If the database doesn't actually load on a non-runestone hosted book check the console for errors. If there is a 404 for sql-wasm.wasm that is a runestone code issue. I *think* that the Runestone code assumes the server will be serving that file from `"/_static"`.

You can get a copy of the file out of here if you want:
https://github.com/sql-js/sql.js/releases/download/v1.5.0/sqljs-wasm.zip

But we can also look at getting Runestone to package that up with other js.

examples/sample-book/rune.xml
```diff
@@ -222,6 +222,24 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
...
+
+        <exercise xml:id="exercise-sql-using-db">
+            <title>An SQL program that uses an SQLite database file</title>
+            <statement>
+                <p>Select all the columns of all the rows in the <c>test</c> database table.</p>
+            </statement>
+            <program xml:id="sql-using-db" interactive='activecode' language="sql" label="sql-using-db" database="sql/test.db">
+                <input>
+                    SELECT * FROM test
+                    ====
```
bnmnetp, line 236: This may work, but technically should't the test code be surrounded by `<test> ... </test>`

rbeezer:
Follow-up: see https://github.com/RunestoneInteractive/rs/issues/351

rbeezer:
Merged this with the `tests` fix.  Not running properly in all casews as described in RS issue noted just now.

PR #2056, opened 2023-09-04
Image description

Alex-Jordan:
Here's what I have for image descriptions. 

FWIW, I would call this ready for review. However I am unable to work on EPUB production. My laptop can't deal with the stage that makes speech strings (at least that is where I think it is failing.) New laptop coming sometime in October, but I just can't to EPUB or Braille production right now. So I'm pretty sure the full `description` comes out bad for EPUB. Unless the EPUB reader knows what to do with `details` and `summary` HTML elements; maybe it does?

The `details`, `summary`, and containing `div.image-description` will need styling.

Commits separated in a good faith effort to help with before/after review of HTML.


Alex-Jordan:
Forgot to mention: awaiting a review from local accessibility staff.

Alex-Jordan:
This is updated (force pushed) with two changes:

* With feedback from Jennifer, put `title="description"` on the  HTML summary element.
* Schema altered to require decorative attribute XOR shortdescription, and if the latter, optional description.

More to say in the discussion thread.

rbeezer:
The test for elements besides `var` doesn't need to use a `count`, since an empty node-set tests as false.  And then it can be done in the match, which means no two-part choose is needed.  See 5ee05152ed425fe62bc18502eac3969ce1ecf100

More reporting on the -dev thread, so wait for that.

rbeezer:
Visual inspection looks real good.  Thanks for all the collateral cleanup.

Any harm in merging this and getting the EPUB icon sorted out afterwards?

Think we should delay broad announcement until Oscar has seminar writeup completed?

Alex-Jordan:
No harm, I think.

Probably someone with interest and a little time available could polish the CSS after it's merged.

rbeezer:
Repackaged and merged.  No braille - that needs more work and I may do it just now.

Thanks very much for the work on this, and for modernizing the "svg wrapper" template.  That has had a tortured history.

Website update initiated just now.

rbeezer:
Someplace I said I would make the `describedby` id linkage more reliable.

Done at: bf2d6f14fc19df300a509706ebca52d7af5c0db4

PR #2059, opened 2023-09-05
Use localization on toc help text

ascholerChemeketa:
Bug fix - TOC button help text was not localized. This does that and moves help text to "title" from "aria-label" so it also works as a tooltip for icon only views of TOC


rbeezer:
Thanks!  Merged, but will wait on a website rebuild.

PR #2061, opened 2023-09-07
Allow any combination of url & source in brandlogo

ascholerChemeketa:
Implements todo in pretext-html.xsl

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -13359,24 +13359,24 @@ TODO:
 
 <!-- Brand Logo -->
 <!-- Place image in masthead -->
-<!-- TODO: separate url and image, now need both or neither -->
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
If no URL, tries to use $baseurl so that image behaves the same as the title it is next to. If that fails it resorts to linking back to the same page.

rbeezer:
Very good.  Thanks!

Some XSL nits:
* I put actions for `xsl:when` onto their own indented lines, as discussed at Drop-In.
* `<otherwise/>` works just as well for a no-action final eventuality.
* A `!=` operator has bit me before.  Maybe an empty node-set and an empty string both become `false` - I don't know.  So I go
```
not( [node-set] = '')
```
made that edit here.

Likely website update before the day is out.

PR #2064, opened 2023-09-10
Schema: Allow additional modular files

oscarlevin:
This PR contains two commits.  The first allows `<backmatter>` to be in its own `<xi:include>`-ed file.  The second allows figures, worksheets, and webwork elements to do the same.  It's not as clear that the second group should be isolated to their own files (although the `pretext new demo` template has them.  In any event, I don't think this hurts since it is still required that these elements are included in the right spot based on the rest of the schema.

### DIFF
--- a/schema/pretext.xml
+++ b/schema/pretext.xml
@@ -70,7 +70,7 @@
         <fragment xml:id="start-elements">
             <title>Start elements</title>
             <code>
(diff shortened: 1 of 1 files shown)


rbeezer:
Got it!  Thanks.  Mostly this is so editors are aware that some file is not really the entire document, right?

When validating the full document, we also need the `XMLBase` pattern for `xml:base` which seems to be added automatically (and silently) by the `xi:include` scheme, or something like that.

Pushed, and waiting on next website update.

oscarlevin:
That's right.  Editors complain in the modular file, since they don't know which file included it.

PR #2067, opened 2023-09-14
Schema: correct "website" contents

oscarlevin:
Assuming the deprecation warning about `<website>` is correct, this should fix the schema for that.


rbeezer:
Thanks!  I should have checked that this was already expressed in current `docinfo`.

rbeezer:
Forgot to say: schema browser, etc, will update next time I update the website.

PR #2070, opened 2023-09-15
Geogebra material customization of presentation

dbrianwalton:
This PR addresses the implementation of additional attributes to interactive[geogebra] so that it is possible to add the toolbar, algebra input, and reset icon, as well as to specify the source material's pixel width for cases where it is different than the default 800. It also changed from a fixed iframe size (that was based on a particular 9:5 aspect ratio) to one that uses the provided aspect ratio. The sample article was also updated in Chapter 14 to describe these options, and the provided example had the reset icon added.

Initiated based on pretext-dev thread https://groups.google.com/g/pretext-dev/c/l8zMPV0WYiA


rbeezer:
Thanks again, @dbrianwalton!

Some edits, I hope you don't mind.

*  I like the `xsl:when` spread across three lines, so I just added newlines.
*  I think `<xsl:value-of select="@foo = 'bar'"/>` would produce text that is `true` or `false`.  But I didn't give it a try.
*  I try to never use upper-case for element or attribute names, so changed from CamelCase to lower-case and hyphens.
*  Also try to save abbreviations for frequently-used element names, so made `srcWidth` into `material-width`.
*  We have an "attr" element for attributes that provides the at-symbol (an abbreviation, immediately contradicting myself!).  Replaced a `c` element one place.

Whew!  Sorry about all that.

dbrianwalton:
Those edits all sound good. I'll file the standard practice for style in my memory for future submissions.

PR #2071, opened 2023-09-16
Schema: Fix incorrect start elements

oscarlevin:
This is totally my fault, in 74b0796b9c1ccbdbe2f952b0c283807a526943e3.  I added a few new start elements so editors wouldn't complain about modular files.  Forgot that `Worksheet` is only in the experimental schema, and mis-capitalized `WebWork`.  The result is that the nightly build of the CLI quits before building since it tries to validate the schema.

### DIFF
--- a/schema/pretext.xml
+++ b/schema/pretext.xml
@@ -70,7 +70,7 @@
         <fragment xml:id="start-elements">
             <title>Start elements</title>
             <code>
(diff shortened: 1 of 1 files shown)


rbeezer:
Well, `WebWork` is consistently wrong everywhere else!  ;-)

In now, schema browser will update later.  Carry on.

oscarlevin:
Thanks.

PR #2075, opened 2023-09-23
HTML: adjust spacing of icons in footer

whaynes:
This pull request removes the width attribute on the PreTeXt logo in the page footer.  

This width is preventing the footer's flexbox from distributing the PreTeXt, Runestone, and MathJax icons evenly.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -12528,7 +12528,7 @@ TODO:
 <xsl:template name="pretext-link">
     <a class="pretext-link" href="https://pretextbook.org" title="PreTeXt">
         <div class="logo">
-            <svg xmlns="http://www.w3.org/2000/svg" width="100%" height="100%" viewBox="338 3000 8772 6866">
+            <svg xmlns="http://www.w3.org/2000/svg" height="100%" viewBox="338 3000 8772 6866">
                 <g style="stroke-width:.025in; stroke:black; fill:none">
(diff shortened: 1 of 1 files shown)


davidfarmer:
I am not seeing how this makes a difference in the footer,
nor that it is not evenly spaced already:

https://pretextbook.org/examples/sample-article/html/derivatives.html

whaynes:
I understand that flexbox's justify-content: space-around is supposed to distribute the items evenly with the left and right gaps half the size of the gaps between the items.

This is what I see with and without the width="100%".  Maybe it is only a problem on Safari?
￼
Will

whaynes:
This is what I see with and without the width="100%".
footers

This is what I see on my iPhone
IMG_7919

rbeezer:
Where are we on this one?  Spacing looks fine to me in Firefox and in Chrome, both wide ans mobile-skinny.  Maybe we can ask for some Safari-testing on the `-dev` list to see if that is the real culprit.

I'm hesitant to think that editing the one SVG image directly is the solution here.

davidfarmer:
In the current sample article, I see the icons equally spaced, with or without the "feedback" button.

So I suggest closing this.  Please reopen and explain what I missed if this still seems
to be a problem.

I tested in Firefox and Chrome.

rbeezer:
Thanks, @whaynes for your work on this.

Yes, I think this is OK, or perhaps only a problem on Safari (which won't be the first time).  Please reopen this if you think it still needs further consideration.

whaynes:
It's a problem I see consistently on my browser (Safari) that annoys me, but I can live with it.

PR #2076, opened 2023-09-24
First pass footnotes to details

ascholerChemeketa:
Still a draft. PR created so you have access to code when I hopefully pick your brain on Tuesday.

Sample with "Inline" styling applied is here:
https://computerscience.chemeketa.edu/ascholer/pretext-test/footnotes/output/sample-book.html
Footnote 3 has some embedded math.

Need to pick your brain about `<xsl:template match="fn" mode="wrapped-content">` and best way to filter unwanted content from footnotes.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -1967,7 +1967,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <xsl:template match="commentary" mode="xref-as-knowl">
     <xsl:value-of select="$b-commentary" />
 </xsl:template>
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, Andrew.  Looks good, and I'll look a little closer before Tuesday.

rbeezer:
I'm wondering if `details` is going to be the right thing here.  We'll have block-level content inside `p`.  Which I think won't validate.

Natural enough to park content at the bottom of the page, and use a pointer with an `id` value (or similar) to associate the mark with the content?  Which is basically how born-hidden knowls worked.

Conceivably you could probably start with a commit that reverts 5bffecdcc297c58f461cba0ab18e959597ea8e2f.

Alex-Jordan:
I lean against `details` for inline content. But do we actually have HTML `p`? We have `div.para` in its place, I think. FWIW, that is a legal parent for a `details`.

ascholerChemeketa:
@Alex-Jordan when @rbeezer and I last talked, we decided to focus on "in place block" implementation that would visually split the details into a block in its original location - so I think that is the current direction.

I confirmed that `details` inside true `p` does not validate. But as Alex points out, we have vanishingly few actual `p` tags. There are many things we stuff into `div.para` that probably won't validate inside a `p`.

rbeezer:
Thanks, @Alex-Jordan, for the reminder.  I keep forgetting that! :-(

And thanks, @ascholerChemeketa, for the follow-up.  Perhaps this will goad me into removing any last `p` floating around...

ascholerChemeketa:
Closing. Replaced with https://github.com/PreTeXtBook/pretext/pull/2095

PR #2079, opened 2023-09-28
Make datafile "questions" be marked as optional in runestone-manifest

ascholerChemeketa:
Datafiles are manifested as questions. Those are counted as "activities" that must be completed in Runestone, but there is no way to complete them.

This sets @optional to True on them, which will cause Runestone to mark them as optional and not count them against students: https://github.com/RunestoneInteractive/rs/blob/10bfd6eb5a3cbb6d6b5e1e5880f004f8691bcf29/components/rsptx/build_tools/core.py#L570

@bnmnetp Should sign off on this before it is merged.


bnmnetp (approved):

rbeezer:
Thanks, @ascholerChemeketa, for the catch and the fix.  And @bnmnetp for the ACK.  Merged, and will update website samples very soon.

PR #2080, opened 2023-09-28
WeBWorK: use 'latest' js folder

Alex-Jordan:
This makes HTML look for `pretext-webwork.js` in the new "latest" folder instead of a versioned folder. Less upkeep for Brad in the assignments page.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -10256,7 +10256,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 <script src="{$webwork-domain}/webwork2_files/js/vendor/iframe-resizer/js/iframeResizer.min.js"></script>
             </xsl:when>
             <xsl:when test="$webwork-reps-version = 2">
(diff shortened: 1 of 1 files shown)


rbeezer:
I believe that @bnmnetp wants to know **exactly** which files have been employed as part of his debuging process.  If `latest` changes over time, then that is not possible.  And the behavior of output will change over time (even without any rebuilding) perhaps for better, perhaps not.  See the discussion I just opened on `pretext-dev` for a proposal.

bnmnetp:
I think we should see where the discussion on -dev goes before we merge this.

Alex-Jordan:
I understand. Also:

> If `latest` changes over time,

That's not a different situation than what we have now. Right now, files in "0.32" or whatever also change over time.

rbeezer:
> I think we should see where the discussion on -dev goes before we merge this.

Roger.

rbeezer:
> That's not a different situation than what we have now.

And I'm suggesting on `-dev` that we discontinue that practice.

PR #2081, opened 2023-09-29
Script: generate svg and png using pyMuPDF

oscarlevin:
This creates a (temporary) new option for the pretext script's `latex_image_generation` function.  If `pyMuPDF` is passed as `True` to the function, then two versions of svg and png images are created for latex-image assets.  The one that will be included in the project will now be generated with the pyMuPDF python library (`pip install pymupdf --upgrade` to get this).  To compare, a `classic-*` version of these files will also be created.

If the new parameter is not set, it defaults to `False` and just he classic method of creating these images is used.

Once this is merged, I will update the CLI to allow users to test this functionality by passing a parameter to `pretext generate`


rbeezer:
Thanks!  (I think the Guide stuff is here, too.  Don't do anything, I can deal with it.  In the morning.)

oscarlevin:
Ah, yeah, I started on that branch by mistake.  grrr.  Thanks for taking care of it.

rbeezer:
There might be some lookups in the "executables" configuration file that should go away, I think I saw them being reported in the `-vv` output.  More on `-dev`

rbeezer:
Merged.  After removing stray whitespace.  ;-)  Let's use a new PR for any additions.

PR #2083, opened 2023-10-02
Fixschematypo2

mdoob:
Change of small typo in "line-numbers" attribute within schema/pretext.xml.

### DIFF
--- a/schema/pretext.rng
+++ b/schema/pretext.rng
@@ -1312,7 +1312,7 @@
         <attribute name="language"/>
       </optional>
       <optional>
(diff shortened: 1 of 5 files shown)


rbeezer:
Dear Michael -   This should be just a **single** commit.  And do you see in the list of three above that the first commit is "OBSOLETE" and was not authored by you?  You need to start with a new branch off of `master`.  That certainly means opening a new PR.

I appreciate contributing an important fix in an automated fashion.  But this is a good chance to figure out to do this in an efficient way.  Thanks.

PR #2084, opened 2023-10-03
Fix typo in schema/pretext.xml

mdoob:
Sorry for the mess. I blindly followed the GFA cheat sheet. My bad choice not helped by my ignorance on
how to fix things.
I don't know how often the pretext.rnc file gets regenerated so I fixed that file by hand.
Also, there is a PR on pretext/projects. 

Apologies again; I know you're really busy.

### DIFF
--- a/schema/pretext.rnc
+++ b/schema/pretext.rnc
@@ -498,7 +498,7 @@
                     attribute width {text}?,
                     attribute margins {text}?,
                     attribute language {text}?,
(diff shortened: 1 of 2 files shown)


rbeezer:
Not sure why this was closed.  But the fix is in.  Thanks for catching that.

There are several derived products for the schema, which I added on (in addition to updating teh schema browser, which will get updated online soon).

PR #2085, opened 2023-10-04
WeBWorK: store localizations for 'Hint' and 'Solution'

Alex-Jordan:
The change here puts localized versions of 'Hint' and 'Solution' into `data-` attributes so that later, I can change `pretext-webwork.js` to use those words instead of trying to reverse engineer them from the HTML, which was fragile and recently broke.


rbeezer:
Merged as-is.  Thanks for the discussion at Tuesday Drop-In.

PR #2086, opened 2023-10-06
Include eg/intro ahead of each eg/exercise in manifest

Alex-Jordan:
This is just to sit here until confirmed by @bnmnetp that this does not run contrary to Runestone philosophy.

When an interactive exercise is within an exercisegroup, this adds the exercisegroup's introduction to the htmlsrc version of the exercise in the manifest. For example, suppose a book has this exercisegroup:

```
Convert the given Celsius temperature to Fahrenheit using the formula F = 9/5 C + 32.
  1. 0 degC
  2. 100 degC
```

Prior to this PR, suppose you put the first exercise into a RS Assignment. Then on the assignments page you just get:
```
   0 degC
```
(Of course you can use the in-context link to see the introduction.)

With this PR, you will see:
```
   Convert the given Celsius temperature to Fahrenheit using the formula F = 9/5 C + 32.
   0 degC
```

I started a section in the Sample Chapter (newly reborn as a book thanks to Brad and Rob) that is for testing specific things about when a WW exercise is used in a RS assignment. I'm not sure that I can actually test what the end result of this will look like in a RS Assignment. But I can see the added content in the manifest htmlsrc, and it looks like I would expect it to look, as far as raw HTML.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.xml
+++ b/examples/webwork/sample-chapter/sample-chapter.xml
@@ -2527,6 +2527,57 @@
                 </exercises>
             </section>
 
+            <section xml:id="section-runestone-testing">
+                <title>Runestone Assignment Testing</title>
+                <p>
+                    This section is specifically for testing when exercises are migrated to a
+                    Runestone Assignment page.
+                </p>
+                <exercises>
+                    <exercisegroup>
+                        <introduction>
+                            <p>
+                                This introduction should appear ahead of the exercise when it shows
+                                up in the Assignment page.
+                            </p>
+                            <tabular>
+                                <row>
+                                    <cell><m>(0,0)</m></cell>
+                                    <cell><m>(0,1)</m></cell>
+                                </row>
+                                <row>
+                                    <cell><m>(1,0)</m></cell>
+                                    <cell><m>(1,1)</m></cell>
+                                </row>
+                            </tabular>
+                            <image>
+                                <latex-image>
+                                    \begin{tikzpicture}
+                                        \draw[fill=blue!20] (0,0) -| (1,1) -| cycle;
+                                    \end{tikzpicture}
+                                </latex-image>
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
@bnmnetp Do you have any thoughts against what this would do?

I'm thinking it is a necessity for exercisegroup exercises to make sense out of context, that we tie the introduction to them. This is a blunt way of doing that. But at the same time, it's simple and does not ask Runestone to do things like go and look up an introduction.

bnmnetp:
This seems like a smart idea, at least for now.... I think this is an issue we want to address in a more structured way long term, but I'd be in favor of merging this.

rbeezer:
This change is bigger than just WW, so I think the new tests should go into Chapter 5 of the **Sample Book**, which is full of Runestone goodies.  I'll make that swap as part of merging this.

rbeezer:
OK, more to this than Runestone v. WeBWorK.

*  Code is very good.  I added comments.  Also renamed the template, since there are too many "intoduction" about and there is teh risk of confusion.
*  I left the Sample Chapter alone, as-is.  No testing.  @bnmnetp  should be alert to comments next.
*  Copied sample chapter bits over to sample book, at the end of Chapter 5.  Expanded, documented, on new commit with my name.  HTML build goes bad since there is no "image-41.svg" for the introduction.  Runestone build is bad since there is no "image-17.svg".  (I think I understand the mis-match, but that is orthogonal to this PR.)  Anyway, used the new device of a `@label` on the **latex-image**.  That solves the problem (once I generate the image and include it in the sample book commit).
*  I suspect the image stuff will be OK on Runestone, since images will be generated anew, just for a RS target.

Updating website right now.

Alex-Jordan:
Maybe both though? Because the sample book is not "bigger" than WW examples, in the sense it has no WW examples.

PR #2089, opened 2023-10-13
Improve switching between working directories

oscarlevin:
This implements a `with working_directory(path)` context that switches to the directory `path`, and then switches back to the previous directory when the tasks in the context are complete (or even if they fail).  This should provide consistent behavior in every situation where there used to be, 
```
owd = os.getcwd()
os.chdir(tmp_dir)
...
os.chdir(owd)
```
The hope is that this will prevent mysterious errors that sometimes pop up for windows and perhaps the Critical 2 error.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -89,6 +89,9 @@ import zipfile
 # regular expression tools
 import re
 
+# contextmanager tools
+import contextlib
+
 # * For non-standard packages (such as those installed via PIP) try to keep
 #   dependencies to a minimum by *not* importing at the module-level
 #   (with justified exceptions)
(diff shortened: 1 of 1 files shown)


rbeezer:
OK, for the record, I've put this through the `pretext/pretext` script for each option that appears to be affected, and everything seems to function just fine.

The diff is really bad - not just gross indentation, but also coincindental matches due to the shifts.  And `--diff-algorithm=patience` does not seem to help.  _C'est la vie._

Hopefully the Windows "No 2" error goes away, or the messages from the context manager help to pinpoint the problem.  Thanks, @oscarlevin , for putting this together.

PR #2091, opened 2023-10-20
Add `requirements.txt` for easy pip install of external libraries

oscarlevin:
Two commits, the second add documentation.  Now users can run `pip install` from the `pretext` directory to get all external libraries.


rbeezer:
Thanks!  I wouldn't have thought to have this around.

One unclosed `cline` that I was very inefficient about locating.  :-(

Online update to The Guide on the next rebuild.

PR #2095, opened 2023-10-24
Change footnotes to details/summary

ascholerChemeketa:
Implementation for footnotes as detail summary.

Goes with:
https://github.com/PreTeXtBook/CSS_core/pull/81

See there for pics

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -1933,7 +1933,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <xsl:template match="commentary" mode="xref-as-knowl">
     <xsl:value-of select="$b-commentary" />
 </xsl:template>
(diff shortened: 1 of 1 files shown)


rbeezer:
Lots of complicated stuff goes into making footnotes, so I restored some "WTH" functionality.

This makes one consequential change (which sort of contradicts our discussion a few days ago).

`span.ptx-footnote__contents` is `div.ptx-footnote__contents`

This is more backward-compatible with what we have notw (such as when there is a cross-reference to a knowl, which is supported, even if a bit odd).

Maybe that means adjustments to the CSS on https://github.com/PreTeXtBook/CSS_core/pull/81 has to change with an upate via a new commit.

It is easy enough to make the `div` a `span` but will be more of a change in places not directly involved in this PR (side-effects).

Beta with these changes, and items to test with https://github.com/PreTeXtBook/CSS_core/pull/80:

https://pretextbook.org/beta/footnotes-20231024

I might fine-tune teh HTML PR a bit more, but this should be good for David to test, modulo div/span.

@davidfarmer - you tell me what you want for a new vwersion number (0.83?) and I'll make that change as I button this up.

rbeezer:
Two small comments:
* My take on relevant CSS is that `span -> div` does not matter, other than maybe removing a `display: block`.
* I have a branch ready to layer on this which will obliterate the (temporary) manufacture of cross-reference knowls for (original) footnote content.

rbeezer:
All done.  I edited this pretty heavily to bring back content creation for cross-reference knowls, but silently melded into the one commit here.  Sorry this one has been sort of slow-motion.  But very glad to get these fake knowls out.  Aah, I have another commit to get rid of that dead code!

PR #2096, opened 2023-11-03
Change cursor to pointer when over footnote numbers

ascholerChemeketa:


rbeezer:
Posted just as I went AWOL.  Thanks, @ascholerChemeketa, for the fix and your patience.

@davidfarmer: not out of the woods yet.  Change is made to CSS in `pretext` repository, as it should be.  But I can't test since we still point to `pretextbook.org`.  Fine now, this is a straightforward fix.  But also not available to all publishers until you copy absolute latest `css/pretext.css` to `pretxtbook.org`.

I will make having HTML output use local copies a higher priority since this will make testing harder and will be a PITA for you.  Can you let me know when I can verify the fix is correct?  Thanks.

PR #2099, opened 2023-11-10
WeBWorK: check select option exists before setting as selected

Alex-Jordan:
This should fix something mentioned today on -dev.

@davidfarmer Do you know what is the procedure now? Should I merge this? Is it better if I ask you to merge it? And then should I go somewhere on the utmost server and pull?

### DIFF
--- a/js/pretext-webwork/2.17/pretext-webwork.js
+++ b/js/pretext-webwork/2.17/pretext-webwork.js
@@ -182,7 +182,7 @@ function handleWW(ww_id, action) {
 					if (/^\\text\{.*\}$/.test(this_answer)) {this_answer = this_answer.match(/^\\text\{(.*)\}$/)[1]};
 					let quote_escaped_answer = this_answer.replace(/"/g, '\\"');
 					const option = body_div.querySelector(`select[id="${answer}"] option[value="${quote_escaped_answer}"]`);
-					option.setAttribute('selected', 'selected');
+					if (option) {option.setAttribute('selected', 'selected')};
 				}
 			}
 		}


rbeezer:
I will likely merge/push in a minute.  How you get it to the right address is not so critical (and out of my hands at any time, right now).  Thanks.

rbeezer:
Its in now.  Thanks.

Alex-Jordan:
I need some guidance about how to put this into production use. I think from @davidfarmer.
* This change has been merged into the `pretext` repo. 
* On utmost, there is still the `.../js/0.33/` folder which has copy of the git repo for JS_core, now retired.

I think I should make the same edit here, manually inside the 0.33 folder. Does that sound right?

Is that all I need to do? Do new builds of HTML rely on locally bundled JS or something and therefore there is nothing more to do?

rbeezer:
Builds do not yet use/assume locally available CSS or JS.  So you are still responsible for what is being served 

That will change but not without clear announcements and instructions.  If a backward-incompatible-something needs a version number change we can still coordinate that, but I'm not available for that right now.  I don't think this needs that.

I think we will eventually get out of the version number business too.

Alex-Jordan:
OK, as in the past, I just manually edited `pretext-webwork.js` in the 0.33, 0.32, 0.31, and 0.3 folders (to match this edit).

Unlike in the past, I am not thinking about staging and committing that change in the JS_core repo housed by each of those folders.

But in practice I rarely followed up with doing those things anyway in the past, so it's more like the past after all :) For now.

@Tanaquil18, do you want to check that the issue has cleared up? I forget where it was mentioned, but this had to do with a problem that had a dropdown select question.

Tanaquil18:
Just checked, and it is working now!  Thanks!

PR #2101, opened 2023-12-22
ensure html build directory has intended permissions

StevenClontz:
Addresses https://github.com/PreTeXtBook/pretext-cli/issues/652


StevenClontz:
See discussion: https://github.com/PreTeXtBook/pretext-cli/pull/658#issuecomment-1885272034 @bnmnetp @oscarlevin

rbeezer:
This is all done, starting from 1acfd1ecbfc88bcd6538465a005070a5678101fb

`distutils` cam back, then went away.

HTML output identical to what we were doing with `distutils`, including permissions (at least on my system).

PR #2105, opened 2024-01-04
Add localization support for Chinese

zhugecunfu:


rbeezer:
Thanks, @zhugecunfu!  This is a great addition.  (PDF/LaTeX will need some font work to support this, but HTML will be just fine.)

Sorry for the delay, I am just coming back from a long unplanned absence from development.

PR #2109, opened 2024-01-21
Knowl rewrite

ascholerChemeketa:
Rewrite of knowl.js to remove JQuery dependency
Incorporates knowl accessibility improvements discussed in this closed PR: https://github.com/PreTeXtBook/pretext/pull/2107

Opened as draft. Others won't easily be able to test this until pretext points at `_static` for js/css.

Tested and working:
- Sage cells
- Basic tricky xref locations: inside tables or mathjax

Needed testing:
- Interactions with Webwork
- Other tricky xref locations

### DIFF
--- a/css/knowls_default.css
+++ b/css/knowls_default.css
@@ -82,6 +82,7 @@ article > [data-knowl]::after, article > [data-knowl]:hover::after, article > [d
     border-radius: 10px;
     padding: 0;
     margin-top: 10px;
+    scroll-margin-top: 60px; /* height of header */
 }
 
 .knowl-output.original {
@@ -109,7 +110,17 @@ article > [data-knowl]::after, article > [data-knowl]:hover::after, article > [d
     background: inherit;
 }
 
+.knowl-output--hide {
+    display: none;
+}
 
+.knowl-output__error .para:first-child {
+    margin-top: 0;
+}
+
+.knowl-output__error a {
+    text-decoration: underline;
+}
 
 .knowl-output .knowl-output {
     border-width: 6px;
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -9032,6 +9032,37 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <p>Foo <dataurl source="data/runningstatisticstemplate.ots" visual="http://example.com">Runners Template</dataurl> Bar</p>
             <p>Foo <dataurl href="http://go.microsoft.com/fwlink/?LinkID=521962">Sample Excel Spreadsheet</dataurl> Bar</p>
             <p>Foo <dataurl source="data/runningstatisticstemplate.ots">Runners Template</dataurl> Bar</p>
+            <p>Testing of output positioning for xref's that are inside containers:</p>
+            <table xml:id="self-referential-tabular-xref">
+                <title>Self-referential Xref In a table</title>
+                <tabular>
+                    <row bottom="medium">
+                        <cell>A</cell>
+                        <cell>B</cell>
+                        <cell>C</cell>
+                        <cell>D</cell>
+                    </row>
+                    <row>
+                        <cell><xref ref="self-referential-tabular-xref" text="type-global" /></cell>
+                        <cell>B</cell>
+                        <cell>C</cell>
+                        <cell>D</cell>
+                    </row>
+                </tabular>
+            </table>
+            
+            <figure>
+                <caption>Xref Inside MathJAX</caption>
+                <p>
+                    <md>
+                        <mrow>{\mathbb Z}_2 \times {\mathbb Z}_2</mrow>
+                        <mrow><xref ref="theorem-FTC" text="type-global"/></mrow>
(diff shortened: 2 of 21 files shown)


rbeezer:
Thanks, @ascholerChemeketa!

> Others won't easily be able to test this until pretext points at _static for js/css.

Right.  That is about half-done, and I hope to gbet it out on Monday.

rbeezer:
Thanks very much @ascholerChemeketa.  I've had a peek.

* New rules: I've asked @davidfarmer to review.  I can likely test and approve small stuff going forward, but David should look at overhauls.
* Authors can provide alternative text for clickable of an `xref`.  So they are not always as simple as `Lemma 1.5.451`.  More specifically, they can have HTML elements.  Thus the `copy-of` instruction you are modeling off of.  **However** now `$content` goes in an attribute, whose value is a string, not HTML.  What to do?  Hard: pass around an alternate text version of `$content`.  Easy and less useful: just say "Reveal a knowl" (or similar).  I'm inclined to merge without this and work it out separately.

Aside: this is an example of where allowing extra flexibility then makes it harder for us to serve some readers.

davidfarmer:
There is a limit to what I can do just by looking at the code.

Can we have a live example for testing?

rbeezer:
https://pretextbook.org/beta/knowl-2024-01-22

Is this still necessary?  I mean the "live" part.  I built this locally and it tests out just fine.

rbeezer:
Built the WW sample chapter and it all seems functional.  Problems inside knowls, knowls inside of problems.  So I'd say if there are any gotchas, they are not pervasive.

ascholerChemeketa:
Yes, testing this one there will be important.

There were similar limits to what I could do where I was just forced to
read the code and had no test examples (e.g. WebWork). I fully anticipate
some issues to fix.

Veering into meta-discussion about process...

Live previews should not be needed anymore for anyone who knows how to
build without the CLI - just pull the PR and do a build of your favorite
book (https://stackoverflow.com/a/30584951).

Building/updating them doesn't sound like a great use of Rob's time, but
they would make "drive by" testing easier, especially for non-devs. They
could be automated through GitHub. We could also add minimal hooks to
github that called a piece of code on pretextbook.org to automatically
build the sample-book for PRs into a predictable folder (
pretextbook.org/development/PR/NUMBER/).

bnmnetp:
@ascholerChemeketa - There is a WeBWorK sample book in the pretext core repo now.

I like the idea of saving rob the hassle by automating the build/deploy of preview/beta books.  If Rob likes the idea, happy to help set up the github actions.

ascholerChemeketa:
Started a thread on `pretext-dev`. Make sure to tune in @bnmnetp

bnmnetp:
😄  👍🏻

ascholerChemeketa:
@rbeezer I think this should address concerns about tags in aria-label. Now streaming the text from the existing title, which should be tag free and have some bonus info. Adds localization of "Reveal" or "Close".

ascholerChemeketa:
Runestone in knowls almost ready... dropped this back to a draft while I finish that

ascholerChemeketa:
Runestone is working now. If runestone component is in a knowl that is being loaded, the existing page is searched for a matching HTML id. If none is present, the component is rendered in knowl. If that id already exists on the page, this message is rendered in the knowl:

_The interactive that belongs here is already on the page and cannot appear multiple times. [Scroll to interactive.](URL)_

Merging should wait on webwork-pretext shifting from `id-ref` born hidden knowls to summary/details based ones.

rbeezer:
Excellent!  Thanks for all the work (and patience) on this one.  Rebuilt the commits, but did not change any code.  One tab to remove.  Merged, pushed, and website samples rebuilt.

*  Would you like to make a "heads up" on -dev?  Just so folks know to consider nightlies carefully if anything seems off?  I can do it if you'd prefer not to, just let me know.
*  I could experiment, but I wonder if we can remove the JQuery JS we are distributing, and then we can get rid of the `js_lib` directory with a couple of file moves into plain 'ol `js`?  (Perhaps after #2108.)
*  Testing Runestone components in knowls.  Since the components can go wide, do we need/want the border of the knowl to go wide now as well?  If so, start a new thread/issue/PR at your leisure for any associated discussion.

rbeezer:
Forgot to say - I did some tests a while ago on the content going to attributes and it seems to be OK.  The `-common` template tries to only deal with **text** so perhaps that is why.  Though I can't say I ever found a concrete definitive explanation.

ascholerChemeketa:
@rbeezer - replying to your JQuery question(s):

There are other pretext js files that use jQuery. So it can't be fully purged. Yet. 
Eventually, it should be either purged or automatically grabbed by a build process - hosting a copy of the `jquery.min.js` file isn't awesome. But for now leaving a local copy of `jquery.min.js` in place means that any scripts that depend on jquery are usable offline. I'd say leave it alone until the JS/CSS bundling gets tackled.

I am pretty sure you can remove `jquery.min.js` and `jquery.sticky.js`

Agree there is no reason to have `js_lib` and `js`. Makes sense to merge the two.

PR #2111, opened 2024-01-25
WeBWorK: make separate js file for 2.18

Alex-Jordan:
This removes two symlinks and adds a complete 2.18 pretext-webwork.js file. Previously it was just a symlink to the 2.17 one.

The other symlink removed here is only used by old builds that are still getting their js from utmost.

Lastly, this makes an edit to the 2.17 version (and 2.18 as well, still just a copy of 2.17) so that if MathQuill is being used and someone clicks to reveal answers, the MQ input field gets hidden.


Alex-Jordan:
BTW, these seem to still be symlinks in pretext itself (that is, until this PR deletes them). It must be the CLI that was copying these and turning them into empty files/losing their link property.

rbeezer:
@StevenClontz and @oscarlevin might be interested in the fact that symlinks were not copied.  Probably does not make a real difference right now.  There are arguments for some of the Python file copying routines about how to deal with symlinks.

rbeezer:
Thanks very much.  I had a heck of a time isolating the MathQuill stuff from the file structure stuff...

Issue #2113, opened 2024-01-26
Make dark mode work well


davidfarmer:
Add known issues with using the browser's built-in dark mode.

Please include a link to the HTML and which browser had an issue.

oscarlevin:
Using Chrome's dark mode doesn't do anything for my book at all.  There is a "Dark" extension that forces sites to use a dark mode, and this does invert the colors.  However, it doesn't invert the colors of SVG files, so they look like black dots on very dark gray.  https://discrete.openmathbooks.org/dmoi3/sec_gt-intro.html

ascholerChemeketa:
This PR is the one that turns on the existing dark css rules when dark mode is set by the OS/browser.

https://github.com/PreTeXtBook/pretext/pull/2114

rbeezer:
Sample article, with #2114 applied, is at https://pretextbook.org/beta/html-dark-2024-01-26

davidfarmer:
In the sample article:
https://pretextbook.org/beta/html-dark-2024-01-26/section-interesting-corollary.html

Checkpoint 4.4 knowl has white background (needs to be dark, not necessarily the
same as the page background).

Same for Checkpoint 4.7 (also need to address the H/A/S in that knowl)

Solution in Exploration 4.2 knowl looks fine (but maybe use a different dark background?)

Aside near Checkpoint 4.10 needs dark background (but definitely different than the
main dark background).

Sage cell in Technology 4.12 has stark white background.  Make less white, or should
Sage cells be light text on dark background?

4.2.6 Solutions for This Subsection:
    white border on left should be less light

rbeezer:
Lots to chew on in the "Graphics" section.  LaTeX-images (tikz) are likely based on an old conversion from PDF to SVG, it is possible they might behave differently if regenerated.

https://pretextbook.org/beta/html-dark-2024-01-26/section-graphics.html

Asymptote images all seem to have a solid-white background.  But transparent images elsewhere don't seem to work so well, things like axes are being washed out.  Not clear to me what we want here, and if we will need to advise authors on constructing their images.

ascholerChemeketa:
Update - 

Oscar and I had a recent discussion about simplifying color support in general as part of that move. Fixing dark mode would dovetail nicely with that.  I will start a discussion on the dev list about colors once we reach that point.

PR #2114, opened 2024-01-26
Move existing dark-mode css to work with css media query

ascholerChemeketa:
This creates duplicates of css rules targeting `.pretext[data-atmosphere="dark"]` and moves them to `@media (prefers-color-scheme: dark)`

Very hacky preview of what existing dark mode rules do.

### DIFF
--- a/css/setcolors.css
+++ b/css/setcolors.css
@@ -450,3 +450,165 @@ body.pretext {
   background: #606;
 }
 
+  
+  /* TEMPORARY -- TESTING STANDARD DARK MODE NEEDS REFACTORING */
+  /* Duplicates of .pretext[data-atmosphere="dark"] .pretext[data-atmosphere*="dark"] rules from above */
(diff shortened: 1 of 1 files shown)


rbeezer:
Wild.  Beta up at https://pretextbook.org/beta/html-dark-2024-01-26

rbeezer:
Is this ready to merge?  Would it be ready with more work?  Should the copy be a move so we don't need to keep two versions in sync?  How does this relate to #2113?  So many questions.

ascholerChemeketa:
This is ready to close without merging. This was a way to share what the existing dark mode code did and not a serious attempt to implement a long-term improvement.

But keep issue https://github.com/PreTeXtBook/pretext/issues/2113 open.

A real solution should wait until CSS bundling is in place. Oscar and I had a recent discussion about simplifying color support in general as part of that move. Fixing dark mode would dovetail nicely with that. Will start a discussion about colors in general once we reach that point.

PR #2118, opened 2024-01-30
Never add hash identifier if same as page name

ascholerChemeketa:


rbeezer:
Beautiful!  And I can even read the code.  ;-)  Thank-you, thank-you, thank-you!  Merged.

PR #2119, opened 2024-02-01
Strip newlines from permalink descriptions

jjrsylvestre:
I noticed that the non-rich-text version of permalinks to items whose title contains math have recently started to have extra newlines in the math part for some reason. This patch strips them.

I haven't tested this extensively, but I can't think of any reason why `@data-description` on a `div.autopermalink` element might need to contain a newline.

### DIFF
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -137,7 +137,7 @@ function permalinkDescription(elem) {
     if ((lastChr == '.') || (lastChr == ':'))  {
         retStr = retStr.slice(0,retStr.length - 1);
     }
-    return retStr;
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, Jeremy.

> for some reason

Well it is the start and end of a `span.process-math` that each becomes a newline.

I'll merge this, but we do allow multi-line titles (like for a `book`), so long-term I'd prefer to understand this better.

So pinging @davidfarmer and @ascholerChemeketa since this is mostly Javascript.  Title to Chapter 3 of the sample article is a good place to test.

rbeezer:
>  seemed like new behaviour

Yes, I think so.

I do like the looks of #2126.  Thanks.

PR #2121, opened 2024-02-02
Pretext-webwork knowl update

ascholerChemeketa:
Changes pretext-webwork to use summary/detail born hidden knowls instead of `id-ref` based knowls. This change enables removal of `id-ref` related code from knowl.js.

Am only able to test against Tested against 2.17 using:
https://pretextbook.org/examples/webwork/sample-chapter/html/sample-ww-chapter-4-2.html#using-hints
But the code replaced appears to be identical across 2.16/2.17/2.18

Unblocks https://github.com/PreTeXtBook/pretext/pull/2109

@Alex-Jordan

### DIFF
--- a/js/pretext-webwork/2.16/pretext-webwork.js
+++ b/js/pretext-webwork/2.16/pretext-webwork.js
@@ -497,7 +497,7 @@ function handleWW(ww_id, action) {
                 ww_container.prepend(iframe);
             }
 
-            iFrameResize({ checkOrigin: false, scrolling: 'omit', heightCalculationMethod: 'lowestElement' }, iframe);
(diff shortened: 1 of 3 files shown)


Alex-Jordan:
I can double check on things next week, but I have faith you have this right.

Can you say a word about the iframeResizer change?

ascholerChemeketa:
The `lowestElement` setting appears to reserve future space for the summary. That leaves a long gap after the titles:
image

`min` only accounts for the spaces used by the titles themselves:
image

When you expand them, the available space is increased appropriately:
image

rbeezer:
Very good!  (Removed three stray tabs.) Merged.  Thanks.

PR #2123, opened 2024-02-09
Guide: edits to CLI project.ptx description

Alex-Jordan:
Yesterday I updated a very old project, our calculus lab manual. Remember things like `autoname="no"`? Captions on a `sidebyside`? I hit several blasts from the past.

Anyway I had some readability trouble with the guide's section on the `project.ptx` file. Totally my own issues with failing to read everything all together and instead only reading the snippets I thought I could get away with. But I thought some editing could help other people who read like I sometimes do. Roughly the changes here are all:

* correct some typos
* state some critical detail at the opening of a section, not only at the end
* attribute values inside a `<c>` instead of in plain text double quotes
* some consistency applied to the attribute descriptions, and move "required if" conditions close to the start of the description

### DIFF
--- a/doc/guide/author/processing.xml
+++ b/doc/guide/author/processing.xml
@@ -373,7 +373,7 @@
             </p>
 
             <p>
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks for the clean-up!

I think `@autoname` is still documented in the sample article!  :-(

PR #2124, opened 2024-02-10
WeBWorK: fix fat commas

Alex-Jordan:
This does two things related to perl "fat commas". One is just to escape the `>` signs in the XML, for consistency sake.

The change here that affects the output is with key value pairs. Long story short, I forgot that with perl key-value pairs, if you separate with a comma, the key should be surrounded in quotes. Without quotes, it creates perl warnings. So back to fat commas at line 3003.


rbeezer:
Thanks.  Merged as-is.

Took me a while to realize I would only see changes in the intermediate file from the first pass before being sent to the server.  So just settled for seeing a few base64 strings changing in HTML.

What's a "fat comma"?  ;-)  Never mind, I consulted Wikipedia.

PR #2125, opened 2024-02-10
WeBWorK: remove MathView css

Alex-Jordan:
As a comment suggests (a comment now being deleted) it is unclear why this CSS file was loaded this way. Maybe there was a temporary, experimental period where we loaded live WW problems by inserting them into the DOM without using an iframe. Then I could see this being used. But the old 2.16- style of live rendering, and also the settled 2.17+ style of a static version that transitions to the live problem both use iframes, and the proper CSS for that tool should be handled correctly now by pretext-webwork.js.

So these lines can go. Note that on a 2.18 WW server, the file here exists but is in a different location. So this is actually  causing console errors (404) at sites using a 2.18 server.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -10247,14 +10247,8 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- ########################## -->
 
 <!-- WeBWorK HTML CSS header -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks for keeping this neat and tidy.  I added a follow-on commit to rename the relevant template to make its purpose clearer.

PR #2126, opened 2024-02-11
Javascript: only strip newlines from math content in permalink descri…

jjrsylvestre:
Not looking for this to be merged right away, just looking for discussion.

Here's a new version of #2119 that only strips newlines from `innerText` of `.process-math`.

Is this the right approach? Feedback from @davidfarmer and @ascholerChemeketa welcome.


jjrsylvestre:
No one seems to have any objections to this, can we merge it?

rbeezer:
Merged!  Thanks.

PR #2128, opened 2024-02-16
Add animations for born hidden and xref knowls

ascholerChemeketa:
This should be ready to roll.

Adds animations back that were lost in the transition to details/summary and then in the conversion of knowls.js away from jquery.

@davidfarmer @Alex-Jordan

### DIFF
--- a/js_lib/knowl.js
+++ b/js_lib/knowl.js
@@ -1,18 +1,113 @@
-// Code for link based knowls
+// Code controlling behavior of xref knowls and born hidden knowls
 
 // Assumes this file is loaded as part of initial page
 window.addEventListener("load", (event) => {
-  addLinkKnowls(document);
+  addKnowls(document);
 });
 
-function addLinkKnowls(target) {
+function addKnowls(target) {
   const xrefs = target.querySelectorAll("[data-knowl]");
   for (const xref of xrefs) {
     LinkKnowl.initializeXrefKnowl(xref);
   }
+
+  const bornHiddens = target.querySelectorAll(".born-hidden-knowl");
+  for (const bhk of bornHiddens) {
+    const summary = bhk.querySelector(":scope > summary");
+    const contents = bhk.querySelector(":scope > div");
+    new SlideRevealer(summary, contents, bhk);
+  }
+
 }
 
-// A LinkKnowl managages a single link based knowl
+// Used to animate both types of knowls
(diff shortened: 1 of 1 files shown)


rbeezer:
Took this for a quick drive around the block.

"xref" knowls, and `hint|answer|solution` always-born-hidden knowls are working great (to my eye).

But a larger block, electively knowled, opens without any animation and then cannot be closed.  In the sample article Checkpoints 4.4, 4.7, 4.14.  Firefox and Chrome (latter might not be super-recent).  A `hint|answer|solution` **inside** these Checkpoints still seems fine.

In the console I see:

```
Uncaught TypeError: this.contentElement is null
    onClick file:///tmp/s/_static/pretext/js/lib/knowl.js:58
    SlideRevealer file:///tmp/s/_static/pretext/js/lib/knowl.js:44
    SlideRevealer file:///tmp/s/_static/pretext/js/lib/knowl.js:44
    addKnowls file:///tmp/s/_static/pretext/js/lib/knowl.js:18
    <anonymous> file:///tmp/s/_static/pretext/js/lib/knowl.js:5
    EventListener.handleEvent* file:///tmp/s/_static/pretext/js/lib/knowl.js:4
knowl.js:58:7
```

ascholerChemeketa:
Fixed. Code assumed that all the BHKs were placing contents in div's, but the exercise element has its content in an article.

Now the assumption is that the content is whatever one element appears right after the `<summary>` inside the `<details>`. I *think* that handles all existing BHKs. Requiring there to be one element that contains all of the content for the knowl seems reasonable - pretty much necessary for styling as well.

rbeezer:
Very good!  Combined commits, merged and pushed.

Your call: heads-up to `-dev` to warn about the change?

There is a "body" template that employs a "body-element" template for making content of many-many things (eg sans title).  I think that abstraction gurantees the single-element assumption.  But to be safe I am going to comment the code a few places to that effect.

rbeezer:
Postscript.  Should an image description animate?  Not a knowl, but is a `details`.  Example:

Figure 9.17 
https://pretextbook.org/examples/sample-article/html/section-graphics.html#figure-sage-exosagevec1

ascholerChemeketa:
@rbeezer Maybe?  Seems reasonable.

How about footnotes? 
Are there other uses of details/summary?

PR #2129, opened 2024-02-18
A few validation/spelling errors fixed

oscarlevin:
I'm working on adding to the development schema and noticed a few obvious errors in the sample article and book.  

I realize some of these might have been added intentionally to test what happens if the schema is not followed.  If that is the case, I can revert those.

Also possible that some of these might be intended to be legitimate, but the schema says no.


rbeezer:
Thanks!  A schema is a great thing to have, no?  ;-)  Merged with no changes.

As much as possible, the sample article should build cleanly, so I try to have all error-provoking stuff just be temporary during actual development (and so never pushed).  There is an example document **full** of problems, but I've not been adding so much to that lately.

PR #2130, opened 2024-02-19
Schema improvements

oscarlevin:
This is the start of a bunch of schema and experimental schema improvements.  Leaving as draft for now, but would love feedback as I continue to work on this.

One major change to the non-development schema is the tweak the description/shortdescription for images.  Previously, if you had `@decorative` with either "yes" or "no", then it was a violation to have a description or shortdescription.  So I remove that rule.  I also allow shortdescription and description to come in any order, with the possible other elements of an image.  And for now, both are optional.

### DIFF
--- a/schema/pretext.xml
+++ b/schema/pretext.xml
@@ -685,6 +685,7 @@
                     UniqueID?,
                     LabelID?,
                     PermanentID?,
+                    Component?,
                     TextParagraph
                 }
             ParagraphLined =
@@ -692,6 +693,7 @@
                     UniqueID?,
                     LabelID?,
                     PermanentID?,
+                    Component?,
                     element line {TextShort}+
                 }
             </code>
@@ -1417,6 +1419,7 @@
             Console =
                 element console {
                     PermanentID?,
+                    Component?,
                     attribute prompt {text}?,
                     attribute width {text}?,
                     attribute margins {text}?,
@@ -1431,6 +1434,7 @@
             Program =
                 element program {
                     PermanentID?,
+                    Component?,
                     attribute width {text}?,
                     attribute margins {text}?,
                     attribute language {text}?,
@@ -1456,6 +1460,7 @@
             List =
                 element ol {
                     PermanentID?,
+                    Component?,
                     attribute cols {"2"|"3"|"4"|"5"|"6"}?,
                     attribute marker {text}?,
                     element li {
@@ -1468,6 +1473,7 @@
                 } |
                 element ul {
                     PermanentID?,
+                    Component?,
                     attribute cols {"2"|"3"|"4"|"5"|"6"}?,
                     attribute marker {"disc" | "circle" | "square" | ""}?,
                     element li {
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks for making more progress on this.

*  A "decorative" image is one that does not need any description and gets an empty @alt attribute (like maybe a feuron to break between sections).  An accessibility thing.  So I think it is **meant** to be mutually exclusive with text descriptions, **in the case where `@decorative="yes"`**.  I think.  Maybe we need counsel from @Alex-Jordan.
*  I think `@axis` and `@grid` are exclusive to JessieCode.  The eternal debate: do we want a specific pattern like a  `SlateJessieCode`  or do we just allow lots of attributes on the understanding they are ineffective when irrelevant?  Drop-In?
*  Do we have an example of a `slate` holding just text?  Am I reading that right?
*  That's ambitious going after `Areas`!

oscarlevin:
> Do we have an example of a slate holding just text? Am I reading that right?

Line 7808 of the sample article has a `slate` in an `interactive` with jessiecode that is just text.

Additional commits:
- Add in finer control for `@decorative`.  So now either `@decorative="yes"` or the value is "no" or the attribute is missing, and there is an optional  `shortdescription`/`description`.  Should one of these be required?
- Separated out a pattern for jessiecode, so when that attribute is present, we allow different attributes and bare text inside the slate.
- Added label everywhere that UniqueID can go, plus in the interior elements of image.

Regarding this last one, should `xml:id` still be allowed for images?  I realize that there isn't ever a need to xref these, but for the purpose of using a  `xpointer` in an `xi:include`'d file?

rbeezer:
> Line 7808 of the sample article has a slate in an interactive with jessiecode that is just text.

Right, forgot that we did this.

> Should one of these be required?

I don't want to enforce accessibility practices through the schema.  (But want to encourage through education!)  Would folks just add empty `description` to feed the validator?

> for the purpose of using a xpointer in an xi:include'd file

I'm not real eager to have the schema in the service of `xpointer` which I think is something of a failed standard/project/implementation.  If an `xml:id` always suggested a possible cross-reference target, then I think we would have enough around to be modular enough.  Could be mistaken, but I'd rather start there and see how it shakes out.  Or we can talk it through in Drop-In.

rbeezer:
Is this still a draft?  Checking that I am not holding things up.

oscarlevin:
I want to make a few more improvements tomorrow, then it should be ready for review.

oscarlevin:
Okay, reorganized my commits into a stable and dev change.  I think this is ready to be merged.  Still more work to do before the sample-article validates without error, but we are closer.

oscarlevin:
One more commit added on that adds component in both stable and experimental areas.

rbeezer:
Finally!  Sorry for the long delay.  This morning's crisis was finding my script for testing the schema...  And right after you did a CLI update - sorry.

Looking real good, this is a huge improvement!  Found at least one gotcha in sample article source that I will fix shortly (may look for others).

Trimmed some whitespace, and blew up one commit into three, wrote commit messages.  But no changes in substance at all.

Website updating right now.

PR #2131, opened 2024-02-24
Pretext: escape backslashes in regex

Alex-Jordan:
With Python 3.12.2, running `pretext` produces these warnings:
```
/Users/alex.jordan/pretext/pretext/pretext.py:2524: SyntaxWarning: invalid escape sequence '\*'
  elif len(re.findall('^[ ]{2},\*apt} #([a-j]+?) ', line)) > 0:
/Users/alex.jordan/pretext/pretext/pretext.py:2526: SyntaxWarning: invalid escape sequence '\*'
  chapter_number = brf_to_num(re.findall("^[ ]{2},\*apt} #([a-j]+?) ", line)[0])
```

The intent here is to escape the asterisk in the regex context. But first the backslash must be escaped or python will produce that warning. I'm guessing might `\*` become `*`, and this regex was not working as intended. Or maybe `\*` becomes `\*`. But this change should have things working as intended and make those warnings go away.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -2521,9 +2521,9 @@ def _split_brf(filename):
             if len(re.findall("^[ ]{2}#([a-j]+?) ", line)) > 0:
                 numbered_chapter = True
                 chapter_number = brf_to_num(re.findall("^[ ]{2}#([a-j]+?) ", line)[0])
-            elif len(re.findall("^[ ]{2},\*apt} #([a-j]+?) ", line)) > 0:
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I think it is just because I upgraded python and it's reporting this where previous python versions sweep it under the rug. Not building braille, building webwork. But python is trying to parse everything in `pretext.py`, catches this, and gives those warnings.

rbeezer:
> it's reporting this where previous python versions sweep it under the rug

As suspected.  Thanks for the confirmation.  Merged.

Pinging @alexeikolesnikov since he is interested in this code.

Issue #2116, opened 2024-01-29
Knowl styling consistency


ascholerChemeketa:
Report from Jeremy Sylvestre email...

Small styling issue:

Go to
https://sites.ualberta.ca/~jsylvest/books/DLA1/section-inverses-theory.html

At the end of the first paragraph there's an xref to Proposition 4.5.1. Click to open the knowl. Then scroll down to the bottom of the knowl. This <proposition> has three proofs (for different substatements withing the proposition).

The styling for the three links to open the proof knowls is inconsistent. (But no such issue if you click "in-context" and go to the actual proposition.)

rbeezer:
Closed via #2134

PR #2134, opened 2024-02-27
Knowl format cleanup

ascholerChemeketa:
Still a draft. Generating PR so @rbeezer can check out the xsl.

### DIFF
--- a/css-obsolete/pretext_add_on.css
+++ b/css-obsolete/pretext_add_on.css
@@ -1460,9 +1460,6 @@ from Jiří Lebl */
 /* padding, not margin, to get colored background (and not be absorbed) */
     padding-top: 0.25em;
 }
-.ptx-content .knowl-footer {
-    display: none;
-}
 
 .ptx-content .exercisegroup > .conclusion {
     margin-left: 1.5em;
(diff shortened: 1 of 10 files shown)


rbeezer:
I think the XSL looks real good at first blush.

ascholerChemeketa:
@davidfarmer @rbeezer 

OK, I think this is ready for real review.
- Closes #2116 by making sure nested knowls are styled as knowls.
- Restores some of the other styling lost in the conversion of knowls to details/summary.
- (Re)Adds animation to both forms of knowls.

Comments:
Born Hidden and XRef knowls both get underlined and have same color. XRef links get the same color as external links and no underline to clearly distinguish that they are links and not knowls.

This does not restore the color stripes that prior to details/summary were on the left side of some born hidden's. IMO those decorations should wait for a deeper dive on color themes & dark mode - hard coding in specific colors complicates those efforts

Born Hidden's have the triangle marker as that is what they looked like even before the details switch. XRef knowls do not. It would not be hard to add triangle markers to them if that was desired. It does nicely communicate that they open something up. But, it can look a little weird when the thing opening up is further down the page.

rbeezer:
Excellent!  Repackaged commits, but otherwise as-is.  Thanks for working through this one!

PR #2137, opened 2024-03-06
Scripts in interactives

ascholerChemeketa:
An initial attempt at implementation of `<script>` in `<interactive>`.

Opening so @rbeezer has a concrete version of the proposed format to evaluate.


rbeezer:
Looks good.  I don't think the template needs to be modal.  Try using just a `match` as

* `match="interactive/script"`
* `match="interactive[@platform='javascript']/script"

The first is good, the latter is better if it is legal (I forget if the test in the filter is OK).

Check that an automatic preview is OK (why not?) - that'll be better with a `@label`.  Then make sure the PDF builds.

ascholerChemeketa:
Selection logic updated.

Checked preview/qr generation - all seems fine. Adding those files to the gen directory.

PDF looks good.

rbeezer:
Really good.  Thanks for working through this one.

Website samples being rebuilt right now.

PR #2138, opened 2024-03-08
XSL - add codelens attribute

bnmnetp:
Sometimes you don't want to have the show in codelens button, even when the language is python because the tracer code does not support turtle graphics or image processing.


rbeezer:
Merged this a few days ago during Drop-In.  Which means (a) I forgot to tag it with the PR number, and (b) I forgot to close it.

At f72655a360eb26bf8ecfe4887ead42215d00483e

Thanks for revealing the hidden features!  ;-)

PR #2139, opened 2024-03-11
Add docs for codelens="no" on activecodes

ascholerChemeketa:
This adds minimal documentation for the `codelens="no"` option on activecodes that was added in #2138.


rbeezer:
Timely!  I was just going to merge another PR.  But this can seamlessly go right on top of the commit with the code.  ;-)

Thanks!

PR #2144, opened 2024-03-19
Prefix tmp directory with `pretext_`

siefkenj:
Prefixing the tmp directory with `pretext_` makes it easier to find when debugging. A similar change was done in the CLI.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -3970,7 +3970,7 @@ def get_temporary_directory():
 
     global __temps  #  cache of temporary directories
 
-    temp_dir = tempfile.mkdtemp()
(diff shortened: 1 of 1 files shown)


rbeezer:
Nice!  i didn't know about that option.  I changed the prefix slightly to `ptx-`, mostly to differentiate it from directories produced by the CLI, which could be helpful when chasing down who is doing what.

Catching up after a few days away - will get to other PR soon.

siefkenj:
Maybe `pretext_core_` could differentiate the two while still having them show up next to eachother when sorting alphabetically?

rbeezer:
Perhaps.  But "core" **is** PreTeXt, not something different.  So maybe the CLI could do `ptx-cli-`.

PR #2145, opened 2024-03-23
Add native DoenetML support

StevenClontz:
Naively seems to work with HTML and PDF builds, but feedback from @rbeezer is appreciated to get this merged in and help @mattboelkins adopt DoenetML for Active Calc.


rbeezer:
One question without any testing:  should `image` element in loading-preview really be an HTML `img` element?

If so, I can silently adjust.  No need to pile on here.

StevenClontz:
Hotfixed it myself. (I'm not shocked that I made that error, but I am shocked that the image *appeared* in my browser even though I used the wrong tag.)

rbeezer:
Looking good.  This is going to be a real good starting point.

*  Too bad Doenet isn't using real XML.  We could use namespaces then.
*  "I am shocked that the image appeared in my browser"  - too funny.
* Needs a QR code for the repository, like the preview image.
* Reason for adding to the no-op `create-iframe-page`?  Those are `interactive` that are **already** `iframe` .  It does not look to me like that is the case here.  (And perhaps due to its position, I think the template is not even being applied.)  I'm getting good "standalone" pages with no additions.
* Do we think a Doenet like your example can have two boxes talking to each other as **PreTeXt** `slate` in a **PreTeXt** `sidebyside`?

StevenClontz:
> Too bad Doenet isn't using real XML. We could use namespaces then.

There is a strict flavor of DoenetML that's XML (I think basically the first processing step is to turn the markup into XML) but I don't think there are many (any) humans who write in it.

> Needs a QR code for the repository, like the preview image.

I'll take care of that.

> Reason for adding to the no-op create-iframe-page?

Honestly that was just me copy-pasting things I didn't understand to model this after e.g. authored Geogebra. I'll clean that up.

> Do we think a Doenet like your example can have two boxes talking to each other as PreTeXt slate in a PreTeXt sidebyside?

Maybe? Can you point to an example of another interactive in the sample article I could model after?

rbeezer:
> Can you point to an example of another interactive in the sample article I could model after?

Ten slates in four `sidebyside` , all in one `sbsgroup`.  ;-)

Figure 13.4 
https://pretextbook.org/examples/sample-article/html/section-interactive-authored.html#figure-animation

If this behaves, then great.  But don't let it hold up this PR.  So holler if you give up, as I will be sitting on this.

> copy-pasting things I didn't understand

Thought so.  ;-)

StevenClontz:
I think this is ready to merge as of d38b157 

I did not end up trying to implement mutliple slates that communicate - DoenetML has its own layout features to support authoring a widget such as Figure 13.4. Maybe it can be done later as an enhancement, but I already cringe to think of the Frankenstein monsters that might be created by mixing and matching the PreTeXt and DoenetML layout engines together...

jaltekruse:
Before we merge this I should give you a better URL for the hosted built version of DoenetML. I put it on our dev server while I was working on prototyping, but we aren't monitoring that server actively, so it isn't a good thing to rely on for this purpose. I should give you a versioned URL on our production server (or maybe publish to a CDN). I'll talk to @dqnykamp and get back to you soon.

StevenClontz:
Converting to draft until DoenetML javascript host is finalized.

rbeezer:
OK, thanks @StevenClontz and @jaltekruse.  I think this is complete, so standing-by.  Feel free to ping me in other ways if it seems I'm not hearing about this one.

StevenClontz:
Cool. If @jaltekruse replies with the updated URL for the doenetml JavaScript, @rbeezer can go ahead with implementation without my intervention. But if I see it first I'll update the PR and mark it as ready.

StevenClontz:
@jaltekruse @dqnykamp Any news on this?

Another possibility: we package up and ship a few other JS libraries for use in PreTeXt HTML. We could add your DoenetML script to that suite, so it's used locally rather than being called from your server or CDN. What do we think of that? @rbeezer?

jaltekruse:
@StevenClontz I'd advise against it for DoenetML for now. The bundle is currently a bit big and would be best to only load when it is needed and not bundle it together with other things. I have done some research on this I am planning to use a service called jsDelivr, which can serve packages out of NPM as simple CDN urls. I should have a URL posted tomorrow, I need to work on wrapping up something else this afternoon.

rbeezer:
> we package up and ship a few other JS libraries for use in PreTeXt HTML

Doing something like this for Runestone is in my queue.  But I think we want to get some experience with Doenet before we go too far, and conditional loading will be a good thing.  So getting it via jsDelivr will work well now, methinks.

rbeezer:
@StevenClontz - find the global `has-*` variables and define one for Doenet.  Then make a simple template that checks that variable before including the script in the head.  There are lots of models to choose from.  This one could be defined in the HTML stylesheet, I think.

jaltekruse:
@StevenClontz Of course my afternoon task from Wednesday took most of Thursday as well... Duane is currently reviewing the changes I made to DoenetML that came up while I was testing with Runestone, mostly upgrading our MathJax version. That is the version that is on the "dev" doenet server, but the changes weren't in the main branch yet as I was still finishing up the Ruenstone integration work to finalize changes on both sides of that. Still hoping to get you a URL today, just want to do a somewhat official release of DoenetML, which we haven't done for this standalone package we need in this case.

StevenClontz:
Any update here? I have authors for TBIL.org wanting to incorporate some of the new https://www.doenet.org/library/ content in our upcoming precal book, but this is a blocker from doing so.

jaltekruse:
@StevenClontz  I did get up a version of this up on NPM and jsdelivr. It is working, but I might want to change the URL in the not too distant future to make some improvements, but I assume that is okay. Sorry about the delay, we've just been trying to figure out some stuff as the Doenet grant is ending, so there are various thing we are trying to wrap up.

```
      <script type="module" src="https://cdn.jsdelivr.net/npm/doenet-standalone-test@0.6.0/dist/doenet-standalone.min.js"> </script>
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/doenet-standalone-test@0.6.0/dist/style.css"></link>
```

StevenClontz:
As long as this resource won't be yanked later, this is great. We can always replace with an upgrade in future versions of PreTeXt.

StevenClontz:
@rbeezer I believe this is ready.

rbeezer:
Thanks @StevenClontz  and @jaltekruse  for all the work and iterations to get this in place.

Let's keep it low-key for a bit, without a huge announcement, so we change it up if we have to?  See how TBIL and Active Caculus fare?  Perhaps we can open it up before the upcoming Doenet workshop?

PR #2147, opened 2024-03-25
Ww 219

Alex-Jordan:
These are more updates to the new `pretext-webwork.js` file for version 2.19. Most of this will not be perceptible to the end use but makes things run better under the hood. One thing that is different for the user is that with this, the tall palette toolbar for entering math will not trigger the iframe to resize. Instead it is scrollable. That toolbar is very tall compared to most exercises and it looks bad when it causes the iframe to resize.

Commits broken down by topic, but if you'd like to squash into one, that's fine.


rbeezer:
Very good!  That's a lot of deletions!

PR #2148, opened 2024-03-28
fix typo

sean-fitzpatrick:
Sorry for the one-character pull request, but I had nothing else to add to it and this seemed better than just leaving "buttom" in place until I came up with something.


rbeezer:
Thanks!

PR #2149, opened 2024-03-28
Ensure Asymptote generation runs in background using local method

sean-fitzpatrick:
In some environments (in particular, Windows), Asymptote defaults to the `-V` option when generating images. The assumption is that the user will want to see their output when it is created.

This isn't ideal when you are building a textbook with 300 Asymptote diagrams. Every single image will open in your web browser, so you can forget about using it until Asymptote generation is done. I suspect this is also not ideal on systems with limited memory.

This adds the `-noV` option to the Asymptote command to ensure that every format, on every OS, is generated silently.

I've tested it on Windows against APEX and it ran as expected. However, someone should check to see if I have chosen the best spot to place the `-noV` option.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -326,7 +326,7 @@ def asymptote_conversion(
                 # 2021-12-10, Michael Doob: "-noprc" is default for the server,
                 # and newer CLI versions.  Retain for explicit use locally when
                 # perhaps an older version is being employed
-                asy_cli = asy_executable_cmd + ["-f", outform]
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks!

*  I moved the option to **follow** the output format, I like that one first when debugging.
*  Tests fine on Linux, but lets watch the list for gotchas on other platforms.

sean-fitzpatrick:
Thanks! I'll test on Windows when I get a chance. Without the -noV option it's mildly annoying to build on Windows: 300 figures means 30 minutes or so of having an image open in your web browser every few seconds!

PR #2151, opened 2024-04-05
WeBWorK: don't put all solutions and hints in the first task

Alex-Jordan:
This was a mistake in the 2.19 WW js I recently added. I thought I was being clever and efficient by doing a certain thing outside of a loop. But it turns out that caused all hints and solutions for a multi-task exercise to be placed wherever the first hint/solution was. So for example, causing the second task's solution to appear within the first task, after the first task's solution.

This fixes that.


rbeezer:
Merged unmolested.

PR #2156, opened 2024-04-20
slideshow: update image template for managed directories

### LINKED ISSUE #2155: External images in slideshows do not consider the publication file
Check the template matching `image` and condition on the existence of a `generated` directory.  Good chance there maybe other similar failures to respect aspects of the publication file.

Obsoletes: https://github.com/PreTeXtBook/pretext-cli/issues/717

oscarlevin:
Replaced the simple `image` template with a tweaked override of the html non-sidebyside image container.  Now images in reveal.js slideshows will behave just like html images, except if they are bare images they can have a `@pause`.

In particular, this will respect managed directories.  Closes #2155 .

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -13227,4 +13227,5 @@ TODO:
     </xsl:if>
 </xsl:template>
 
+
 </xsl:stylesheet>
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -5969,7 +5969,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     </xsl:variable>
     <xsl:variable name="layout" select="exsl:node-set($rtf-layout)" />
     <!-- div is constraint/positioning for contained image -->
-    <div>
+    <div class="image-box">
         <xsl:attribute name="style">
             <xsl:text>width: </xsl:text>
             <xsl:value-of select="$layout/width"/>
@@ -5981,8 +5981,6 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <xsl:value-of select="$layout/right-margin"/>
             <xsl:text>%;</xsl:text>
         </xsl:attribute>
-        <!-- fragment for reveal.js pauses -->
-        <xsl:apply-templates select="." mode="class"/>
         <xsl:apply-templates select="." mode="image-inclusion"/>
     </div>
 </xsl:template>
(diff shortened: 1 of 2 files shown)


rbeezer:
Long post just now on `-dev` responding to general strategies in play here.

Without any testing, a question.  How do the images behave with an `image-box`, etc.  The sort-of automatic scaling that RevealJS does always gives me fits when I **use** a presentation.  I can usually manage decorative images (e.g. a textbook cover), but other times I need to use my browser's zoom-in or zoom-out to temporarily get a single slide usable.  Any observations?

rbeezer:
Shoulda also said.  Thanks for hopping on this one.  I was going to address the managed directories (and not much more!), so thanks for being more comprehensive.  ;-)

kcrisman:
> Without any testing, a question. How do the images behave with an `image-box`, etc. The sort-of automatic scaling that RevealJS does always gives me fits when I **use** a presentation. I can usually manage decorative images (e.g. a textbook cover), but other times I need to use my browser's zoom-in or zoom-out to temporarily get a single slide usable. Any observations?

Can you give an explicit example of that (if you happen to have one on the web, this is a request not a demand)?  I've not had that problem, but perhaps my custom code dealing with sizing meant I accidentally avoided that problem.

oscarlevin:
I did very little use testing, only made sure that the htm was doing better.  This all came about because I wanted to put a qr code in my presentation yesterday and realized that it was always getting blown up too big, no matter what width I was adding.  

I agree the auto resizing is annoying.  I need to zoom out on the title page often, and then zoom back in when there is an image, since that is the best way to make the image *smaller*, when the text gets larger.

oscarlevin:
Second attempt, using a modal template to set the `@class` on the img as either `image-box` or `fragment` in the reveal.js conversion.  I'm not sure which is better; both approaches work as intended.

No need for any sidebyside modifications, as the reveal.js conversion already handled that correctly.

oscarlevin:
Third time's the charm I think.  Using `<xsl:apply-imports>` makes it easy to wrap any image that has a pause attribute with a span with `@class="fragment"` and then include the image template from `pretext-html`.

rbeezer:
Much better!  ;-)

*  Why a `span` and not a `div`?  Won't the block-level `div.image-box` be illegal inside the inline-level `span.fragment`?
*  Rather than two `xsl:if`, I'd suggest a `choose/when/otherwise`.
*  Experiment for the benefit of your XSL-fu:  inside the `match` add `and (@pause = 'yes')` and then have the template simply do the wrapping straight-away with no conditionals inside the body of the template.  I think that test is OK inside the match (but could be mistaken, so test first).

oscarlevin:
Ah, that's much cleaner.  Should be ready for squashing.

rbeezer:
Looking good.  Thanks for working through this one.  The addition of an extra `div` made this a lot cleaner.  I hope the XSL-fu has increased.

While I'm at it see: #2159

PR #2158, opened 2024-04-21
Modify ToC scroll so it doesn't change tab location

ascholerChemeketa:
While playing with a screen reader, I figured out that the js library function `scrollIntoView` sets the target item as the current location for tab purposes. Thus when the ToC has to scroll to reveal an item, that item becomes the current tab stop. If the user hits tab they are already deep in the menu instead of starting from the top of the page and immediately hitting the skip link.

This modifies the ToC scroll code to not rely on that library function

### DIFF
--- a/js/pretext.js
+++ b/js/pretext.js
@@ -12,11 +12,13 @@
  */
 
 function scrollTocToActive() {
+    //Try to figure out current TocItem from URL
     pagefilename  = window.location.href;
     pagefilename  = pagefilename.match(/[^\/]*$/)[0];
+    pagefilename = pagefilename.split("#")[0];
     possibletocentries = document.querySelectorAll('#ptx-toc a[href="' + pagefilename + '"]');
     if (possibletocentries.length == 0) {
-        console.log("linked below a subsection");
+        //linked below a subsection
         pagefilename  = pagefilename.match(/^[^\#]*/)[0];
         possibletocentries = document.querySelectorAll('#ptx-toc a[href="' + pagefilename + '"]');
     }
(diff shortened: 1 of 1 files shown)


rbeezer:
I think the undesirable behavior is not universal.  With Firefox I get the "Skip to Main Content" when expected, while on Chrome I do not.

Sample Article, Section 19 has a cross-reference to Appendix B.  Following that causes a scroll of the ToC.  In Firefox a Tab produces the skip-link.  In Chrome it does not.  Doesnn't mean we shouldn't fix this so Chrome works better.

Question:  I'm not quite following the empty `possibletocentries` and the comments about `subsection`.  Do you mean `subdivision`?  The first is a specific element/division, the second is the generic situation of nested divisions.  We can make a ToC to different depths and we can chunk the HTML to different granularities.  So I don't see a `subsection` as special/distinctive.

If you edit this, there is a comment where `tab` needs a space after it.  Thanks for fixing up this accessibility FUBAR.

ascholerChemeketa:
Rebuilt that function to clarify what I believe to be the original intent of that logic.

Tested against ToC's from depth 2-5 and it appears to work correctly for all.

ascholerChemeketa:
@rbeezer Looks like I didn't tag you when I de-draft statused this.

rbeezer:
> Looks like I didn't tag you when I de-draft statused this.

Well, I'd seen it, but had lost track of it since there haven't been too many PRs lately.  ;-)

Thanks for the further work on this one.  Its in now.  May be tomorrow before the website sees a rebuild.

rbeezer:
Section 19 of the sample article has a cross-reference to Subsubsection 4.2.1.  I guess it is too much to expect that the 4.2 entry of the ToC becomes active?  I guess it would mean looking at ancestors of the hash location until there was some sort of match.  Would it help if the link/xref had some hints in it (provided by the XSL)?

ascholerChemeketa:
Yes, that deserves a fix. The existing code highlights the lowest match and the "page level" match. There, the hidden subsubsection is being highlighted and the subsection is skipped.

I can easily mark the entire path from low level to high level (i.e. subsubsection/subsection/section). That commit is currently on my fork here:
https://github.com/ascholerChemeketa/pretext/commit/2cb12bf719421631d395e737a1e8c5c108e16040
I can do PR or you can just grab that code and smash it into this one.

Anything more nuanced (picking what to highlight based on the toc display depth) would likely benefit from more structure or involve some ugly hakery.

But I'm thinking it might be useful to consider active highlighting based on what part of the book is currently being viewed. Right now if you scroll around on a page like that one the ToC's "active" item rapidly becomes misleading. But that deserves its own conversation.

rbeezer:
Soundx good.  Both an addition and a subsequent conversation.

> I can do PR 

Yes, please?  A bit easier for me and we have a record here.  Thanks.

PR #2160, opened 2024-05-08
2-col-list-space-fix

ascholerChemeketa:
Fixes 2 column lists having a smaller left column.


rbeezer:
Thanks!  Merged, and rebuilding website now.  More on the precipitating thread.

PR #2162, opened 2024-05-13
Update Czech translation

jirilebl:
Update the Czech translation


rbeezer:
As always, thanks for keeping this current!

PR #2163, opened 2024-05-16
WeBWorK: replace explanation essay boxes with a message

Alex-Jordan:
This makes it so that PG "explanation essay boxes" behave like regular essay boxes for interactive rendered WeBWorK problems.


rbeezer:
Thanks!  Website update in a little bit.

PR #2164, opened 2024-05-16
proof of concept hacks to implement Doenet without iframes

StevenClontz:
Mostly for a conversation with @jaltekruse - I expect this will be closed and (possibly) reimplemented more strategically in a separate PR.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -198,6 +198,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- fix right now.                                                     -->
 <xsl:variable name="b-has-sfrac"        select="boolean($document-root//m[contains(text(),'sfrac')]|$document-root//md[contains(text(),'sfrac')]|$document-root//me[contains(text(),'sfrac')]|$document-root//mrow[contains(text(),'sfrac')])"/>
 <xsl:variable name="b-has-geogebra"     select="boolean($document-root//interactive[@platform='geogebra'])"/>
+<xsl:variable name="b-has-doenetml"     select="boolean($document-root//interactive[@platform='doenetml'])"/>
 <!-- 2018-04-06:  jsxgraph deprecated -->
 <xsl:variable name="b-has-jsxgraph"     select="boolean($document-root//jsxgraph)"/>
 <!-- Every page has an index button, with a link to the index -->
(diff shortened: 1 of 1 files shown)


xsl/pretext-html.xsl
```diff
@@ -9693,6 +9695,16 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
...
+<!-- For more complicated interactives, we just point to the page we generate -->
+<xsl:template match="interactive[@platform='doenetml']" mode="iframe-interactive">
+    <div>
+        <xsl:apply-templates select="." mode="html-id-attribute"/>
+        <xsl:apply-templates select="." mode="permid-attribute"/>
+        <xsl:apply-templates select="." mode="size-pixels-attributes" />
+        <xsl:apply-templates match="slate[@surface = 'doenetml']"/>
+    </div>
+</xsl:template>
+
```
StevenClontz, line 9707: Here we overwrite the template that creates an iframe and just render the HTML itself. But this breaks asset generation for capturing preview screenshots - the headless browser explicitly looks for an iframe element, which it doesn't find due to this hack.

xsl/pretext-html.xsl
```diff
@@ -9816,7 +9828,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/doenet-standalone-test@0.6.0/dist/style.css" />
     <script onload="onLoad()" type="module" src="https://cdn.jsdelivr.net/npm/doenet-standalone-test@0.6.0/dist/doenet-standalone.min.js"></script>
     <script>
-        <xsl:text>function onLoad() {window.renderDoenetToContainer(document.querySelector(".doenetml-applet"))}</xsl:text>
+        <xsl:text>function onLoad() {document.querySelectorAll(".doenetml-applet").forEach(e=>window.renderDoenetToContainer(e)}</xsl:text>
```
StevenClontz, line 9831: There's something buggy (with doenet-standalone?) here where if I have multiple applets on the same page, sometimes (often) the second applet isn't rendered correctly, see:

image

rbeezer:
I'm going to close this.  I use this space to keep track of what is active and/or needs needs merging and it sounds like this is not the case with this right now.

Of course, it will still be available for collaboration as part of the "closed" group.

PR #2167, opened 2024-06-05
Refactor preview_images to use thread instead of subprocess

oscarlevin:
This fixes the issue of interactive preview images being unable to be generated on Windows.  

Instead of starting http.server in a subprocess, this now starts it in a thread.  GET messages from the server are sent immediately to log.debug.

Also fixed a trailing semicolon and a unneeded use of `os.path.join` when creating the URL.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -2090,7 +2090,7 @@ def preview_images(xml_source, pub_file, stringparams, xmlid_root, dest_dir):
             # First index contains original baseurl of hosted site (not used)
             for preview_fragment in interactives:
                 # loaded page url containing interactive
-                input_page = os.path.join(baseurl, preview_fragment + ".html")
+                input_page = baseurl + "/" + preview_fragment + ".html"
                 # filename of saved preview image
                 filename = preview_fragment + "-preview.png"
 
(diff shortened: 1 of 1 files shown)


oscarlevin:
I should say, it would be good to also test this on linux/mac in addition to the testing I've done on Windows.

rbeezer:
Thanks, @oscarlevin!  I will test on Linux here shortly.  Perhaps @Alex-Jordan could do a test on a Mac?

rbeezer:
OK, testing on Linux with sample article.  Short answer: looks good.

Messages all look informative and rational (at full verbosity).  Correct files generated, and visually look good.  File sizes look to be within epsilon of "old" versions in repository.

I'm going to ping @Alex-Jordan off-list.

rbeezer:
Forgot the caveat.  Using `pretext/pretext` script, and I did not check how current (old?) my Playwright is.

Which prompts me to create: https://github.com/PreTeXtBook/pretext/issues/2168

oscarlevin:
Accidentally pushed and accidental merge of Master, so force-pushed a reversion.

rbeezer:
Merged without any Mac testing.  Fingers crossed.

Removed extraneous whitespace.  Isolated unrelated small fixes from the real changes.

Thanks for backing up our cross-platform guarantees.

PR #2172, opened 2024-06-16
WeBWorK: clear solution label regardless of div or span

Alex-Jordan:
This area of the pretext-webwork.js code clears the label like "Solution" that WW puts as the solution clickable (so that PTX can replace it with a localized version). Previously it was a `div`, but now it's a `span`. This change in code will clear it regardless of what HTML element it is.


rbeezer:
Merged!  Thanks.

PR #2173, opened 2024-06-17
add 'debug.style.browser' stringparam to show style browser

oscarlevin:
This implements a simple javascript addon that can be included via the string param "debug.style.browser".  When active, it shows a bar at the bottom of the page where you can pick which color, inner style, and outer style you want to view the page in.  Could be used by a "publisher" to select which of the styles they want to adopt, or an author viewing the sample article to see what is available.  And of course, developers can ensure that their styles look good with various combinations of the other two elements.

### DIFF
--- a/js/style_browser.js
+++ b/js/style_browser.js
@@ -1,0 +1,114 @@
+window.onload = function() {
+  // get link elements
+  var linkElements = document.getElementsByTagName("link");
+  // get links with style_, colors_ and shell_ css:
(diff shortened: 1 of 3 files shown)


rbeezer:
Love it!  Merged.

* I set default value to an empty string.  Easier to deprecate.  Which really isn't important for a debug switch.
* I made the comment appear before the script.
* You are getting good at JS.

Advertise on `-dev`?

PR #2175, opened 2024-06-17
LaTeX: add \par\medskip for paragraph following case or paragraphs

### LINKED ISSUE #2174: No separation between case/paragraphs and a following paragraph in LaTeX output
In LaTeX output:

For source structure
````
proof
  p
  case
  case
  p
````
The final `p` is not a new paragraph but instead just continues on the last line of the preceding case.

Similarly, for source structure
````
paragraphs
  p
  p
p
````
there is a significant amount of vertical space preceding the `paragraphs` but no vertical space between the final `p` of the `paragraphs` and the new `p` outside of the `paragraphs`, making that final `p` look like it is part of the `paragraphs`.

jjrsylvestre:
Fixes #2174.

### DIFF
--- a/xsl/pretext-latex.xsl
+++ b/xsl/pretext-latex.xsl
@@ -6766,8 +6766,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- TODO: maybe we could look backward at the end of a paragraph       -->
 <!-- to see if the above scenario happens, and we could end gracefully. -->
 <xsl:template match="p">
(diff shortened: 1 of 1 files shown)


jjrsylvestre:
Ran sample article through both master branch and this branch, only difference between the two generated `.tex` files (besides generation times) was a single extra `\medskip` near the end of section 20, where the only instance of `paragraphs` followed immediately by `p` occurs.

rbeezer:
Merged!  Thanks.

PR #2177, opened 2024-06-19
Refactor handling of ol marker attribute to allow custom markers in html

jjrsylvestre:
Analyse `@marker` for `ol` at the assembly stage and "augment" source tree with the results.

For LaTeX output, processing the sample article produced an identical `.tex` file as when processed using the master branch.

For HTML output, use custom item counters based on `html-id` for each `ol`, then style these separately according to the marker. Style for every `ol` in the entire source tree needs to go in every chunk file because of the possibility of nesting content from other "chunks" in knowls.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -4539,7 +4539,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     <li><p>Purple</p></li>
                 </ul></p>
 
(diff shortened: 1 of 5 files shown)


jjrsylvestre:
Pulled the trigger a little too soon. Comparing section 11 in the LaTeX and HTML versions of the sample article, noticed that start-at-0 is broken in HTML output.

Will re-submit soon.

jjrsylvestre:
OK, fixed 0-based lists in HTML, now re-open and force-push.

jjrsylvestre:
Only 198 lines of list styling added to each file of the sample article. ;)

rbeezer:
Really nice!  Thanks for bringing HTML up to the typographical standards of LaTeX/PDF!

Combined into one commit.

PR #2179, opened 2024-06-20
Implement @direction="cycle" for <case>

jjrsylvestre:
I think this one's ready to go.

Diff of `sample-article.tex` between branch and master is identical except for material this commit adds to the sample article along with resulting numbering changes after that added material.


rbeezer:
Can you do `CYCLENUMBERSEPARATOR` with an abstract template?

jjrsylvestre:
Do you mean so that that special word is only hard coded in one place
instead of several? Or is there some way to eliminate that altogether? I
couldn't think of any other way to "return" two different but related
values from a template...

rbeezer:
Wonderful!  Thanks.  (i'd commented on the magic string and then failed to send it.)

The abstract approach will be easier to maintain, I hope you like it.

Repackaged, but no content changes.  May add to author guide right now.

PR #2181, opened 2024-06-21
Partial Order Parsons Implemented & Documented

ascholerChemeketa:
Implementation of markup for Parsons problems to allow for partial order autograding.


rbeezer:
How do you feel about using `@name` where you have used `@tag`?  Totally cosmetic, but important for authors' efficiency/sanity?  We have a `<tag>` element (to make consistent documentation of PreTeXt itself) and `@name` gets used many other places, sometimes in a very similar way.  It'd help authors be more likely to "guess" the right attribute name.

ascholerChemeketa:
That makes sense.

Change made and documentation typo fixed.

rbeezer:
Consolidated and merged.

A `cline` in the block named `a` had markup (`nbsp`) which was washed away in HTML, appeared in PDF, and raised a warning during processing (no makup allowed in `cline`).  So I made an executive decision to remove it.  Holler if that was wrong.

Thanks for getting to this - I never finished my attempt.

ascholerChemeketa:
@rbeezer That was the correct decision. I believe that was an artifact I introduced while trying to help Kevin figure out how to force an empty line into the Parsons block. My apologies!

rbeezer:
>  My apologies!

No problem.  That's why we do reeviews.  Glad it was handled smoothly.

PR #2182, opened 2024-06-21
FITB: static substitutions

dbrianwalton:
Restored the two dynamic exercises in sample article and provided the necessary substitution file for the static build to work.


rbeezer:
Thanks!

PR #2183, opened 2024-06-21
FITB: Update dynamic exercise markup scheme

dbrianwalton:
The original PR for dynamic exercises used a markup that had too many elements as attributes. Going forward, to easier facilitate definition of new options, wanted to transition from attributes to child elements.

In addition, there were some unexpected issues relating to the use of the original number and string (regexp) checkers when used in combination with the new markup. This was addressed in a separate PR to Runestone that Brad merged today.
The markup requires that new Runestone build for interactives to function properly.

The sample article problems have been verified to continue to work in both html and static form.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -8896,15 +8896,58 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <exercises>
                 <title>Dynamic Fill-In</title>
 
+                <introduction>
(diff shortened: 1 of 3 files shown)


dbrianwalton:
@bnmnetp and @rbeezer : The reason the FITB problems were missing Runestone css application was missing one layer of #div so this fixes the appearance discrepancy as well.

rbeezer:
Merged.  Thanks for the improvements.

PR #2185, opened 2024-06-25
use searchparams rather than magic URL for preview

StevenClontz:
See discussion at https://groups.google.com/g/pretext-dev/c/UtJRHp_mgPU/m/bazSzI8FBQAJ

This enhancement allows for deploying a PreTeXt document to a URL that includes `/preview/` without automatically injecting the `newstuff` classes. E.g. if my book is named `Calculus Preview` and I deploy my work to `https://my-great-calculus-book.org/preview/` I may not want the `newstuff` classes.

It also allows for injecting the `newstuff` classes into any desired PreTeXt document, by appending `?ptxpreview` to the end of the URL, without having to make a separate deployment for a "preview" version.

### DIFF
--- a/css-obsolete/pretext_add_on.css
+++ b/css-obsolete/pretext_add_on.css
@@ -2709,32 +2709,6 @@ body.standalone.worksheet .ptx-content section.worksheet > .heading + .para {
     overflow: auto !important;
 }
 
-/* Preview of annual edition */
-
-.pretext main .para.newstuff {
-    border-right: 6px solid #6f6;
-    background:  #dfd;
-}
-.pretext main li.newstuff {
-    border-right: 6px solid #66f;
-    background:  #eef;
-}
-.pretext main li.newstuff .para.newstuff {
-    border-right: none;
-    background: inherit;
-}
-.pretext main article.newstuff {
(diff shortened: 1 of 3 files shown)


StevenClontz:
@davidfarmer

davidfarmer:
This is fine with me.  I am not aware of a current use case for the previous "preview" feature.
Probably some half-baked experiment that should have been temporary.

StevenClontz:
Thanks David. If that's the case, of course another solution is to remove these lines altogether.

davidfarmer:
Yes, just removing that "feature" makes sense.

rbeezer:
Well, I thought that was a really neat feature.  But we will have a commit that retains it for anyone who wants to bring it back.

rbeezer:
Just looked at this.  The `newstuff` CSS should be removed as well, no?

StevenClontz:
Done in 4311c9b

rbeezer:
Consolidated and merged.  Thanks!

* The `css-obsolete` is not in use (I think).  I removed changes there, since I think the whole thing should be dropped.  But will need a review first.  Not now.
* Removed an extraneous blank line.
* Maybe some discussion on `-dev`.

PR #2186, opened 2024-06-28
Make pymupdf the default tool for generating svg and png

oscarlevin:
from latex-image.  The original method is still available, behind a pymupdf switch that now defaults to true.


rbeezer:
This is looking good, though I have not yet taken it for a drive.  Thanks for preserving the old behavior.  I think it needs two related things, and one follow-on:

* Documentation should stop talking about installing/using `pdf2svg` and `PyMuPDF` should be added as a dependency for the `pretext/pretext` script running in a Python virtual environment.
* Lots of project files and `pretext.cfg` files in the repository pointing at `pdf2svg`.  These need to change so the website update script won't be impeded.
* I'm only likely to update SVG images for the sample article as a test.  A **subsequent** PR could update others in the repository.

oscarlevin:
Added two commits to address your first two bullets.  Also added a new issue to eventually completely remove `convert` (#2188 )

rbeezer:
Thanks, Oscar!  Sorry for the long delay.  I thought this was a bit more involved than it really is/was.  Merged as-is, but for one whitespace character.

I'll suggest @Alex-Jordan remove the final dependency on `convert`.

Great to get rid of all those instructions in the Guide about one-off software installations.

PR #2187, opened 2024-06-30
Auto add input element to programs that only contain text

ascholerChemeketa:
Based on recent support discussion...

If a program lacks an **input** element around the text inside it, auto-supply the **input**. This tries to be very conservative and only does so if there are no other children in the parent element. (e.g. if you have a **tests** in the program you will not get an automatic **input**.)

I am quite sure that the XSL selection could be done better. Took this on as an XSL exercise, let me know how it should have looked. :)

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -10060,14 +10060,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
             <p>And some self-referential XSL:</p>
 
+            <!-- input element intentionally ommited here. Pretext processor should correct. -->
             <program language="xslt" margins="15%">
-                <input>
-                <![CDATA[<xsl:template match="biblio" mode="number">
-                    <xsl:apply-templates select="." mode="structural-number" />
-                    <xsl:text>.</xsl:text>
-                    <xsl:number from="references" level="any" count="biblio" />
-                </xsl:template>]]>
-                </input>
+            <![CDATA[<xsl:template match="biblio" mode="number">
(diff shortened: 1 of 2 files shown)


ascholerChemeketa:
@rbeezer Yes, `program[not(input)]` is much cleaner. I started trying to match on something like that but ran into conflicts with the `program/@interactive` match below it. My solution was to match on the text. But with fresh eyes I figured out how to get them to coexist.

How much do I care about whitespace changes like this?

Orignal hand written:
```
<program ...>
    <input ...>
    ...contents...
    </input>
</program>
```

Autogenerated input element:
```
<program ...><input>
    ...contents...
</input></program>
```

If I do care about having the created input element look nicely spaced, any tips?

rbeezer:
No person ever sees that **intermediate** XSL file, so space between elements that is just visual formatting (and not textual important content) is irrelevant to the conversions.

Your question made me think we might to consider that `xsl:strip-space` is in play for `program` as part of this change, but after reading Lenz, I decided not (and did not even check the status of `program` at all).

I think this does explain why that space went missing (very early on in processing!).

Carry on!

ascholerChemeketa:
@rbeezer I think this is ready to go. I pushed a cleaned-up commit history.

Thanks for the insight into your process for testing assembly output.

rbeezer:
Merged.  Thanks for taking this on.

Got some whitespace change on a `section` of the sample article being formed as a search document for Lunr.  Couldn't really track it down, but it seems harmless.

Added a follow-on commit to the schema. (Where `program` probably needs more work after all the extensions for Runestone.)

PR #2190, opened 2024-07-05
Script: create managed folders when they don't exist

Alex-Jordan:
Like the commit message says, this creates the generated and external folders when they do not exist. In particular if a project doesn't commit the generated folder, then when you first clone the repository you won't have this folder and things won't build until you create it. But with this commit that annoyance is relieved.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -4342,8 +4342,8 @@ def get_managed_directories(xml_source, pub_file):
             missing_dir_error = " ".join(
                 [
                     'the directory "{}" implied by the value "{}" in the',
-                    '"source/directories/@{}" entry of the publisher file does not',
-                    "exist. Check the spelling, create the necessary directory, or entirely",
+                    '"source/directories/@{}" entry of the publisher file does not exist and could',
+                    "not be created. Check the spelling, create the necessary directory, or entirely",
                     'remove the whole "source/directories" element of the publisher file.'
                 ]
             )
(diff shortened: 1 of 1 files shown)


rbeezer:
I tried the `pathlib` module once long ago, and it was a disaster.  I do not intend to start introducing it now.  See uses of `os.path` elsewhere.

Alex-Jordan:
Now with `os`.

rbeezer:
Thanks for the changes.  Looking good visually.  Should merge soon.

rbeezer:
Well, I thought this would be straightforward.  But testing makes me think otherwise.

Our strategy is to copy an author's material into a temporary directory and manipulate it there (perhaps as part of the copy).  Not sure we write things into an author's source, and I don't think we should be touching that really.  And perhaps some issues with permissions will result.

So how about:

* Soften existing failure (directory does not exist) to a warning: "Did you mean to have such a directory?  We don't see it.  Perhaps you need/want to add it?"
* In `copy_managed_directories()` react to `None` by creating an empty `external` or `generated` (respectively) in the temporary working space.  Presumably it won't be touched anyway.
* We'd need to survey if `getting` these directories is always followed by the `copy`.

oscarlevin:
For what it's worth, the CLI ensures these, and all other required directories, exist.

Alex-Jordan:
> Not sure we write things into an author's source

The purpose of the `generated` folder is for pretext to write things into it. All long before copying to a scratch folder. So it seemed OK to write the folder itself.

I will withdraw this. I don't have bandwidth to pursue the other approach you outlined.

PR #2191, opened 2024-07-05
Schema: add myopenmath to experimental schema

oscarlevin:
I think this is the intended behavior for MyOpenMath problems.  Just experimental for now.


rbeezer:
Needed a comma between the two (new) attributes.  Otherwise, unchanged.  Thanks for keeping this evolving.

PR #2192, opened 2024-07-05
Fix wrapping issue for standalone-worksheet print versions

rhinopotamus:
Per Andrew Scholer in [this google groups discussion](https://groups.google.com/g/pretext-support/c/i32M1zm4SUc/m/kgaF26TdBAAJ), this is a css edit to fix an issue where text goes beyond the right edge of the page when a standalone worksheet is printed in portrait mode.

### DIFF
--- a/css/pretext_add_on.css
+++ b/css/pretext_add_on.css
@@ -2618,15 +2618,6 @@ body.standalone.worksheet .ptx-content .onepage .workspace.squashed.tight {
 body.has-sidebar-left.mathbook-loaded.standalone.worksheet .ptx-page .ptx-main {
     margin-left: 0;
 }
-body.standalone.worksheet .ptx-page > .ptx-main .ptx-content {
-    max-width: 1030px;
-    width: 1030px;
-    margin-right: 0;
-}
-body.standalone.worksheet.a4 .ptx-page > .ptx-main .ptx-content {
-    max-width: 1001px;
-    width: 1001px;
-}
 
 body.standalone.worksheet .ptx-content .goal-like {
     border: none;


rbeezer:
@davidfarmer - can you take a peek at this one?  As the creator of worksheet printing.  See a recent post on `pretext-support`.

rhinopotamus:
(Aside which I'm happy to make into a separate issue if that's helpful: 
Because of the way `pretext-latex.xsl` builds a `\newgeometry{}` command from a `<worksheet>`, you can control margins in the print build. However, the web build doesn't know anything about margin attributes of a worksheet. 
I feel like it should be possible to pass margin attributes to the css similarly, but I don't know enough about xslt / css / etc. to figure this out immediately.)

mitchkeller:
As I said in reaction to Andrew's suggestion of this on the Google Group, I don't think this is the right long-term fix. Hopefully @davidfarmer can make this a priority in a little over a week during the Steubenville workshop, since worksheets are high on the agenda there.

davidfarmer:
I suggest accepting this, and when Mitch and I are in the same place next week, we can investigate further.

rbeezer:
Merged.  Thanks very much @rhinopotamus!

> when Mitch and I are in the same place next week, we can investigate further

So this may see some further iteration soon.

rbeezer:
> (Aside which I'm happy to make into a separate issue if that's helpful:
Because of the way pretext-latex.xsl builds a \newgeometry{} command from a <worksheet>, you can control margins in the print build. However, the web build doesn't know anything about margin attributes of a worksheet.
I feel like it should be possible to pass margin attributes to the css similarly, but I don't know enough about xslt / css / etc. to figure this out immediately.)

Yes, a new issue would he helpful, so this suggestion does not get lost.  Thanks.

PR #2193, opened 2024-07-05
Don't add MOM when extracting pg

oscarlevin:
This fixes the circular dependency for books that have both WeBWorK and MyOpenMath.  Authors must generate the ww-representations file first, then ensure they have fetched any MyOpenMath problems, all before trying to build or generate other assets.  

The CLI already uses `ensure_webwork()` before building.  It can easily add an `ensure_myopenmath()` as well.  That way asking for other html assets before myopenmath has been fetched would be caught gracefully.

Alternatively, the assembly xsl could be further modified so that other assets can be generated without first fetching the `mom.xml` files.  Unless of course those xml files could contain assets that need to be generated, in which case it is good this isn't possible.

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -2411,7 +2411,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:apply-templates select="@*" mode="representations"/>
         <!-- Now bifurcate on static/dynamic.  PG problem creation should not fall in here. -->
         <xsl:choose>
-            <xsl:when test="$exercise-style = 'static'">
+            <xsl:when test="($exercise-style = 'static') and not($b-extracting-pg)">
                 <!-- locate the static representation in a file, generated independently -->
                 <!-- NB: this filename is relative to the author's source                -->
                 <xsl:variable name="filename">


rbeezer:
I'm thinking there should be some variable like `b-extracting-mom` and then we do not try to incorporate any non-essential XML files/content.  Some more general scheme for when we have a third problem type producing static versions off on some server.  And it would remove a "you gotta do this first" we are getting her.

Don't do anything, I'd like to merge this as an improvement, and think more carefully than above about the general situation.

rbeezer:
I went ahead and merged this as a stopgap measure so other things can proceed.  I'll think more about a proper scheme to prevent these clashing dependencies.

oscarlevin:
It seems like the issue with MOM in particular is that it is added for all "static" uses, but that includes the generation of static resources.  So perhaps a "generating-assets" variable of some sort?

rbeezer:
Right.  I think each type needs its own variable, switched to `true` when its own extraction stylesheet is run.  Then there is an overall variable (the logical or of the individual ones) which prevents trying to incorporate material that may not yet be extracted and is not part of the current extraction.  (Which makes sense to me, but that may not be very clear to read!)

PR #2194, opened 2024-07-06
Guide: correction

Alex-Jordan:
The guide has `qrcode` where it should be `qr-code`.

### DIFF
--- a/doc/guide/publisher/publication-file.xml
+++ b/doc/guide/publisher/publication-file.xml
@@ -150,15 +150,15 @@
 
         <subsection xml:id="common-qrcode-image">
             <title>QR code image</title>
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks!  Merged.

But I sort of wish it had been `qrcode` as the attribute name (similar to non-hyphen versions elsewhere).  Do you think we could do a clean deprecation of `qr-code` and move to `qrcode` for this too?

Alex-Jordan:
I can maybe do that. Do we have a deprecation scheme for publisher file attributes already? If now, maybe I would add that.

rbeezer:
> Do you think we could do a clean deprecation 

Correction really, see #2198

PR #2195, opened 2024-07-06
LaTeX: remove hidelinks argument

Alex-Jordan:
I get a warning from the `=true` that this has right now. Consulting documentation, it's not supposed to take an argument.


rbeezer:
Excellent, one more `Warning` gone.  ;-)  Two notes:

* Only used for "print" PDF, so not part of normal testing (took me a minute to figure that out).
* I added in a few `tex.stackexchange` links as code comments.

Issue #2189, opened 2024-07-03
\fillintext dependency on \real package issue


dbrianwalton:
The LaTeX macro \fillintext defined in pretext/xsl/pretext-latex.xsl has a dependency on the \real macro that comes from the LaTeX package `calc`. This package must be a dependency of at least one of the conditional packages that are loaded because the macro sometimes is defined and sometimes not. See the pretext-support discussion here: https://groups.google.com/g/pretext-support/c/jUxiSXdDjEU

I *think* the correction is to add a line to `\usepackage{calc}` to the standard LaTeX preamble. Is there a reason this should be conditional? The possibly related line `\usetikzlibrary{calc}` is always included, so I'm guessing the above package should as well.

rbeezer:
Closing as resolved. #2196 ("LaTeX: remove `\fillintext` dependency on 'calc' package") removed exactly this — `\real` is provided by `calc`, and that dependency is gone from the LaTeX conversion. It's on master.

*Claude Opus 4.8 (1M context), acting as a review assistant for Rob Beezer*

PR #2196, opened 2024-07-06
LaTeX: replace use of \real in \fillintext

Alex-Jordan:
This addresses #2189 by removing `fillintext`'s dependency on `calc`.


Alex-Jordan:
Stand by, there is still another reliance on `calc` to remove...

Alex-Jordan:
OK, now this is working with no `calc` package. Both `\real` and `\heightof` are from `calc` and are not used anymore.

rbeezer:
Thanks!  I feel much better with that dependency gone/minimized.  Two small changes to the XSL in order to get slightly different formatting of the LaTeX file.

* A newline into the LaTeX file right after `\newlength{\fillinheight}` (as previously)
* A % at the end of the line `\setlength{\fillinheight}{\baselineskip}\addtolength{\fillinheight}{1.2pt}` so the main macro does not get spurious whitespace bleeding through (as with other lines in the macro definition)

PR #2197, opened 2024-07-07
Bundle Runestone files into _static on build

ascholerChemeketa:
This implements what I believe @bnmnetp wants for Runestone assets in PreTeXt books - local copies in _static.

The PTX HTML build outputs links for RS assets as per this logic:
1) If the user has a stringparam set indicating they are developing runestone, the pretext script outputs links that will match those output by RS's webpack config in dev mode. The assumption is the user will be building their own RS assets to `_static/runestone`

2) Otherwise, the normal process is followed to find the current version of RS - look at debug flag; query server; default to what PreTeXt specifies. 

   Once the version is identified, pretext script attempts to download it and unpack it to `_static/runestone`. Future builds that need the same RS version will not redownload the assets.

   If the download is successful or files already exist, generate links that point to the production version of tiles in `_static/runestone`. 

  If the download can't be done, generate links that point to the Runestone CDN.

Has a partner PR in Runestone:
https://github.com/RunestoneInteractive/rs/pull/497

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -3556,6 +3556,51 @@ def _runestone_services(params):
     altrs_version = services.xpath("/all/version")[0].text
     return (altrs_js, altrs_css, altrs_cdn_url, altrs_version)
 
+# todo - rewrite other code that does similar things to use this function?
+def get_web_asset(url):
+    """Get the contents of an http request"""
+    try:
+        import requests
+    except ImportError:
+        msg = 'The "requests" module is not available and is necessary for downloading files.'
+        log.debug(msg)
+        raise Exception(msg)
+
+    try:
+        services_response = requests.get(url)
+    except requests.exceptions.RequestException as e:
+        msg = '\n'.join(['There was a network problem while trying to download "{}"',
+                            'and the reported problem is:',
+                            '{}'
+                            ])
+        log.debug(msg.format(url, e))
+        raise Exception(msg.format(url, e))
+
+    # Check that an online request was "OK", HTTP response code 200
+    response_status_code = services_response.status_code
+    if response_status_code != 200:
+        msg = '\n'.join(["The file {} was not found",
+                            "the server returned response code {}"
+                            ])
+        log.debug(msg.format(url, response_status_code))
+        raise Exception(msg.format(url, response_status_code))
+
+    return services_response.content
+
+def download_file(url, dest_filename):
+    """Write a web asset to a local file"""
+    contents = get_web_asset(url)
+    try:
+        import pathlib
+        output_file = pathlib.Path(dest_filename)
+        output_file.parent.mkdir(exist_ok=True, parents=True)
+
+        with open(dest_filename, 'wb') as f:
+            f.write(contents)
+        print(dest_filename, "downloaded")
+    except Exception as e:
+        raise Exception("Failed to save download", dest_filename)
+
 def html(
     xml, pub_file, stringparams, xmlid_root, file_format, extra_xsl, out_file, dest_dir
 ):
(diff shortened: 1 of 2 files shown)


rbeezer:
Working visually on fone.  Looks good in principle.  Please advise on coordination with the RS PR.  (Unless it is there, will go look now.)

bnmnetp:
I believe the two can be done independently.  but this one should definitely go first. @ascholerChemeketa do you agree?

ascholerChemeketa:
Consulted with @bnmnetp on Discord. Agreed that for now better to leave RS files directly in `_static`. 

Updated to do so.

This does mean we are declaring RS files "special" - they essentially own the top level of `_static`. Anything else that gets bundled into `_static` will need to be in a subdir (like pretext already is) to avoid possible collisions.

bnmnetp:
I have merged RunestoneInteractive/rs#497 So feel free to merge away here.

dbrianwalton:
The corresponding RS PR has a comment about the webpack...xml (services file), but I don't see the creation of that file in the current flow described on https://runestone-monorepo.readthedocs.io/en/latest/javascript_feature.html 

It appears that the current Runestone workflow does not create the webpack_static_imports.xml file that would be needed -- or is that also no longer going to be needed?

dbrianwalton:
To clarify my comment, let me describe my understanding of the *old* and *new* methods.

The current/old (before debug.rs.dev) approach to working with the Runestone files in a development setting is to call `npm run build`, then `python scripts/dist2xml.py test`, and then copy contents of `runestone/dist` into `output/web/_static/test` (albeit with the `output/web/` controlled by the project/publication data).

I think the new approach uses `npm run build` with an extra option that pushes the built files directly into a folder, with that folder specified as the `output/web/_static` folder. PreTeXt will separately add additional `js` and `css` files specific to PreTeXt into the same `_static` folder.

Whenever the pretext.py needs to access instances in temporary directories, such as when my code needs to generate the dynamic_substitutions, we need a hook to pull the Runestone files for the single-page files to work properly. I am ultimately not concerned about whether the Runestone files are at the `_static` folder level, but if the only place they are stored is in `output/web/_static/`, then that folder is also going to contain the `js` and `css` files the PreTeXt also pushes there.

It would be nice if there was something comparable to the generated or external directories that are automatically copied over. Even if there was a string-param that pointed to `interactives/runestone/dist` or something. (I do appreciate the idea of building runestone directly to the output folder so that we don't need to rebuild PreTeXt every time while working on the Runestone javascript.)

rbeezer:
Looking at this now.  Recent comments are really for @bnmnetp ?

ascholerChemeketa:
Short version... there are going to have to be some changes to pretext.py to enable what @dbrianwalton wants to do. It is indeed a new use case. I don't **think** it will fundamentally change what is done in this PR, so I would suggest merging it so he can easily build a new PR on top of it.

Longer responses to comments coming in a few...

dbrianwalton:
I don't know if the comments are for @bnmnetp or @ascholerChemeketa . Having the RS files eventually get to the root of output/web/_static seems to be what Brad needs. I'm trying to think through how this relates to static extraction if those files also need to be copied to temp directories—I'm trying to work through this in parallel.

This implementation assumes that someone externally inserts the RS files into that output directory (part of the Runestone npm run build step). I just want to raise the issue that this doesn't work well for temp directories. Maybe just need another stringparam to say where to pull from? I can do this later if needed.

ascholerChemeketa:
I just replied to the email chain with my "long" response.

@dbrianwalton 

The more I think about it, the more I like the quick and dirty version - just non-recursively copy _static to the temp folder used by build. Currently it does require a successful HTML build first that places the RS files in `_static`. But it would not be hard to modify `dynamic_substitutions()` to grab the files from the RS CDN and place them into `_static` if they are not there.

I think trying to cache the files into a `generated` like directory is going to cause more issues than it fixes.

rbeezer:
Thanks for doing a very nice job on this.  It looks great.  Questions and comments in decreasing order of importance.  I'll likely dig into this now, remove some whitespace, cosmetic edits, etc and merge.  So I can make later changes or you can, but not here right now, I think.

Lots to chew on.  Sorry.  We could chat at Drop-In Tuesday (won't make Thursday, it appears).

*  I like how you prevent re-loading the webpack stuff.  But I worry a bit about inspecting the final destination (not the ever-changing temp directory).  And especially with the CLI frequently taking matters into their own hands.  These downloads are authors and publishers doing builds.  Maybe not a great load on the RS servers?  Less than millions of readers getting a point release into their cache?  We could ping the CLI on `-dev` to see if this final resting place is OK to interrogate?  (Perhaps they trash it anyway from build to build?)
*  Does `dist-1.2.3.gz` need to hang around?  Could a test for a new version interrogate the version number in `webpack_static_imports.xml`?
*  Just thought of this one - would several versions of these files accumulate in `_static`?  I think the CLI does a "clean" before deploying, but others may not realize the accumulation?  Like my website-update script.  ;-)
*  There seems to be a corresponding `*.gz` file for almost (but not all) of the files.  Are they needed?  Can they be filtered out when unpacking?
*  The CLI likes to talk about "assets".  I might change references to Runestone assets to be "Runestone services" to avoid confusion.
*  "Override the Runestone manifest" - PreTeXt makes **another** (critical) manifest file for RS.  So this would be a bit confusing.  Renaming to "Runestone services file", or similar.
*  Do you think it would help readability to move the RS download routine into a function of its own?  Maybe too many parameters to pass?  Though that might clarify/hide a few things about what is involved.  Maybe.
*  I might do additional work to name the Runestone services file back to its RS name.  Less confusion.
*  I consider the absence of `_static/runestone` as technical debt.  I hope we can convince @bnmnetp  to not have this be a non-performing loan on our books.  ;-)

ascholerChemeketa:
The hardest question in my mind is to what degree things in the output directory are treated as a cache or not. 

The CLI defaults to cleaning it on each build, but that can be overridden. I think that is the right approach. Leaving the existing output in place and overbuilding it should be something done when debugging a book (rebuilding parts of it) or debugging pretext itself.

That does mean more traffic for the CDN as the check for "do we have the files in `_static` will fail on a full build.

How many book builds are there in a day? Hundreds on average? Thousands on a busy day? I guess that is a question for @bnmnetp. Also, once books are bundling their own RS assets, every book page that is not on Runestone's server but using RS interactives will no longer ping the RS server for the files on page load, so that should be less traffic.

@dbrianwalton's need for the RS files during static generation would complicate things. But the least complex option is to have `dynamic_substitutions` do a similar file check. If the output directory exists and wasn't wiped, those files get used. Otherwise, it ends up re-downloading the files. Asset generation is likely even less common than book building, so that doesn't sound awful.

If the CDN traffic is an issue, we could cache RS files to some `generated` like directory. But that is going to be its own can of worms. 

* `dist-1.2.3.gz` file is just serving as "proof of existing copy of RS files". Yes, that could go away and `webpack_static_imports.xml` could be interrogated for the version.
* `gz` files could be filtered out and would not be hard to do. But maybe better to not even include them in the bundle that RS produces? Their use case is only on a webserver server that is configured to assume they are there. They belong in the build for the live version of files on the RS CDN, but maybe not in the version designed for use for inclusion in a book. The number of authors that will happen to be serving their book in an environment configured to use those files will likely be small. And they represent ~6MB of the ~14MB total size of the `dist.1.2.3.gz file`. Pruning them would significantly reduce the amount of data sent from the CDN.  @bmntp - does your deployment process on RS depend on the final `.gz` having all of those individual `.gz` files?

All the verbiage notes make sense.

@rbeezer Let me know if you want me to update the `dist-1.2.3.gz` part of the PR.

rbeezer:
Switched the PreTeXt side to use the `webpack_static_imports.xml` filename, at 569cfd62b6ea59e349a56d920e13621dbb53a652

Got rid of the `pathlib` module use, since I abhor that module, at 527e723f1a213713a4196dbde48d990c937d174f.  :-(

Done for now.  So, @ascholerChemeketa you are free to make some of the discussed improvements.

PR #2198, opened 2024-07-07
Publisher variables: change common/qr-code to common/qrcode

Alex-Jordan:
This follows up on #2194. It changes the correct publisher file element to use from `qr-code` to `qrcode`.

It does not deprecate `qr-code`. I could relatively easily do a one-off where someone's `common/qr-code/@image` would be recognized and used. I was hoping instead to have/make a general tool for retiring publication file elements and attributes, but it's too big of a project and I can't afford time on that.

`qr-code` was never actually in the publisher file, it was only "official" in that it was demonstrated in the sample article. And until yesterday's merge of #2194, even the documentation didn't direct people to `qr-code`. There was a `-announce` message but it was directly people to the bad documentation.

So it's not unreasonable to just end support ungracefully for `common/qr-code/@image`. But if you would prefer, I could do the one-off extended support right now.

### DIFF
--- a/doc/guide/publisher/publication-file.xml
+++ b/doc/guide/publisher/publication-file.xml
@@ -150,15 +150,15 @@
 
         <subsection xml:id="common-qrcode-image">
             <title>QR code image</title>
-            <idx><h>common option</h><h>qr-code</h></idx>
-            <idx><h>qr-code</h><h>common option</h></idx>
+            <idx><h>common option</h><h>qrcode</h></idx>
+            <idx><h>qrcode</h><h>common option</h></idx>
 
             <p>
                 QR codes for videos and interactive elements can have an image placed in their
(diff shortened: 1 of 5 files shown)


rbeezer:
Really nice!  Including using the new publisher variable Python routine, and fixing some typos.  ;-)

Yes, lets just view this as a "correction", not a "deprecation".

Would you be willing to make a -announce post?  You could quote the original announcement, explain the discrepancy, and say that "qrcode" won out.  Probably not very important to do this, so don't sweat it.  Or tell me to do it.

Thanks for picking this off while it was fresh.

Rob

rbeezer:
I've got a warning all figured out.  I'll add once at a keyboard.

rbeezer:
Appears as a deprecation warning, but that's OK.  See c599da56dfb9e18e84c381efeacd8ce04337d67a

PR #2200, opened 2024-07-07
LaTeX: restore parskip as well as parindent for parboxes

Alex-Jordan:
There is already a mechanism to restore the ambient `parindent` to when a `parbox` (or -like?) environment is used. This extends that to also restore the `parskip`.

For example, my projects tend to have zero `parindent` but then have a 0.5pc `parskip` to identify paragraph separation.


Alex-Jordan:
Yes, but I thought that was the situation with `\parindent` too. I didn't find that PTX sets it explicitly, yet this mechanism is here to recover it when needed. But are you saying this is only there because parboxes zero that out?

rbeezer:
> But are you saying this is only there because parboxes zero that out?

Yes, but `tcolorbox` really (at least how I think of it) .  And `parindent` is perhaps only to support customizations.

Merged!  And I added a follow-on commit to explain the reasons for these (I think).

PR #2201, opened 2024-07-08
Add tests to program schema

ascholerChemeketa:
I noticed that `<tests>` was missing from `<program>`.

Not sure if that should change the notation or implementation for `<input>`. `<input>` is only really optional if there are no `<tests>` in the program.

### DIFF
--- a/schema/pretext.xml
+++ b/schema/pretext.xml
@@ -1440,9 +1440,9 @@
                     attribute highlight-lines {text}?,
                     attribute interactive {"codelens"}?,
                     (
-                        element input {text} | text
-                    )
-                    element tests {text}?
+                        element input {text},
+                        element tests {text}?
+                    ) | text
                 }
             </code>
         </fragment>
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
@rbeezer Yell if you want me to add some kind of warning/error on a program with no input that has tests. If so I'll append it to this.

rbeezer:
This needed some changes, it had syntax errors (the pipe and the parentheses-grouping were mixed up, I think).  See corrected corrected version.  ;-)

Really appreciate the help with the schema, but perhaps we should get you set up with some of the tools to test contributions as they are created.

ascholerChemeketa:
Ahhh, looks like I was missing ( ) around my construct.

Yes, being able to test would likely increase my odds of offering helpful commits. :)

Installed what `schema/build.sh` appears to need but now getting assorted java errors - will bug you at a drop in.

PR #2204, opened 2024-07-10
Parsons runnable

ascholerChemeketa:
Complete, but likely in need of some cleanup based on @rbeezer advice.

Goes with https://github.com/RunestoneInteractive/rs/pull/500

Known questionable features:
* Schema attempt
* Modifications to text utils. They are in as separate commits. 
  * One fixes an issue I was running into where if you called count-length on an empty string it recursed forever.
  * Other is more drastic - it allows you to exempt the start or end of a string from newline removal in sanitize-text. Intent is to keep authored blank lines like this:
    ```
    <element-containing-text that is sanitized with @preserve-start>
  
      I clearly want 1 newline
    ```
    while removing the first one that you get by putting your content on a new line:
    ```
    <element-containing-text that is sanitized with @preserve-start>
      I do not want a newline
    ```
    The implementation does cause issues if you try to use it and you formatted your pretext with no newline for the content. If `preserve-start` is applied to this it will actually remove the first line:
    ```
    <element-containing-text that is sanitized with @preserve-start>This will go away because I format code like an animal
    ```
* `program-preamble` and `program-postamble` used to be inside `program`. But that meant I couldn't reuse the `Program` schema, so I moved them out of the program. Not sure if that is dumb or a good example of "let the schema guide you".
* Example uses a datafile, which is maybe wasteful. I wanted to make it clear you could use all the normal activecode attrs like datafile.


rbeezer:
First commit seems fine.

Second commit: checkout to this point, build sample article as HTML before and after checkout.  Lots of stuff goes away.  "Extra" author-provided LaTeX macros in almost every file.  But also huge chunks of code (entirely).   I cannot quickly see where the problem lies.

Suggestions:

* Can the runnable Parsons go in now, and fine-tuning blank lines be a seperate PR/process?  Maybe put recursion-halt on its own, too?
* I'd think some (much?) of this logic should go in the `trim-start-lines` and `trim-end` templates, not all in `santize-text`.  Keep the templates modular and on point.  You'll need to pass along the `preserve-` variables, which is unsightly and laborious, but right now I see that as the cost of keeping the logic itself clean.

ascholerChemeketa:
OK, sorry about that, I thought I was getting a clean diff.

I reduced this to only have the Parsons changes. Will do the text manipulation changes to enable blank lines as a separate PR.

ascholerChemeketa:
Ignore the scarry series of force-pushes. I had rebased onto https://github.com/PreTeXtBook/pretext/pull/2197 while testing and pushed that. Just rebased back to head of main.

rbeezer:
This was finished off at #2236.

PR #2211, opened 2024-07-18
name anonymously duplicate elements

### LINKED ISSUE #1993: update author guide for visual attribute
https://pretextbook.org/doc/guide/html/topic-url.html still says 

Compare `<url href="https://example.com/">Demo Site</url>`(https://pretextbook.org/doc/guide/html/topic-url.html#p-763)which will render ...

but [pretext-announce](https://groups.google.com/g/pretext-announce/c/cTrDYzwBIqU/m/Px9V0kKnAgAJ) says that the visual is now mandatory, and I get plenty of 

> error: text not allowed here; expected the element end-tag

because of this.  I presume the Author guide should explicitly disallow those, correct?

oscarlevin:
This removes any duplicate anonymously named elements in service of the new reference documentation workflow.  

Additionally, we clean up the schema for references, which will close #1993.

Finally, the new naming suggests an improvement for how to build the experimental schema, which is demonstrated in Frontmatter section.  A later PR will do similar improvements for the other experimental sections.

### DIFF
--- a/schema/pretext.xml
+++ b/schema/pretext.xml
@@ -70,7 +70,7 @@
         <fragment xml:id="start-elements">
             <title>Start elements</title>
             <code>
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, @oscarlevin, et alia!

For the record, first commit gets `url` right and eliminates several (false) errors when validating sample article.

Second commit resutls in no change in validation output for the sample article.

PR #2214, opened 2024-07-18
HTML: Geogebra slate to match material dimensions

dbrianwalton:
This PR implements ability for Geogebra applet  to specify dimensions for material-width and material-height that are internal dimensions of the slate used by the applet.

Based on trying to resolve the discussion: https://groups.google.com/g/pretext-support/c/ETIbKs5h9I8

I have not yet revised any documentation, and am wondering if `#interactive[@geogebra="material-id"]` should be revised to use the applet/slate approach if material dimensions are provided. (I don't know if there is a good reason to preserve the URL approach.)

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -9657,7 +9657,14 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         </xsl:apply-templates>
     </xsl:variable>
     <xsl:variable name="ggbMaterialHeight">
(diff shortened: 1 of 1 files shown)


rbeezer:
Note to self:  once ready, edit `@srcWidth` to `@material-width`

rbeezer:
Merged as-is.  With a correction that may have been incorrect for a long, long time, in a follow-on commit.

Not sure about URL/applet and interactice/slate.

And not sure about heights and aspect ratios.

Are you able to suggest in the Google Group what Anna should use?  She's right in front of me at the workshop, so I can help her.

Thanks!

PR #2215, opened 2024-07-18
Fitb update and documentation

dbrianwalton:
Prepared documentation for the fill-in-the-blank exercise. While doing that work realized some modifications to the proposed markup would be beneficial, so made changes to that as proceeded.

Includes implementation of the automated solution for fillin problems based on given answers.   Feedback elements are assumed to be structured for now — need to add ability to identify text that needs to be surrounded by <p/> and this requires additional conversation.

Also note that the math object javascript requires Runestone services version that includes Runestone PR #504, which was merged July 16.


bnmnetp:
@dbrianwalton For Rob and I the canonical place to test these things for Runestone is in the sample book.  So adding some of these in the sample book would be great so we can ensure they work as well in that environment as in the plain html.

rbeezer:
Merged.  Thanks.

(Had to fix a `test="solution[@include-automatic='no'"` missing a bracket.)

PR #2217, opened 2024-07-18
Information about LAD Custom Publishing

kcrisman:
This is another print on demand service.  It operates a little differently from the other big options, in ways that might make it more desirable for some authors (and probably less so for others).

### DIFF
--- a/doc/guide/publisher/print-on-demand.xml
+++ b/doc/guide/publisher/print-on-demand.xml
@@ -40,7 +40,7 @@
         </li>
     </dl></p>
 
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, KDC!  Was not aware of this one.  Merged as-is.

kcrisman:
> Thanks, KDC! Was not aware of this one. Merged as-is.

I've used them before for course packs, and it turns out they were helpful with this too (though I haven't sold many yet!).  Having a phone option is nice.

PR #2218, opened 2024-07-19
add vertical space after sidebyside

sean-fitzpatrick:
This adds a 10pt gap after a sidebyside.

In some environments the current setting looks OK, because there is indentation of the paragraph after a sidebyside.
But in `example` (and perhaps some other environments), if you have a `sidebyside` containing two figures with captions, there is no identation of the following paragraph, and it is hard to tell where the caption ends and the paragraph begins.

Adding the gap makes it clear where the sidebyside ends and the following paragraph begins.


sean-fitzpatrick:
The lack of indentation was my fault, but it still looks better (to my eyes) with the extra separation between the sidebyside and the following text.

rbeezer:
In!  Thanks.

PR #2219, opened 2024-07-19
@workspace fix

mitchkeller:
`pretext-latex.xsl` currently only defines the length `\workspacestrutwidth` when the file contains `exercise[@workspace]`. To support `@workspace` living on other possible `worksheet` descendants, this changes to define the length any time `@workspace` occurs in the document. Two rudimentary worksheets are added to the sample article: one with two `activity` elements each with `@workspace` and one where `@workspace` only appears on `activity/task`.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -14280,6 +14280,61 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     </sidebyside>
                 </page>
             </worksheet>
+            <worksheet label="worksheet-activity-no-task">
+              <page>
+                <activity workspace="1in">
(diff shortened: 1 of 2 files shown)


mitchkeller:
I will add that the `worksheet` I put in don't actually assist with testing things here as part of the full sample article. The sample article has `exercise[@workspace]` already, so `\workspacestrutwidth` is being defined. The test worksheets on their own will fail to build to PDF without [cebec80](https://github.com/PreTeXtBook/pretext/pull/2219/commits/cebec80dcfdfb83867ba82aec5349f36586c8d40).

rbeezer:
Nice commit messages - you get a rare true merge.  ;-)

Thanks, these will be useful.

PR #2220, opened 2024-07-19
FITB: Add examples to sample-book

dbrianwalton:
I noticed @bnmnetp Brad's comment about examples in the sample-book and added a variety of different examples. The old examples are still there.

I was partially through an update when I saw the last PR was accepted, so I hope my merging of the master in the commit history isn't problematic. Checking the diff of this branch with master has a few lines that it claims are different but I really can't see what is different. (Maybe just a different history of which commit they changed on because of my merge?)


dbrianwalton:
See PR#2224, which includes these changes but additional and is rebased off recent master.

PR #2223, opened 2024-07-19
Fix deprecations in sample-slideshow

sean-fitzpatrick:
Building the sample slideshow results in two deprecation warnings:
1. The `minified` option is obsolete
2. Lists were still using `label` instead of `marker`

This fixes both deprecations.


sean-fitzpatrick:
I've added images to the slideshow. But to build successfully we are going to need managed directories. I will make those changes, but I'll hold off on pushing until Rob signs off on the change.

rbeezer:
Thanks for these improvements, Sean.  I'm trying to keep up.

*  Tabs in two different commits.  This literally **tripled** the time it took me to get this reviewed and cleaned up.
* The CLI is not part of development workflow.  So both images are missing and are not going to be generated.  Error messages below. 
* Merged locally since it was so painful to get rid of the tabs.  I'll wait a bit before pushing.
* Please make a **new commit** with the images as soon as possible so I can close this.

Thanks!

sean-fitzpatrick:
Sorry about the tabs! I think I must need to update the VSCode extension. 

I've got 30 seconds of internet left before takeoff. How do you want directories structured? Do you want the generated images generated and committed? 
(I guess maybe, since pretext/pretext doesn't generate images yet for slides)

rbeezer:
Yes, generate and commit in hte "usual" places.

rbeezer:
"Images can be included from external sources by placing them in the assets folder."

We don't assume the CLI, so there may not be an `assets` folder.

rbeezer:
Yes, please start a new PR (which is really what I meant to say before).

> someone will need to review and edit my hacking attempt

Expecting me to edit is not always the most productive approach.  Work to minimize this.

Best to assume managed directories for the Python.  Actual images can come before and be adjusted later.  Right now the sample has missing images.

PR #2224, opened 2024-07-20
FITB: Combined addition of examples to sample-book and expand use to tasks and project-like

dbrianwalton:
This PR should be easier to compare since it was rebased from a more recent master. I hope it is okay that I combined two different pieces of the upgrade into one request.


rbeezer:
Thanks for continuing to evolve this!  Merged as-is.

Doing an interactive rebase with an edit on each commit can get clean-up changes into the commits where they originated.  Its a nice technique to have in your arsenal.  I can walk you through iit at a Drop-In sometime.

I think the static version could use some more work.  Some double minus-signs, some raw LaTeX bleeding though, plus a cubic polynomial that is just a constant 2 (just looks suspect to me).

PR #2225, opened 2024-07-20
Slideshow styling update

ascholerChemeketa:
Fixes to revealjs styling.

Changed the container for `sagecell-sage` elements.  Having them as `pre` was causing all kinds of slideshow styling issues. The [sage examples](https://sagecell.sagemath.org/static/about.html?v=0141c2bee781881e735b5e98a913a0de5f2b2efc011158c7627de0318c02f6cc569f07beefc2eca4bdba7e55ec59235a4ba78ce37797ae52d0978386182bd413) show a div for that container. Tested this against sample-article - does not appear to adversely affect HTML builds.

Added Prism highlighting for code and a code sample.

Pushed this PR independent of ongoing CSS bundling as I know others are working on slides. Bundling will slightly change how the CSS gets written but can be done independently of this.

I think this is ready, but happy to tweak anything.

### DIFF
--- a/examples/sample-slideshow/sample-slideshow.xml
+++ b/examples/sample-slideshow/sample-slideshow.xml
@@ -181,6 +181,10 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                     </li>
                 </dl></p>
             </slide>
+        </section>
+
+        <section>
+            <title>Blocks and code</title>
 
             <slide>
                 <title>Blocks: <tag>definition</tag></title>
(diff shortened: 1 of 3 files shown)


sean-fitzpatrick:
I have some image examples ready for you to use in testing, but the PR is still in progress.

Perhaps the easiest way to test image generation would be to use the CLI with `pretext-new-slideshow` and then copy the contents of `sample-slideshow.xml` in #2223 into `main.ptx`, and do `pretext build web`.

There's some weirdness with Asymptote that needs to be figured out.

ascholerChemeketa:
@sean-fitzpatrick I added some rules that should target the Asymptote images.

@rbeezer The more features that get added that require duplicating code from the HTML stylesheets, the more it becomes clear that reuse via bundling is going to make this cleaner. Styles will need to move to a separate CSS file or if there is a strong desire for a one file solution, we would have to add something to dump the content of a known CSS file into a `<style>`  element in the document.

sean-fitzpatrick:
Thanks! I made some progress on getting pretext/pretext to build slides but I'm missing something since I still don't get an output file. 

Images are copied correctly and the reveal stylesheet is applied, but I'm not getting an html file. 

My flight connection is a little shorter than I thought so I don't think I'll figure it out today

rbeezer:
Thanks very much!  I tend to remove the "extended" commits in the commit message - if the rationale/commentary is important I think it is better as code comments to be read in context.

rbeezer:
Forgot to say: yes, moving into a file is fine!  Whatever you think is best.

PR #2226, opened 2024-07-21
Handle pub variable with no value

ascholerChemeketa:
If a pub variable results in an empty string the get_publisher_varaible function chokes. 

This provides `None` for missing values. Defaulting to `""` might require less code for handling but also be more likely to result in sneaky errors based on assuming a non-empty string came back.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -4524,7 +4524,7 @@ def get_publisher_variable(xml_source, pub_file, params, variable):
     with open(temp_file, 'r') as f:
         for line in f:
             parts = line.split()
-            pairs[parts[0]] =  parts[1]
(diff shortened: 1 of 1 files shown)


rbeezer:
Nice catch!  Had not thought that through.

I've spent a long time on this one.  Mostly because there is an existing test on the return value being an empty string.

I agree with everything you have said.  But I think I want to keep this function true to what it says it does - the publisher variable may indeed be an empty string, and in XSL-land that is frequently used for the same sorts of things `None` is.  (And this is not simply to avoid changes to behavior or employment.)  So I have addressed the bug, by returning an empty string.

I've kept your name on this - I hope that is OK.

PR #2227, opened 2024-07-21
Add description of the @attachment  for shortanswer

bnmnetp:


rbeezer:
Thanks for getting this in!

* Consolidated to one commit.  I think the first is already in the code, and I don't duplicate the merges.
* Corrected some grammar (capitalization mostly).
* The `tage` element makes an empty element, switched to `tag`.
* Added to the schema!  ;-)  On a new commit.

PR #2228, opened 2024-07-22
Slideshow: Add devscript support

sean-fitzpatrick:
Changes to `pretext/pretext` and `pretext/pretext.py` to add support for reveal.js slides.

It is working for me, but my Python is not great, so another set of eyes is needed.

Note that I didn't add a reveal-zip target. I'm not sure if that would be wanted.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -3635,7 +3635,50 @@ def html(
             log.info("zip file of HTML output deposited as {}".format(derivedname))
     else:
         raise ValueError("PTX:BUG: HTML file format not recognized")
+def revealjs(
+    xml, pub_file, stringparams, xmlid_root, file_format, extra_xsl, out_file, dest_dir
+):
+    """Convert XML source to reveal.js HTML file"""
 
+    # to ensure provided stringparams aren't mutated unintentionally
+    stringparams = stringparams.copy()
+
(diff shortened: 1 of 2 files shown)


pretext/pretext.py
```diff
@@ -3635,7 +3635,50 @@ def html(
...
+def revealjs(
+    xml, pub_file, stringparams, xmlid_root, file_format, extra_xsl, out_file, dest_dir
+):
+    """Convert XML source to reveal.js HTML file"""
+
+    # to ensure provided stringparams aren't mutated unintentionally
+    stringparams = stringparams.copy()
 
+    # Consult publisher file for locations of images
+    generated_abs, external_abs = get_managed_directories(xml, pub_file)
```
kcrisman, line 3647: This line will make me very happy 😄

rbeezer:
Thanks, Sean!  Very close.  Needed to build the outpt file name (`derivedname`) for input to the `xsltproc()` call.  I switched that up.

Some stuff near the end was then not really necessary until we have an option to zip everything together, especiually in hte case where we want to copy over the necessary resources (the actual Reveal JS and CSS).  So I got rid of that for now.  Perhaps something similar will come back.

See new issue, https://github.com/PreTeXtBook/pretext/issues/2235

Issue #1890, opened 2023-01-02
Allow slide numbering in revealjs


kcrisman:
In the place where the revealjs conversion initializes itself and things like the navigation mode, it would be pretty easy to allow slide numbers as well, e.g. after https://github.com/PreTeXtBook/pretext/blob/6e1ee142defa0885eabfcd4d36a9f19eee91f44e/xsl/pretext-revealjs.xsl#L294 one could add

```
<xsl:text>  slideNumber: true,&#xa;</xsl:text>
```

Most likely this would be better to do as a publisher switch or author option than to hard-code it in.  However, it's certainly quite useful to have around!

kcrisman:
See #2231 .  I think it's ready, but it's possible there will be discussion about which element it should be an attribute of, etc.

PR #2231, opened 2024-07-22
Add slide number option to revealjs

kcrisman:
This adds an option to use the revealjs slide numbering.  Note that with the default "grid" option this creates ordered pairs, so I didn't make it the default option; I don't mind the ordered pairs personally, but I figure others might find them distracting.  With "linear" slides it would be fine for anyone, I think.

### DIFF
--- a/doc/guide/publisher/publication-file.xml
+++ b/doc/guide/publisher/publication-file.xml
@@ -740,8 +740,9 @@
 
             <p>The<cd>
                 <cline>/publication/revealjs/appearance</cline>
-            </cd>element can have the following attribute:<ul>
+            </cd>element can have the following attributes:<ul>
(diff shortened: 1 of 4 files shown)


kcrisman:
@ascholerChemeketa @sean-fitzpatrick you may be interested in this potential option.

rbeezer:
PreTeXt controls numbering, not the output formats.  Necessary for consistency across outputs and for when a subset is built.  The exception is about half of the laTex, since its behavior is the model.

kcrisman:
> PreTeXt controls numbering, not the output formats. 

Perhaps there is a misunderstanding about what this enables.  There is no connection whatsoever to numbering of theorems etc.; this simply enables (as an option) the tiny numbers that every slide deck format I know of supports.  This is a very heavily used option outside of PreTeXt (including Beamer), and a standard option for revealjs, so it seemed like a reasonable thing to support *as a non-default option*, especially since no one commented otherwise on #1890.

Issue #1907, opened 2023-01-18
Geogebra interactives communication with slates.


whaynes:
The attached files show how I changed the geogebra template to allow communications with another slate. 

It required loading the geogebra applet as a promise, and supplying a way to add a javascript listener function.   Also this demonstrates an alternate display for the interactive's instructions as an aside, which is nicely hidden until the reader needs it.

I hope that this is useful, and that this ability can be built into the core xslt.

Thanks, 
Will Haynes

[ggb-demo.zip](https://github.com/PreTeXtBook/pretext/files/10449115/ggb-demo.zip)

rbeezer:
Closing as implemented. The GeoGebra applet now loads as a `Promise` (the `GGBApplet` construction in `pretext-html.xsl`), the mechanism demonstrated here for communicating with another slate; #2234 added a GeoGebra promise example to the sample article. Thanks @whaynes.

*Claude Opus 4.8 (1M context), acting as a review assistant for Rob Beezer*

PR #2234, opened 2024-07-24
Geogebra promise

whaynes:
This loads geogebra applets as javascript promise to permit communication with slates.  

There's a call to a function with a hard coded name 'listeners()' on line 10003, which may not be the best or right way to do this.

I also have added an example to the sample article.  

This addresses an issue I made last year:  https://github.com/PreTeXtBook/pretext/issues/1907


rbeezer:
Really nice!  And thanks for the great example for the sample article.

Only one change of consequence: I renamed the directory with your JS to use your last name.  A bit easier for me to stay organized with multiple contributors.

Had to remove some whitespace from the JS and sample article.  I'm on a quest to have all our contributors understand what a problem this can become and what a fiddle it is to clean-up.  Maybe with a bit of an attitude, but here's the explanation:

https://pretextbook.org/doc/guide/html/git-whitespace.html

Thanks again, great stuff!

Rob

whaynes:
Sorry about the whitespace problems.  I struggle with them all the time.   

I have often thought that what is needed is a 'canonical' PreTeXt formatter that can be run from the command line or scripts, so that contributors can run it to tidy up the source in a consistent way.  It should enforce tabs vs spaces, and put newlines before selected tags like sections and paragraphs, but not before inline tags.  I think that VSCode can do some of this, but every user can customize its rules, and it can't be run from the command line.  

Will

PR #2236, opened 2024-07-27
Allow parsons whitespace

ascholerChemeketa:
Broken out of https://github.com/PreTeXtBook/pretext/pull/2204. Based off of that PR as it updates the logic in pretext-runestone to use the new feature. Only the last two commits are new to this PR. If you prefer it based off main HEAD with duplicative commits in pretext-runestone.xsl, let me know.

Moved logic to `trim-start-lines` and `trim-end`. New logic enabled removing the infinite-recursion on empty string modification that the prior attempt had.

Allows `sanitize-text` calls to preserve "intentional" whitespace. Intentional defined as white space that follows the first newline (for preserve start) or proceeds the final one (for preserve end).

Now with clean diffs!


rbeezer:
I tested and merged **just** the one commit expanding the `sanitize-text` template.  I'll pick up the rest here soon - no need to do any rearrangement (I think).

One small edit.  When `=` is an attribute specification, I write it "tight".  When used for a logical test, I put spaces around it.  Makes certain global searches a bit easier/reliable.  So made that change one place.

rbeezer:
Lots of repackaging.  But no major changes.  Updating website right now.  Thanks for this one!

*  Added some code comments about liberalization of `activecode-language` template.
* Three syntax errors in schema ttempt.  :-(

Do we expect to be runnable outside of Runestone servers?  I got the problem right (first try!) but did not see a "run" button.
Screenshot 2024-08-04 at 17-21-50 SB Parsons Exercises

PR #2237, opened 2024-07-28
Ptx slate

whaynes:
This PR creates a #slate to wrap and render PreTeXt content within #interactives.  This gives the resulting HTML an @id so that it can be addressed by javascript.

A couple of thoughts:

1.  I need an @id on an ancestor of the content -- it could be another element like #sidebyside or #interactive, but @labels aren't permitted on #sidebysides, and the #html:div for the #interactive doesn't actually wrap the #sidebyside.
2. `<xsl:apply-templates/>`  will render any content.  This probably should be restricted to a reasonable subset by the schema.   #tabular and #p for sure, but probably others.
3. `<xsl:call-template name="latex-macros"/>`  is needed so that the macros are available in the iframe.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -8284,7 +8284,8 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                         </slate>
 
                         <sidebyside width="35%" margins="10%">
-                            <tabular label="ggb_2__17_table1" halign="right">
+                            <slate surface="ptx" label="ggb_2__17_table1">
+                            <tabular halign="right">
                                 <col width="10%" halign="center"/>
                                 <col width="10%" halign="center"/>
                                 <col width="10%" halign="center"/>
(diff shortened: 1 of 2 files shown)


rbeezer:
@whaynes - thanks for getting this going.  A few changes...

* `latex-macros` really need to be on the entire #ifram page we build.  Had not considered that we would need that in the first versions of `#interactive`.  So I added that as a PreTeXt-wide change.
* I changed the `@surface` to be `pretext`.  Generally I try to save abbreviations for really common items (`p`, `m`).
* Used a template to place the `@id` onto the `#div`.  That'll prefer `@label` but also use `@xml:id`.  Not sure whicvh we should promote for authors to use.
* Expanded the narrative in the sample article, which will remind me of certain aspects when I get around to writing documentation.

Couple of commits with my name - your two still have your name, plus small changes I made.  Thanks again!

whaynes:
Thank you.   I think you can safely remove the label hack at commit 2c466d40 now.

rbeezer:
> you can safely remove

Aah, I'd already forgotten about that.  Thanks for the reminder.  Done at d06e6737b61e9098681b15e9e853125b4c2030c1

PR #2239, opened 2024-07-30
FITB: correct attributes on fillins relating to static representation

dbrianwalton:
Addresses the wrong reference in the fillin that should be a cubic (was referencing the wrong object) and fixes `@mode` for fillins that were supposed to be math so that the LaTeX was within `<m>`.


rbeezer:
Thanks for updating these.  I repackaged this as two commits, one for each sample document.

The "substitutions" file is unchanged?  I re-generated it and formed a diff, but that was not helpful to me.  Seems like the one currently in the repository has been re-formatted?  I get a new version with no line breaks (which is fine, since it is not meant to be human-readable).  But the diff is worthless.

dbrianwalton:
The second PR that deals with ability to do reductions and simplifications
will have some changes to the substitutions.

rbeezer:
Thanks.  Going there now.

PR #2240, opened 2024-08-01
Fitb math reductions

dbrianwalton:
In order to address the issues relating to funny-looking math in the static version, the majority of the issues arose from generating formulas from other formulas. An experienced human would know that simplifications were required after the manipulations, but the code as then implemented just left things as they were. This would give things like x--5 which should be changed to x+5 or x^(5-1) which should change to x^4.

The btm-expressions library that drives the underlying javascript math objects had been designed with reduction rules in mind, but this part of the library had been neglected. I have updated the library and submitted a Runestone pull request to install the updated library. (See https://github.com/RunestoneInteractive/rs/pull/520)

This is the parallel pull request on the PreTeXt side that implements calls to the reduce commands. The sample-article and sample-book examples were adapted to refer to a new `@reduce` attribute and the documentation was also updated with a brief explanation.

The PR also includes the logic for pulling Runestone services to be local following the recent merge for html. However, for identifying a development folder I also refer to a new stringparam debug.rs.dev.folder in combination with debug.rs.dev and the contents of that folder are copied into the tmp_dir's _static folder.

This should not go live until Runestone merges the updated library. Otherwise, PreTeXt will be referencing features that don't currently exist.


dbrianwalton:
@bnmnetp has merged the corresponding Runestone PR. This should be ready now.

rbeezer:
Thanks!  Merged as-is (after a rebase).

I still see a `x+-3` in the LaTeX?

dbrianwalton:
Rob,

I'll hunt down the source of that failure to simplify and update in the
near future.

Thanks,
Brian

PR #2241, opened 2024-08-02
log deprecation messages as warning

StevenClontz:
This change will help users see when they are using deprecated markup.


rbeezer:
Thanks!

PR #2242, opened 2024-08-04
Deprecate metadata elements in a sidebyside

Alex-Jordan:
See lower in thread, starting 8/5/2024.

~For example if a `sidebyside` has two "panel" children, but first there is an `idx`, the count for the actual panel children will be off. This fixes that.~

### DIFF
--- a/xsl/entities.ent
+++ b/xsl/entities.ent
@@ -50,7 +50,7 @@
 <!-- (Include this comment in count!)                                   -->
 
 <!-- typical items in blocks that get special treatment -->
-<!ENTITY METADATA "title|subtitle|caption|index">
+<!ENTITY METADATA "title|subtitle|caption|idx|index">
 <!ENTITY METADATA-FILTER "self::title or self::subtitle or self::caption or self::idx or self::index[not(index-list)]">
 
 <!-- Subdivisions and runs of paragraphs (like in a proof)  -->
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -1300,6 +1300,10 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <!-- so eliminated in this phase             -->
 <xsl:template match="@permid" mode="repair"/>
 
+<!-- 2024-08-05: remove metadata elements from a sidebyside, -->
+<!-- which have not been schema-compliant since circa 2017   -->
+<xsl:template match="sidebyside/*[&METADATA-FILTER;]" mode="repair"/>
+
 <!-- ########### -->
 <!-- Assembly ID -->
 <!-- ########### -->
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -5498,9 +5498,7 @@ Book (with parts), "section" at level 3
 
     <!-- Number of Panels -->
     <!-- count the elements destined for panels  -->
-    <!-- Metadata banned, roughly 2017-07, now pure container  -->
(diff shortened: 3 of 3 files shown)


rbeezer:
The schema does not allow any metadata for a `sidebyside`.  Nor should it.  It is a container for layout purposes and does not have any identifying information (title, number, etc).  I don't think an `idx` should produce an entry in the index that is a knowl that says "sidebyside" - that won't make any sense to a reader, I think.

Alex-Jordan:
Roughly at line 5499 you have:

```
    <!-- Number of Panels -->
    <!-- count the elements destined for panels  -->
    <!-- Metadata banned, roughly 2017-07, now pure container  -->
    <!-- Retain filter for backward compatibility              -->
    <xsl:variable name="number-panels" select="count(*[not(&METADATA-FILTER;)])" />
```

Do you want to continue backward compatibility? It's not doing that if you don't also have the change in this PR. An instance of that in ORCCA is what led to this.

I could change this PR to adjust this block if that's better.

rbeezer:
Well, it has been 7 years...

What do you think about a non-fatal runtime warning (`PTX:ERROR`) if there is metadata?  "Results will be unpredictable"  I'd announce on `-announce`.

Comments in the code about the hiostory could be placed in both places to help those of us with faulty memories.

Then again, there is not a lot of harm in maintaing backward-compatibility.  Except to tolerate bad habits.  Let me think some more.

Alex-Jordan:
I don't really have an opinion about it. For my recent ORCCA PDF build, there are a lot of uncommitted edits to pretext repo files (along with custom XSL in the project itself) that I made to get the PDF to come out in ways that I wanted it to. Still using `pretext/pretext/pretext`, not the CLI.) And now that I'm done, I'm just trying to go through all those things and identify which things would serve the greater good, and offer them up as PRs. Today I saw these two short things (this one and #2243). But I'm happy to withdraw this or change it in whatever way you would prefer.

Larger PRs to come including Tacoma work. But only whatever initially seems like it servers the wider good and not just ORCCA.

rbeezer:
An idea, which is nearly equivalent.  Kill `sidebyside/*[&METADATA-FILTER;]` in the `repair` phase of assembly.  Then code can assume they are not ever present.

Maybe create a deprecation warning?  With identical match.  Dated 7 years ago (but placed with most recent).  Say they are being "ignored".

Not worth an announcement?  Anybody who has these is likely getting odd results anyway?

Alex-Jordan:
> Not worth an announcement? Anybody who has these is likely getting odd results anyway?

So if a person had this situation and tries to build PDF, it's not just odd results that happen—the LaTeX fails to build. What happens (in for example an `idx`, `p`, `image` situation) is that when the second panel is being processed, the XSLT looks for the *3rd* width from the layout parameters. There is no 3rd width, so the tex is written with nothing where the width should be. And then LaTeX compilation fails because there needs to be a number there.

I know not everyone builds PDF.

When I come back to this I'll work it up as you described.

Alex-Jordan:
This PR now deprecates metadata elements in a sidebyside.

- adds `idx` to the `&METADATA;` entity, which was conspicuously missing. Note `&METADATA;` entity is used nowhere before this commit.
- kills `sidebyside/*[&METADATA-FILTER;]` in assembly.
- cleans up corresponding code in -common.
- adds a deprecation warning that is post-dated to 2017.

rbeezer:
Very good!  Wasn't that fun?

*  Thanks for catching the missing `idx`!
*  Broke out the typo fix on its own commit.
* The `count(*)` had two trailing delimiters, so this would not execute.  Fixed it silently.

PR #2245, opened 2024-08-06
LaTeX: avoid losing negative space in multicolumn exercisegroup

Alex-Jordan:
There is an issue that is difficult to reproduce, as it only arises under rare conditions. If you have a multicolumn exercisegroup, and the content of an exercise is a certain width that just barely fills up the available width, it leads to unwanted horizontal and vertical space.

I could try to mock up a PTX example, but I'm hoping that is not needed. This did come up in ORCCA, with out-of-the-box exercisegroup handling. Here is where I posted about this on tex.se. For that post, I took PTX tex output and pared it down to an MWE. You can see the unwanted space in the items where the content is a bit too wide:
https://tex.stackexchange.com/questions/722329/unwanted-space-in-tcbraster

The accepted answer is to replace `\hspace` with `\hspace*`. Here is an explanation of the difference:
https://tex.stackexchange.com/questions/89082/hspace-vs-hspace

What is happening is that when the content is an unfortunate width, it leads to LaTeX killing the negative space that we have here to outdent the number. And `\hspace*` is a construction that refuses to let space like that be eaten up.


rbeezer:
Well, that tex.se post is very interesting reading.  ;-)

Looks like a good change.  I'll likely drop in a code comment to explain the `*` so that nobody removes it later (including this forgetful somebody).

rbeezer:
Great catch.  Added comments on a separate commit, as promised.  Thanks!

PR #2247, opened 2024-08-12
Updated knowl.js for MathJax compatibility of hidden content 

geoff-cox:
The following PR is based on [the suggestions](https://github.com/mathjax/MathJax/issues/3266#issuecomment-2273886865) provided by MathJax.

Removed line 315, which previously called MathJax.typesetPromise() before the knowl's visibility was updated. MathJax could not accurately measure the dimensions of elements while they were still hidden (display: none).

Updated line 60 to delay the call to MathJax.typesetPromise() until after the knowl becomes visible, but before the unrolling animation begins. This ensures that MathJax has the correct layout context for typesetting, preventing layout shifts and improving the smoothness of the animation.

I verified that knowls containing mathematical content is rendered correctly without layout shifts and ensured that the unrolling animation remains smooth and that no visual glitches occur during the transition.

### DIFF
--- a/js_lib/knowl.js
+++ b/js_lib/knowl.js
@@ -57,7 +57,7 @@ class SlideRevealer {
       this.triggerElement.setAttribute("open","");
       this.contentElement.style.display = '';
       // Trigger the animation to expand or collapse the knowl.
-      // We delay the MathJax typesetting until the knowl is visible to ensure proper measurements
+      // Delay the MathJax typesetting until the knowl is visible to ensure proper measurements
       // are taken, but before the unrolling begins. This helps avoid layout shifts and ensures
       // smooth animation with correctly sized content.
       MathJax.typesetPromise().then(() => window.requestAnimationFrame(() => this.toggle(true)));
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, @geoff-cox!  I was just going to catch-up and encourage you to do this.

I'll try to take it for a spin, but input from @davidfarmer and @bnmnetp would be very helpful.

rbeezer:
Took this for a drive and it looks good.  Cross-reference knowls with math seem fine.  Turned on lots of born-hidden knowls for the sample article and things like a Parsons problem with math inside a born-hidden knowl seem to be fine as well.  Would like to have @davidfarmer check-off before merging.

rbeezer:
Pinging @davidfarmer and @bnmnetp for input or check-off.  Thanks.

davidfarmer:
Looks okay.  Sorry it took so long.

bnmnetp (commented):
Looks good to me.  This is a standard trick to have some animation control.

rbeezer:
Merged, with code as-is.  Consolidated to a single commit with a one-line commit message.

Thanks for fixing this up - seems to be working great.  Sorry for the delay, this one needed extra review (which I lost track of).

PR #2249, opened 2024-08-20
Fix typo resulting in missing example titles

### LINKED ISSUE #2248: `EXAMPLE-LIKE` titles missing in latex
At some commit between CLI version 2.5 and 2.6, the styling for example-like blocks in latex changed resulting in the title not showing up.

oscarlevin:
A relatively recent fix to the latex tcb styles included a missing "app" on line 3332 of `pretext-latex.xsl` which results in the title of an example-like not appearing in pdfs.

Will close #2248


rbeezer:
Merged as-is online.  Thanks for making the PR.  Please advertise?  I'll be offline for a while.  Thanks!

PR #2251, opened 2024-09-06
Remove extraneous line in documentation

mafitch:
Documentation had an error: extraneous line showed up in appendices that blocked compilation.


mafitch:
Also added a fix suggested by David Lippman to address MOM images having a width="102%" or similar issue.

rbeezer:
Thanks, @mafitch!  Sorry I was not able to help with this one until now.

Combined min/max fix, so down to just two commits, and edited commit messages (as I do).  So you will need to pull to get changes and should delete your original branch proposing changes (commit hashes have changed).

More commentary on the original thread...

PR #2253, opened 2024-09-10
Force web requests during build to timeout after 10 seconds

ascholerChemeketa:
To address issue encountered when RS servers were not responding.


rbeezer:
Thanks very much!  I'd thought to do this but had not hopped on it so quickly.

PR #2254, opened 2024-09-11
Slideshow: ensure svg in sidebyside uses full panel width

sean-fitzpatrick:
This updates the `pretext-revealjs.xsl` stylesheet to write an extra CSS rule into the HTML file produced for a slideshow.

For SVG images within a sidebyside (which will include any generated images from latex-image) this ensures that the SVG width is equal to the width of the sidebyside panel containing it.

Credit to @ascholerChemeketa for the CSS rule.
One question for @rbeezer : do we want this to apply to all image formats, and not just SVG? If so, it is an easy edit to remove the filter in the rule that restricts it to SVG.

### DIFF
--- a/xsl/pretext-revealjs.xsl
+++ b/xsl/pretext-revealjs.xsl
@@ -243,6 +243,10 @@ dfn {
     float: left;
     padding-right:2em;
 }
+
+div[style*="display:table-cell"] img[src$=".svg"] {
+    width: 100%;
+}
           </style>
 
         </head>


rbeezer:
Prepared to merge this.

> do we want this to apply to all image formats, and not just SVG?

I'd think so.  I can edit - like this?

`img[src$=".svg"]` -> `img`

Would this be better/easier if I first put a `sidebyside` class on the overall output element?  And the rule would look more semantic?

sean-fitzpatrick:
My understanding from @ascholerChemeketa is that this is the correct edit if you want all image formats to scale to the sbs width. 

By sidebyside class, do you mean a CSS class? When we discussed this at drop-in, we did note that slides need a lot more CSS attention, to add more/better classes. 

I think this would be a good idea, but it's hard for me to say how much work this would be.

rbeezer:
Thanks, @sean-fitzpatrick!

1.  Removed condition on SVG, so generalized to all images.

2.  Added CSS comments to explain the purpose of the rule.

3.  Added a `.sidebyside` class on the overall `div` just like HTML output has.  Maybe rule can be adjusted and my comments (2) are no longer necessary?

Shouldn't we have an example in the sample slideshow using a small image so that this stays fixed?

sean-fitzpatrick:
I can add an example. Perhaps not tonight though. Long week...

sean-fitzpatrick:
@rbeezer if I am adding an example to the sample slideshow, do you have a preference between an image from a source file, and a generated image?

rbeezer:
I don't use the CLI in testing, so a `@source` image of a known/concrete size would be most useful.  If we need/want to test generated images, we can do something explicit for that.

sean-fitzpatrick:
There are some images already in the sample slideshow, that I added in Ohio.

One of those was a sidebyside with an SVG generated from `latex-image`, and a WebGL generated from `asymptote`.

But now I'm more puzzled, because the two sidebyside panels there don't seem to be taking the width they're assigned.

PR #2256, opened 2024-09-12
LaTeX: fix space after division exercise and glossary item titles

Alex-Jordan:
For me with LaTeX 2023, working from the master PTX branch, I get things like here:

Screenshot 2024-09-12 at 4 50 14 PM

where there is no space between the title and the bodytext. I do not see this happening at [the canonical example at pretextbook.org](https://pretextbook.org/examples/webwork/sample-chapter/sample-chapter.pdf#page=13). So I guess it is something about the LaTeX distribution.

The change I am making here addresses the issue for me, and seems semantically correct. Instead of having the space be in the title, the space is moved to the `after title` key. I reviewed structures like these and only found three instances for flavors of divisional exercise and for glossary items. Then I get like:

Screenshot 2024-09-12 at 4 51 16 PM


rbeezer:
OK, change makes good sense, and seems fine with sample article using **my** (older) LaTeX installation.

Very close to doing the Ubuntu LTS upgrades on the three machines I support, that's an every two-year process.  So perhaps my TeX installation will be more modern very soon.  Thanks for getting out in front of this one!

PR #2258, opened 2024-09-18
Updated MyOpenMath documentation in Guide and sample-article

mafitch:
Added details about MOM problems that may be problematic. Added use of parameters in sample-article.

### DIFF
--- a/doc/guide/author/overview.xml
+++ b/doc/guide/author/overview.xml
@@ -540,14 +540,14 @@
         <title>MyOpenMath Exercises</title>
         <idx>myopenmath exercise</idx>
 
-        <p>MyOpenMath is a hosted online homework system with tens of thousands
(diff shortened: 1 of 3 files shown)


rbeezer:
Thanks, @mafitch, that'll be a big help.  And it is always desirable (required?) to have tests in the sample article of new features (`@params`).  (I didn't know that had been added!)

*  Repackaged as two commits, as I do.
*  Tried regenerating the changed problem for the static versions in the repository (even if I expected no real change).  I don't use the CLI in development, so don't regenenerate automatically.  See `pretext-dev`.
*  Maybe you want your commits attributed with a more long-term email address, not `Mark Fitch <58183892+mafitch@users.noreply.github.com>`?
*  Will update Guide online soon, but not immediately.

Thanks again, looking forward to more!

PR #2259, opened 2024-09-19
Change p.intertext into div.para.intertext

ascholerChemeketa:
It looks like the `<p>` generated for intertext paragraphs missed the shift to `div.para`.


rbeezer:
Thanks!  That's a great catch.  `intertext` is sort of a weird thing, modeled after LaTeX...

PR #2260, opened 2024-09-25
Cline area text fix

ascholerChemeketa:
Possible fix for the issue raised on discussion list of quotes in cline/areas getting "smartified"


ascholerChemeketa:
Fix works in web. PDF output appears fine with or without patch... I think static version already is fine.

PR #2261, opened 2024-09-25
Docinfo default programming language

ascholerChemeketa:
This should allow authors to specify a default value for `@language` on things like programs and Runestone interactives.

Added the setting to docinfo on the sample-book and sample-article, but commented out in sample-article (where there are more pseudocode samples we don't want to assume are python).

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -122,6 +122,9 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
             <!-- Expects favicon/favicon-32x32.png, and favicon/favicon-16x16.png,       -->
             <!-- possibly in external images directory, when specified in publisher file -->
             <favicon/>
+            <!-- If provided, the default programming-language will be used for -->
+            <!-- any programsor runestone interactives that lack @language      -->
+            <!-- <defaults programming-language="python"/> -->
         </html>
 
         <!--
--- a/examples/sample-book/bookinfo.xml
+++ b/examples/sample-book/bookinfo.xml
@@ -72,5 +72,11 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     \usetikzlibrary{arrows,matrix}
     \usetikzlibrary{snakes}
     </latex-image-preamble>
+    
+    <html>
+      <!-- Any program or Runestone interactive without a @language will -->
+      <!-- assumed to be python                                          -->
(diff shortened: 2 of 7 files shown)


bnmnetp:
@ascholerChemeketa -- it would be ideal if this was written to the manifest, so it can be stored in the database so that the scratch activecode on the instructor and assignment pages use the default.

ascholerChemeketa:
@bnmnetp There is a pubfile entry that controls the scratch activecode. It feels like that is what should get written to the manifest for determination of instructor/assignment page behavior. 

If you agree, I can do that as a separate PR.

bnmnetp:
OK, that makes sense.

ascholerChemeketa:
Did a grep for all @languages in `pretext-runestone` and `runetsone-static` modified them to use the `active-language` template. 

With one exception in `runestone-static`:
```
<xsl:template match="areas" mode="static-areas">
    <xsl:choose>
        <xsl:when test="cline">
            <!-- code, so make a "program" structure -->
            <program>
                <xsl:copy-of select="@language"/>
                <input>
                    <xsl:apply-templates select="cline" mode="static-areas"/>
                </input>
            </program>
        </xsl:when>
        <xsl:otherwise>
            <!-- regular text, this will match a default template later -->
            <xsl:apply-templates select="*" mode="static-areas"/>
        </xsl:otherwise>
    </xsl:choose>
</xsl:template>
```

I'm not sure I understand that template well enough to pick the right action... let @langauge slide through, or try to look up active-language.
Any advice @rbeezer ?

rbeezer:
Something is not quite right with languages and Parson problems.  Huge before/after diff with the sample book.  Perhaps due to
* `natural` being a language
* And/or `not(@language)` being different than `($active-language = '')`
Or something like that.  Not sure the "active-language" template is meant to return an empty string?  Or maybe I changed some behavior with a switch away from named templates?

It was a bit of work to peel this off this PR.  Can you either (a) tell me a small fix I can make and meld in, or (b) put this on a new PR?

Thanks!

ascholerChemeketa:
OK, will review the `active-language` and its interaction with `natural` and do a new PR with that part. You can close this one.

BTW - according to the Parsons crew at SIGCSE, they are `Parsons` problems, even in the singular. No apostrophe.

rbeezer:
Yes, I got this straight once but couldn't remember.  Are they problems, questions, or exercises?  ;-)

Thanks for the work to get the default language in place.

PR #2262, opened 2024-09-26
PreFigure additions to sample article and prefigure-conversion

davidaustinm:
This PR adds two prefigures to the sample-article, one with annotations.  It also adds the prefigure executable into the function prefigure-conversion and copies the output into the destination directory.  This assumes the user has added the line "prefigure = prefig" to pretext/user/pretext.cfg so perhaps this could be incorporated into pretext/pretext/pretext.cfg.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -2980,15 +2980,64 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 <p>There is nothing useful here just yet, testing preparations for integration of PreFigure images.</p>
 
                 <image>
-                    <prefigure label="one">
-                        <foo/>
-                    </prefigure>
+                  <prefigure label="one">
+		    <diagram dimensions="(300,300)" margins="5">
+		      <definition>f(t,y) = (y[1], -pi*y[0])</definition>
+		      <coordinates bbox="[-1,-3,6,3]">
+			<grid-axes xlabel="t"/>
+			<de-solve function="f" t0="0" t1="bbox[2]"
+				  y0="(0,2)" name="oscillator"
+				  N="200"/>
+			<plot-de-solution at="x" solution="oscillator"
+					  axes="(t,y0)" />
+			<plot-de-solution at="xprime" solution="oscillator"
+					  axes="(t,y1)" stroke="red"
+					  tactile-dash="9 9"/>
+			<legend at="legend" anchor="(bbox[2], bbox[3])"
+				alignment="sw" scale="0.9" opacity="0.5">
+			  <item ref="x"><m>x(t)</m></item>
+			  <item ref="xprime"><m>x'(t)</m></item>
+			</legend>
+		      </coordinates>
+		    </diagram>
+                  </prefigure>
                 </image>
 
                 <image>
-                    <prefigure label="two">
-                        <bar/>
-                    </prefigure>
+                  <prefigure label="two">
+		    <diagram dimensions="(300,180)" margins="5">
+		      <coordinates bbox="(0,0,10,6)">
+			<define-shapes>
+			  <circle at="A" center="(4,3)" radius="2"/>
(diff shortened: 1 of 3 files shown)


rbeezer:
The `"source"` stanza is probably OK as-is.  How about initiating a new `"svg"` stanza (as an `elif`) with the additions you are making?

(I'll be able to smash the commits together, so just push a new one on top of the two on your branch already).

davidaustinm:
pretext_conversion now imports prefig, which can be installed with "pip install prefig"

rbeezer:
Built SVG and PDF seemingly with no problem.  Did not attempt to explore results with  diagcess.html.

Could not install the liblouis Python bindings in my virtual environment.  Just gave it a naive go, perhaps I will ask on the `liblouis` list.

Was/Is `rsvg-convert` necessary to build PDF?  I thought I saw that somewhere.  It is already on my system (imagine that) so I didn't do anything extra.

*  When building SVG, seems that an annotations file is built for the tangent line diagram. Named `two.xml`.  Now, if I elect to build source files (into the same directory), a new `two.xml` clobbers the annotations.  I'd do something like `two-annotations.xml`, but I'd guess that means some work on your end independent of just this problem.

*  Is `tactile` a better format name than just `braille`?  It is more than just braille, no?

*  The  pretext/pretext  script should already have most of what you need to make PNG (from PDF by using PyMuPDF?).  See about line 664 (with this PR applied) where that happens in the `latex_image` drill.  Not sure if we make PNG straight from SVG or not.  (We need PNG for EPUB that go to Kindle.)

*  Made new `@label` that will be a bit easier to track.

*  Tabs in the two new examples.  I converted them to spaces.  Now the indentation is a mess and needs to be straightened out.  Hint, hint.

*  I'll see about at least getting the right thing going to include SVG into HTML output.  Might be nearly trivial.

*  I get an error that the `louis` object is not defined when I ask for braille (see below).  Can't tell if there is a error about not loading the module that I'm not seeing, and a better message is needed?  Or?  Guess that depends on the whole virtual environment bit.



```console
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 848, in <module>
    main()
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 612, in main
    ptx.prefigure_conversion(xml_source, publication_file, stringparams, args.xmlid, dest_dir, args.format)
  File "/home/rob/mathbook/mathbook/pretext/pretext.py", line 302, in prefigure_conversion
    prefig.engine.pdf('tactile', pfdiagram)
  File "/home/rob/mathbook/python-virtual/rab/lib/python3.12/site-packages/prefig/engine.py", line 56, in pdf
    filename = build(format,
               ^^^^^^^^^^^^^
  File "/home/rob/mathbook/python-virtual/rab/lib/python3.12/site-packages/prefig/engine.py", line 42, in build
    core.parse.parse(filename, format, publication)
  File "/home/rob/mathbook/python-virtual/rab/lib/python3.12/site-packages/prefig/core/parse.py", line 53, in parse
    mk_diagram(element, format, publication,
  File "/home/rob/mathbook/python-virtual/rab/lib/python3.12/site-packages/prefig/core/parse.py", line 19, in mk_diagram
    diag.place_labels()
  File "/home/rob/mathbook/python-virtual/rab/lib/python3.12/site-packages/prefig/core/diagram.py", line 210, in place_labels
    label.place_labels(self,
  File "/home/rob/mathbook/python-virtual/rab/lib/python3.12/site-packages/prefig/core/label.py", line 233, in place_labels
    position_braille_label(label, diagram, ctm, background_group, 
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rob/mathbook/python-virtual/rab/lib/python3.12/site-packages/prefig/core/label.py", line 342, in position_braille_label
    space = louis.translateString(
            ^^^^^
NameError: name 'louis' is not defined
```

PR #2264, opened 2024-10-03
Deprecate `<titlepage>` and `<colophon>`

oscarlevin:
This PR will (eventually) deprecate the `<titlepage>` element, allowing an author to write the contents of a titlepage inside frontmatter.

Current progress:

1. Only for HTML.
2. Have not "assembled" the new correct source, but instead have lots of or's when the elements of a titlepage are called.
3. Deprecation message.

Will work on the assembly fixes next.  Then fixes for LaTeX and epub and...?

Happy to change what I have here.  Quickly realizing there are at least four ways to accomplish any task in xslt.


oscarlevin:
Completely undid my first commit except for the titlepage deprecation warning.  What this does:

1. Adds deprecation warnings when author includes a titlepage or colophon in frontmatter.
2. Uses the repair pass of assembly to *create* a titlepage and colophon using the appropriate elements either in the frontmatter or in frontmatter/titlepage and frontmatter/colophon (for backward compatibility) so the other templates can continue to build as they have been.  

I think this approach makes sense since some outputs really do have "pages" for titlepage and colophon and setting these as elements in the xml helps with that.  We just don't want authors to have to worry about which frontmatter elements go in each.

If this is good, then I add some code comments to the titlepage/colophon elements in the other style sheets, update the schema, update the sample documents, and update the documentation, all either in this PR or in follow-on ones.

rbeezer:
Very interesting, and lots to think about.  Can we pause and discuss during the second hour of Tuesday Drop-In?  Might be easier.

Food for thought: do think `titlepage` and `colophon` are a good idea, or not?  ;-)

davidfarmer:
I like the idea of removing these, because they can be automatically assembled from structural elements,
and a typical author will not do the work to learn what is supposed to be in these components.

oscarlevin:
I'm available all those times to hop on zoom.  Pick your favorite.

rbeezer:
Lets do Monday at Noon Pacific?  "dropin" Zoom should be open, or can you shoot me a link?

oscarlevin:
Sounds good.  See you then on the usual dropin link.

oscarlevin:
I have most of what we discussed working, although I could use a hint: how can I ensure that the invented colophon page is put in the table of contents (navigation)?

rbeezer:
Dang.  This is going to be whack-a-mole.  :-(

How about: `repair` phase leaves behind a totally empty `colophon`.  It can have a `@label`.  We get pages, navigation, etc just as always.  But authors (and developers!) don't really need to know about it.  Maybe one big code comment someplace is good enough for developers to know the history and intent?

`@label` or (`@xml:id`) could be from author's current source for their backward compatibility (HTML links).  Else we write `@label="front-colophon"`, or similar, as part of `repair`?

rbeezer:
Well, that idea was not so great either.  How about:
```
frontmatter
    metadata
    titlepage
        titlepage-stiuff  
    colophon
        colophon-stuff
```
*  `metadata` has all the information in one place, not in any order. (Better name?)
*  `titlepage` and `colophon` are just as they are now.  100% backward compatible.  Everything still works.
*  `titlepage-stuff` and `colophon-stuff` are empty elements.  Generators, like `<today/>`.  They produce stuff and templates can control how much and for which document types and for which conversions.

oscarlevin:
In what way was your penultimate idea not great?  I've implemented it with the latest forced push (698f5ae).

If I understand your last comment, it is basically what I have, except with `<metadata>` instead of what I just left as `<titlepage>`.  The main addition is the clever `-stuff` empty elements that developers can easily plug into to say which things go where.  I assume that your suggested structure is generated by assembly, not by the author.

Some ai-assisted ideas for the container name: `publication-details` or `pubdeets` or `pubdata` or `pubinfo` (but don't want to confuse with the publication file), `bibliographic-information` or `biblinfo` or `bibliodata`, `citation-information` or `cinfo` or `cite-info`....  And now that all the bad ideas are done, we can find one we like.  

I do want to remember that in an article, `titlepage` doesn't really fit either.  Also, there are no "big" frontmatter items, so the need to have a container for the bibliographic information is less urgent.

I can be at dropin early.  No worries if this reaches you too late for that to work.

rbeezer:
> I can be at dropin early. No worries if this reaches you too late for that to work.

I'll get on about 30 minutes early.

oscarlevin:
Today's work in latest commit.  Only minimal testing, but hoped to get your eyes on it for suggestions.

rbeezer:
Visual scan of most recent commit, no actual testing, and not the full list of commits.

I like `bibinfo`, modeled after `docinfo`.

What do you think of a variable `$bibinfo` defined by `$document-root/frontmatter/bibinfo`?

Looks like `repair` will give an author a `titlepage` and `colophon` (in a book) forevermore, even if they don't want them?  The match could be on `frontmatter[not(bibinfo)]`, which would use a `bibinfo` as a signal of updated /new source.  We'd need to make `bibinfo` a required child of `frontmatter` I guess, perhaps empty. (?)

Literal `<frontmatter>` created inside `repair` template can be `<xsl:copy>`.  Unsure if there is an advantage to the latter.

Abstract templates:  I think I'd be inclined to issue a warning.  "The conversion you are building up needs an implementation of..."  The template for `colophon-items` will preserve the document order, which seems undesirable.

Code style - I'm (mostly) in the habit of constructing text nodes in the result tree with `xsl:text`.  Maybe a bit verbose, but I like the resulting indentation.  For example,

```xsl
<xsl:attribute name="label">front-colophon</xsl:attribute>
```
as
```xsl
<xsl:attribute name="label">
    <xsl:text>front-colophon</xsl:text>
</xsl:attribute>
```
I plan to be at Thursday Drop-In, at least from the start.

rbeezer:
Maybe a better signal for the necessity of `repair`:

```xsl
match="frontmatter[titlepage[not(titlepage-items)]]"
```

oscarlevin:
> What do you think of a variable `$bibinfo` defined by `$document-root/frontmatter/bibinfo`?

Great idea.  That's ready to go.

> Looks like `repair` will give an author a `titlepage` and `colophon` (in a book) forevermore, even if they don't want them? The match could be on `frontmatter[not(bibinfo)]`, which would use a `bibinfo` as a signal of updated /new source. We'd need to make `bibinfo` a required child of `frontmatter` I guess, perhaps empty. (?)

Currently `colophon` will only survive if it is already present (and in a book).  

`titlepage` is could easily be turned off.  But what should happen in html on the `frontmatter.html` page, where titlepage information is currently displayed, in each case.  For that matter, what should happen in latex when there is no titlepage? 

> Literal `<frontmatter>` created inside `repair` template can be `<xsl:copy>`. Unsure if there is an advantage to the latter.

I'll experiment.

> Abstract templates: I think I'd be inclined to issue a warning. "The conversion you are building up needs an implementation of..." The template for `colophon-items` will preserve the document order, which seems undesirable.

Done!  

> Code style - I'm (mostly) in the habit of constructing text nodes in the result tree with `xsl:text`. Maybe a bit verbose, but I like the resulting indentation. For example,

Done!

I'll wait to push the latest until I figure out how to deal with author info at the start of the document in the absence of a titlepage.

rbeezer:
> But what should happen in html on the frontmatter.html page, where titlepage information is currently displayed, in each case. For that matter, what should happen in latex when there is no titlepage?

As long as `title` stays as a (required) child of `article` or `book` then there is always something to place in output that is front-matter-ish.

oscarlevin:
I still feel like I'm doing something dirty, but I have html doing something different at least when there is no authored titlepage.  See pretext-html.xsl at #L486-L488, #L593-L595, and #L976-L993.

Besides this and some other cleanup, I think this is about ready.

oscarlevin:
Implemented the changes we discussed today.  Should be ready to go.

rbeezer:
Testing has gone well and will merge soon, once I can report more fully here.

rbeezer:
Great work, thanks for sticking with this one.  Changes:

* The pre-processor was killing the **back** colophon as well as the front one.  I prefixed both `match` for cleaning out old versions with `frontmatter/` - not strictly necessary for `titlepage`, but in the name of symmetry.
* Two instances of `xsl:call-templates` that should be singular.  (They were not being exercised anywhere, but I removed a `titlepage` and it provoked an error.)

Some improvements I can see (and I am happy to do).  Don't attempt thse now, I've got things organized to merge.

* Use `$bibinfo` consistently (19 places in two stylesheets have `frontmatter/bibinfo`).
* Swap in the named template for an `h2` title in HTML for even the case where there **is** a `titlepage`.  Not much to be gained by two versions of the same code for constructing the title.
* I think the pre-processor repair can be modularized by moving changes to `titlepage` and `colophon` elements from the main template into the two templates that are now killing the old versions.  Slightly simpler (e.g. `xsl:if` can just be handled by template matching), clearer in three bite-sized templates, and the two templates killing old versions get a better purpose.  I'll need to test carefully.

Could be a few hours before I can get this merged, but feel free to prep a message for `-announce`.  Planning to be at Drop-In.

oscarlevin:
That all sounds great.  I look forward to seeing how you modularized the pre-processor.  I'll prep an announcement and a CLI update to incorporate this when it is ready.

rbeezer:
Merged.  Will update website rendered examples soon.  You are clear to post on `-announce`.

rbeezer:
> The front matter stuff is not going into the material used by Lunr to do search.

More precisely, the diff on HTML output shows that it is some text from the (front) colophon that is going missing.  The search document for a colophon is still being generated but its content (names in credits, for example) is simply an empty string.

rbeezer:
The PR that won't die...  ;-)

> I think the pre-processor repair can be modularized

Done at 25d6f670e8eb4a4266a949225993524e5fda732d.

Using `xsl:copy` mean there is not a new namespace attribute added.  Maybe irrelevant, but less clutter.  Using a `node()`, rather than `*` on the duplication of **other** front matter stuff means some source comments come along (maybe in the wrong place now!  no matter)  and some text/whitespace nodes come along to make the pre-processor XML marginally easier to read.

**Question:** do we need to produce the `front-colophon` label for authors who do not provide one?  Is this a left-over?  Since author's source still allows a `frontmatter/colophon` there is a place for an author to do it (and we should encourage that).  A crappy ID will be made no matter what and that will encourage an author to create a better identifier.

rbeezer:
> Swap in the named template for an `h2` title in HTML

Done at 35c3b21b9b97b270fa14971f2057068b71726dd7.

rbeezer:
> Use `$bibinfo` consistently 

Done at 4779998e8df1341cdbb0df50b3a0fc0e4c9f0f11

PR #2265, opened 2024-10-04
Enable latex style xsl via publication file

oscarlevin:
Added publisher variable as `publication/latex/latex-style`.  Added variable to `publisher-variables.xsl` and enabled reporting  in `report-publisher-variables.xsl`. 

LaTeX conversion in `pretext/pretext` looks for that variable, but only uses it if extra xsl was not specified (a warning is logged if both extra xsl and the publisher variable are used).  

Added to the documentation (I put it at the top, since this seems like the most "overall" latex option, but am happy to organize differently).

### DIFF
--- a/doc/guide/publisher/publication-file.xml
+++ b/doc/guide/publisher/publication-file.xml
@@ -208,6 +208,17 @@
             See <xref ref="pdf-print"/> for a more general overview of this conversion with more details.  This includes some existing options that will eventually migrate to the publication file.</p>
         </introduction>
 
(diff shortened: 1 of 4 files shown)


rbeezer:
Great idea and great execution!  Merged with only edits to the commit message.

(That existing `extraction_xslt` variable is a real misnomer!)

I've put this to work in production with the automated builds of the Guide as a PDF for the website.

Would you like to announce on `pretext-announce`?  Don't need to say much, just add patting yourself on the back and point to the documentation.

PR #2267, opened 2024-10-16
Default language expansion

ascholerChemeketa:
This adds support for default programming language to parsons, hparsons, and codelens.

Also adds a few examples/tests and schema entry.


ascholerChemeketa:
Did some cleanup and a bug fix to Codelens trace generation. I think this is set.

ascholerChemeketa:
@rbeezer Added the RS manifest default language. To do that with existing code, I allow `get-programming-language` to match anything (in this case it needs to match `book`). I know you like narrow matches, but in this case I think being permissive is better.

It means you can use it anywhere and get a string that is either the @language or the default or an empty string.

If `get-programming-language` fails to match, then any attempt to use it that fails (including indirectly through the `xxx-language` templates below it) returns the entire node that came in (matches some copy template elsewhere I assume). Which is really odd behavior for something that is supposed to just return the name of a programming language.

ascholerChemeketa:
Looking into a possible other manifest info. Converting to draft until done.

ascholerChemeketa:
False alarm... thought codelens was missing `data-language`, but it doesn't need it. 

I believe this is set.

rbeezer:
> fails to match, then any attempt to use it that fails

General approach, which is **not to suggest it needs to be this way here**:

Make a narrow match template that you want.  Then make a catch-all template (`select="*")` that uses `xsl:message` to say, "Oops, something went haywire" and yields **no result**, rather than the built-in universal default template that is producing all the content that you find odd behavior.

ascholerChemeketa:
> Make a narrow match template that you want. Then make a catch-all template (`select="*")` that uses `xsl:message` to say, "Oops, something went haywire" and yields **no result**

That makes sense, thanks.

ascholerChemeketa:
Updated to include moving the docinfo attr

rbeezer:
OK, assembly is a very brittle thing and needs to be treated very carefully.  I've done some experiments and now understand the recursion.  I know you are doing whatever it takes to avoid that, and I suspect I can make a hack that would avoid it in a bit cleaner way.  There are some red flags here that make me nervous.

*  The lookup table for various names of programming languages should not live in `-assembly`.
*  The structures in `docinfo` are only interrogated in `publisher-variables.xsl` and then lead to the creation of global variables.  I'd like to maintain that approach/separation.

I think the following might fit better in the current architecture.

*  During assembly for dynamic formats, just pass along the `@language` (if present) and deal with it in `pretext-runestone.xsl`.
*  During assembly for static formats, similarly just pass along the `@language` (if present).  Each static conversion may translate in different ways, so the templates in play may be different, but they can each deal with the possibility of a default.
*  Then a global variable can be "safely" created in `publisher-variables.xsl` at the "right" time, and the lookup table can stay in `-common`, without too much (any?) impact on `-assembly`.  The variable would be a default language, or empty if none is given.

If I'm missing something, or if this needs more discussion, we can work through it at Drop-In tomorrow?

Rob

ascholerChemeketa:
I figured out why the `pretext-runestone-static.xsl` was trying to use the default language - the presence of an `@langauge` is used to determine `$b-natural`, which changes the rendering of components. So we need to know if there is a default language in the affected templates.

This duplicates code from `pretext-common.xsl` (ugly), and assumes that if there is docinfo it is at `/docinfo` (possibly a problem?).

Fingers crossed, but prepared for this PR to rise again from the grave just in time for Halloween.

rbeezer:
OK, seeing clearer now.  In time for Halloween.

* There could be several `docinfo` tagged for different versions.  So existing query might find several.  And it might be possible that the publisher chooses a version so that there are **no** `docinfo`.  I can solve this by making available a new collection of source trees that is earlier than the `assembly` versions and later than the `original` versions.
* Problem is that Parsons problems have two variants, computer code and natural language.  And right now, effectively, natural language is the default if not specified.  One solution: if you want a default **programming** language, you must explicitly specify natural language if you have instances of that.
* And/or maybe natural language is the default default-language (for Parsons).  Is this not current status?
* **OR**, my current preferred alternative.  We have `default-program-language` which is only effective for `program` (listing, activecode, activecode exercises).  Then we have `default-parsons-language` (or something more generic?) where natural language is a possibility.
* Do we need a `none` or `pseudocode` language for `program`?  This might be current behavior when no language is given.

ascholerChemeketa:
I feel like we should optimize for use in the common case, which would seem to be one language for both activecode and Parsons. So if there is a `default-program-language` and no other information, we should assume it applies to Parsons as well. I'm fine with adding a `default-parsons-language` for someone who wants to have all their programs in `c++` and Parsons in pseudocode, but that feels like a less common use case.

That doesn't change needing to know if there is some kind of default in play when building the Parsons.

The various language lookup templates return an empty string if the language is unknown. Most of the templates that use it already are looking for that, so `language="none"` should behave OK even without an explicit entry in the language table.

Will hope to catch you in breakout re: the `docinfo` lookup. But if I don't, should I be waiting on your implementation of the new source tree?

rbeezer:
OK, an earler tree from the assembly pass, which should respect versions, but not set off recursion.  Should be self-explanatory at 1d7bedd7afba697d702e1ebb5be79a2455f03ffb

ascholerChemeketa:
Updated to use your new docinfo access.

Added parsons default that can be used separately from programs default.

rbeezer:
Before/after testing looks good, dynamic (HTML) and static (LaTeX).  One question before I really do super-careful testing (w/, w/out, edits in sample book).

In `xsl/pretext-runestone-static.xsl` for horizonatl Parsons.  Seems to be querying `docinfo` (and incorrectly, from the document root, and not via new intermediate tree).  Shouldn't it be using the modal `get-programming-language` template like for "regular" Parsons?

ascholerChemeketa:
Good catch. I missed cleaning up that copy/paste after you added the new hook.

Fixed.

rbeezer:
```xsl
 <xsl:variable name="default-language">
     <xsl:apply-templates select="$document-root" mode="active-language"/>
 </xsl:variable>
```
is very suspicious.  (Yes, I know what it is doing.)

It might be cleaner, overall, to define two global variables (programs, parsons) that are defaults given in `docinfo` once and for all (instead of repeated queries), where an empty string means they are undefined.  An alternative is two templates that return the defaults.

If `mode="active-language"` had a restrictive match, then this would be exposed sooner.  I'll be at Drop-In today.

rbeezer:
OK, I just spent 30 minutes packaging this up so I could test, assuming it was ready to merge.  Put changes to samples last, checked out code-only changes, so no defaults being set.  Built sample book for Runestone.  There are changes.

Here is the manifest version, which is the easiest to read.

Rob

```diff
         <label>ActiveCode: Listing 5.1.8</label>
         <htmlsrc>
-          <div class="ptx-runestone-container">
-            <div class="runestone explainer ac_section ">
-              <div data-component="activecode" id="PTXSB_2_statistics">
-                <textarea data-lang="python" data-timelimit="25000" data-audio="" data-coach="true" style="visibility: hidden;" id="PTXSB_2_statistics_editor" data-question_label="" data-codelens="true">loan_amount = [1250.0, 500.0, 1450.0, 200.0, 700.0, 100.0, 250.0, 225.0, 1200.0, 150.0, 600.0, 300.0, 700.0, 125.0, 650.0, 175.0, 1800.0, 1525.0, 575.0, 700.0, 1450.0, 400.0, 200.0, 1000.0, 350.0]
+          <pre class="program">
+            <code class="language-none">loan_amount = [1250.0, 500.0, 1450.0, 200.0, 700.0, 100.0, 250.0, 225.0, 1200.0, 150.0, 600.0, 300.0, 700.0, 125.0, 650.0, 175.0, 1800.0, 1525.0, 575.0, 700.0, 1450.0, 400.0, 200.0, 1000.0, 350.0]
 
 country_name = ['Azerbaijan', 'El Salvador', 'Bolivia', 'Paraguay', 'El Salvador', 'Philippines', 'Philippines', 'Nicaragua', 'Guatemala', 'Philippines', 'Paraguay', 'Philippines', 'Bolivia', 'Philippines', 'Philippines', 'Madagascar', 'Georgia', 'Uganda', 'Kenya', 'Tajikistan', 'Jordan', 'Kenya', 'Philippines', 'Ecuador', 'Kenya']
 
 time_to_raise = [193075.0, 1157108.0, 1552939.0, 244945.0, 238797.0, 1248909.0, 773599.0, 116181.0, 2288095.0, 51668.0, 26717.0, 48030.0, 1839190.0, 71117.0, 580401.0, 800427.0, 1156218.0, 1166045.0, 2924705.0, 470622.0, 24078.0, 260044.0, 445938.0, 201408.0, 2370450.0]
 
 num_lenders_total = [38, 18, 51, 3, 21, 1, 10, 8, 42, 1, 18, 6, 28, 5, 16, 7, 54, 1, 18, 22, 36, 12, 8, 24, 8]
-</textarea>
-              </div>
-            </div>
-          </div>
+</code>
+          </pre>
```

rbeezer:
I found the problem.  This is the first commit of any consequence in my repackaging, and alone produces the same diff given above.  In other words, I know where it lies, but not why.

```diff
diff --git a/xsl/pretext-common.xsl b/xsl/pretext-common.xsl
index 44d9d59b..c7c1feda 100644
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -8020,16 +8020,42 @@ Book (with parts), "section" at level 3
 <!-- Define the key for indexing into the data list -->
 <xsl:key name="proglang" match="language" use="@ptx" />
 
+<!-- Define variables for default active language - will be picked up by -->
+<!-- RS manifest and can be a different string than the raw language. -->
+<xsl:variable name="default-active-programming-language">
+    <xsl:if test="$version-docinfo/programs/@language">
+        <xsl:for-each select="document('')/*/mb:programming">
+            <xsl:value-of select="key('proglang', $version-docinfo/programs/@language)/@active" />
+        </xsl:for-each>
+    </xsl:if>
+</xsl:variable>
+
 <!-- Determine programming language to use. First choice is @language     -->
 <!-- on current element. If that is not available, check docinfo default. -->
 <!-- "exercise" might be a Runestone interactive (programming) exercise.  -->
-<xsl:template match="program|exercise" mode="get-programming-language">
+<xsl:template match="program" mode="get-programming-language">
     <xsl:choose>
         <xsl:when test="@language">
             <xsl:value-of select="@language" />
         </xsl:when>
-        <xsl:when test="$docinfo/html/defaults/@programming-language">
-            <xsl:value-of select="$docinfo/html/defaults/@programming-language" />
+        <xsl:when test="$version-docinfo/programs/@language">
+            <xsl:value-of select="$version-docinfo/programs/@language" />
+        </xsl:when>
+    </xsl:choose>
+</xsl:template>
+
+<!-- For a parsons, use @language, default parsons language, or default -->
+<!-- programming language in that order of preference.                  -->
+<xsl:template match="*[@exercise-interactive = 'parson' or @exercise-interactive = 'parson-horizontal']" mode="get-programming-language">
+    <xsl:choose>
+        <xsl:when test="@language">
+            <xsl:value-of select="@language" />
+        </xsl:when>
+        <xsl:when test="$version-docinfo/parsons/@language">
+            <xsl:value-of select="$version-docinfo/parsons/@language" />
+        </xsl:when>
+        <xsl:when test="$version-docinfo/programs/@language">
+            <xsl:value-of select="$version-docinfo/programs/@language" />
         </xsl:when>
     </xsl:choose>
 </xsl:template>
```

ascholerChemeketa:
We decided that the schema for the default in `<docinfo>` in the original PR were ill-advised. So you do need a content update there to get the same output: 

You probably have this in your `bookinfo.xml`:
```xml
    <html>
      <!-- Any program or Runestone interactive without a @language will -->
      <!-- assumed to be python                                          -->
      <defaults programming-language="python"/>
    </html>
```

You now need:
```
    <!-- Any program or Runestone interactive without a @language will -->
    <!-- assumed to be python                                          -->
    <programs language="python"/>
    <!-- Any parsons problems will be assumed to be natural language   -->
    <!-- If not set, this would default to programs@language           -->
    <!-- If neither is set, parsons will default to "natural" language -->
    <parsons language="natural"/>
```

rbeezer:
OK, I'll try that in a bit.  Thanks.

rbeezer:
> So you do need a content update there to get the same outpu

I don't think that is it.  Changes here have an updated `bookinfo`.  I'm seeing the same thing.

ascholerChemeketa:
Hrmmm. On b60901a665f008d03d1ab27ea2d326d583e9dd2b I see the expected:

```html
          <div class="ptx-runestone-container">
            <div class="runestone explainer ac_section ">
              <div data-component="activecode" id="PTXSB_2_statistics">
                <textarea data-lang="python" data-timelimit="25000" data-audio="" data-coach="true" style="visibility: hidden;" id="PTXSB_2_statistics_editor" data-question_label="" data-codelens="true">loan_amount = [1250.0, 500.0, 1450.0, 200.0, 700.0, 100.0, 250.0, 225.0, 1200.0, 150.0, 600.0, 300.0, 700.0, 125.0, 650.0, 175.0, 1800.0, 1525.0, 575.0, 700.0, 1450.0, 400.0, 200.0, 1000.0, 350.0]
```

Maybe try a quick build with my commit?

If that works, something must be different between that final commit in my chain and what is in your repackage.

If that doesn't, maybe we can look at it next drop in.

rbeezer:
OK, I was avoiding the changes in **content** for the sample book, which also meant I was avoiding the change in the specification in `docinfo`.  So even though I saw that `docinfo` change in the vicinity, my testing was ignoring it (by mistake).  So I split that commit, and now "no harm" testing looks OK.  I still need to do some testing with the new content in the sample book.  But now the ball is in my court.

rbeezer:
All done.  Thanks for sticking with this one.  Documentation needs an update?

ascholerChemeketa:
Great!

There is some documentation in the Sample Book. I'll expand on that with the docs in the Program refactor that is currently in the works.

PR #2268, opened 2024-10-16
Add p to remarks if missing around contents

ascholerChemeketa:
Currently, if a remark-like does not have a `<p>` around contents, that content never appears in the output. This adds a paragraph if missing while making sure that if there is a title it is placed outside the `<p>`

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -1030,6 +1030,17 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     </xsl:copy>
 </xsl:template>
 
+<!-- Add input element around text in program when missing -->
+<xsl:template match="&REMARK-LIKE;[not(p)]" mode="repair">
+    <xsl:copy>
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
Yes, you are right, that would not be good.

Alternate approaches:
- Wrapping any child text nodes in `<p>`? 
- Only doing the fix if there are no children other than the text node?

It is caught in schema validation, but silently dropped in conversion.

Alex-Jordan:
I'm not opposed to the goal, but trying to compensate for an author not
following schema could be a slippery slope that could lead to more
maintenance and convenience feature requests. Lots of nitty gritty cases to
consider. I'm going to bow out of this thread :)

davidfarmer:
I have always viewed "content is exactly one p" as a special case:
it is silly to force the author to wrap that content in a p.
(But don't prohibit them from doing it.)

I think we already allow this for `li`.

But I don't really care, because in the future authors will not
write any tags.

ascholerChemeketa:
Closing this after discussion in drop-in.

I think it would be nice to have, but there are too many issues to solve as a small one off.

PR #2272, opened 2024-10-22
Listing captions above programs

ascholerChemeketa:
This implements the "Runestone activecode steals caption of listing it is in" so that explaining text is placed inside RS box.

@rbeezer
It also moves listing caption above the listing in HTML output to match CMOS style. I don't trust my understanding of the LaTeX to  update that output path. We could break this out separate, but it does improve the readability of listings, which benefit from some context before the reader hits a wall of text.

### DIFF
--- a/css/pretext_add_on.css
+++ b/css/pretext_add_on.css
@@ -91,8 +91,12 @@
     white-space: pre;
 }
 
-.ptx-content figure > figcaption:first-child {
-    margin-top: 1.5em;
+.ptx-content figure.figure > figcaption{
+    margin-top: 1em;
+}
+
+.ptx-content figure.listing > figcaption + * {
+    margin-top: 1em;
 }
 
 .ptx-content figcaption + .named-list-content {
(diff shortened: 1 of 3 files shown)


rbeezer:
I like the move of the heading/caption to the top.  I'll need to mirror that change in LaTeX and braille.

I'm less enthusiastic about splitting the caption off from the heading and into the blue box.  It feels "random" and not like a caption anymore.  I would much rather see some work to accomodate Runestone components with more PreTeXt-y CSS for a less bold look-and-feel.  Then maybe the heading/cpation would feel more attached and less disjoint.  (That's a big project, but one I'd really like to see.)

If you feel strongly about the split, maybe host a live version of 5.1, post on -dev, and suggest a comparison with current website version.  Solicit opinions.

rbeezer:
Original discussion on `-dev` was about titles.  Code here seems to be about captions.  They are different, and the code is likely as confused as the history.  So getting this straight has never been easy.  Look for a design discussion on `-dev`.

rbeezer:
> code is likely as confused

**Existing** code!  Not PR code.  ;-)

ascholerChemeketa:
Closing in favor of:
https://github.com/PreTeXtBook/pretext/pull/2299

PR #2274, opened 2024-10-22
Create schema for publication files

oscarlevin:
This creates a schema for publication files, consistent with the documentation as of 2024-10-22.  Uses the same style (xml to rnc to rng and xsd) as the pretext schema.  Named everything `publication-schema` to avoid confusion with the file `publication.xml` which is the actual publication file for the schema.

`build.sh` updated to include this.


oscarlevin:
force pushed a fix for a typo

rbeezer:
Merged.  Thanks for doing this.

*  Testing revealed that some changes to the main PreTeXt schema in the `*.xsd` file, for the previous `bibinfo` never got comitted.  I got those in first on a new commit.  Didn't investigate why that was...
* And your typo-fix was exposed to me since it didn't make it into the derived files (I think).  Silently melded in.

PR #2276, opened 2024-10-22
Listing filenames

ascholerChemeketa:
Basic implementation of filenames for listings. No fancy download links, but I'm not convinced they are needed as anywhere a link would be active the reader can copy/paste into an editor.

It is quite likely there is a more ideal LaTeX implementation.

Built off of https://github.com/PreTeXtBook/pretext/pull/2272 - suggest resolving it first.

### DIFF
--- a/css/pretext_add_on.css
+++ b/css/pretext_add_on.css
@@ -91,8 +91,12 @@
     white-space: pre;
 }
 
-.ptx-content figure > figcaption:first-child {
-    margin-top: 1.5em;
+.ptx-content figure.figure > figcaption{
+    margin-top: 1em;
+}
+
+.ptx-content figure.listing > figcaption + * {
+    margin-top: 1em;
 }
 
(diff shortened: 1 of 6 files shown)


rbeezer:
Took this for a quick drive on top of #2272.

I'd rather see some visual/typographical separation between the filename and the caption.  What do you think of placing it in parentheses, **after** the caption concludes, and style it with a monospace font?

Perhaps LaTeX needs to pass this as an argument to the environment, so style writers can have some control over placment and look?

ascholerChemeketa:
How about monospace bold with parens but before the caption? In a quick tour of programming texts, if the filename is listed, it tends to look like part of the title of the block.

ascholerChemeketa:
Closing in favor of https://github.com/PreTeXtBook/pretext/pull/2299

That does NOT include @filename on listings. But title serves the intended purpose.

PR #2277, opened 2024-10-22
Add stdin attribute to activecode

ascholerChemeketa:
stdin attribute didn't make the initial implementation for activecode. This interfaces with existing RS code to provide input to a program.


rbeezer:
This looks a lot like #2232, which I have a lot to say about, and never quite got it all down.  Briefly, I'm not sure this should be an attribute, and I think maybe the structure of `program` needs a serious re-think.  `input` isn't really so great as the name of a child.

ascholerChemeketa:
I'd add `tests` isn't a great name for "extra code that may or may not be tests to append to the reader's code. :)

Closing this.

rbeezer:
> tests isn't a great name

Exactly.  ;-)

PR #2278, opened 2024-10-24
p to div-para in license

ascholerChemeketa:
License HTML is generating `<p>` instead of `<div class="para">` which breaks if non-phrasing HTML (like a `<details>` element inserted by a PTX `<url>`)


rbeezer:
Got it!  Thanks for the catch and the fix.  Eventually this will migrate to the website examples.

PR #2279, opened 2024-10-28
PreFigure changes to sample article and added PNG support

davidaustinm:
Edits to the sample article to put #prefigure elements in their own namespace and fix indentation (hopefully!).  Also added support for PNG conversion of PreFigure source and changed the format name from "braille" to "tactile" for tactile PreFigure diagrams.  

Necessary changes to several XSL stylesheets are in the second commit.

For the time being, compiling PreFigure source that is inside a namespace will require grabbing the PreFigure nightly build.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -2996,62 +2996,63 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     <li>Production of tactile versions, perhaps available via archive links, will be another addition.</li>
                 </ul></p>
 
-                <!-- Indentation needs to be cleaned up by a knowledgeable PreFigure author -->
-                <image width="70%">
-                  <prefigure label="prefigure-tangent">
-            <diagram dimensions="(300,300)" margins="5">
-              <definition>a=1</definition>
-              <definition>f(x) = exp(x/3)*cos(x)</definition>
-              <coordinates bbox="[-4,-4,4,4]">
-            <grid-axes xlabel="x" ylabel="y"/>
-            <graph at="graph" function='f' />
-            <tangent-line at="tangent" function="f" point="a"/>
-            <point at="point" p="(a, f(a))">
-              <m>(a,f(a))</m>
-            </point>
-              </coordinates>
+                <image width="60%">
+                    <prefigure label="prefigure-tangent"
+                               xmlns="https://prefigure.org">
+                        <diagram dimensions="(300,300)" margins="5">
+                            <definition>a=1</definition>
+                            <definition>f(x) = exp(x/3)*cos(x)</definition>
+                            <coordinates bbox="[-4,-4,4,4]">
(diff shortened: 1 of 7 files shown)


rbeezer:
Thanks, David.  I have significant edits in place for this right now, so please respond with instructions where asked.  Do not push any new commits since I won't be able to deal with them.

* I told you to put namespace declarations late, but I moved them up a notch or two to be contiguous with the others.  ;-)
* The `prefig.engine.png()` call had a `dpi=100` that was raising an error.  I removed it so I could proceed.  Copy/paste error?  Or?
* I'm getting "extra" output.  For example, creating PNG images gets me SVG images as well.  I don't know what the first argument in `prefig.engine.png('svg', pfdiagram)` is all about.  Can you suggest an edit that will change this, or is it part of PreFigure proper?  Not isolated to PNG production.
* I had to "enable" `all` as an option for PreFigure production (it **was** present in a line that was commented out).  But then the `if/elif` really is not doing what you want.  Only the first clause executes and `all` only produces SVG.  So I just removed it all.  Let's add support for `all` back in once this PR is done.

I'll be at Tuesday Drop-In.  But if you have a simple suggestion for "extra" output, that'd help and I can get my half-baked review completed.  Thanks.

davidaustinm:
-  I wondered about the location of the namespace declarations so thanks for clarifying.
-  dpi=100 is a cut and paste error.  Sorry about that.
- We will generate an SVG that is not explicitly requested when the format is "png", "pdf", or "tactile".  I left those in since I noticed you used in the sample article the SVG that was produced from the tactile diagram.  To remove, you will need to glob on *svg in the output directory:
for file in glob.glob(temp_dir + '/output/*.svg'):
    os.remove(file)
- Arg.  Yes, I saw that in the latex_image_conversion and copied it without remembering we were using elif rather than a sequence of if's.

Sorry for the sloppy stuff.

rbeezer:
Got it.  Thanks for the quick reply.  Back to it in a few.

rbeezer:
Alright, changes noted earlier, plus now used provided stanza to remove SVG files **and** XML annotation files.

*  Without looking at PreFigure code, I'd submit the first argument of `engine.png()` and similar is maybe not needed?  And maybe PreFigure should manhandle its own temporary files so consumers do not need to?
* You could implement `all` with a series of `if`.  The efficiency of mutually exclusive `outformat` is of no concern to me here.
* Wanna tackle EPUB?  We prefer SVG, but Kindle limits them, so we make a PNG version, too.  if I remember right.

Thanka for working through this one!  Its great to have PF so tightly integrated.  Alert the CLI?  Write something for `-announce`?

davidaustinm:
I think you're right about the `engine` stuff.   There are more options in the standalone version, but not so much within PreTeXt.

Thanks for the update and for your work on this.   Let's talk tomorrow about next steps.  I'm definitely up for working through EPUB.  I'd also like to get some things like LaTeX macros dropped into the PF source files.

PR #2280, opened 2024-10-29
Allow activities to have exercise-interacives

ascholerChemeketa:
Add activity to the list of `exercise`ish containers that can have interactives.


rbeezer:
`&PROJECT-LIKE;` is an "entity", and these are defined in  `xsl/entities.ent`.

In particular, this one is defined to be `project|activity|exploration|investigation` so adding `activity` here shouldn't have any effect.  Do you see the addition making any (beneficial) changes in the HTML output?

PR #2281, opened 2024-10-30
PreFigure intermediate file clean up, prefigure-preamble added

davidaustinm:
This adds a few LaTeX macros and a #prefigure-preamble into the sample-article, along with a third #prefigure diagram that can make use of them.

Also, the prefigure_conversion function is cleaned up a bit, changing elif to if.  The nightly prefig build (not prefigure!) will clean up after itself when building pdf, png, and tactile versions.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -67,7 +67,12 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         % comments anywhere get stripped, with or without faux % comment \% characters
         \newcommand{\indefiniteintegral}[2]{\int#1\,d#2}
         \newcommand{\testingescapedpercent}{ \% } % just testing
+        \newcommand{\bvec}{{\mathbf b}}
+        \newcommand{\vvec}{{\mathbf v}}
         </macros>
+        <prefigure-preamble>
+            <line stroke="blue"/>
+        </prefigure-preamble>
         <!-- We control the appearance of the text of cross-references here           -->
         <!-- (these are the "xref" elements).  Because this changes the words         -->
         <!-- that appear in your output, it is a decision with stays with the source. -->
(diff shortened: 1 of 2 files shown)


davidaustinm:
If everything goes into the same directory, there will be a name conflict as a diagram whose label is 'foo' will make foo.pdf for both formats, 'svg' and 'tactile'.  Should we put the tactile versions in a `tactile` directory under the destination directory?

rbeezer:
>  Should we put the tactile versions in a tactile directory under the destination directory?

Sure.

Another gotcha.  I'm making a publication file, and the Python is trying to make images from it.  Seems to be non-fatal, but shows in the log messages.

rbeezer:
Here's my best version of a generated publication file.  Neither the blue nor the macros seem to have effect.

```xml
<?xml version="1.0"?>
<prefigure xmlns="https://prefigure.org"><line stroke="blue"/><macros>
\newcommand{\definiteintegral}[4]{\int_{#1}^{#2}\,#3\,d#4} 
\newcommand{\myequation}[2]{#1\amp =#2} 
\newcommand{\indefiniteintegral}[2]{\int#1\,d#2}
\newcommand{\testingescapedpercent}{ \% } 
\newcommand{\bvec}{{\mathbf b}}
\newcommand{\vvec}{{\mathbf v}}
\newcommand{\lt}{&lt;}
\newcommand{\gt}{&gt;}
\newcommand{\amp}{&amp;}
</macros></prefigure>
```

davidaustinm:
Is this in a file pf_publication.xml in the same directory as the source
files?

I didn't think about trying to compile the publication file.  Won't be hard
to fix.

davidaustinm:
This should be fixed if you grab the prefig nightly that I just built.

rbeezer:
> if you grab the prefig nightly that I just built.

Yes, that's much better.  Maybe I'll finalize the publicatiion-file-creation tomorrow.

rbeezer:
Sample article stuff went in now that #2283 is done.  So all three parts of this are merged.  Thanks!

If you are ready to advertise, send me a subject line and text for `-announce` and I will post.

rbeezer:
Had to add SVG and PDF versions of the projection example.  I don't rely on the CLI to auto-genenerate these sorts of files, so they are commited into the repository.

davidaustinm:
Thanks for clarifying the generation of images.  I wasn't sure how to handle that, but the same issue will show up in the EPUB conversion.

This all looks really good.  Thank you for all the work you've put into it!  I will hold off on publicity for just a bit to see where we're at with the diagcess stuff.

rbeezer:
>  I will hold off on publicity

Strictly on your schedule.  But yes, having annotations in place will be a great feature to have at roll-out.

PR #2283, opened 2024-10-30
PreFigure format "all" support in prefigure_conversion

davidaustinm:
"all" is added as an implemented format for prefigure source in `pretext`

`prefigure_conversion` inside `pretext.py` reworked for format "all".  With this format choice, tactile PDF goes into `destination_directory/tactile`

### DIFF
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -607,7 +607,7 @@ def main():
     if args.validate:
         ptx.validate(xml_source, out_file, dest_dir)
     elif args.component == "prefigure":
-        if args.format in ["source", "svg", "pdf", "png", "tactile"]:
+        if args.format in ["source", "svg", "pdf", "png", "tactile", "all"]:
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks, I needed this for the conclusion of #2281.

*  Just realized there will always be a `pf_publication.xml` since the collection of LaTeX macros will never be empty (`\lt`, `\gt`, `\amp`).  So the removal should always succeed. (No harm as-is anyway.)
* The copy tree stuff at the end was already in, and also in this commit.  Easy merge conflict to resolve, but we should keep an eye on it.

Thanks for moving this along!

davidaustinm:
The copytree call at the end of `prefigure_conversion` looks correct to me.

rbeezer:
Thanks!  Always a bit of uncertainty when I need to resolve a merge conflict in a PR.

PR #2284, opened 2024-10-31
PreFigure integration with EPUB conversion

davidaustinm:
Updates to the `pretext-epub.xsl` stylesheet to support the inclusion of PreFigure source in the EPUB converstion.  The source for a PreFigure diagram is also included in `epub-sampler.xml`.

Using the included publication file, the `pretext/pretext` script built the requested SVG image from PreFigure source and placed it in `gen/prefigure`.  Building epub-svg from the sample EPUB document included the image appropriately in the EPUB output.  

This felt a little too easy, but we seem to be getting to the point where we can treat a #prefigure element in the same way as #latex-image, which is all I've done here.

### DIFF
--- a/examples/epub/epub-sampler.xml
+++ b/examples/epub/epub-sampler.xml
@@ -225,6 +225,40 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                         </description>
                     </image>
                 </figure>
+
+                <p>A Scalable Vector Graphics (SVG) image produced from <prefigure/> source with the <c>pretext</c> script.</p>
+                <figure xml:id="figure-prefigure-tangent">
+                    <caption>The graph of a function and a tangent line</caption>
+                    <image width="60%">
+                        <prefigure label="prefigure-tangent"
+                                   xmlns="https://prefigure.org">
+                            <diagram dimensions="(300,300)" margins="5">
+                                <definition>a=1</definition>
+                                <definition>f(x) = exp(x/3)*cos(x)</definition>
+                                <coordinates bbox="[-4,-4,4,4]">
+                                    <grid-axes xlabel="x" ylabel="y"/>
+                                    <graph at="graph" function='f' />
+                                    <tangent-line at="tangent" function="f" point="a"/>
+                                    <point at="point" p="(a, f(a))">
+                                        <m>(a,f(a))</m>
+                                    </point>
+                                </coordinates>
+
+                                <annotations>
+                                    <annotation ref="figure"
+                                                text="The graph of a function and its tangent line at the point a equals 1">
+                                        <annotation ref="graph-tangent" text="The graph and its tangent line">
+                                            <annotation ref="graph" text="The graph of the function f" sonify="yes"/>
+                                            <annotation ref="point" text="The point a comma f of a"/>
+                                            <annotation ref="tangent" text="The tangent line to the graph of f at the point"/>
+                                        </annotation>
+                                    </annotation>
+                                </annotations>
+                            </diagram>
+                        </prefigure>
+                    </image>
+                </figure>
+
                 <p>Here we drop a small reference to an equation
                 earlier, notably <xref ref="zeta-2" />.</p>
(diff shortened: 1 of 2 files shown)


rbeezer:
Visual looks good, should test very soon.

rbeezer:
> This felt a little too easy,

C'mon!  ;-)  That's a testimony to your work (and a bit of mine).  Enjoy it.

Merged, since this make an EPUB that "works".  But a job for me and a job for you - nothing is ever 100% easy.

* PF namespace was bleeding through into the output (like on the literal `style` element holding some MathJax CSS).  (Which turns up in a before/after diff.)  Adding `pf` to `exclude-result-prefixes` made that go away.
* Building `epub-kindle` (which prefers PNG images), I see the PNG are being tagged as being SVG.  Not a problem here, more general and won't hold this up.
* I use [epubcheck](https://www.w3.org/publishing/epubcheck/) for validation.  I think my local copy is via an Ubuntu package.  May be a (slow) online version.  For the SVG version, it must be checking the SVGs themselves too.  Boatloads of errors, all very similar to the below.  Ordered pair is (line, column) offsets.  I think you want to get this sorted.  (A few pre-existing CSS errors and a sageplot SVG with other problems.)

```
ERROR(RSC-005): epub-sampler.epub/EPUB/xhtml/generated/prefigure/prefigure-tangent.svg(12,85): 
Error while parsing file: attribute "type" not allowed here; expected attribute "aria-activedescendant", 
"aria-atomic", "aria-autocomplete", "aria-busy", "aria-checked", "aria-colcount", 
"aria-colindex", "aria-colspan", "aria-controls", "aria-current", "aria-describedby", 
"aria-details", "aria-disabled", "aria-dropeffect", "aria-errormessage", "aria-expanded", 
"aria-flowto", "aria-grabbed", "aria-haspopup", "aria-hidden", "aria-invalid", 
"aria-keyshortcuts", "aria-label", "aria-labelledby", "aria-level", "aria-live", 
"aria-modal", "aria-multiline", "aria-multiselectable", "aria-orientation", "aria-owns", 
"aria-placeholder", "aria-posinset", "aria-pressed", "aria-readonly", "aria-relevant", 
"aria-required", "aria-roledescription", "aria-rowcount", "aria-rowindex", 
"aria-rowspan", "aria-selected", "aria-setsize", "aria-sort", "aria-valuemax", 
"aria-valuemin", "aria-valuenow", "aria-valuetext", "class", "clip-path", "clip-rule",
"color", "color-interpolation", "color-rendering", "cursor", "display", 
"externalResourcesRequired", "fill", "fill-opacity", "fill-rule", "filter", 
"focusable", "image-rendering", "lang", "marker-end", "marker-mid", "marker-start", 
"mask", "ns:type", "onactivate", "onclick", "onfocusin", "onfocusout", "onload", 
"onmousedown", "onmousemove", "onmouseout", "onmouseover", "onmouseup", "opacity", 
"pointer-events", "requiredExtensions", "requiredFeatures", "role", "shape-rendering", 
"stroke", "stroke-dasharray", "stroke-dashoffset", "stroke-linecap", "stroke-linejoin", 
"stroke-miterlimit", "stroke-opacity", "stroke-width", "style", "systemLanguage", 
"tabindex", "text-rendering", "transform", "visibility", "xml:base", "xml:lang" or 
"xml:space" (with xmlns:ns="http://www.idpf.org/2007/ops") or an attribute from another
 namespace
 ```

davidaustinm:
I see those errors too with epubcheck.  I include an attribute `@type` in the output SVG that indicates the meaning of a graphical element.  For that element, we have `@type="vertical grid"` indicating it is part of the vertical grid.  It was mainly used for debugging earlier in the project but doesn't serve a purpose anymore and can be removed.  It's strange because it fails on the second grid line and not the first.  

There is a second batch of three errors because three different elements are assigned the same value of `@id`.  This is a problem I understand and thought I had fixed a while ago.  Again, this won't be a problem but will take a little more work.  Not sure I can get to it today.

EPUB is pretty rigorous in validating SVGs.  It's like the wild west inside a browser.

rbeezer:
>  I include an attribute @type

I wonder if the Wild West move of using `@data-type` will appease the validator?  Or if there is something similar for SVG?

PR #2285, opened 2024-10-31
Update basics-part.xml

Tanaquil18:
remove outdated info about example solutions having to be knowled.


rbeezer:
Great!  Thanks for updating this.  (I don't even remember implementing that option!)  It'll be visible on the next website update within a day or two.

PR #2287, opened 2024-11-01
Cleanup of PreFigure SVG for type and id attributes in EPUB

davidaustinm:
This only updates the PreFigure SVG used in building the `epub-sampler` so that `epubcheck` no longer reports errors from the SVG.  

I just made a new nightly build of prefig that will produce the same SVG.

### DIFF
--- a/examples/epub/gen/prefigure/prefigure-tangent.svg
+++ b/examples/epub/gen/prefigure/prefigure-tangent.svg
@@ -9,38 +9,38 @@
   </defs>
   <g id="grid-axes">
     <g id="grid" stroke="#ccc" stroke-width="1">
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks.  Merged as-is.

For good measure, I updated the SVGs in the sample article with the 2024-11-02 nightly, at 9700d20df2131ba78e4d767a08067d6ed1094e99

rbeezer:
And I forgot to say: we are back to known errors from the validator.

davidaustinm:
That's good news.  Just out of curiosity, I put in some `@data-type` attributes, and the validator is still happy.  I'm not sure that's a useful attribute anymore so I'll leave it.  MathJax drops in an impressive number of attributes that pass the validator as well.

PR #2289, opened 2024-11-07
tactile PreFigure diagram converted to PNG in sample-article

davidaustinm:
Three changes to the sample-article:

- added a PNG version of the tactile diagram prefigure-diffeq in sample-article/media/tactile
- modified the PDF version of that diagram for better font handling
- edited sample-article.xml to load the PNG rather than an SVG version of that diagram

That image requires the braille29 font, and problems with sizing of the braille labels appear in the SVG when that font is not present.  The same issue appeared in the PDf version of that image.


rbeezer:
The sample article was not specifying the SVG, or any other format.  No extension means pick the best format.  So adding `.png` means that is all that will ever be used, across formats.  Which is probably good here, no relying on fonts, etc.  So I removed the PDF, it won't be used.

The provided PNG has a lot of padding.  I increased the width to 100% (from 70%).

Merged, but the surrounding text needs adjustment (it is now inaccurate).  Won't rebuild the website right away.  Make a new PR if you'd like to remove the padding on the PNG image.

> That image requires the braille29 font

Which image?  ;-)

PR #2292, opened 2024-11-12
better static representation of matching exercise

oscarlevin:
This inserts an extra "gutter" column in the tabular for the static version of a matching Runestone exercise, to make it obvious that the premises and responses might not be matched up already.  Note that the solutions version doesn't have this, so it looks like they really are matched up.


rbeezer:
Thanks!  Yes, that does look better.  Likely still room for improvement.

I assume you wanted the solution to get the same treatment?  I added that silently.

The `<p><nbsp/></p>` hack is now enshrined in code.  :-(  _C'est la vie_.  The blind will really not like it.

oscarlevin:
I did not mean for solutions to get this, since there the two halves are matched.  Not a big deal though.  I'll think of some other ways this might be better displayed.

PR #2295, opened 2024-11-15
add banner warning to -classic conversion

oscarlevin:
No changes to the standard latex conversion.  No diffs between this and the previous -classic latex output.


rbeezer:
Very good.  First divergence.  ;-)

Eventually I think this new template matching on `/` is only needed to insure the warning happens.  So can probably go away once the warning is not needed.

PR #2297, opened 2024-11-18
initial modularization of preamble

oscarlevin:
Before getting too far into it, I wanted to check that this is heading in the right direction.  If so, I'll keep going and add on to this PR.


oscarlevin:
Do you mean think about adding things like the xcolor-style hook or about removing them?

I will use the "no-space-before-ending-slash" convention.

rbeezer:
> Do you mean think about adding things like the xcolor-style hook or about removing them?

Just "think about them".  Are there better ways?  Is the current arrangement best?  Et cetera.  Seems like your head will be into those ideas and you are going to bump into whatever is there already.  Maybe your "be ready for each journal" work will suggest new approaches.

oscarlevin:
I'm not claiming that these chunks are the best way to break things up, but it does preserve the order and mostly separates different sorts of things into different templates (a big exception being the last group of miscellaneous tweaks).

No change in output for sample-book or sample-article, other than date/time of generation.

PR #2298, opened 2024-11-19
Program update / add missing features

ascholerChemeketa:
I believe this is ready for review.  I just rebased to master and regrouped commits.

Testing is going to be fun as this requires new Runestone code (https://github.com/RunestoneInteractive/rs/pull/582) and some of it only works on a Runestone server.


ascholerChemeketa:
@bnmnetp merged https://github.com/RunestoneInteractive/rs/pull/582

This should be ready for testing PTX side.

bnmnetp:
I merged, but I have not yet created a new runestone release.  I will do that tomorrow.

bnmnetp:
runestone 7.4.6 is now out.

rbeezer:
The organization here is improved - thanks.  I think (based on not much care) that the "view source" stuff should have been an independent, or later, PR.

A refactor needs to clearly make no changes in output.  It looks to me like I should be able to checkout the "Deprecate and fix program/input" commit and there should be no changes to the output of the samples.  (Yes?)

LaTeX output, both sample book and sample article, show many hunks of code going missing.  Some appear to be inside `solution` to exercises.  I didn't investigate further, but maybe a static version of a Parsons programming exercise provides a solution which is not yet migrated from `input` to `code`?

If changes are needed, please rebase them into the commits you already have and force-push.  I've got about an hour invested in understanding the current organization and don't want that to change.

ascholerChemeketa:
Will go check into those.

Just to clarify...
I should make any new commits be fixups to the existing ones? Or do you want the new commits to be individual commits and just reordered into the existing sequence?

rbeezer:
> I should make any new commits be fixups to the existing ones?

Yes, meld additions into current organization, no new commits.  Force push once ready-to-go.  Thanks.

ascholerChemeketa:
There was indeed a bug in constructing the program text for latex.

Did a force push that changed 58f7ca1 - Static representations use code instead of input and include visible pre/postambles

The datafile issue was due to some "view-source" related code that got accidentally separated up into different commits (that were on either side of the commit you were looking at). I moved the code that accidently got included in an early commit to with the rest of the view-source stuff.

rbeezer:
Thanks for the updates.  May be offline tomorrow after Drop-In (contractors), but I'll try to get to this ASAP.

rbeezer:
The sample book does not build with this PR.  Looks like infinite recursion with some template that has variables `$first-char` and `$pad`.  If the PR is trimmed to not include any new content in the sample book, all's well (did that first).  So I suspect some new content is provoking a bug.

I had about an hour invested in running this through tests before encountering this bug.  That will all need to be redone when changes are made.  Please do the basic tests necessary to make sure this is ready to go.

ascholerChemeketa:
@rbeezer 

I get the frustration, but it was testing and is working... I just checked out my last commit in the PR branch (https://github.com/PreTeXtBook/pretext/pull/2298/commits/c3e01d44dc8c7c9d9d5cb630c1e0ffed89acad1d) and built sample book. HTML and latex both built fine.

If I rebase that branch to your latest commit on master (8558c528ba32eadc4e1083666fe38aaa84c82dd1) I get the error you are seeing.

Did you test before rebasing? ;)

The issue must be coming from some interaction between the PR and code added since I last updated it. Or possibly git is auto-merging some changes that should be mutually exclusive.

Will see if I can figure out what that interaction is before drop in tomorrow.

rbeezer:
> If I rebase that branch to your latest commit on master (https://github.com/PreTeXtBook/pretext/commit/8558c528ba32eadc4e1083666fe38aaa84c82dd1) I get the error you are seeing.

Dang.  Sorry about that.  Yes, I rebase routinely.  I guess at worst, a `git bisect` is in order.

See you in the morning.

ascholerChemeketa:
Here is the issue:
In `pretext-view-source.xsl`, in 
```xsl
<xsl:template match="*" mode="view-source-widget">
```
this variable:
```xsl
<xsl:variable name="original-element" select="$original-labeled//*[@original-id = $the-element-id]"/>
```
Picks up 2 copies of the element - one each from both of the version files.

So two copies of the source are included in the annotation. There are two very long Java programs in the book. Doubling the size of them is enough to choke the text processing recursion.

Have you tried building the annotated book since versioning the sample-book? I am guessing it will have the same issue.

rbeezer:
Switched to a post-version tree, as discussed in Drop-In.  Now the sample book builds without errors.  Nothing pushed, more details later.

PR #2299, opened 2024-11-20
Code listings

ascholerChemeketa:
This implements title (not caption!) at the top of listings in HTML. Latex left as an exercise for the reader.

It also implements program/console pairs in listings. The console will only be displayed if the program does not become and activecode - so you can author a sample run that shows in static output but not when the user can just run the code themselves.

Currently, I steal the listing number to put a "subtitle" on the console that looks like:
Sample run of Listing 2.3.1

There is enough markup around it and the rest of the content that a theme should be able to do fancy visual things, including hiding that chunk of text, with the program/subtitle/console.


rbeezer:
> Latex left as an exercise for the reader.

I can attempt that, but it means a delay as it goes into the queue.

> It also implements program/console pairs in listings.

This sounds like a different change (though I have not looked closely).  Please split this PR into two discrete tasks, I can't manage reviews, changes, repackaging, LaTeX additions when a PR does two things.

> Currently, I steal the listing number to put a "subtitle" on the console that looks like:

On pretext-dev, I wrote:

> (b) I'm reluctant to stick an author with a defalt title/heading that may not be what they are really after.

> There is enough markup around it and the rest of the content that a theme should be able to do fancy visual things, including hiding that chunk of text, with the program/subtitle/console.

We don't always have this flexibility in other formats.

rbeezer:
OK, blown out into several more commits.  You know, to increase your personal count.

*  I think we have `\textbf` happening twice for titles.  Not your fault, wider than just `listing`.
*  Schema had some accumulated syntax errors.  I got thos fixed.
*  CSS only applies to legacy themes?  Not sure if that is bitrot or intentional.  Examine the two commits for details.
*  You might double-check the commits for the sample documents (I did myself, but it'd be good to have another look).
*  Other "figure-like" have a schema pattern called `MetaDataAltTitle` (which I switched to).  We don't want a `title` to be optional.

PR #2301, opened 2024-11-22
Create example workbooks for testing

mitchkeller:
This PR is partially for discussion as requested at https://groups.google.com/g/pretext-dev/c/YAPaSa0FcNU/m/VfVNv0z9AAAJ

I have created the two sample workbooks, one as an article and one as a book. The pretext/pretext script doesn't throw errors or warnings. The book version actually comes out in good shape. The article version has some formatting issues for the pages in LaTeX, likely because the article document setup doesn't have sufficient geometry configurations made. Numbering in the article is bad, since there are no `<section>` elements, so you get numbers starting with 0, but in the book things look fine.

Publication files undoubtedly have too much stuff in them, but I just wanted to get something up and running for discussion and inspiration rather than getting things perfect at this point.


rbeezer:
Thanks, Mitch.  Of course, being champion did not imply it was up to you.  So a special thanks.

As championee, I just merged this.  With little testing, given its provisional nature.  One of the (few) things I like about GitHub is how READNE.md gets picked up for online browsers, so I added a minimal disclamatory one on a follow-on commit.

I've found a cascade of .gitignore really confusing sometime.  I might move yours up into the root one.  Yes, I know bewing relocatable is a good thing.

More on the -dev thread.

PR #2302, opened 2024-11-25
Update rune.xml

Tanaquil18:
fixed non-bijective matching example to avoid a function that could belong to two different categories.  It might be worth noting that as I understand it, the proposed expanded (terrific!) functionality still won't handle that case.


rbeezer:
Thanks very much, @Tanaquil18!  Had not noticed that twist.

Do a `git log` on your local repo at this branch.  You'll see your comment here in the discussion as part of an "expanded" commit message.  Not sure how that happens, but I routinely scrub them from commits.  One succinct line, plus the actual changes, should be enough for someone to understand.  Website update won't be immediate.

PR #2303, opened 2024-11-25
Fix static output program-language lookup

ascholerChemeketa:
Found this while debugging https://github.com/PreTeXtBook/pretext/pull/2298

`get-programming-language` from `pretext-common` is not available in `pretext-runestone-static`. So the `get-programming-language` template call there is not working correctly.

This fix works... but it ugly in that it duplicates templates from `pretext-common` into `pretext-runestone-static`. What is the best way to do this right? A new xsl file with the language stuff that gets included from both places?

### DIFF
--- a/xsl/pretext-runestone-static.xsl
+++ b/xsl/pretext-runestone-static.xsl
@@ -51,6 +51,34 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <!-- matches here are fine with a *[@exercise-interactive='foo'], -->
 <!-- as a convenience.                                            -->
 
+<!-- These get-programming-language templates duplicate logic from pretext-common  -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Just to be clear, this is needed so that #2298 is correct?  I have some ideas about alternatives, which we can discuss at Drop-In.

ascholerChemeketa:
The current behavior (pre 2298) is incorrect.  This fix is independent of that PR - I just noticed the issue while working on the new PR.

See you in drop-in.

rbeezer:
> This fix is independent of that PR

Got it, thanks.

rbeezer:
OK, this is in as a temporary fix.  I'll want to split out "natural" form "language" as discussed in Drop-In. Thanks.

PR #2304, opened 2024-11-26
Add optional @continuation attribute to #console and #input

skiadas:
- This allows an optional "continuation" attribute to the console and input tags, to accompany the corresponding prompt attribute. 
- If present, the continuation attribute will be used as prefix for subsequent lines of a multi-line input.


rbeezer:
@skiadas:  looks real good.  Thanks for the careful PR.  Partially offline this morning, on fone, so just a visual review for now.

#2298 is going to move `input` to `code`.  Conflicts like this rarely happen!  Just sit tight - I have to get that other long-running PR resolved first.

rbeezer:
First, apologies for the long delay.  That's not normal.  A variety of factors at play, including my mistake about `input` vs. `code`.

When I produce LaTeX for the sample article (not the minimal example), I see lines going missing before and after this PR is applied.  I think it is an `input` with no command (text).  Two examples

```diff
@@ -8928,7 +8928,6 @@ A specialized version of a program listing is an interactive command\slash{}resp
 pi@raspberrypi ~/progs/chap02 $ (*\consoleinput{gcc -Wall -o intAndFloat intAndFloat.c}*)
 pi@raspberrypi ~/progs/chap02 $ (*\consoleinput{./intAndFloat}*)
 The integer is 19088743 and the float is 19088.742188
-pi@raspberrypi ~/progs/chap02 $ (*\consoleinput{}*)
 \end{console}
 \tcblower
 \end{listingptx}%
```

```diff
@@ -10695,7 +10692,6 @@ x.means
 pi@rpi ~$ (*\consoleinput{gcc -o intAndFloat intAndFloat.c}*)
 pi@rpi ~$ (*\consoleinput{./intAndFloat}*)
 19088743 (integer) and 19088.742188 (float)
-pi@rpi ~$ (*\consoleinput{}*)
 \end{console}
 \end{sbspanel}%
 \end{sidebyside}%
```

I think the `tokenize()` is getting a totally empty string (no newline) and is not making any `token`.  Could verify with 

```
<xsl:message><xsl:value-of select="count(str:tokenize($sanitized-text, '&#10;')/token)"/></xsl:message>
```
Could be better to go back to recursively stripping out lines.  Or maybe add a newline at the very end of the sanitized text (just tried this, no luck, so maybe I haven't analyzed this fully).  I'll let you decide where the edge case is and how to address it.

skiadas:
Oooh, oops, I was trying to bring myself in line with the current pretext repo and in the process may have deleted the proposed changes. Hm I thought we had already merged these changes in though?

skiadas:
Going to retry to open this now that I added new version, not sure it will work.

skiadas:
Ok that seems to have worked and this should be my updated files. As you pointed out, whitespace-only inputs (including completely empty input) are skipping the for loop. I simply added an if-clause to directly handle this case just before the for-loop. Empty input blocks feel a bit weird to me, but I suppose they are needed, to show just a prompt.

I'll see if the HTML output exhibits similar behavior then add a similar fix.

rbeezer:
Git-tastrophe averted?  ;-)  Yes, HTML was fine.  I'll have another go at this.  Much less of a delay this time!

rbeezer:
Very nice - that's a good feature to have now.  Some notes:

* `<xsl:text></xsl:text>` can be shortened to `<xsl:text/>`
* For whatever reason, I/we do characters in hex.  So `&#xa;` rather than `&#10`.  makes searching easier if they are all the same.
* Moved the example from the **minimal** example into the **sample article**.
* Otherwise, did my usual repacking of the commits.

Would you like to write a short blurb saying what is available and pointing to the documentation?  I'll put it up on `pretext-announce`.  Thanks.  (Website rebuilding right now, give it 10 minutes.)

skiadas:
> Git-tastrophe averted? ;-) 

Hehe, yep, and I finally got to actually use `git reflog`. For blurb I think the two lines in my first comment on this pull request look about right. Were you thinking something longer?

rbeezer:
> finally got to actually use git reflog.

Saved me more than once!

PR #2305, opened 2024-11-29
add a PreFigure sidebyside into the sample-article

davidaustinm:
This adds two annotated PreFigure diagrams in a `sidebyside` in the sample article.  I don't think it's ready to be merged, but I wanted to illustrate some ways in which PreTeXt and diagcess are not getting along.

1. The `@widths` and `@margins` attributes of the `sidebyside` seem not to be fully respected, and the result of changing the values of those attributes doesn't seem entirely predictable.
2. The relative size of the two diagrams changes when the dimensions of one of the SVGs changes, which I don't believe should happen.  The version here looks alright (after a bit of fiddling), but changing `@dimensions="(300,200)"` from `(200,200)` in the first `diagram` will change the relative size of the two diagrams and sometimes one diagram will bleed over into the other.  
3. I'm wondering if the navigation instructions could be placed in the margin.  It's a relatively long piece of text "Diagram Exploration Keyboard Controls" and when it breaks into two lines under one of the diagrams, the vertical alignment is affected.  In a `sidebyside`, we shouldn't need it twice.


davidaustinm:
PS:  Grab the prefig nightly before compiling.

rbeezer:
*  I experimented some, mostly for naught, since I forgot you need a server to see the annotations behaving.  But not enough to see what you describe as bad behavior.
*  `sidebyside` panels have CSS for "flex" layout.  I really can't say how this might be interacting with the diagcess JS.
*  I made a "beta" which is live.  Perhaps you can test it in your browser of choice, and then post on `-dev` asking for help?  (And alert Volker to join the post?) https://pretextbook.org/beta/prefigure-sbs-2024-12-01
*  I agree we certainly do not need multiple instructions inside `sidebyside` and `sbsgroup`.  Let's get it down to one set of instructions first, and then we can think about moving it to the margin.  (Mention this in your post?)  Let's do this somewhere else, not on this PR.

Psst: marking this as a draft will help me.  And when it is ready to go, please provide enough image files for all the different output formats (we don't use the CLI for development).

davidaustinm:
I'm changing this from a draft to ready for review since I think it's ready for distribution in the sample article.  This will provide a test case for a few PreText/PreFigure integration issues.

I've added all the necessary output formats for the PreFigure diagrams and made some small changes in the PreFigure source for one of the diagrams to take advantage of some new features in the language.

I'm not totally confident in my git skills so please advise if something is amiss.  I'm happy to revise, if needed.

rbeezer:
Merged as-is, but combined into one commit.  Thanks!  This will be useful.  Maybe I'll refresh the website in teh morning.

> I'm not totally confident in my git skills so please advise if something is amiss. I'm happy to revise, if needed.

I had to resolve a merge conflict when I rebased onto the current tip.  Not sure that is **anybody's** "fault."  In a perfect world, PR authors would rebase before submitting for review, but almost nobody does.

PR #2306, opened 2024-12-02
Keywords support

oscarlevin:
This still needs a lot of work, but sharing in case @rbeezer has time and wants to redirect.  Mostly still trying to figure out where things should live in the xsl.

So far, only working on HTML conversion.  Added title support in `pretext-common`.

Need to think more about the how keywords interact with other components.  In a book, keywords should go in colophon, I think.  In article, they should go *after* the abstract, so not in the titlepage, even though in html the abstract is shown on the "first" page (if there is chunking).

### DIFF
--- a/css/pretext.css
+++ b/css/pretext.css
@@ -489,6 +489,27 @@ https://yoshiwarabooks.org/mfg/MathModels.html */
     display: inline;
 }
 
+.ptx-content .keywords {
+    margin-top: 1em;
+}
+.ptx-content .keywords > .title {
+    font-weight: 600;
+    line-height: 1.125em;
+    display: inline;
+}
+.ptx-content .keywords > .title::after {
+    content: ".\2009\2009\2009";
+}
+.ptx-content .keywords > .title + .para {
+    display: inline;
+}
+
+.ptx-content .support {
+    display: block;
+    margin-top: 1em;
+}
+
+
 /*       -----      */
 
 
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -227,6 +227,15 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
                 <!-- Can set date manually or use the "today" element -->
                 <date><today /></date>
+                <keywords>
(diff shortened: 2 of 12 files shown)


rbeezer:
A quick visual review looks good.  Comments:

*  Support and affiliation in schema will be on some other commit, I presume?
*  When there is an `@authority`, lets use the language/title of the authority.  But generic `Keywords` should be localized.  Ask in Drop-In about procedures for new entries.
*  `test="position() != last()"` is a red flag since `position()` counts **everything** like comments in author's source.  Better is `test=following-sibling::keyword` to see if the comma/separator is needed.  (Or some logical variant.)

Location in XSL is not such a big deal, and location in the output should follow your instincts.

oscarlevin:
I've just finished a draft of the keywords element.  Working in latex and html.  Things that could be improved are how to deal with primary/secondary MSC's, and maybe how to deal with the `@variant` of these: currently I opted for `@authority="msc" @variant="2020"` instead of `@authority=msc2020`.  

@rbeezer: Would you rather we tackle these one family of bibinfo at a time, or have all the new stuff together?  Note, the schema in this PR has a mish-mash of stuff that I put in there mostly so I don't forget about it, so I need to clean that up regardless (and put it on its own commit, last ;)).

rbeezer:
Let's do **two** final commits: schema mish-mash AND then keywords only.  If you can separate them?

oscarlevin:
Done.  Keywords is ready for review.

oscarlevin:
I've been working on `<support>` and I think it would fit nicely in with this PR.  If you haven't started looking at this, just give me the thumbs up to tack it on here.

rbeezer:
Go ahead.  I was just coming around.  I'll do affiliation instead and you can handle the rebasing necessary.

oscarlevin:
This is ready for review now.

rbeezer:
Excellent!  How long did it take you to make an un-numbered footnote in LaTeX.  ;-)

*  "Authored" got misspelled when you moved it.  :-(  Fixed.
*  I took out the CSS, since there is a big CSS contribution coming.  Trying to minimize disruptions.
*  I put in commented-out placeholders for new localizations - I'm going to add that now.
*  No extraneous whitespace!

Great progress, keep it coming!

PR #2307, opened 2024-12-03
added new prefigure PNG with caption

davidaustinm:
This replaces the original PNG of the tactile diagram in the sample article with a new one that includes a caption.  One explanatory sentence is also added to the paragraph above.


rbeezer:
Merged as-is.  Thanks for the update.  Website will be refreshed sometime soon, but not immediately.

PR #2309, opened 2024-12-09
Bugfix single line program

ascholerChemeketa:
I just discovered the recent program update doesn't handle the case where the contents of `<program><code>` does not include any newlines. Samples:

`<program>print("Hello")</program>`
`<program><code>print("Hello")</code></program>`

`substring-before-last` is being used to clean up the trailing whitespace, but that template returns `''` if there is no newline. So currently, the code in those samples is dropped.

First commit adds a parma to `substring-before-last` that allows the caller to ask for the $input back if there is no match. That generates clean diffs for sample-book and sample-article in all formats.

Other commits use that to fix pretext-html and adds a test case to the sample article.

Sorry about this one, I missed that this test case was no where in sample-book or article and thus not getting checked.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -10688,6 +10688,10 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 </program>
             </listing>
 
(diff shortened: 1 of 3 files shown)


ascholerChemeketa:
Force pushed with new template approach instead of a flag on existing template

ascholerChemeketa:
Just rebased to recent master and retested.

Still has clean diffs except for the one new test in sample-article.

rbeezer:
Very good.  Merged as-is.

Sorry - I have no good excuse for why this took this long...

PR #2311, opened 2024-12-10
Iframe resize message handling

ascholerChemeketa:
This updates lti.frameResize to work for all iframes, not just My Open Math

@bnmnetp - It includes your Cay Horstmann sample with you as the author of the commit. 

There are artifacts created in the Horstmann after it resizes - scrollbars appear that then go away after you click on the iframe. I did lots of testing, this appears to be an issue on their side with rendering into the new area. Anything that I could figure out to do on our end to try to fix it was insanely hacky (first resize with 30 extra pixels of width, then a few ms later rerender at correct width). In the simple iframe test I added to the same section there are no lingering scrollbars. So for now, no hackery. 

@drlippman - this builds on your MoM code. I didn't really change it and it all still works with samples in Sample Article. I was wondering why you change the style.width and style.height of the iframe instead of the iframe's width and height attributes. If there is some magic there I am missing, I would love to know about it.

### DIFF
--- a/examples/sample-book/ext/splice/splice-resize.js
+++ b/examples/sample-book/ext/splice/splice-resize.js
@@ -1,0 +1,19 @@
+const bodyEl = document.body;
+bodyEl.style.background = 'red';
+
+const growBtn = document.createElement('button');
+growBtn.textContent = 'Grow';
+document.currentScript.parentElement.appendChild(growBtn);
+
+growBtn.addEventListener('click', () => {
+  const currentHeight = bodyEl.clientHeight;
+  const newHeight = currentHeight + 100;
+  bodyEl.style.height = `${newHeight}px`;
+  window.parent.postMessage(
+    {
+      subject: 'lti.frameResize',
+      height: newHeight,
+    },
+    '*'
+  )
+});
--- a/examples/sample-book/rune.xml
+++ b/examples/sample-book/rune.xml
@@ -4398,6 +4398,24 @@ TEST_CASE( "Test the add function" ) {
         </exercise>
     </reading-questions>
 
+    <section xml:id="splice-integration">
+        <title>Splice Integration</title>
+
+        <p><url href="https://cssplice.org/">SPLICE</url> is a project to supply documentation and infrastructure to help with adopting shared standards, protocols, and tools for web-based learning tools. The project has designed a <url href="https://cssplice.org/slcp/index.html">protocol</url> for an embedded iframe to communicate with its host page.</p>
+
+        <p>One aspect is the ability of an iframe to ask for a new size. Below is a test of that capability.</p>
+
+        <interactive xml:id="splice-resize-example" platform="javascript" width="60%" aspect="2:1" source="splice/splice-resize.js"/>
+
+        <p>And here is a more complex example that uses various parts of the grading protocol.</p>
+
+        <figure xml:id="horstmann-codecheck">
+            <caption>CodeCheck <c>iframe</c></caption>
+            <interactive xml:id="interactive-horstmann-codecheck" iframe="https://codecheck.io/files/wiley/ch-bj4cc-c06_exp_6_105" width="100%" aspect="1:1" />
+        </figure>
+    </section>
+
+
     <section xml:id="atomic-video">
         <title>YouTube Video Embedding</title>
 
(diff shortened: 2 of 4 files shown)


drlippman:
No magic. My understanding is that it's just a matter of personal preference, with some people preferring to keep all styling in CSS.  I probably got that code from somewhere else originally. Feel free to change it if it's more consistent with other practices in pretext.

ascholerChemeketa:
@drlippman Thanks!

bnmnetp:
@ascholerChemeketa - so it looks like this will end up in BOOK/_static/pretext/js/lti_iframe_resizer.js. -- which means that I should include it on assignment pages to ensure that problems resize in that arena as well.

ascholerChemeketa:
It could be moved somewhere else, but yes, that is the current location. 

Eventually, it would be nice to get all of the PreTeXt related JS bundled up better. Perhaps into two piles - the stuff needed by RS assignment pages (knwols, this, etc...) and the stuff that isn't (toc, permalinks, etc...). 

I suppose you could bake in a copy of the logic to the server-generated pages instead of doing that.

bnmnetp:
No, I already load things from `.../BOOK/_static/` so its not a stretch to add it.

I agree that a bundle of PreTeXt stuff needed by the RS assignment pages would be great.

I don't think baking it in to the pages is a great idea.

rbeezer:
Do we know if this interacts with fram resizing for WeBWorK?  There is an example in the distribution, need to run makefile for problems processed by server **then** make HTML.  I think this is a good test example:

https://pretextbook.org/examples/webwork/sample-chapter/html/section-the-quadratic-formula.html

Checkpoint 1.2.2
Do part (a) correctly and watch part (b) reveal itself and resize.

bnmnetp:
I don't know, but we should strongly encourage WeBWorK to adopt this protocol if they don't, as the lti_reseze message is already supported by other LMS.

Alex-Jordan:
If someone is able to explain to me what this is about, perhaps I can comment on this and WeBWorK.

Status quo. A WeBWorK server hosts a build of this project: https://github.com/davidjbradshaw/iframe-resizer. There are certain features that we like and/or are necessary. One of those is that you can tag certain elements to be monitored for where their bottom is, and make sure to include those in the iframe height calculation. Even when they are things that would normally be OK to extend below out of the page. This comes up with the feedback buttons, which in 2.19 can produce a tall popover that can dip below the iframe edge without having tagged that popever's div as one of these things to monitor.

With PTX, the web page with the embedded exercise loads `iframeResizer.min.js` from where the WW server is serving it. The page within the iframe also loads `iframeResizer.contentWindow.min.js` from the same place. These two things go together.

Feel free to take discussion to the forum if it's not right to continue about this here.

bnmnetp:
OK, it sounds like this isn't something we need to worry about for WeBWorK, although both are doing similar things.  Using the `postMessage` interface to communicate from child to parent and vice versa.   

I haven't seen the code that @ascholerChemeketa is working on, but this seems a lot more complex and more likely fully featured that what the SPLICE project has provided.

ascholerChemeketa:
Tested against the webwork sample - it works correctly.

Yes, @bnmnetp is right. The `iframeResizer` library is a much more capable version of what the SPLICE protocol provides (which is in turn borrowed from the lti protocol that people like Canvas all agree on.) 

So we have two different protocols running to resize elements in an iframe. They should not interact unless a page included as an iframe was trying to use the `iframeResizer` library AND emit `lti.frameResize` messages. In a perfect world, we would not need both. But for now I think we wo.

rbeezer:
*  Small merge conflict due to head/foot caches.  Got it straight, I think.  At least the GROW example functions well.
*  New `interactive` need new QR codes and automatic preview images.  Yes, GitHub is not a cache.  But my sanity is precious.  I added them.
*  Brad added a bunch of stuff like this already.  Now we have two "SPLICE Integration" sections.  Can you harmonize them into one?  Watch out changing `@xml:id` (which should really be `@label`?) or generated bits will have name changes.
*  Lots of stray whitespace in `js/lti_iframe_resizer.js`.  It'd be good to clean that out while in the neighborhood.

Thanks for this.  Resizing `iframe` has been a long-running fiddle.

PR #2313, opened 2024-12-11
Count pad length speedup

ascholerChemeketa:
When building sample-book for html, xsltproc identifies count-pad-length as the template that accounts for the second most time ~12-15% of the total.

This reimplements count-pad-length to use counters as parameters instead of strings to avoid lots of string building and slicing. That provides ~3-4x speedup. 

Second commit changes the left-margin template to only pass in the first line to count pad length. Which prevents count-pad-length from passing copies of a possibly giant multiline string to itself over and over.
That is another ~3-4x speedup

count-pad-length now accounts for ~1-1.5% of the total time.

Generates clean diffs on the sample book & article.


rbeezer:
Very nice.  Yes, a sliding index is much more sensible.  And I'm seeing a time savings that is a double-digit percentage.

One cosmetic edit: I made `cur-index` into simply `index`.

PR #2314, opened 2024-12-11
escape-json-string speedup

ascholerChemeketa:
escape-json-string accounts for ~%5 of an HTML build of sample book. Mostly generating the `lunr-pretext-search.js` file.

This refactors that template to do one pass of replace using node-sets of search and replace items instead of multiple replace passes with strings.

Only about 10-15% speedup on execution time for this one :( 
But it is a real speedup.

Clean diffs on sample-book


rbeezer:
I've not read the documentation of `str:replace()` in a long time.  So probably never recognized it worked this way.  Much cleaner.

When I build the sample book for a Runestone server, I'm seeing run times that are as much as 20% **longer** with this change.  What were your experiences with timing and profiling?

rbeezer:
Merged this.  Not sure why my timings are so unreliable (too lazy to fire up the profiler...).  There's a ubiquitous LaTeX template that could use this treatment.  (Ignore my previous comment.)

PR #2315, opened 2024-12-11
Affiliation for authors, editors, and contributors

oscarlevin:
As a step toward better frontmatter, this allows for an author (or editor or contributor) to have their `<department>`, `<institution>` and `<location>` wrapped in an `<affiliation>` element.  This will allow for multiple affiliations and referencing common affiliations later.

The current PR results in no changes to the latex output of either the sample article or sample book.  HTML output looks unchanged as well.

`<location>` was already a valid child of `<author>` for `<contributor>`, so I moved that to also be allowed for `<author>` and `<editor>` rather than introducing a `<address>`.  

The first commit is just schema.  It duplicates some of the changes in #2306, so either this or that might create a conflict depending on which is merged first.  I'm happy to resolve that if it occurs.

### DIFF
--- a/schema/pretext.rnc
+++ b/schema/pretext.rnc
@@ -197,21 +197,21 @@
             Bibinfo =
                 element bibinfo {
                     (
-                        (Author, Author*, Editor*)
+                        (Author+, Editor*)
                         |
-                        (Editor, Editor*)
+                        (Editor+)
                     ),
                     Credit*,
                     Date?,
                     ColophonCredit*,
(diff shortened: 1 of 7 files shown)


rbeezer:
Very good.  Exploded it out a bit, and dropped a whitespace character in the schema, but no code changes.  #2306 will need some rebasing, I suspect.  Thanks!

oscarlevin:
Awesome.  I'm almost done with additions to #2306.

PR #2316, opened 2024-12-11
Additional splice examples for sample-book

bnmnetp:
This also contains a project.ptx file that works and a working project.ptx file for the webwork testing book.


rbeezer:
I could cherry-pick the SPLICE examples.  I'm a bit perplexed by the rest.

Big merge conflict when I try to rebase, all the codelens attribute stuff.  Maybe that should be a different PR?  Anyway, I generally don't need nor want merges as part of a PR.

The CLI project files have things like:
```
debug.rs.dev="yes"
```
```
<stringparam key="debug.rs.services.file" value="file:////Users/bmiller/rs/bases/rsptx/interactives/runestone/dist/webpack_static_imports.xml" />
```
Aren't those going to cause problems for others?

bnmnetp:
For now, go ahead and cherry pick the SPLICE examples and close.  I'll get a fresh fork and do the update to the project.ptx file separately.

rbeezer:
Tried to cherry-pick the examples.  The `interactive` all have the same `@xml:id`.  And would likely do better with a `@label`.

They'll need screenshots and QR codes for the repository, since we don't use the CLI to build these.

A small lead-in paragraph explaining what these are (URL to DSA?) would be helpful.

rbeezer:
Since I'm such a nice guy ;-) I added the SPLICE examples, built and added the necessary images so the PDF would build, and added a bit of explanation (which @bnmnetp maybe should check).

Other parts of this are now on #2327.

Website rebuild will happen in the morning.

PR #2317, opened 2024-12-12
Eliminate code formatting recursion issues

ascholerChemeketa:
This should come after:
https://github.com/PreTeXtBook/pretext/pull/2313

This replaces the text utilities used for code formatting with versions that should not cause recursion issues. With all of them in place, we can now process a 2000 line, heavily indented code chunk with a `--maxdepth` of < 200. At this point, the thing most likely to cause a recursion depth error is the `printIndented` code in `xml-to-json`.

These changes are all more or less time-neutral.

Clean diffs on sample book.


ascholerChemeketa:
@rbeezer I think I'm done messing with text-utils for now.

rbeezer:
I like it.  Great improvements.

rbeezer:
> I think I'm done messing with text-utils for now.

So soon?  Seems like you are just getting up to operating temperature...

PR #2318, opened 2024-12-13
Speedup xml-to-json indent

ascholerChemeketa:
The `"indent"` template in xml-json appears to have time complexity that is exponential in the height it is called from. And it gets called many times in the production of a blob of json. It accounts for ~5% of the xsltproc build time of the sample book. In documents with lots of short pages, it can account for 25+% of the build time.

The template is only used to make the indentation for the JS that gets added to pages for sage cells and mathjax.

This patch turns it into a non-entity in terms of build time and generates clean diffs on sample book.

### DIFF
--- a/xsl/xml-to-json.xsl
+++ b/xsl/xml-to-json.xsl
@@ -212,9 +212,7 @@ done`
 
     <xsl:template name="indent">
         <xsl:param name="depth" select="'0'" />
-        <xsl:for-each select="//*[position() &lt;= $depth]">
-            <xsl:if test="position() &lt;= $depth"><xsl:text>  </xsl:text></xsl:if>
-        </xsl:for-each>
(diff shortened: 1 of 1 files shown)


rbeezer:
That was a good template to target!  ;-)  We have a `duplicate-string` template, so I experimented with that.  For the sample article (more Sage maybe, definitely more MathJax):

* Before: 6.47%
* PR: 0.04%
* "duplicate-string": 0.24%

(The last assumes I'm reading the profiler right.)  The slightly faster version would eventually stop adding indentation, so at least would have required some sort of warning when that threshold was reached.  The change will always be correct.

If the extra 0.2% is too much to bear, we could rip out the XML-to-JSON drill and replace with "formatted" `xsl:text` elements, but that would need discussion on `-dev`.

Thanks for all these speed-ups!

PR #2319, opened 2024-12-13
Speedup and fix serialize-namespace

ascholerChemeketa:
After other speedups, `serialize-namespace` now accounts for ~30% of the time reported by xsltproc.

The offender appears to be the ancestor scanning. This replaces it with a list of known offenders that tend to leak into HTML output.

This produces clean diffs in the sample-book, but it clearly could change behavior... but I'm not sure what the original intent was. 

I can also get clean diffs by just never calling `serialize-namespace` from `serialize`. 

This was added as part of "Watermarks in HTML and LaTeX (PR #1018)" but I can't see what it 

@Alex-Jordan can you provide insight?

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -1875,8 +1875,10 @@ Book (with parts), "section" at level 3
 <!-- original element declaration. And then serialize what is left.                      -->
 <xsl:template match="*" mode="serialize-namespace">
     <xsl:for-each select="./namespace::*">
-        <!-- test taken from http://lenzconsulting.com/namespace-normalizer/normalize-namespaces.xsl -->
-        <xsl:if test="name()!='xml' and not(.=../preceding::*/namespace::* or .=ancestor::*[position()>1]/namespace::*)">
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
@rbeezer 
After our discussion, I made some changes. This version only tries to copy namespace if the `authored-source` param is set.
It also changes the logic for figuring out the namespace to compare current vs parent.  This correctly handles the prefigure example from the annotated sample article that was not being produced by the unmodified version of `serialize-namespace`

ascholerChemeketa:
@rbeezer 
After our discussion, I made some changes. This version only tries to copy namespace if the `authored-source` param is set.
It also changes the logic for figuring out the namespace to compare current vs parent.  This correctly handles the prefigure example from the annotated sample article that was not being produced by the unmodified version of `serialize-namespace`

rbeezer:
Clever!  I like it.  I did an annotated sample book a few days ago.  Took about 12 minutes.  Now about 30 seconds.  The accumulation of all the speed-ups the past few days.  Thanks.

PR #2320, opened 2024-12-13
variable to hold common head and end of body content for HTML

ascholerChemeketa:
This makes a variable to hold the head and end of body contents inserted into each page that is common to all pages generated for a document.

It provides a decent speedup to the `file-wrap` template, but also serves as a way to allow different page types to leverage the same list of insertions. I have not extended this to `standalone-page` or `create-iframe-page`, but we could do so. They would be a subset of `file-wrap-simple-XXX` varaibles.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -6842,12 +6842,12 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 <!-- A hook could go here for some extras       -->
                 <!-- ########################################## -->
                 <xsl:call-template name="pretext-js" />
-                <xsl:call-template name="knowl" />
                 <xsl:call-template name="fonts" />
                 <xsl:call-template name="font-awesome" />
                 <xsl:call-template name="diagcess-header" />
                 <xsl:call-template name="css" />
                 <xsl:call-template name="runestone-header"/>
+                <xsl:apply-templates select="." mode="knowl" />
             </head>
             <body>
                 <!-- potential document-id per-page -->
@@ -10721,7 +10721,6 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:call-template name="myopenmath-js"/>
     <xsl:call-template name="syntax-highlight"/>
     <xsl:call-template name="pretext-js" />
-    <xsl:call-template name="knowl" />
     <xsl:call-template name="fonts" />
     <xsl:call-template name="hypothesis-annotation" />
     <xsl:call-template name="geogebra" />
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
The diffs for this are not clean as one or two page specific items had to be shifted as did the search js that is only part of the full page and not simple.

But if you examine them, you should see a few chunks that just shifted relative order.

ascholerChemeketa:
Force pushed to rebase against master post Oscar's update

rbeezer:
I'm getting bit by what I am forever warning everybody else about.  :-(

The `knowl` template is a named template.  But it applies the `type-name` template, which uses context to localize the button text ("Evaluate" in English).

*  The `knowl` template should move to a `match="*"` and an `xsl:apply-templates`.  My fault: I think I know how that crept in a while ago.
*  Then it should be obvious that this template can't be rolled up into a massive pre-built variable.
*  Sample article produced a warning.  Is there no Sage in the sample book?  Perhaps not.  The sample article has everything but the kitchen sink, or close to that.
*  I'd love to see this approach enacted for the other page-construction templates.  Hint, hint.

ascholerChemeketa:
OK, will go look into those and implement for `standalone-page` or `create-iframe-page`.

ascholerChemeketa:
Updates with requested changes.

Notes:
- `standalone-page` does not appear to have had mathjax previously, it looks like it should so I added it.
- `standalone-page` had the runestone ethical ads, which doesn't look like it belongs there. Removed it.

Diffs are even messier now as factoring out the templates into subsets required reordering templates that were called in different order in different places. :(

rbeezer:
I had this one in my sights.  Is there a reason it was closed?  I was not neglecting it.

ascholerChemeketa:
Sorry - was doing local pruning and got overzealous. Closure was unintentional.

rbeezer:
Rebased on master, and got one merge onflict, which I think I handled OK, and I think is unrelated to next item.

PR has
```
<xsl:apply-templates select="$document-root//ol" mode="ol-marker-style"/>
```

Existing code defines
```
<xsl:template match="ol-marker" mode="ol-marker-style">
```

Produces boatloads of unfomatted list content.  Which is typical when a modal template has nowhere to go.

I think maybe some of Jeremy's work slid in here in the meantime.  Maybe #2402, #2377, #2177.  But I have not explored.

Are you willing to rebase (again!) and clean-up the bitrot (which is all my fault!)?  Thanks.

davidaustinm:
This is a request to add the diagcess library to the head of a slideshow and then initialize it.

ascholerChemeketa:
I found the chunk of list styles you were talking about and fixed that. The rest appears good.

Force-pushed after rebase and fix to existing commit

rbeezer:
Very good!  Merged as is, but for one whitespace hunk.  I'm going to put a small warning on `-dev` just due to all the movement.

PR #2321, opened 2024-12-14
check for sage before doing lots of checks

ascholerChemeketa:
Simple check to see if there are ANY sage elements on the page before doing the 20+ detailed scans of the page.

Small but real efficiency win.

Clean diffs on sample book.


rbeezer:
Added an extra efficiency - we already have a `b-has-sage` global variable.  ;-)  **THANKS!**

PR #2322, opened 2024-12-14
Better handling of duplicate checks for label/xmlid

ascholerChemeketa:
This consolidates the logic for `duplication-check-label` and `duplication-check-xmlid` into one place.

That logic is implemented to speed it up. Instead of trying to select duplicates of each node, this constructs a sorted list of the attributes being compared and then searches them linearly. Old logic was essentially n^2 where n is the number of nodes being compared, new logic should be something line nlogn + m^2 where m is the largest number of identical values.

The commit to `hello-world` is just for your testing purposes - you can leave it on the cutting room floor. It demonstrates that the logic still works, and the speedup for a document with lots of labels.


rbeezer:
Another win!  Thanks, I like it.  The old template had some odd sporadic behavior when I was first building it.  I won't miss it.  I like this a lot better.

PR #2323, opened 2024-12-16
Fix unconverted close input tags

ascholerChemeketa:
This fixes two `</input>` tags that did not become `<\code> tags in 09371e5bfeac67c4d4dbe4d2ab090e4d041e4789.

Apologies for letting those slip through!


rbeezer:
Got it.  Thanks.  I don't think anybody (including me) has noticed.  ;-)

Rebuilding website now...

PR #2291, opened 2024-11-12
Css bundling

ascholerChemeketa:
Draft for @rbeezer to poke at the XSL.

Look for `@Rob` in `publisher-variables.xml` for my failed attempt to set an `html-theme` variable via key(). The workaround version is right below.

The `publication-structural.xml` file is set to build the new `default-modern` themes via `<css theme="...">`. Anything that has `<css style="...">` will be built using old CSS with a warning. It can be used without node, including changing the palette (see `pretext.py` for test palette options). Any other theme (`salem`, `denver`, `tacoma`) will require node is installed and doing `npm install` one time from the `script/cssbuilder` directory.

You can rebuild just the theme with `-c theme` on the pretext script. 

This RS PR is related:
https://github.com/RunestoneInteractive/rs/pull/575
It is not required, but it enables dark mode on RS components and other minor adjustments to make RS components look more like the PTX around them.

Needs:
- [x] Oscar is still doing some testing/development
- [x] More documentation
- [x] Check on CLI - Oscar says it should be ready to go
- [x] Rebase - integrate any very recent CSS
- [x] Update other examples/guide to use new themes?

PR #2324, opened 2024-12-16
CSS update prebuilts 2024-12-16

ascholerChemeketa:
This just updates the prebuilt files with recent CSS changes.

Once css-bundling PR is applied, this should only have diffs in `css/dist/`


ascholerChemeketa:
Integrating into #2291

PR #2325, opened 2024-12-16
Node instructions update

ascholerChemeketa:
Updated guide instructions for installing node.

This is based offof the last commit in css-bundling as it updates the changes already in that PR. Diffs should be minimal once that PR is applied.


rbeezer:
Grabbed the final three commits here.  Thanks for the attention to the documentation.  Website updates perhaps later tonight.

PR #2326, opened 2024-12-16
EPUB missing image warning

ascholerChemeketa:
Currently, if the packaging_tree that is built for an EPUB ends up with images that are missing a `sourcename` it results in a broken build. This is happening when I try to build the `sample-book` with `-f epub-svg`.

This produces warnings for each image that does not end up with the right information but allows the build to continue.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -3397,10 +3397,13 @@ def epub(xml_source, pub_file, out_file, dest_dir, math_format, stringparams):
     # position image files
     images = packaging_tree.xpath("/packaging/images/image[@filename]")
     for im in images:
-        source = os.path.join(source_dir, str(im.get("sourcename")))
-        dest = os.path.join(xhtml_dir, str(im.get("filename")))
-        os.makedirs(os.path.dirname(dest), exist_ok=True)
-        shutil.copy2(source, dest)
+        try:
+            source = os.path.join(source_dir, str(im.get("sourcename")))
+            dest = os.path.join(xhtml_dir, str(im.get("filename")))
+            os.makedirs(os.path.dirname(dest), exist_ok=True)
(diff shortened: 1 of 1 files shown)


rbeezer:
> results in a broken build

In what way?  I get an EPUB that is readable by `calibre` and  `epubcheck` (though the former is not always pretty and the latter has errors).

It appears to me that `@sourcename` really shouldn't be empty.  Might it be better to get to the root of the problem?

ascholerChemeketa:
It looks like the cause of the images missing expected attributes are the ones injected into the source tree in the representations pass:
https://github.com/PreTeXtBook/pretext/blob/8a0b43ccf90be07afc280e5fadcc3cf71e4869e0/xsl/pretext-assembly.xsl#L2769

Should I turn this into an issue based on those images not meshing well with the `packaging-info` template?

rbeezer:
> It looks like the cause of the images missing expected attributes are the ones

I was suspicious of that, and went looking, and I think I got distracted by some other rabbit holes...

> Should I turn this into an issue based on those images not meshing well with the packaging-info template?

Yes, please!  Don't get rid of this - the warning will be good for when we mess this up again.  ;-)

ascholerChemeketa:
Issue created for root issue:
https://github.com/PreTeXtBook/pretext/issues/2345

rbeezer:
Merged.  A bit of economizing in the code, and an expanded error message (mentioned GH issue).  Thanks for this.

Will scribble later today on #2345 with some insights just devloped.

rbeezer:
The root cause here has been addressed.  See d37568e52128aa34be581728c710f384ae5d2f68 and perhaps some nearby commits.  I removed the reference to this issue from the error message.

PR #2328, opened 2024-12-18
Cache a copy of runestone services xml file and use that for checking…

ascholerChemeketa:
The html function in pretext.py is using the runestone tar file as its check for whether to use the existing runestone files in `_static`, but it deletes that file. So it never decides to use the existing install.

This writes a copy of the runestone services XML file to the output directory and uses that file to check for an existing install.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -1063,7 +1063,7 @@ def dynamic_substitutions(xml_source, pub_file, stringparams, xmlid_root, dest_d
     if "debug.rs.dev" not in stringparams:
         # See if we can get Runestone Services from the Runestone CDN.
         #  "altrs" = alternate Runestone
-        altrs_js, altrs_css, altrs_cdn_url, altrs_version = _runestone_services(stringparams)
+        altrs_js, altrs_css, altrs_cdn_url, altrs_version, services_xml = _runestone_services(stringparams)
         # Previous line will raise a fatal error if the Runestone servers
         # do not cooperate, so we assume we have good information for
         # locating the most recent version of Runestone Services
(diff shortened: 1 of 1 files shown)


oscarlevin:
As I'm reading this, we are talking about a small xml file that will now be included in the output, and therefore deployment, of html/runestone builds.  I guess there isn't anything wrong with that, but it seems like storing this in the common ~/.ptx directory the CLI already uses for this sort of stuff would make sense.  As would keeping the actual rs services files, for that matter.

ascholerChemeketa:
@oscarlevin 

I'm agnostic on the file location. But I am pretty sure none of the logic in the pretext script and `pretext.py` know nothing about `~/.ptx`. So I'm inclined to leave this in `_static` unless Rob wants to merge `~/.ptx` into the fold as part of those.

By actual services file, do you mean the .tgz package? Rob wants it gone in the final output (which makes sense). So absent some other cache location, there is nowhere good to put it.

rbeezer:
Another win!  I was able to insert this into the middle of work I am doing in this area (and wanted to include).

One little gotcha:
```diff
-        services_record_output_path = os.path.join(dest_dir, "_static", services_record_name)
+        services_record_output_path = os.path.join(tmp_dir, "_static", services_record_name)
```
You might not have seen that if you didn't trash the eventual `dest_dir` after each test run.

Not pushed yet, I've got more work surrounding it, but it should show here eventually (soon?).  Closing now.

rbeezer:
Pushed now.  There was some more confusion with filenames and if the temporary or output directories were being written to or read from.  Refactored, so now hopefully clearer.  Announcement on `-dev` shortly.  Thanks for your work on this one.

PR #2330, opened 2024-12-18
Add performance audit note to strip-leading/trailing-whitespace

ascholerChemeketa:
Just adds comments that I checked for optimizations on these templates and found none.

Kind of dumb... but probably will keep me from redoing the work in 6 months when I look at them and say "Why didn't I try to speed those ones up?"


rbeezer:
Yes, I will often go back and attempt the same thing six months later.  Thanks for improving the code.

PR #2331, opened 2024-12-18
Limited lifespan for element deprecation warnings

ascholerChemeketa:
The deprecation warnings are expensive to process in a large book. For many authors, checking 10-year-old warnings is moot. But for a long-neglected book, doing so might be important.

This modifies the deprecation-message calls to only process the source if the warning is less than some `max-age`

Currently `max-age` defaults to 5. There is a new stringparam `author.deprecations.max-age` that can be used to set something different - either `none` or an integer number of years worth of checks to run.

Build time for sample book with different `max-age` values:
```
no max      : 30 seconds
5 years     : 22 seconds
2 years     : 18 seconds
0 years     : 16 seconds
```

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -32,7 +32,8 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     xmlns:date="http://exslt.org/dates-and-times"
     xmlns:exsl="http://exslt.org/common"
     xmlns:str="http://exslt.org/strings"
-    extension-element-prefixes="pi exsl date str"
+    xmlns:dyn="http://exslt.org/dynamic"
+    extension-element-prefixes="pi exsl date str dyn"
     xmlns:mb="https://pretextbook.org/"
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
@rbeezer If you like the approach, and do want the really old deprecations. I'll go pull them out of history and add them back in a new commit. It looks like they are in a mix of formats, so I won't do that unless you say "go".

rbeezer:
Reviewing on fone.  How about giving an author only two choices: "recent" and "all"?  We can't make it too easy to squelch these and then pay the price with difficult support questions.

rbeezer:
I can't immediately see the reason for `dyn:evaluate`.  I should be at Drop-In in a bit.

ascholerChemeketa:
Force pushed with changes discussed in Drop in. stringparam now only offers a boolean choice between current warnings or all warnings.

rbeezer:
Fantastic!  Added a little follow-on commit to error-check the parameter.

Several messages coming on `-dev`.  Thanks.

PR #2332, opened 2024-12-19
Work sheetprinting 1

davidfarmer:
Adjusted the length of a worksheet in letter and a4, so that the worksheets in the sample article print as expected.

Also deleted old code and fixed trailing whitespace.

I had to make the numbers smaller than previously mentioned in pretext-dev.


rbeezer:
Very good!  Thanks for the maintenance.  Limited test works great in Firefox on Linux.

Will update website examples just now.  13 minutes.

**"Worksheet Printing is Back!"**  Announce on the recent `-dev` thread, or more widely on `-announce`?  Your call.

PR #2334, opened 2024-12-19
"level" template transition next step

ascholerChemeketa:
This attempts to move towards only using levels determined by the "augment" pass.

Marked as a draft until I can check in with @rbeezer about it.

PR #2335, opened 2024-12-20
Html toc single build

ascholerChemeketa:
The ToC is a significant chunk of each HTML page and is mostly the same on every page. This PR precomputes the ToC into a variable and then on each page augments that with the necessary `active` and `contains-active` classes.

Reduces xsltproc time by ~15-20% for sample book. For books with lots of smaller pages savings can be 35%+.

Clean diffs on sample book.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -11553,6 +11553,85 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     </div>
 </xsl:template>
 
+<!-- Table of Contents per-page customization      -->
+<xsl:template match="*" mode="customized-toc-items">
+    <!-- get a copy of the toc-cach that we will decorate -->
+    <xsl:variable name="toc-contents-rtf">
+        <xsl:copy-of select="$toc-cache-rtf"/>
+    </xsl:variable>
+    <xsl:variable name="toc-contents" select="exsl:node-set($toc-contents-rtf)"/>
+
+    <!-- get the unique id of the current page -->
+    <xsl:variable name="uid" select="@unique-id"/>
+    <!-- use that to find the ToC node for that page -->
+    <xsl:variable name="this-page-node" select="$toc-contents//*[@uid = $uid]"/>
+    <!-- ancestor list will allow us to identify when we are in the path to the page -->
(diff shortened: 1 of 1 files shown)


rbeezer:
That is brilliant!  Cleaner code, no-change diffs, and a speed-up.  (And a nice sequence of commits.)

Will update website shortly, a few more PRs to do.  Keep 'em coming!

oscarlevin:
Uh-oh.... this seems to have broken revealjs builds.  I get a lot of:

```
error: * PTX:BUG:   an element ("frontmatter") does not know its level
error: * PTX:BUG:     asking if a "frontmatter" is a leaf, for a document that is not a "book" nor an "article"
``` 

starting at the commit that these were merged.

ascholerChemeketa:
It looks like it still generates the show correctly, but it does produce those scary-looking messages.

They all come from where the ToC is being built into a variable.

The quick and dirty fix is to add a variable that pretext-revealjs.xsl sets and pretext-html checks while building that variable. That may be part of the right fix as well - there is no reason that pretext-revealjs.xsl needs the ToC generated. It does feel a little ugly to have the more generic sheet (pretext-html) rely on hints from the most specific sheet, but that may be unavoidable.

It does feel like there is a bigger issue here. Shouldn't we be able to provide the answer to "is-leaf" and "level" for "frontmatter" in a slideshow? If that was fixed the bug reports would go away. That seems like it has implications for https://github.com/PreTeXtBook/pretext/pull/2334

rbeezer:
Reviewed the PR and the RevealJs stylesheet.  Not seeing it.

> Shouldn't we be able to provide the answer to "is-leaf" and "level" for "frontmatter" in a slideshow?

Might just be a matter of defining a level for `slideshow`? And/or `slideshow/section`.

I'll get back to this in a bit, when I am able.

rbeezer:
>  there is no reason that pretext-revealjs.xsl needs the ToC generated

So we override the `toc-cache-rtf` global variable in the derived worksheet.  On it.  Push in a minute.

rbeezer:
Fix in at: 2fb780c35d73aa2461219a6413b596788b1c5f23

Thanks @oscarlevin for the heads-up and @ascholerChemeketa for the hints.

oscarlevin:
Can confirm that the most recent commit fixed the issue.  Thanks!

PR #2337, opened 2024-12-24
fix revealjs function call

oscarlevin:
This corrects an outdated function call in the revealjs build function of the pretext script.  

Note that there is no CSS being copied at all any more, but this should be fine as the revealjs css appears to be completely hard-coded into the page.


rbeezer:
Fix is in.  Nice job with the automated catch, and thanks for the fix.

PR #2338, opened 2024-12-26
Translated 'Code block' and 'Keywords'

valelav:


rbeezer:
Thanks very much for updating this!

PR #2339, opened 2024-12-27
bugfix - large logo image on safari

ascholerChemeketa:
Fix for the issue reported here:
https://groups.google.com/g/pretext-support/c/S1R3xQ_kPuo/m/CI0Xh1l0BgAJ?utm_medium=email&utm_source=footer

I don't have a Safari browser available, but the same issue is apparent with the Epiphany browser (also webkit) and this fix works there.


rbeezer:
Went ahead and merged this - more on -support.  Thanks!

(Did a quick force-push to get commit message right.)

PR #2342, opened 2024-12-29
Add a result cache to get_publisher_variable to store processed XML data

ascholerChemeketa:
Calling `get_publisher_variable` multiple times from Python redoes the possibly expensive work of running the XSL report and parsing it. This saves the results of that work for use in future calls to the function.

On APEX Calculus, this and https://github.com/PreTeXtBook/pretext/pull/2341 reduce the build time of APEX from 20m+ to ~2.5m.


rbeezer:
Really good!  Merged as-is, only with a new commit message.

Curious to know why an attribute on a function might be superior to a module-level variable, such as our `__temps`?  That could wait for Drop-In, no need to answer here.

Thanks for YASU.

PR #2343, opened 2024-12-30
CSS: better handling of coloraide python library

ascholerChemeketa:
coloraide library is used for contrast-checking custom color choices but it was not added to `requirements.txt`.

This adds the library and makes its absence a non-fatal warning.


rbeezer:
Thanks for the quick fix!

PR #2344, opened 2024-12-30
CSS: fix SCSS palette variable name

ascholerChemeketa:
Late in the SCSS development process, I started to rename a variable from `$palette` to `$color-scheme`. That never got finished and ended up on an alt branch which got merged in during a late-stage rebase. This is causing attempts to control the palette from PTX to not work.

This restores the original name and fixes the issue.

Tested against multiple themes and color options.


rbeezer:
Merged.  Thanks.

PR #2347, opened 2024-12-31
EPUB: add css.map file to the manifest

ascholerChemeketa:
@rbeezer - this adds the epub.css.map file to the epub and its manifest. 

I've opened up the epub file and confirmed that it is in the right place. Calibre still reports not being able to find the file. epubcheck says it can't parse it. I *think* that epubs just can't deal with sourcemap file.

Sharing in case you want to test more. If not, feel free to close the PR - recreating the work will be trivial when someone is ready to do an EPUB deep dive.


rbeezer:
Thanks for running through this!  We can always bring it back if/when we get serious abut styling EPUB.

PR #2348, opened 2024-12-31
add optional external runestone methods for the CLI to hook into

oscarlevin:
This slightly refactors the runestone services logic, creating a new function that does the downloading of the services xml file, and adding hooks to the `html` and `dynamic_substitutions` functions.  PreTeXt expects this function to either be `None` or to have keyword arguments, including a `format` that can be either "xml" or "tgz" to distinguish whether it should just do the services query (xml) or get the entire tgz file.

Behavior of the pretext script should be unchanged when `ext_rs_methods` is `None`.

Happy to rename variables if you have a better suggestion.

### DIFF
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -719,7 +719,7 @@ def main():
         ptx.datafiles_to_xml(xml_source, publication_file, stringparams, args.xmlid, dest_dir)
     elif args.component == "dynamic":
         ptx.dynamic_substitutions(
-            xml_source, publication_file, stringparams, args.xmlid, dest_dir
+            xml_source, publication_file, stringparams, args.xmlid, dest_dir, None
         )
     elif args.component == "theme":
         ptx.update_theme(xml_source, publication_file, stringparams, dest_dir)
@@ -734,6 +734,7 @@ def main():
                 extra_stylesheet,
                 out_file,
                 dest_dir,
+                None
             )
         elif args.format == "html-zip":
             # no "subtree root" build is possible
@@ -746,6 +747,7 @@ def main():
                 extra_stylesheet,
                 out_file,
                 dest_dir,
+                None
             )
         elif args.format == "revealjs":
             ptx.revealjs(
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -979,7 +979,7 @@ def tracer(xml_source, pub_file, stringparams, xmlid_root, dest_dir):
 #  Dynamic Exercise Static Representations
 #
 ################################
-def dynamic_substitutions(xml_source, pub_file, stringparams, xmlid_root, dest_dir):
+def dynamic_substitutions(xml_source, pub_file, stringparams, xmlid_root, dest_dir, ext_rs_methods):
     import asyncio  # get_event_loop()
 
     # external module, often forgotten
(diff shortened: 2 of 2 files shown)


oscarlevin:
Just force-pushed a fix so that dynamic_substitutions uses the same hook, and for the pretext script to use the correct signatures.

rbeezer:
> Just force-pushed

You read my mind.  I meant to comment yesterday, but it got away from me.

OK, no more time today to investigate, but in the interest of moving this along.  I think Brad uses some `debug.*` Runestone features for testing, perhaps with the CLI.  Will the setup here still allow that, without you duplicating the debug logic in the `external_*` function?

oscarlevin:
This now includes the fix for `debug.rs.version`, whose presence will prevent the external method from being called.

oscarlevin:
Just added a commit that changes the timeout for the requests.  Instead of 10 seconds for both connecting and reading the contents, this now timeouts after 1 second for the initial connection, but still allows 10 seconds to complete the task.  That way, if there is no response, we know sooner.

oscarlevin:
Heads up; I'm going to refactor this PR to make it easier for the external method to give back the correct info.

oscarlevin:
All done.  Now a single commit with the complete changes described in the comments above.

rbeezer:
Thanks.  Merged as-is, with just commit message edits.  Sorry for the (long) delay.

oscarlevin:
I see three reasonable ways to implement these hooks (as well as the ones for image generation, which would actually be simpler).

1. As they are now.
2. Call the hook `rs_method`. Test if it is `None`, if so, then set it equal to the default function.  Don't do any checking later, since `rs_method` is what we want it to be.
3.  Same as 2, but using the default function as the default for a keyword argument.  As in `def _runestone_servces(stringparams, rs_method = query_runsteon_services):`

You have said that you don't like default keyword arguments, so perhaps Option 2 is the best way to go.  I don't know whether that or Option 1 is more easy to understand, but at least you see what the argument is doing right away.

Happy to implement this if you think it would be worth doing.  Of course my options above are slightly more complicated for the rs stuff since there are really two different methods hidden in this one, but that shouldn't be a blocker.

oscarlevin:
Also, FYI, I just released 2.14 that has this and the corresponding CLI functions.  Should be very useful for our friends in Africa.

rbeezer:
> Should be very useful for our friends in Africa.

Yes!  Definitely.

rbeezer:
Some (personal) clarity while thinking overnight.

Two places in previous commit in `pretext/pretext` where `None` was added as a parameter.  If you are new to the code, chasing down what that does might be mildly mysterious.  Change those to something like `ptx.status_quo_rs_stuff`.

Now, in `pretext.py` define a method `status_quo_rs_stuff`.  Near `html()` or near related RS support methods.  Document it **very heavily**.

This way, if a third consumer comes along (not you, not me), it is clear how they can provide a different override.  And no keywords with defaults.  And status quo is encapsulated in `status_quo_rs_stuff`.

And maybe before starting, `ext_rs_method` could get a cosmetic name change?  I think of it as less "external" and more "optional" or "overrideable" or "custom".

PR #2349, opened 2025-01-01
add hooks for external image generator functions

oscarlevin:
Only applies to sage, asymptote, and latex-image.  

For each of these, I essentially factored out the inside of the `for image in...` part into its own function, and then either call that function or call the `ext_converter` function if it is not none.   In some cases, this required a little massaging.  In particular:

- Removed the `devnull = open(os.devnull)` in favor of just setting `stdout="subprocess.DEVNULL` as needed (so that we don't have to pass `devnull` to the `individual_*` functions)
-  Moved the conditional import statements inside the `individual_*` functions.
- Set default values for some parameters in the asymptote conversion.  The logic here could be further refactored I think: set up all these various commands inside a dictionary so we don't have to pass so many arguments to `individual_asymptote_conversion` function.  I didn't want to tinker too much though.
- For keeping track of failed latex_images, I wrapped the call to `ext_converter/individual_latex_converter` in a try/except, and add the image to `failed_images` in the except case.  Then were we previously added the image to the `failed_images` just raise an exception.

Also fixed signatures in other places that call these converters, both in `pretext.py` and `pretext`


rbeezer:
Looking good.  Thanks.

*  Did this first, since it looked easier to understand.
*  Built SVGs for the three components affected.  Diffs, before/after, look OK, but amazingly different for `sageplot`, even if two runs are done with no change in the code.
*  Perhaps obvious.  But I like the external functions and the internal functions having the same signature.  I bet you could do an if/else that set a variable to one function or the other and then just have one call to that "variable" function?  That'd enforce the signature.
*  If you find yourself repeating parts of stock `individual_*()`, lets try to factor those back into a single instance in the common code?
*  I wonder if we can do some sort of minimal check after the conversion call.  An `assert`?  That the promised file was built?  "foo.png" (say) exists?  Could turn-in the variably-named function as not doing its job?
*  I know you didn't want to get carried away.  Thank-you!  But do you think it is time to do away with the PyMuPDF switch (and associated alternatives, two old executables)?

Perhaps I can get #2348 in the next couple days.  Will comment there in a minute.

oscarlevin:
> * Built SVGs for the three components affected.  Diffs, before/after, look OK, but amazingly different for `sageplot`, even if two runs are done with no change in the code.

Very strange.  No ideas.

> * Perhaps obvious.  But I like the external functions and the internal functions having the same signature.  I bet you could do an if/else that set a variable to one function or the other and then just have one call to that "variable" function?  That'd enforce the signature.

Definitely like this idea.  I had thought of using `converter = individual_*` as a default argument, which would remove the if/else completely.  But to avoid default arguments, we could just set `indificual_* = ext_converter_method` at the top if that is not None.

> * If you find yourself repeating parts of stock `individual_*()`, lets try to factor those back into a single instance in the common code?

Agree.

> * I wonder if we can do some sort of minimal check after the conversion call.  An `assert`?  That the promised file was built?  "foo.png" (say) exists?  Could turn-in the variably-named function as not doing its job?

Nice idea.  

> * I know you didn't want to get carried away.  Thank-you!  But do you think it is time to do away with the PyMuPDF switch (and associated alternatives, two old executables)?

I was planning on asking about that at drop-in.  I think this would be a nice first step at general cleanup.  I'll send a PR shortly.

PR #2352, opened 2025-01-03
Script: improve error catching and documentation for individual image generation

oscarlevin:
As discussed, this will re-raise the import error from `individual_latex_image_generation`.  Also includes a few doc-string-like comments at the start of the affected functions to make it more clear what `ext_converter` should do.


rbeezer:
Well, I needed this to debug my re-creation of my Python virtual environment.  All better now.  Thanks.  ;-)

Merge conflict with the old pyMuPDF-removal, I think I got that resolved properly.  Works fine with WeBWorK sample chapter (the **other** PR I was working with!).

PR #2355, opened 2025-01-04
Fix issues with wide content in elements with @cols

ascholerChemeketa:
This makes `@cols=X` in either a list or an exercisegroup a suggestion that wide content can ignore.

Given:  AA  BB CCCC DD EE and 5 cols of space we will now show:
```
AA_BB
CCCC_
DD_EE
```
If B was left out it would be:
```
AA___
CCCC_
DD_EE
```

### DIFF
--- a/css/components/helpers/_cols.scss
+++ b/css/components/helpers/_cols.scss
@@ -6,14 +6,15 @@
   justify-content: start;
 }
 
-// allow a selector to appear in columns
 // see lists and exercises for sample use
-
 @mixin allow-cols($el, $col-gap: 2em) {
   @for $i from 2 through 6 {
     .cols#{$i} > #{$el} {
-      width: calc(100% / $i - #{$col-gap});
-      max-width: calc(100% / $i - #{$col-gap});
+      // columns are a suggestion enforced by a minimum size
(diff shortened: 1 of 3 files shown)


jjrsylvestre:
Works well for me! Thanks for this Andrew.

rbeezer:
Prepared to merge this, but Sample Article needs work.  LaTeX macros should not be defining environments (with begin/end pairs), we want to limit them to be short "one-liners" that enforce consistent notation, etc.  So I'd prefer not to violate best practices in the Sample Article.  And the `m` element is meant for inline content.

Can the `p` and `m` and system construction be replaced by an `md` that will provide as much stress?  There are several systems (with ampersands for alignment) elsewhere in the sample article (early Section on mathematiics?).  One hunk of whitespace where new content begins.

A force-push would be great.  Thanks.

ascholerChemeketa:
Force pushed with cleaned up long math samples and removed whitespace.

rbeezer:
Excellent!  I think MathJax is making those displays a bit compressible?  In any event, love the behavior.  Thanks for redoing the LaTeX.

(Not in a good place to update the website sample article right now.)

PR #2356, opened 2025-01-04
HTML: Don't include pretext_search.css

ascholerChemeketa:
CSS for native-serach is now baked into theme.css for all themes. An include for `pretext_search.css` is no longer needed/wanted. At best it 404s, at worst it picks up files left over from older builds.


rbeezer:
Merged - thanks for the quick fix.

PR #2357, opened 2025-01-06
Script: report ImportError during latex image conversion

jjrsylvestre:
If `ImportError` is raised during `individual_latex_image_conversion` then the current latex image file is thrown on the `failed_images` pile and conversion moves on to the next file. After processing (and failing on) all files, the script reports that conversion failed for all files but doesn't say why. (It says to refer to the log files, but in this case no log files will have been generated.)

If a required module was not available for the first file it will certainly not be available for all subsequent files, so we might as well bail out on the first `ImportError` and report it to the user.


oscarlevin:
Indeed.  See #2352

jjrsylvestre:
Oops, my bad --- didn't think to check list of pending pull requests.
Closing in favour of #2352.

PR #2358, opened 2025-01-09
CSS: Darkmode improvement for WW controls

ascholerChemeketa:
Removes hardcoded colors (and background image) for WW controls to make them handle dark mode better.

Does NOT fix the contents returned by the WW server. That will need more significant changes to respect dark mode.

What I know about WW is limited to building the sample chapter to bug check PRs. Someone more knowledgable than me should sign off on this. @Alex-Jordan ?

### DIFF
--- a/css/components/interactives/_webwork.scss
+++ b/css/components/interactives/_webwork.scss
@@ -1,5 +1,7 @@
 // TODO - needs refactoring and dark mode update
 
+@use 'components/helpers/buttons-default' as buttons;
+
 /*  WW problems */
 .ptx-content .wwprob table.attemptResults {
   margin-left: 2em;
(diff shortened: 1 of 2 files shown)


rbeezer:
Went ahead and merged this.  "Activate" button (which was the motivation?) is only just barely visible for me.  Better on hover, but maybe still needs work.  It'd be good to see if @Alex-Jordan can send back "better" HTML so this improves?

OK, not sure what happened.  "Update pre-built themes" went missing?  Terminal history is a ways back, but I don't see it being skipped.  So maybe that is why I'm not seeing any improvement.  I can try to cherry-pick that commit later, but I'm done for this session now.  Please advise.  Thanks.

rbeezer:
Fetched the branch/PR again.  Only picks up one commit.  Hmmm.

I was able to fetch the missing commit by its hash (never done that before).  Content looks OK.

Cherry-pick'ed it onto a new branch.  Would not apply cleanly and it didn't seem to be something to fix by hand.

Can you do the prebuilt theme over again with this work in place and on a new PR?

I'm mystified.

ascholerChemeketa:
@rbeezer Sorry, did a force push here before I saw your note. Going to go make a new PR...

rbeezer:
See #2362 for more.

Alex-Jordan:
Finally replying to this. The HTML in question that makes those buttons (Activate, Check Responses, Reveal, etc) is not produced by a webwork2 server, or PG, or a PG renderer. It is either produced by pretext's `pretext-html.xsl` in the case of "Activate", or by `pretext-webwork.js` in the `js/` folder tree. In both cases, it's just a button with a class `webwork-button`. Well, OK the ones other than "Activate" have a little inline styling for `margin-right`, but that is not relevant here.

So whatever badness there is for these buttons in dark mode comes from the CSS that goes along with the `webwork-button` class. Probably this PR and the related one have cleared that up, but I haven't checked yet.

PR #2359, opened 2025-01-10
CSS: Bugfix MathJax top pixel clipping in Chrome

ascholerChemeketa:
Fix for issue described here:
https://groups.google.com/d/msgid/pretext-support/348ae3aa-bc5a-4273-ab58-03b1f985fbe3n%40googlegroups.com?utm_medium=email&utm_source=footer

Root issue appears to be Chrome doesn't always reserve enough space for the rescaled text that MathJax produces. This adds a few pixels of fudge to account for that.

### DIFF
--- a/css/components/elements/_math.scss
+++ b/css/components/elements/_math.scss
@@ -7,6 +7,7 @@
 
 .displaymath mjx-container[jax="CHTML"][display="true"] {
   margin: 0 0 0 0;  // container is going to apply margin, so remove it from mjx-container
+  padding-top: 2px; // Chrome sometimes clips the top 1-2 pixels of math after it has been scaled - provide room for that content
 }
 
 // ?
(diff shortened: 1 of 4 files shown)


rbeezer:
Am I right that changes to the pre-built theme are going to need to be made sequentially, both as PRs are formed and as merged?

For schema changes, I often update derived files as I merge.  Do we need that workflow?

ascholerChemeketa:
@rbeezer Sent an email re: the pre-built updates.

rbeezer:
> Sent an email re: the pre-built updates.

Thanks - I'll reply when I run out of time for PRs!

rbeezer:
Merged.  Thanks.  Glad that was fixablle.

Normally, I'm adverse to messing with the fine details of MathJax, but this seems workable, and no real visible problem with math in Firefox.  Still I've made a note that maybe we can revert tis once MJ4 comes to town.

PR #2360, opened 2025-01-12
CSS: Custom theme lets pub file control focused TOC

ascholerChemeketa:
Legacy themes allow the pub file to control whether the ToC is "focused". In modern themes that is controlled by the theme. 

This allows the pubfile to control the ToC "focused" mode when `theme="custom"` is specified.


rbeezer:
Merged.  Thanks for the quick response.

PR #2361, opened 2025-01-13
Webwork CSS hidden-content fix

ascholerChemeketa:
Webwork had repurposed a `.hidden-content` rule that was part of the old knowl code to hide items that should not be displayed.

That rule got phased out. This adds new rules so that class has the intended effect when applied to the div's that WW tries ot hide.

### DIFF
--- a/css/components/interactives/_webwork.scss
+++ b/css/components/interactives/_webwork.scss
@@ -115,3 +115,10 @@ label.incorrect .feedback {
 .ptx-content .webwork-button {
   @include buttons.ptx-button;
 }
+
+//.hidden-content used to be used for knowls and got repurposed for hiding webwork contents
+// now just used by webwork
+.problem-buttons.hidden-content,
+.problem-contents.hidden-content {
+  display: none;
+}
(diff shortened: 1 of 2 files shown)


rbeezer:
With a rebase attempt, I get
```
Merge conflict in css/components/interactives/_webwork.scss
```

Not prepared to deal with that right now.  Maybe we wait until I get `npm` going?

ascholerChemeketa:
Did a rebase + force push if you want to try it again pre npm.

rbeezer:
Merged as-is.  Thanks.  And thanks for the assist with the fresh prebuilt update.

Noticed for the first time that the WW-Activate "Loading" splash screen is a bit off vertically.  Can't imagine it is due to this change? Advise and I'll circulate further if necessary.

Alex-Jordan:
When I go here:
https://pretextbook.org/examples/webwork/sample-chapter/html/sample-ww-chapter-4-2.html

I don't see the Loading being off. That page says it was built on Jan 2, so it doesn't have several recent changes to CSS. Is that instance supposed to be a more recent build?

rbeezer:
> That page says it was built on Jan 2

No website updates for a while, you'll need to build your own to participate in debugging for now.

PR #2364, opened 2025-01-14
Generate better stack traces for XML errors in included files

ascholerChemeketa:
Currently, if there is a syntax error (e.g. tag mismatch) in an included file, no information is provided about what file has the issue.

Authors writing a book page by page can figure out the location, but it can be maddening after using a tool to do a batch conversion of edit. 

This makes sure the trace ends with a pointer directly to the offending file:
```
  ...
  File "/mnt/f/Programming/thinkcpp-rs/pretext/Chapter1/program.ptx", line 20
lxml.etree.XMLSyntaxError: Opening and ending tag mismatch: p line 17 and li, line 20, column 10
```

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -4070,7 +4070,25 @@ def xsltproc(xsl, xml, result, output_dir=None, stringparams={}):
     # lxml.etree.XMLSyntaxError: Excessive depth in document: 256 use XML_PARSE_HUGE option
     huge_parser = ET.XMLParser(huge_tree=True)
     src_tree = ET.parse(xml, parser=huge_parser)
-    src_tree.xinclude()
+    try:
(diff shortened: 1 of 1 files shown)


rbeezer:
This will be a really nice improvement.  But I'm seeing some kind of trouble.

Built sample book.  Error reporting is much improved with an intentional error added.  Then did a before/after HTML build with no errors.  All `xref` seem to not work.  As if the `@xml:id` vanished.  Also seeing some changes where stuff with an `xml:base` is going missing.  I understand `xml:base` to be part of the `xi:include/@href` mechanism, but otherwise don't have much insight.  Not sure I have much more guidance.

rbeezer:
Love it.  Thanks for the workaround.  I'm going to advertise on -dev shortly.

PR #2365, opened 2025-01-14
Add xmlid_root option for braille conversion

oscarlevin:
@davidaustinm noted that building a braille book especially could benefit from restricting to just a chapter.  This adds that option in the same way that the xmlid_root is done in the other conversions.

Note, I have not tested that this actually works with the rest of the braille conversion.  Understand it might take a while to get this merged.

### DIFF
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -781,9 +781,9 @@ def main():
                 dest_dir,
             )
         elif args.format == "braille-emboss":
-            ptx.braille(xml_source, publication_file, stringparams, out_file, dest_dir, "emboss")
+            ptx.braille(xml_source, publication_file, stringparams, args.xmlid, out_file, dest_dir, "emboss")
         elif args.format == "braille-electronic":
-            ptx.braille(xml_source, publication_file, stringparams, out_file, dest_dir, "electronic")
+            ptx.braille(xml_source, publication_file, stringparams, args.xmlid, out_file, dest_dir, "electronic")
         elif args.format == "epub-svg":
             ptx.epub(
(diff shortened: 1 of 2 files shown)


oscarlevin:
Upon closer inspection, I think more work needs to be done, specifically in the xsl.  Converting to draft for now.

rbeezer:
`xmlid_root` was designed to reduce the scope for building assets when isolating them in source.  It was never meant for producing subsets of output formats.  So I'd really rather not go further in this direction and not tie ourselves in knots making this even more complicated than it is already.

We have made some real improvements in speed for HTML.  Does braille need similar attention?  (Note that the braille conversion is not even complete right now, it still needs some tricky situations worked on.  That should be the priority.)

oscarlevin:
That all makes sense, and I wasn't sure whether this would just work with the right string parameters or not (now I know).  

The use case here is to be able to provide accessibility offices one chapter at a time instead of a huge file for the entire book.  So upon further reflection, some sort of chunking option would be more appropriate.  I'll close this and think about that other direction.

rbeezer:
I think setting the chunk level to 1 is supported.  And only that level.  Give it a go?

I agree that being able to emboss/distribute by chapters iis very desirable.

rbeezer:
> And only that level.

Other than zero.  Which might be worth testing while you are at it.

davidaustinm:
I set the chunking level to 1,
`PTX:INFO    : braille chunking level (from publisher file) set to "1"`
and got back a single BRF, which is identical to the original one PTX created.  

Setting level="0" gave the same thing, while level="2" (I won't be deterred) reported that it was using level="0" instead.

PR #2366, opened 2025-01-14
HTML: display ToC on root page

ascholerChemeketa:
On the root page of a document (e.g. sample-book.html) there is no ToC entry that corresponds to the current page. Thus nothing is considered active and the entire ToC is hidden.

Solution here is to force the root ToC item to always think it "contains-active". That way the root level of the ToC is forced to display.

Clean diffs on sample-book other than sample-book.html which has the desired extra class on the top ul in ToC.


rbeezer:
I think I had noticed this, but had not thought to report it as a bug.  :-(  Too busy doing something else in that moment, iirc.  Thanks for the catch and the fix.

Am I right that this always goes on a `ul` which has no ancestor that is a `ul`?  If so, would it not be cleaner to test for that situation as the definition of `$is-root-ul`?  We could avoid a single-use `xsl:param`?  If so, PR encouraged.  ;-)

Maybe a condition like
```xsl
"count(ancestor-or-self::ul) = 1"
```

PR #2367, opened 2025-01-14
FITB: Process statement using no mode (instead of mode="body")

### LINKED ISSUE #2294: FITB: Elements not being retained within fillin exercises
Reports from Chrissy and Andrew state that some elements do not appear inside exercises that are created with the fillin structure.

Chrissy's report thread: https://groups.google.com/g/pretext-dev/c/-AfefZJx-I8/m/c-VNNATVAAAJ
Andrew's report thread: https://groups.google.com/g/pretext-dev/c/wmvrpWwOABM/m/DpvBbkrjCAAJ

Chrissy reported issues with images sometimes not appearing:

> I went ahead and converted my fill-in exercise to use the new markup, but something kind of weird happened.  I have an image in the statement of the problem, and I had been under the impression that image could be a child of figure or statement, but not <p>.  So it was a child of the statement in the old markup. 
> 
> When I switched to the new markup, the image did not show - only the short description text did.  I put a paragraph around it, and it did show, and build with no errors or warnings. It also works inside a figure, but I didn't really want the numbering. 

and issues with links inside the fillin exercises:

> Here's the live version with the old markup:  [Exercise 1.: How a matrix is indexed](https://tanaquil18.github.io/matrixtheory/sec-matrices-elimination.html#matrix-indices)
> When you enter any number other than 6 or -2, the feedback prompts you with an xref to where this notation was defined, in the Definition of a Matrix.  In the old markup, the link that goes to that definition.  
> 
> When I switch to the new markup, the link is just to the section, not the definition specifically.  The live version with the new markup is  [Exercise 2.: How a matrix is indexed](https://tanaquil18.github.io/matrixtheory/sec-matrices-elimination.html#matrix-indices2)
> 

Andrew reported additional issues:

> According to the schema, exercise statement should allow for elements like <program>, <image> etc... And those elements work fine in other exercises.
> 
> But in fillins, most of those are getting dropped in the JSONification of the statement element.
> 
> Is this a known issue?
> 
> I didn't do an exhaustive test, but program,  pre, and image all are dropped. Tables do make it through.

I do plan to dig into this, and am not yet sure if this is an issue on the PreTeXt side — it could be due to working on XML stubs in the XSL processing — or on the Runestone side — the RS javascript HTML insertion process may have side effects.

Submitting this issue so that it is formally tracked.

dbrianwalton:
This should resolve #2294. Also corrects some logic and typos relating to default correct tests.

### DIFF
--- a/xsl/pretext-runestone-fitb.xsl
+++ b/xsl/pretext-runestone-fitb.xsl
@@ -107,7 +107,7 @@
                     <xsl:text>"problemHtml": </xsl:text>
                     <xsl:call-template name="escape-quote-xml">
                         <xsl:with-param name="xml_content">
-                            <xsl:apply-templates select="statement/*" mode="body" />
+                            <xsl:apply-templates select="statement/*" />
                             <xsl:if test="$b-dynamics-static-seed">
                                 <div>
                                     <xsl:attribute name="id">
@@ -124,7 +124,7 @@
                     <xsl:text>,&#xa;"solutionHtml": </xsl:text>
                     <xsl:call-template name="escape-quote-xml">
                         <xsl:with-param name="xml_content">
-                            <xsl:apply-templates select="solution/*" mode="body" />
+                            <xsl:apply-templates select="solution/*" />
                         </xsl:with-param>
                     </xsl:call-template>
                     <!-- Add packages that need to be loaded as javascript -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks for the tune-up!  Merged with no code changes.  Clean diff with sample book.  Two notes:
*  Two topics here, two commits would be preferable.
* I like to put spaces around logical operators (`string-length()>0`).  Makes certain grep'ing actions easier and more reliable.

rbeezer:
You would have had an easier time if I'd thought to do the following:  ;-)

https://github.com/PreTeXtBook/pretext/issues/2372

dbrianwalton:
Thanks for the tips moving forward.

PR #2368, opened 2025-01-15
Webwork loading interstitial position fix

ascholerChemeketa:
Fix for Webwork loading interstitial not being centered in parent. Noted here:
https://github.com/PreTeXtBook/pretext/pull/2361


rbeezer:
Looking good!  Thanks.

PR #2369, opened 2025-01-15
HTML: better XSL for ToC on root fix

ascholerChemeketa:
Implements suggestion from https://github.com/PreTeXtBook/pretext/pull/2366 to get rid of single-use param.


ascholerChemeketa:
Clean diffs on sample article.

rbeezer:
Realized late last night my test/condition on #2366 was not correct.

I think `not(ancestor::ul)` is equivalent to `count(ancestor::ul) = 0`.  I'll merge this soon, but not right now, should you care enough to make a change.

Thanks for the update - much cleaner.

rbeezer:
Beautiful.  Thanks.

PR #2370, opened 2025-01-15
Script: do not use prebuilt default-modern when node is available

ascholerChemeketa:
This implements proposal 1 from https://groups.google.com/g/pretext-dev/c/j9K4ttA-bns

I think it is the least controversial. It should have no impact on users, But it makes things much less confusing/fiddly if you are editing/testing CSS changes on books that use default-modern.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -3706,11 +3706,12 @@ def _build_custom_theme(xml, theme_name, theme_opts, tmp_dir):
     error_message = "Node.js is required to build themes other than default-modern. Make sure it is installed and in your PATH"
     try:
         import subprocess, json
+        node_exec_cmd = get_executable_cmd("node")
         # theme name is prefixed with "theme-" in the cssbuilder script output
         full_name = "theme-{}".format(theme_name)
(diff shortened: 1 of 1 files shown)


rbeezer:
New laptop, I don't have node.  I'll give this a whirl.  Wrong, I've got some "nvm" thing/cruft in my home directory.  v22.  I know that is old.

So I get a **HUGE** diff versus pre-PR.  May just be line breaks, but it is hard to tell.  Seems to render OK, visually.

If I break my node-executable setting, I get pre-built just fine.  So PR is functioning as desired.

BUT - do we want to deal with people like me who have old stuff around building their CSS?  Any way to enforce a version for node?  Do we want a version-minimum for SCSS?  Is it package.json we need/want?  Do we already have it?  Sorry to be so node-adverse.

rbeezer:
OK, back to some CSS PRs.  Diff is better this time - totally different files.  I think the EPUB/Kindle cruft was lurking around before.  Thanks for the explanation -  I've lost track of who is building what and when.  Now we have the CLI in the mix!  ;-)

Merged as-is.  Thanks.

PR #2371, opened 2025-01-16
CSS: add min node and cleanup cssbuilder/package.json

ascholerChemeketa:
Adds a min node version to the package.json for CSSBuilder and removes some default cruft from the file.


rbeezer:
Merged as-is, with no testing.  Thanks!

PR #2374, opened 2025-01-18
CSS: Documentation update for cssbuilder script

ascholerChemeketa:
I noticed that I had let some of the docs get out of date. This updates the samples for invoking `cssbuilder` by hand with a JSON blob to customize the results. This technique might be handy for you @rbeezer when you go to update the website build script to show more design variants.

No functional changes. One small change to the actual source to make code intent clearer but it produces exactly the same result.


rbeezer:
Thanks for the maintenance!  Let's make sure these are two commits (doc and code), no matter how trivial one part might be.

PR #2375, opened 2025-01-19
add alt text to footer logo images

oscarlevin:
I was running some accessibility tests and got complaints about images without alt-text.  This adds them to the logos in the footer.

Note, for the PreTeXt logo, created as an `<svg>` element, there are lots of suggestions for how to get something like alt-text.  This was one of them, and with it, WAVE doesn't complain.


rbeezer:
Very good - I'll get this in shortly.  We have a `@decorative` attribute on `image` that writes an empty `@alt`, but I guess explicit text here will be better, since we know what it is.

rbeezer:
> I was running some accessibility tests 

A brief report on our (new) `pretext-a11y` list would be a very good thing.  Hint, hint.  ;-)

PR #2376, opened 2025-01-19
CSS ACCESSIBILITY: ensure 'skip to main content' link has proper contrast

oscarlevin:
WAVE complains about a contrast issue with the "skip to main content" link.  This keeps the link above the window until it gets keyboard focus (as previously) but also has the same background color in either case.

As suggested by: https://webaim.org/techniques/css/invisiblecontent/#skipnavlinks.


rbeezer:
Looks good in principle.

But there seems to be a whole lot of extra CSS stuff here.  Old stuff about column widths.  Changes for EPUB and Kindle (which should not be affected).  I didn't check, but almost like this was built off some older commit?  It did rebase just fine.

I think this is an example of how CSS changes are making some PRs more complicated?  @ascholerChemeketa would be able to say for certain (and maybe I'm missing something).  In any event, this didn't look right enough to merge right now.

oscarlevin:
I guess I don't understand where we are in the correct workflow for CSS changes.  I rebased, then rebuilt the default-modern theme (and removed changes to epub and kindle).  So this should just have the fixes for the 'skip to main content' link, and the updated built theme.

ascholerChemeketa:
On fone.

Looks good here, I only see the expected changes to SCSS and CSS.

I think we should leave .map files out of commits. See my -dev post for logic.

rbeezer:
OK, merged as-is.  I think the EPUB and Kindle CSS are lagging for some reason, they don't seem to currently have the go-wide columns stuff.  So I don't think that was your problem.  Sorry for any confusion.

But CSS is changing so fast, we need to be sure PRs are built at the current tip.  I might update everything right now, given I have the CSS build process going.

Thanks for the accessibility work!

PR #2377, opened 2025-01-20
HTML: simplify styles for custom ol markers

jjrsylvestre:
@ascholerChemeketa Can you take a look?


rbeezer:
Really good!  A lot of work to make up for an HTML shortcoming.  Thanks for slimming this down so much.  Added a commit with theme updates, which is a new activity when merging (i.e. you were not meant to do it as part of the PR).  Otherwise, as-is.

One comment:  `generate-id()` in `mode="ol-marker-class"` has given me trouble in the past.  I forget the specifics.  Definitely very different strings on different OSes.  Should be fine for per-build use, but lets just keep an eye on it.  We could decide to roll-our-own serial numbers quite easily.

jjrsylvestre:
How about just indexing by a simple counter instead of `generate-id()` ? Adds an `xsl:for-each` to the code, but it's run only once and on what should be a relatively small number of nodes (the number of `ol[@marker]` in the document).

[Diff of updated code](https://github.com/PreTeXtBook/pretext/compare/master...jjrsylvestre:pretext:html-marker-css-id)

Let me know if you'd like a new PR.

jjrsylvestre:
(Actually the additional `xsl:for-each` would run only on the number of _unique_ `@marker` in the document.)

jjrsylvestre:
Updated to use `xsl:apply-templates` instead of `xsl:for-each`. (Not sure what the difference if performance would be... but it produces identical output on the sample article.)

Same link as in previous comment to see updated code.

rbeezer:
Thanks, Jeremy, this will be a good improvement.  At first blush, I think you might be working too hard.  Would the following outline work?

1.  Make a variable that is a node-set of just one element per class.  I think you already do this somewhere else, but it might take some extra XSL.  Might be a good example in the creation of an index.  This might be the tricky part.  Only items not the same as their predecesor?
2.  In a `for-each` over tht naode-set/variable context changes to the "current" item.  Then `position()` reliably reports the serial number with a simple `value-of`.

I can search for examples if desired.

jjrsylvestre:
Getting closer? Sorry for the code churn --- when I only hack on xsl in spurts once or twice a year, it takes a while to get my brain back into a space where I don't overthink things. ;)

Generate index node set.
- Realized I could just do the Muenchian grouping directly on the `$document-root//ol[@marker]` node set.
- Seems to me more efficient to just record `position()` at the time of index-set generation rather than re-compute it every time a custom class needs to be added to an `ol` in the output. (Though probably in practise the difference in efficiency is effectively nil.)
````
<xsl:key name="marker-key" match="ol|ol-marker" use="@marker"/> <!-- never used on node sets that contain mix of ol and ol-marker -->
<xsl:variable name="ol-markers">
    <ol-markers>
        <xsl:apply-templates select="$document-root//ol[@marker and count(. | key('marker-key', @marker)[1]) = 1]" mode="ol-markers"/>
    </ol-markers>
</xsl:variable>

<xsl:template match="ol[@marker]" mode="ol-markers">
    <xsl:element name="ol-marker">
        <xsl:copy-of select="@format-code"/>
        <xsl:copy-of select="@marker"/>
        <xsl:copy-of select="@marker-prefix"/>
        <xsl:copy-of select="@marker-suffix"/>
        <xsl:attribute name="classname">
            <xsl:text>ol-marker-</xsl:text>
            <xsl:value-of select="position()" />
        </xsl:attribute>
    </xsl:element>
</xsl:template>
````

Output CSS file.
````
<xsl:template name="ol-marker-styles">
    <xsl:variable name="ol-marker-nodes" select="exsl:node-set($ol-markers)" />
    <xsl:if test="$ol-marker-nodes//ol-marker">
        <exsl:document href="ol-markers.css" method="text">
            <xsl:apply-templates select="$ol-marker-nodes//ol-marker" mode="ol-marker-style" />
        </exsl:document>
    </xsl:if>
</xsl:template>

<xsl:template match="ol-marker" mode="ol-marker-style">
    <xsl:text>ol.</xsl:text>
    <xsl:value-of select="./@classname"/>
    <xsl:text> &gt; li::marker { content: &quot;</xsl:text>
    <xsl:value-of select="./@marker-prefix" />
    <xsl:text>&quot;counter(list-item,</xsl:text>
    <xsl:apply-templates select="." mode="html-list-class" />
    <xsl:text>)&quot;</xsl:text>
    <xsl:value-of select="./@marker-suffix" />
    <xsl:text> &quot;; }&#xa;</xsl:text>
</xsl:template>
````

Add custom class to `ol` in output when needed.
````
<xsl:template match="ol|ul">
    ...
        <xsl:attribute name="class">
            ...
            <xsl:variable name="ol-marker-class">
                <xsl:choose>
                    <xsl:when test="self::ol/@marker">
                        <xsl:variable name="marker-value" select="./@marker" />
                        <xsl:for-each select="exsl:node-set($ol-markers)">
                            <!-- Should be only one match since ol marker -->
                            <!-- index node set contains no duplicates    -->
                            <xsl:value-of select="key('marker-key', $marker-value)[1]/@classname"/>
                        </xsl:for-each >
                    </xsl:when>
                    <xsl:when test="self::ol and $mbx-format-code = 'a' and @ordered-list-level = '1'">
                        <xsl:text>lower-alpha-level-1</xsl:text>
                    </xsl:when>
                </xsl:choose>
            </xsl:variable>
            <xsl:if test="not($ol-marker-class = '')">
                <xsl:text> </xsl:text>
                <xsl:value-of select="$ol-marker-class"/>
            </xsl:if>
            ...
        </xsl:attribute>
    ...
</xsl:template>
````

rbeezer:
> Sorry for the code churn

I like code churn.  ;-)  Helps you think clearer the next time.  And I know how it is getting into the mindset after six months away!

Yes, that's looking a whole lot more streamlined.  Let's put some code on a new PR - hard to find this once it is closed.

One suggestion: perhaps the guts of defining `$ol-marker-class` could be a separate template.  Just for readability and less clutter in the other parts of matching `ol|ul`?

Issue #2378, opened 2025-01-21
Stop tracking CSS map files


rbeezer:
@ascholerChemeketa : could you organize removing the `*.map` files as a PR?  I could likely do it, but probably better you do it right the first time.

Especially, update the `.gitignore` in the root directory of the repository?  Thanks.

rbeezer:
Closed via #2379.  Thanks @ascholerChemeketa!

PR #2379, opened 2025-01-21
CSS: Remove css.map files from git and ignore in future

ascholerChemeketa:
PR for https://github.com/PreTeXtBook/pretext/issues/2378

I assume we are OK with leaving them in the history. They aren't all that big, so trying to retcon them out of existence would likely be more trouble than it was worth.


rbeezer:
> I assume we are OK with leaving them in the history.

Yes, I don't think I have ever tried to excise anything.  I've been pretty conscientious about what gets tracked.

rbeezer:
Great.  Thanks for this.  Merged as-is.  More or less, since there were merge conflicts in the `*.map` files.  Hopefully now a thing of the past!!!  ;-)

PR #2380, opened 2025-01-21
CSS: Improve google font request. Use newer API, request italic

ascholerChemeketa:
Improve google font rendering by updating web font API request to v2:
https://groups.google.com/d/msgid/pretext-support/MTAwMDAyMi5iZWV6ZXI.1737492553%40quikprotect

Also, request true italic font version.

### DIFF
--- a/css/fonts/_fonts-google.scss
+++ b/css/fonts/_fonts-google.scss
@@ -12,19 +12,19 @@ $monospace: 'Inconsolata, Consolas, Monaco, monospace;' !default;
 // lists are 1-indexed
 
 $body-font: list.nth(string.split($body, ','), 1);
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged.

I added a changed prebuilt default-modern (via cssbuilder script), which I think was the right thing to do.  Despite being a derived product, I think I like to see the effect of SCSS changes migrating to actual CSS.  So I like that we are doing this.

Inline headings and titles (blocks, exercises, etc.) look very different to my eye.  A little bolder (perhaps solving original problem) but also not as tall (font size?) and a bit crowded horizontally (letter spacing).  Might be correct, but a visual change nonetheless.

Warn people on the bold/alert thread?  With instructions for seeing changes (no website examples yet).  Or wait and see who notices and when?  Merged on the assumption we can adjust.

PR #2381, opened 2025-01-23
CSS: Remove ptx-logo from sidebar; refactor colorado themes

oscarlevin:
This eliminates the pretext logo from the sidebar.  A few related tweaks improve the TOC for default-modern, and a refactoring of the TOC for denver and greeley were needed, both of which look better now.

Tested with all themes and all widths on chrome, firefox, and epiphany (webkit) on linux.  

Note, only committed the SCSS files; you will need to `npm run build` from `scripts/cssbuilder` to try out `default-modern`.

### DIFF
--- a/css/components/page-parts/_toc-basics.scss
+++ b/css/components/page-parts/_toc-basics.scss
@@ -3,22 +3,6 @@
 $sidebar-width: 240px !default;
 $nav-height: 36px !default;
 
-@mixin ptx-logo {  // need space to disappear behind footer - pretext logo fills that space
-  &::after {
-    // Apply logo as a mask so background-color can change it. It is a separate document
-    // so no other way to have styles on page affect it.
-    content: "";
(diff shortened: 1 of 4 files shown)


rbeezer:
This has two differeent sets of changes: a general one (no logo) and some theme-specific ones (Colorado).  (The semicolon in the commit message is a clue!)  Can you make that two commits?  I started, but was not so confident about the split.  A force-push here is fine.  And SCSS files only, is good, we've stop tracking a bunch of stuff and ``pretext/pretext` is rebuilding default-modern with node available (my situation).  Thanks.

oscarlevin:
I should have been clearer; the refactor of the denver and greeley theme was needed in order to remove the logo; without the theme specific changes the logo is needed to not lose the bottom of the toc below the footer.  It just happened that tacoma and salem have a toc that is independent of the footer, so no additional changes were needed in those themes.

oscarlevin:
Thought about it some more and figured out a way to break these up such that the first should still work independently (although I haven't tested that extensively).

rbeezer:
> I should have been clearer;

Sorry, I misunderstood.  No wonder I could not disentangle.  ;-) Wasn't important enough to re-code, but I'll run with the new version once I can get back to this.

rbeezer:
Merged.  Thanks for the re-write.  Glad to see all the clean-up of old workarounds.

PR #2382, opened 2025-01-23
CSS: Build all themes as defaults when doing untargetted build

ascholerChemeketa:
This is the complement of @oscarlevin's https://github.com/PreTeXtBook/pretext-cli/pull/907

When `npm run build` is invoked in cssbuilder, all themes will be built into the dist folder. CLI will use those if a legacy theme is requested or cssbuilder is unavailable.

### DIFF
--- a/script/cssbuilder/cssbuilder.mjs
+++ b/script/cssbuilder/cssbuilder.mjs
@@ -103,15 +103,15 @@ function getTargets(options) {
     // -------------------------------------------------------------------------
     // Web targets - pretext assumes output name will be 'theme-XXX'
     // Legacy targets
-    { out: 'theme-default-legacy', in: path.join(cssRoot, 'targets/html/legacy/default/theme-default.scss'), autobuild: true },
-    { out: 'theme-min-legacy', in: path.join(cssRoot, 'targets/html/legacy/min/theme-min.scss'), autobuild: true },
-    { out: 'theme-crc-legacy', in: path.join(cssRoot, 'targets/html/legacy/crc/theme-crc.scss'), autobuild: true },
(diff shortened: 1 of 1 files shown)


rbeezer:
And then I am tempted to track all five "new" themes in the repository.  Partly because it is easier to add five changed themes to a PR/commit that to add one and exclude four.  Other than philosophy, any problems with that?  It'd clearly show effects, like Oscar's recent changes to **only** the Colorado family (#2381).

ascholerChemeketa:
My main objection isn't having the prebuilts. It is using them unless strictly necessary for a "Look, I made something appear on a webpage!" workshop session. (And wouldn't that normally be happening via the CLI?)

Eventually, I think we want things like https://github.com/PreTeXtBook/pretext/pull/2377 built into theme.css so that they are available out of context (RS assignment page). Creating hacky ways to make that available in a prebuilt sounds like a bad idea.  Which means that for the custom styles to work well, the author can't use a prebuilt.

Detecting when a user needs a custom build for reasons directly related to the theme (palette, colors) or implicitly (custom list markers), or failing to detect that and getting questions about why some styling isn't working, is going to be way worse than occasionally having to point people to the two commands needed to install node via nvm or fnm.



Yes, having them there as a sanity check of what files a SCSS change hits would be nice. As long as we figure out a commit workflow for them that works well for everyone.

rbeezer:
> As long as we figure out a commit workflow for them that works well for everyone.

It's no problem for me to regenerate on each commit touching the CSS.

rbeezer:
Merged this in advance of a spate of CSS PRs.  ;-)

Follow-on commit c3c7451e1c73009f9a98524ae76c715f9396d285 commits the themes to the repository, with the intent that they are part of the history of changes.

PR #2384, opened 2025-01-23
Runestone: Allow horizontal-parsons in containers other than exercise

ascholerChemeketa:
Currently most RS interactives have a template like this that places the component:
```xsl
<xsl:template match="*[@exercise-interactive = 'parson']" mode="runestone-to-interactive">
```

But horizontal parsons have `exercise[...` instead of `*[...` and thus only can appear in exercises and not activities, etc...

This was reported as a bug in RS discord.

This template already limits which elements can get the `@exercise-interactive` https://github.com/PreTeXtBook/pretext/blob/5ed521fd3f5fafadb989e07c2b8637c3523bf3fa/xsl/pretext-html.xsl#L4226

So this PR removes the exercise restriction on the `"runestone-to-interactive"` template

### DIFF
--- a/xsl/pretext-runestone.xsl
+++ b/xsl/pretext-runestone.xsl
@@ -1211,7 +1211,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 
 <!-- Parsons Problem (Horizontal)-->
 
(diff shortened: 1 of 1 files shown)


rbeezer:
Good catch - thanks!

* If you can alert the original reporter, that'd be great.
* If this was less obvious, or more subtle, I'd ask for an example in the sample book exposing the bugand the fix.  But this one is pretty straightforward, so no need.

PR #2385, opened 2025-01-23
CSS: Improve loading for static vs variable google fonts

ascholerChemeketa:
Fixes handling of PT Serif that was made worse by https://github.com/PreTeXtBook/pretext/pull/2380

@rbeezer You can bundle updated default-modern if you like. I *believe* we are in transition to CLI including prebuilt everything and being able to drop those from the repo.

### DIFF
--- a/css/fonts/_fonts-google.scss
+++ b/css/fonts/_fonts-google.scss
@@ -7,28 +7,47 @@ $body: 'Open Sans, Helvetica Neue, Helvetica, Arial, sans-serif' !default;
 $heading: 'PT Serif, Times New Roman, Times, serif' !default;
 $monospace: 'Inconsolata, Consolas, Monaco, monospace;' !default;
 
+@function is-variable-font($font-name) {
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
@oscarlevin If this goes through quickly it might be nice to get both it and #2380 in the same CLI release to avoid intermediary state.

oscarlevin:
I'll make sure both are present in the next release

rbeezer:
> being able to drop those from the repo

HTML is everybody's favorite (and usually first) output.  So I do not want a `node` installation to be required for initial use of the `pretext/pretext` script.  So at a minimum, a usable theme must be present in the repository.  So it is good if the CLI will build all themes, but that should not mean the repository can be incomplete.

rbeezer:
Thanks for the quick adjustment.  We'll never know how long it takes folks to notice.  :-(

Merged as-is, with update to default theme.

PR #2387, opened 2025-01-23
Sbsgroup spacing

ascholerChemeketa:
Fixes sbsgroup in figure spacing issue identified here:
https://groups.google.com/g/pretext-dev/c/calPA-mk1c0/m/lOUWhcxFBQAJ?utm_medium=email&utm_source=footer

Here is the sample Vlerio provided if you want to test against it in Hello World or something:
https://github.com/valelav/sbstest/blob/ee0e03d2b6a089165a1fd311219443dc5f7e4ca6/source/sec-section-name.ptx#L8

Will also see effect of the change at this sample in SA:
[Figure 14.4](https://pretextbook.org/examples/sample-article/html/section-interactive-authored.html#figure-animation) 
https://pretextbook.org/examples/sample-article/html/section-interactive-authored.html#figure-animation

Clean against latex build of SA
HTML build shows expected changes, no impact to legacy styles

### DIFF
--- a/css/components/chunks/_sidebyside.scss
+++ b/css/components/chunks/_sidebyside.scss
@@ -1,9 +1,10 @@
-.sidebyside {
-  width: 100%;
+// Space all rows in a sbsgroup, no other styling
+.sbsgroup > *:not(:first-child)
+{
+    margin-top: 1.5em;
+}
 
-  .sbsgroup {
-    width: 100%;
-  }
+.sidebyside {
 
   .sbsrow {
     display: flex;
--- a/css/components/elements/_figures.scss
+++ b/css/components/elements/_figures.scss
@@ -10,7 +10,7 @@ figure {
 figcaption {
   margin-left: auto;
   margin-right: auto;
-  margin-top: 2px;
+  margin-top: 0.5em;
 
   code.code-inline {
     white-space: pre;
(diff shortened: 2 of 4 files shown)


ascholerChemeketa:
Aside for @rbeezer...

I did a stab at aligning figures in a sbs at the division between the image and caption. I can make it work for simple cases (two equally sized panels) by specifying a 2-row grid layout on the sbsrow and then snapping eachsbspanel/figure/image and sbspanel/figure/caption into the grid, bypassing the sbspanel and figure. I'm pretty sure it *could* be made to work more generally, but it would be some significant changes to the HTML that would not be compatible with legacy styling. Changes would involve lots of pulling widths/margins up from panels to the row to specify the grid's column dimensions. Juice doesn't feel worth the squeeze for now.

rbeezer:
Thanks!  Merged as-is.  More on the originating post.

rbeezer:
> I did a stab at aligning figures in a sbs at the division between the image and caption. 

Thanks very much!  That's on my wish-list of really hard (impossible?) improvements.  So I know it is hard.  We once had two rows.  I think we dropped it when we had captions/titles above, captions below, etc, etc.  Maybe three rows. You are right, not really worth a huge push right now.

PR #2388, opened 2025-01-24
Script: Improve cssbuilder messaging when node is available but packages not installed

ascholerChemeketa:
Another tweak to handling default theme use.

Do not try to build if using default modern when there is no `node_modules` folder in `script/cssbuilder`. Use default and provide instructions to `npm install`

Better message if build is required (non-default theme) and packages are not installed.


rbeezer:
Tested by renaming my `node_modules` directory.  Checks out fine.  Thanks!  Merged as-is.

PR #2389, opened 2025-01-28
Script: add warning if louis import fails for braille

oscarlevin:
In case the user doesn't have LibLouis and its python binding installed, this will provide more helpful feedback.


rbeezer:
New machine, and I don't have liblouis setup yet.  PR works great!  ;-)

It'll take me a bit before I can test **with** liblouis installed.

rbeezer:
Finally got `liblouis` all installed.  Works great.  Thanks.

I made some edits to the error message.  Generally, I try to avoid concrete URLs for the Guide, since there is no real good procedure for detecting when they change.  Search there should get the job done.

PR #2390, opened 2025-01-28
HTML: only create brandlogo link if @url is present

oscarlevin:
This refactors the `brand-logo` template to only create a link around the brandlogo image if there is a `brandlogo/@url` present.  If `brandlogo` is in source without a `@url`, this just puts the image located at `brandlogo/@source` (this is a required attribute of `brandlogo` according to the schema.

Why the change?  Currently, if no brandlogo is in source, we create an empty link which gives an error according to the WAVE accessibility tester.

Previously, there was a CSS "Sigma" in the location of the brand-logo which linked to the current page or `baseurl` or something like that.  But this didn't make much sense, as the link opened in a new tab anyway.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -13167,28 +13167,32 @@ TODO:
 
 <!-- Brand Logo -->
 <!-- Place image in masthead -->
+<!-- We either create a link with an image, or just an image. -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Very good!  Thanks.

PR #2391, opened 2025-01-29
LATEX: provide amsart style for a generic ams journal

oscarlevin:
A lot of work on `pretext-latex-classic.xsl` to get the `latex/pretext-latex-amsart.xsl` custom latex style producing a latex file with `\documentclass[amsart]` and its expected environments working with even the sample article.  

There will be cleanup needed eventually of the `-classic` stylesheet, and an upcoming PR to add a `@journal` variable for the publisher file will give a publisher finer control.  But this is working now so is a good jumping-off point for that work.

### DIFF
--- a/xsl/latex/pretext-latex-amsart.xsl
+++ b/xsl/latex/pretext-latex-amsart.xsl
@@ -1,0 +1,41 @@
+<?xml version='1.0'?>
+
+<!--********************************************************************
+Copyright 2025 Oscar Levin
+
+This file is part of PreTeXt.
+
+PreTeXt is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 2 or version 3 of the
+License (at your option).
+
+PreTeXt is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
+*********************************************************************-->
+
+<!-- Conveniences for classes of similar elements -->
+<!DOCTYPE xsl:stylesheet [
+    <!ENTITY % entities SYSTEM "../entities.ent">
+    %entities;
+]>
+
+<!-- Identify as a stylesheet -->
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0"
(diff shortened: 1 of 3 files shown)


oscarlevin:
Spacing issues are resolved.  Also threw back in the \typeouts.

rbeezer:
Diff is improving!  I know the whitespace is a drag - but it leads to really subtle visual differences that folks (publishers!) do notice.

1.  Lots of blank lines, which could make paragraphs.  If you want visual spacing (which I do not think is advisable) it should at least be a line with a percent in the first position.  I do like the markers for the ends of non-environment things.  Very common example:
```tex
%% end paragraphs

% end of subsubsection: A Subsubsection in a Subsection in an Appendix

% end of subsection: A Subsection in an Appendix

% end of appendix: A Structured Appendix

```

2.  How will an `xref` to a division know where to go (needs a `\label{}`?)

3.  `proof` and `cases` have some leading spaces/whitespace.  Can you use a regular expression to find lines leading with a space?  I sorta think there should not be any.

oscarlevin:
Nice catch on the labels on sections; I had it in paragraphs, not sure how I missed it for the rest.

Fixed proof cases, which needed a fair amount of work.  That should be the last of the leading whitespace (outside of preformatted stuff).

For now, I'm acquiescing on the blank lines.  But you haven't seen the last of them!  Since the goal of these is to produce human readable latex, it might be worth the effort to use blank lines the way most people do when writing latex.  But for now, I'll stick with the `\par`.

rbeezer:
Thanks, I'll see if I can do another test drive tomorrow.

rbeezer:
Titles of cases are getting leading space(s).  Maybe a single non-breaking space now that I cut/paste.  `#xA0` character.  Visible in PDF.
```diff
+\noindent\textit{ Why Not Try This?}\label{questioning-case}
```

Maybe a percent after the `\label` in a division lead-in?

Introductions to divisions don't really have any heading of any sort?  (A "starred" division in generic LaTeX?)
```diff
-\begin{sectionptx}{Section}{Graphics}{}{Graphics}{}{}{section-graphics}
-\begin{introduction}{}%
+\section{Graphics}\label{section-graphics}
+% Introduction
```

Warnings about missing/unsupported divisions might caution about targets of `xref` placed elsewhere that could go missing?

A block with a title gets a heading, parentheses with title having a period, then a bold period.  haven't checked if that is what we want?  (Checkpoint 4.6?)

Footnotes should now be better behaved.  Maybe we unwind the drill to pull them out of blocks?

Are assemblages being rendered right?  Look a bit off in Subsection 4.5.

Are graphics being sized?  All look wide in Section 10.  But side-by-sides look good.

xref to "References" in Section 21 is not working.

Now!  That looks like a research article!  PDF builds without complaint.

Merging, though some of the above should be addressed.  Pinging @davidfarmer

PR #2392, opened 2025-01-29
CSS: fix nested tasks not being indented

oscarlevin:
See https://groups.google.com/g/pretext-dev/c/zMSyZxpnLOM.

### DIFF
--- a/css/components/chunks/_exercises.scss
+++ b/css/components/chunks/_exercises.scss
@@ -47,7 +47,10 @@
   }
 }
 
-
+// subtasks are indented
+.exercise-like > .exercise-like.task {
+  margin-left: 20px;
+}
 
 
 


rbeezer:
Merged.  Thanks!  A tiny bit more info coming on the lists.

rbeezer:
Did we expect this to **not** make any changes to ` css/dist/epub.css`, `css/dist/kindle.css`?

rbeezer:
Right.  I noticed no such update.  Is one needed?  Will tasks be indented in EPUB now?

(Got changes to WW fillin for EPUB on #2393, which made me wonder about this.)

@ascholerChemeketa ?

ascholerChemeketa:
A change in that file is not currently expected to create changes in the `epub.css`.

epub.css is generated from https://github.com/PreTeXtBook/pretext/blob/master/css/targets/ebook/epub/epub.scss

Currently that just grabs:
https://github.com/PreTeXtBook/pretext/blob/master/css/targets/ebook/ebook-common.scss

Which includes: 
https://github.com/PreTeXtBook/pretext/blob/master/css/components/_pretext.scss

That file brings in most of the micro level styling - "elements".

The "chunks" files - which are the larger content blocks that tend to vary more from theme to theme currently are not brought into the ebook.css. Some of them likely should be, others should not (asides). Currently, the code errs on the side of leaving things out. We need to figure out what macro styles really belong (and work) in an EPUB.

Until there is time for a full EPUB audit, we could do one of three things:

- Leave CSS as is.
- I can make a PR that adds more of the "chunks" into the build. Essentially erring on the side of too much CSS for ebooks.
- We could consider ebook/kindle with new styles experimental and I could make a PR to have them use old stylesheets for now.

rbeezer:
> Currently, the code errs on the side of leaving things out. We need to figure out what macro styles really belong (and work) in an EPUB.

Thanks for the explanation.  Not even sure we have a nested `task` example to test in EPUB anyway.  ;-)  Though maybe the sample article builds now.  I'm fine with a go-slow and get-it-right progression.  Let's keep it on our radar, wait for pure HTML to settle down (!?), and bring discussions to `pretext-dev`.  Perhaps an open season on requests sometime later.

PR #2394, opened 2025-01-29
Javascript: fix permalinks for tasks

jjrsylvestre:
Permalink code for `task` made an assumption about HTML structure that was no longer true. Changed to be a little more robust.

### DIFF
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -98,7 +98,7 @@ function permalinkDescription(elem) {
         if ((nodeName == 'ARTICLE') && (elem.classList.contains('exercise')) ) {
             typeStr = "Exercise";
         } else if ((nodeName == 'ARTICLE') && (elem.classList.contains('task')) ) {
-            typeStr = elem.parentElement.firstElementChild.getAttribute('data-description');
+            typeStr = elem.parentElement.querySelector(':scope > .autopermalink').getAttribute('data-description');
             numberSep = "";
         } else {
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged, thanks very much.  More discussion on the lists.

oscarlevin:
Is it possible that this broke something else?  After building my book with the latest nightly, I don't see any permalinks after the first task in a multipart webwork question.  See for example: [Paragraph](https://discrete.openmathbooks.org/dmoi4/sec_seq-exponential.html#extracted-webwork-209-1-1-3), 
 which is the last working permalink on that page.

ascholerChemeketa:
Looks like there is an issue:
```
Uncaught TypeError: Cannot read properties of null (reading 'getAttribute')
    at permalinkDescription (pretext_add_on.js?x=1:101:82)
    at pretext_add_on.js?x=1:347:52
```

That points to:
`typeStr = elem.parentElement.querySelector(':scope > .autopermalink').getAttribute('data-description');`

If that querySelect fails (like in Oscar's sample), the resulting error kills the entire `window.addEventListener("load",function(event) {` callback.

The old code should have had a similar issue. But maybe since it was less well targetted, it less often ran into the issue.

The root problem is a lack of error handling anywhere in the giant chain of changes made by that one event handler. In addition to making sure this current line, it would be nice to add in better error handling more generally. Maybe for starters a big try...catch inside the loop on ~316 that generates each permalink. That way one failed attempt doesn't derail the whole process.

PR #2396, opened 2025-01-30
Script: Custom themes can set arbitrary options in pub/html/css

ascholerChemeketa:
Currently only known options are passed on from pub/html/css to the cssbuilder script. So in a theme that does not make use of `secondary-color`, any attempt to set a value in the pubfile will never be seen by the builder.

This loosens that for `theme="custom"` as we don't know what options make sense for it.

Example need: 
Soundwriting project wants a custom theme (their own scss file), but wants to be able to set different `secondary-color` values from different pub files.

Without this change, there would need to be 2 different scss files and each pub file would point to a different scss file.

### DIFF
--- a/xsl/publisher-variables.xsl
+++ b/xsl/publisher-variables.xsl
@@ -2376,8 +2376,8 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <xsl:template name="get-theme-option">
     <xsl:param name="optname"/>
     <xsl:choose>
-        <!-- Must be an option in the theme -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged!  Thanks.

Perhaps this should go in the next documentation update?  Custom theme writers can add any option they want?  Maybe @oscarlevin wants to know about this for schema-writing purposes?

ascholerChemeketa:
Good call. Once I make your final custom Sound Writing theme I'll do a doc update that describes custom themes and links to our as a sample.

PR #2397, opened 2025-01-30
Javascript: move scrollTocToActive after dynamic TOC logic

jjrsylvestre:
Auto-scrolling of TOC on page load appears to be broken. Here's a fix.

### DIFF
--- a/js/pretext.js
+++ b/js/pretext.js
@@ -58,10 +58,6 @@ window.addEventListener("DOMContentLoaded",function(event) {
        thetocbutton.addEventListener('click', () => toggletoc() );
 });
 
-window.addEventListener("DOMContentLoaded",function(event) {
-       scrollTocToActive();
-});
-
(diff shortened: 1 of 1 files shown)


jjrsylvestre:
Hmm, it seems to only be broken in my project, not sure why. Page load time seems to have something to do with it. Though in any event I think the pull request is benign/harmless at worst.

jjrsylvestre:
No, with further investigation I've determined that it really is a bug in the situation that the TOC is set to be pre-expanded in the publication file. In that case if `scrollTocToActive()` is run before the dynamic expanding TOC logic, then `scrollTocToActive()` computes the wrong value to assign to the TOC's `scrollTop` property because the TOC geometry isn't finalized yet.

rbeezer:
Merged, as-is.  Thanks.

This seems to forever be a little tricky.  Let's keep an eye out once it gets out into the wild.

PR #2398, opened 2025-01-31
HTML: Bump mermaid to v11

ascholerChemeketa:
Bumps version number of mermaid to 11. v11 has some new options and diagram types as well as lots of bug fixes.

No true breaking changes in 10->11. (A few defaults for diagrams changed but they aren't actually breaking.)

All the samples in SB still work. 
https://github.com/mermaid-js/mermaid/releases/tag/v11.0.0


rbeezer:
Got it!  Thanks for keeping this up-to-date.

PR #2399, opened 2025-02-01
Improve display of navbar buttons and footer logos in mobile view

ascholerChemeketa:
Addresses:
https://groups.google.com/g/pretext-dev/c/NXRWnX2XKkk/m/zwHgdqLzDwAJ?utm_medium=email&utm_source=footer

I have three commits on the branch, but they could be squashed into one "better mobile" commit

@oscarlevin I made an update to parts-paper since they override some of what the basic navbar does - do you want to check it out? Once we hit the navbar/mobile breakpoint, it tries to undo the custom location of the toc-toggle. Third commit turns off the large side padding on the .ptx-content-footer once we reach that size so that those buttons don't end up squashed.

### DIFF
--- a/css/components/page-parts/_footer.scss
+++ b/css/components/page-parts/_footer.scss
@@ -40,8 +40,12 @@ $nav-height: 36px !default;
 @media screen and (max-width: $navbar-breakpoint) {
   .ptx-page-footer {
     // prevent icons from spreading too much
-    gap: 50px;
+    gap: 20px;
     justify-content: center;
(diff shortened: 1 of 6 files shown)


oscarlevin:
Looks good to me.

rbeezer:
Merged, as-is, but for one character of whitesapce.  Thanks for the quick reactions on this one.

Intro says three commits, but I see only two.  Seems to checkout just fine, so went with it.  Thanks.

PR #2400, opened 2025-02-01
Add theme names as comments; elevate log messages for building css

oscarlevin:
Two minor commits to make debugging css theme issues easier.

1. Elevate a couple `log.debug` messages to `log.info` for the step when either copying or building a css theme.
2. Add a comment to each theme that SASS will put in the minified `theme.css` so it will be clear what theme is being used from the browser inspector.

These should help with an issue I'm running into with building on runestone where it doesn't seem to be using the requested theme.

### DIFF
--- a/css/targets/ebook/epub/epub.scss
+++ b/css/targets/ebook/epub/epub.scss
@@ -1,3 +1,4 @@
+/*! Theme: epub */
 @use '../ebook-common';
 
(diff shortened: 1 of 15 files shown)


rbeezer:
Very good!  Thanks.  One edit -- capitalized "CSS" in one log message.  Otherwise as-is.

PR #2402, opened 2025-02-03
HTML: index custom ol marker styles by counter instead of generate-id()

jjrsylvestre:
As discussed in comments of #2377

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -211,14 +211,14 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- Here we assume there is at most one                      -->
 <xsl:variable name="the-index"          select="($document-root//index)[1]"/>
 
-<!-- ol markers                                                 -->
-<!-- Make a master list of all author-supplied ol marker styles -->
+<!-- ol markers                                                        -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Quick visual scan while I have a minute.  Looks a lot simpler!

Second template with `mode="ol-marker-class"` has a `choose` with a single `when`.  Can't that become a simpler `if`?

If yes, just say so, and I'll make the edit as part of testing.

rbeezer:
> Can't that become a simpler if?

Answered my own question.  Seems so.

rbeezer:
Thanks!  Much better.

I put the one-case conditional on a follow-on commit.  Does not change output for sample article. So we can revert it, if necessary, and I take the blame.

`generate-id()` gives me strings like "id13" (other OSes may be different).  Since I was getting one less style with the PR, I did the following.  I *temporarily* added in an "id" to the marker class names.  Now I got a really clean diff, before/after the PR.  But makes it obvious that pre-PR rule/class 12 and 14 are identical!  And now they are not.  So not a bugfix per se, but certainly an improvement.

> when I only hack on xsl in spurts once or twice a year

Clearly, we need to find more things for you to do, so we can have some more quality contributions like this one.  Thanks.

PR #2403, opened 2025-02-03
Manually manage publisher-variable report lifespan

ascholerChemeketa:
@bnmnetp found an issue with the storing of the publisher variables report. Runestone uses the CLI as a library (as opposed to firing it up in a separate process and communicating with that). And the CLI uses pretext as a library. So any state in the pretext python is carried from book build to book build.

Currently, the publisher variable report is automatically stored after being generated/parsed. 

This is one option for a quick fix. It turns over management of the cached pub-var dict to consumers of publisher variables. Anything that needs a pub variable now make a call to `get_publisher_variable_report` and gets back a full dict. They query that with `get_publisher_variable(dict, key)`. To avoid repetition, an entry point (`html`, etc...) can do that call and pass off the dict to helpers (`build_or_copy_theme`).

Issues: 
- Not terribly efficient. A run of the CLI may involve multiple script entry points that each require hitting rerunning the generation/parsing.

@oscarlevin Looks like this would require adjustment to one call to `get_publisher_variable` in the CLI.

Marked as draft pending discussion.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -2112,7 +2112,8 @@ def qrcode(xml_source, pub_file, stringparams, xmlid_root, dest_dir):
     # Establish whether there is an image from pub file
     has_image = False
     try:
-        image = get_publisher_variable(xml_source, pub_file, stringparams, 'qrcode-image')
+        pub_vars = get_publisher_variable_report(xml_source, pub_file, stringparams)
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged.  There was a comment added to the very end of the file about a cache of publisher variables, and then no real code.  Looked like cruft, so I removed it.

Added a follow-on commit with just a bit about module-level variables being dangerous.

PR #2404, opened 2025-02-04
Allow pub_file (now pub_file_info) to be a dictionary of pub_vars

oscarlevin:
This is an add on to the draft PR of @ascholerChemeketa (#2403) which would allow the CLI or other front ends to be in charge of maintaining the publisher variables dictionary.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -2605,6 +2605,7 @@ def braille(xml_source, pub_file_info, stringparams, out_file, dest_dir, page_fo
     if isinstance(pub_file_info, dict):
         pub_file_path = pub_file_info.get("pub_file_path")
     else:
+        assert isinstance(pub_file_info, (str, None))
         pub_file_path = pub_file_info
     if pub_file_path:
         # parse publisher file, xinclude is conceivable
@@ -4644,9 +4645,10 @@ def get_managed_directories(xml_source, pub_file_info):
     generated = None
     external = None
     # In case pub_file_info is a dictionary, get the pub_file's path
-    if isinstance(pub_file, dict):
+    if isinstance(pub_file_info, dict):
         pub_file = pub_file_info.get("pub_file_path")
     else:
+        assert isinstance(pub_file_info, (str, None))
         pub_file = pub_file_info
     if pub_file:
         # parse publisher file, xinclude is conceivable
@@ -4834,7 +4836,7 @@ def add_pub_file_to_params(pub_file_info, stringparams):
             stringparams["publisher"] = pub_file_info
 
 
-def get_publisher_variable_report(xml_source, pub_file, params):
+def get_publisher_variable_report(xml_source, pub_file_path, params):
     """Parse the pubfile and return a dict containing the variables"""
 
     # IMPORTANT: to report the value of a (computed) publisher variable,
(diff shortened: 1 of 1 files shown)


oscarlevin:
Okay, this is all fixed up as a single commit.  I think it is quite clever, but understand if you think it's *too* clever.

Summary of changes:
- The big idea: the `pub_file` variable that is in the signature of almost every function here is changing from a string (a path to the publication file) to either a string or dictionary.  
  - If it is a string, then everything works as before.  
  - If it is a dictionary, then one of the keys in the dictionary will be `"publisher"` with value the string-type path to the publication file.
- To remind us of this, variable name has changed from `pub_file` to `pub_file_info`.
- There are three ways we currently use the `pub_file`:
  1. Add it's path to stringparams.  This is now done with the function `add_pub_file_to_params()` which checks if the type is a dictionary or string and does the right thing accordingly.
  2. Use the `get_publisher_variable_report()` to create a dictionary of pub_vars and then `get_publisher_variable()` to get the value of a particular variable.  All instances of the former have been replaced with `ensure_publisher_variables()` which either does nothing (if already a dictionary) or calls `get_publisher_variable_report()`.  
  3. Use lxml to parse the publication file directly.  This happens in two places: as part of the braille conversion and part of the `get_managed_directories()` function.  I think in both cases these could be refactored to use the `get_publisher_variable()` function, but until then, we define `pub_file` to be the path appropriately.

Why all this?  It is to address the Runestone use case of starting a worker to runs pretext, through the CLI's API.  Previously this was a problem because as soon as the `pub_vars` dictionary was created, that would either persist between book builds, or (core) pretext would need to rebuild the dictionary for every step of building a book (generating each asset and building output).  With this change, the CLI can create `pub_file_info` as a dictionary once for each book and use that in all the build/generate steps.

rbeezer:
Will try to fix this with `pub_file` -> `pub_file_info`.  But if you see this immediately, you can checkoff

```
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 855, in <module>
    main()
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 579, in main
    dest_dir = get_destination_directory(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 144, in get_destination_directory
    generated, _ = ptx.get_managed_directories(xml_source, pub_file)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rob/mathbook/mathbook/pretext/pretext.py", line 4647, in get_managed_directories
    if isinstance(pub_file, dict):
                  ^^^^^^^^
UnboundLocalError: cannot access local variable 'pub_file' where it is not associated with a value
```

oscarlevin:
Ah yes, that is likely the result of a bad resolution to a merge commit.

rbeezer:
Thanks.  EPUB is not building.  Some information is passed between stylesheets with string parameters, so maybe a mutation is being lost?  Just a guess.

The `$chink-level` variable is set **late** because it depends on structure which could change with versions.  Not sure if that is relevant or just an artifact.

```
PTX:ERROR   : * PTX:BUG:   the $chunk-level variable is empty, so must have been set improperly, perhaps when overridden in a conversion.  The "containing-filename" template is likely to act incorrectly.  Please report me
```

oscarlevin:
I'll look into it

rbeezer:
Since this needs discussion...

On a first read, the `braille()` and `get_managed_directories()` have an `if/else`.  The `if` is a `dict` and the `else` is everything else (should be `str` or `None`).  My firt reaction was to put in more safeguards, convert `else` to `elif` and use a final `else` that provided some sort of warning (maybe setting `pub_file` to `None` to allow execution to proceed, or not).

Then I encounted some `assert`.  Good.  I can't tell (I tried) if the current `assert`s will ensure the `if/else` I saw to always be "safe".  Is there maybe a more obvious location or choke-point where an `assert` can check: `dict`, `string` or `None`?  Maybe `ensure_publisher_variables()`? Or maybe put more `assert` where I saw the `if/else`.

Very clever, so take the above with a grain of salt, I may not be appreciating every nuance.

oscarlevin:
I likely did not consider the need for the `None` case.  I'll play around with this more, and if needed, we can discuss tomorrow.  Thanks for the comments.

rbeezer:
Yes, I was going to suggest a discussion tomorrow.  I've got some vague ideas.

oscarlevin:
Closing this in favor of speeding up `get_publisher_variable_report` by subsetting the assembly stylesheet.

PR #2405, opened 2025-02-04
Script: fix html watermark svg generation

ascholerChemeketa:
Fix for:
https://groups.google.com/g/pretext-support/c/6M-o-Q8Mb64/m/5rl1-edMAAAJ?utm_medium=email&utm_source=footer

SVG was being serialized from XML without its namespace. Also moved placement to ptx-main which covers more of the viewport when there is not much content on a page.


rbeezer:
Excellent!  Thanks.

PR #2407, opened 2025-02-04
Update sublime editor

mafitch:
Updates help documentation with current status of Sublime text

### DIFF
--- a/doc/guide/appendices/editors.xml
+++ b/doc/guide/appendices/editors.xml
@@ -19,9 +19,9 @@
         <author>Dave Rosoff</author>
 
         <introduction>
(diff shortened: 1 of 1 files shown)


kcrisman:
I hope no one will jump on me for pointing out one niggle.  You've removed other references to Atom, but in the @oscarlevin old text which you moved to emphasize VS Code it still mentions it (because of its previous position in the doc below Atom).  It's not grammatically wrong or anything, but thought you might be intending to remove all such references. HTH.

mafitch:
That is now fixed.

kcrisman:
👍

rbeezer:
Thanks, @mafitch!  Merged, with content as-is.

I **love** having multiple discrete compact commits.  But here. I think all the changes are related and about the same place.  So I smashed them all together (which is **way** easier than splitting out one big commit into multiple commits!).

PR #2408, opened 2025-02-05
Javascript: add try/catch logic around permalink code

oscarlevin:
...to prevent one failure from breaking page.  It looks like when #2394 was merged, it introduced an error on webwork tasks.  When one permalink didn't compute correctly, that would break the entire loop that builds all the permalinks.  As suggested by @ascholerChemeketa, this will simply break the one offending permalink and keep building the rest.

Definitely a stopgap until permalinks move out of javascript.

### DIFF
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -314,55 +314,60 @@ console.log("this is e", e);
         this_url = window.location.href.split('#')[0];
         permalink_word = "&#x1F517;";
         for (var i = 0; i < items_needing_permalinks.length; i++) {
-            this_item = items_needing_permalinks[i];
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged, as-is.

> until permalinks move out of javascript.

Where are we on that?  (I've forgotten.)  Is the ball in my court?

oscarlevin:
> > until permalinks move out of javascript.

> Where are we on that? (I've forgotten.) Is the ball in my court?

I don't think the ball is in your court.  But I do think we should chat about it with interested parties at some point soon (perhaps after @ascholer is done with lti stuff).

PR #2410, opened 2025-02-05
LaTeX: move instances of 'before upper app' to last in line

Alex-Jordan:
This has to do with the thread "a research article in PreTeXt" in the pretext-dev group.

`before upper app` settings for tcolorbox styling should come at the end, lest some other styling use `before upper` and obliterate whatever `before upper app` had appended.

### DIFF
--- a/xsl/pretext-latex-common.xsl
+++ b/xsl/pretext-latex-common.xsl
@@ -2948,9 +2948,9 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:if test="$b-pageref">
         <xsl:text>\label{#3}</xsl:text>
     </xsl:if>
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged.  Small merge conflict with glossaries, owing to the new `\notblank` from some other recent PR.  I think I fixed that up properly.

The various styles came last, so as to provide a method for over-writing default choises (much like CSS).  This change means there is no opportunity to do something different with the "parstyle".  But perhaps that is the way it shouild be?

Alex-Jordan:
> This change means there is no opportunity to do something different with the "parstyle". But perhaps that is the way it shouild be?

Hmm, that does sound like a concern. Maybe we (plus @oscarlevin) can find a better solution. Or maybe it's OK like you suggest.

Issue #2412, opened 2025-02-06
100% width latex-images in naked images are not 100% width in PDF


Alex-Jordan:
This is a subtle spacing issue with naked latex-images in PDF. (It may also be an issue with non-naked images, and not-latex-images, but I haven't investigated.)

A naked image using latex-image in a division will typically be authored in a way that ends like:

```
            ...
        \end{tikzimage}
    </latex-image>
</image>
```

This ends up in PTX-generated LaTeX as:

```
    ...
\end{tikzpicture}
}%
\end{image}%
```

There is a new line there at the end of `\end{tikzpicture}`. This newline is responsible for a small amount of horizontal space to the right of the picture. So if the picture is 100% width, it actually will not quite be flush with the right edge of the surrounding text body. Here you can see this with a naked image/tikz-image I put into the minimal article:

<img width="676" alt="Image" src="https://github.com/user-attachments/assets/e03f3bd3-8862-4f67-9dcf-6086f1315286" />

If I manually alter the tex, adding a `%` so we have:

```
    ...
\end{tikzpicture}%
}%
\end{image}%
```
then the image comes out at 100%.

I had trouble coming up with an automated way to get a `%` sign in there like that. The author's actual content has a text node like `...\end{tikzimage}&#xa;    `, so it's a question how you would insert that `%` there. And you wouldn't want to insert it before other newlines in the author's code. And counting the last line break is not reliable either since they may have additional line breaks at the end.

Maybe it's an unfortunate "author education" situation where they should author the `%` signs?

rbeezer:
Recursively strip final whitespace characters?

We'll get rid of every stray character in LaTeX one day...

rbeezer:
And you might even find an existing template in the file full of such things (pretext-text-utilities.xsl).  Or else write a general enough one to go there.

Alex-Jordan:
Is that sort of stripping something to do in `-assembly`?

rbeezer:
I think I would do it in the extraction and in the Latex.  Maybe a template in common.  Which is not to say assembly is bad, but there is a lot going on there.

Alex-Jordan:
Ah, I was already doing something in assembly and didn't see this first. Check out #2413 and let me know what you think.

rbeezer:
Thanks, Alex, this was a great catch.  Resolved at #2413.

PR #2413, opened 2025-02-06
Assembly: trim trailing whitespace in latex-image code and cap off wi…

Alex-Jordan:
…th % sign

This is to address #2412.

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -1422,6 +1422,15 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     </xsl:copy>
 </xsl:template>
 
+<!-- 2025-02-06: remove trailing whitespace in the last text node of latex-image -->
+<!-- and cap off with a % to prevent unwanted space in output                    -->
+<xsl:template match="latex-image/text()[last()]" mode="repair">
+    <xsl:call-template name="strip-trailing-whitespace">
+        <xsl:with-param name="text" select="."/>
+    </xsl:call-template>
+    <xsl:text>%</xsl:text>
+</xsl:template>
+
 
 <!-- ############################## -->
 <!-- Killed, in Chronological Order -->


rbeezer:
If in assembly then `repair` is the place.

There are other places we fix up text and whitespace, so let me think on this.  In other words, wait on a review and don't write any new code.  Thanks.

rbeezer:
A lot of thinking about this one.  And my gut tells me that we shouldn't be in the business of cleaning up lenient source in `-assembly`.  Mostly we preserve lots of whitespace in source to keep changes to a minimum when examing other effects.

Look in `-common` and search on the variable named `back-cleaned`.  Why doesn't the trailing whitespace go away here?  Then it would just be a matter of adding a percent sign in the "extract" template and in `pretext-latex.xsl`.  I think.  I'll go experiment right now, so sit tight for a bit.

rbeezer:
Don't apply PR.  Build `assembly-static` and `latex`.  Add several newlines to source of a `latex-image`.  Build both again, and check diff.

Newlines survive assembly (they are present still).  LaTeX is the same, meaning the new lines get stripped.

There *is* an added, single newline at the end of the authored LaTeX/TikZ/etc, which I think is owing to the use of the `sanitize-text` template (a poor design decision maybe, but a dangerous one to unwind).

So my thought is that a **guaranteed** final newline should be stripped and a percent sign and a newline be added.  I think that newline is stripped other places.

There might be an opportunity to move the template matching "latex-image" into `pretext-common.xsl`, and reduce (eliminate?) the dependence on `pretext-latex.xsl` (`pretext-latex-common.xsl` now, I think) inside `extract-latex-image.xsl`.

Alex-Jordan:
It's been too long for me to remember for sure what I used there. I am sure I would have tried `\fbox{...}` around the `\begin{tikzpicture}...\end{tikzpicture}`. I may have also tried some sort of tikzpicture-specific option after Googling.

rbeezer:
OK, the desired percent sign is in.  I can see it helping with a test image I messed around with.  For another image ("TikZ Electronics", unwound from its `figure`) it didn't seem to matter as it was fine without it seemed.

I now appreciate that TikZ adds padding (and I could **not** find an easy switch to turn it off), while our cropping removes that.  So there is some infidelity between PDF and HTML, especially obvious at 100% width.

Some education, or some TikZ (simple) suggestions, are probably in order, but I went 'round and 'round for too long and gave up.

Thanks for working this one out.  By the time I'd reviewed the PR it was just as easy to go ahead and make the change.  Reminded me of some dead code for `latex-image/label` that also needs attention.

Updated sample article images.  PDFs are 4x bigger.  :-(

Issue #2414, opened 2025-02-06
`sidebyside/video` not appearing in PDF


Alex-Jordan:
According to schema, a `video` is a legal child of a `sidebyside`. However in PDF, such a `video` is completely missing. For example on page number 159 of the sample article, right before section 26 "Open Problems", the source has a sidebyside with two videos. But there's nothing there in output.

Theory that I have not tested: the `video`s there are converted to `sidebyside` in `-asssembly` (a sidebyside to show their preview and a QR code). But now they exist as a `sidebyside` within a `sidebyside`, so are ignored.

I could work on this but I'm not sure what is supposed to happen. Maybe form a `stack` with the preview image and the QR code?

rbeezer:
Short answer, schema lags.

QR code and all is too much to pack into a print sidebyside.

So the only thing to do is to fix schema and remove examples.

Alex-Jordan:
No objection here, but I'll share some thoughts I had in the meantime.

- Is it a matter of time before my phone's camera app does OCR and recognizes a URL I can tap? In other words, is part of the raison d'etre for QR codes going to fade? If so, a `stack` with an `image` and a URL underneath, even wrapping lines, is maybe not so bad.
- In the meantime, I was wondering how a preview image with an embedded QR code might look. Imagine the preview image, but something like the upper right is overlaid with a QR code square. Or the QR code is overlaid in the center of the preview image. Or flop that, and have a bit larger QR code with the video's preview image in the center.
- Schema might change to disallow `video` as a sbs panel. But the schema will still allow `figure` as an sbs panel, and that figure could have a `video`. I guess that will succeed in a technical sense with conversion to `sidebyside/figure/sidebyside` and then the preview and QR code images within. But the actual look of it may be pretty bad if so much is crammed into an already narrow figure panel.

rbeezer:
Closing — the construction behind this report is no longer permitted, which resolves it by removing the pattern rather than rendering it.

As of `6bf9978f` (2026-07-22) the schema precludes `audio`, `video`, and `interactive` as descendants of a `sidebyside`, at any depth. The reason is the very mechanism behind this bug: in static output (PDF/print) each of these media elements is itself represented as a `sidebyside` — a preview image, a QR code, and links — and a `sidebyside` may never nest inside another. A `video` placed as a panel therefore had nowhere to render and silently dropped.

[...]

Issue #2406, opened 2025-02-04
LATEX: `\setparstyle` needs some tweaks


oscarlevin:
@davidfarmer shared an example of a `sidebyside` that was overflowing in LaTeX due to a `\setparstyle` in the `sbspanelstyle`.  This might also relate to an issue with paragraphs indents in proofs.

In any event, we should sure this up.  Need to be careful about paragraphs inside a sidebyside though.

I can take a stab at this, unless @Alex-Jordan has a quick fix.

Alex-Jordan:
Is there a place I can see the example of this?

Alex-Jordan:
OK, I figured out this is part of @davidfarmer 's recent post about a journal article.

Alex-Jordan:
I believe the issue here is not about `\setparstyle`. Rather it is that for an `image` (or a `tabular`) in a `sidebyside`, there should be no indentation. I am about to open a PR that does that.

PR #2415, opened 2025-02-06
LaTeX: do not indent before an image or tabular within a sidebyside

Alex-Jordan:
This concisely addresses an issue that led to #2406.

I have tested this with `sidebyside`s that have `image` and `tabular` children and it fixed things. I also tested all the other schema-compliant children of a `sidebyside`, including within a `stack`, and nothing else seems to need any attention. Except I did discover issue #2414.


rbeezer:
Excellent!  Merged as-is.

Testing revealed that Figure 26.26 of sample article was quite bad.  Table in first panel had indentation, which overlay it onto graph in the second panel **plus** the graph was partially in the right margin **plus** then an entire line of text got moved onto the next page as a result.  ;-)

Curiously January 29 sample article PDF online does not show this behavior?  I didn't try to chase that down, nor look to see what changes might also have been in effect recently.

PR #2416, opened 2025-02-11
Script: extend time given for cssbuilder to finish its work

ascholerChemeketa:
pretext.py currently allows for 10 seconds for cssbuilder to do its work. (Normally should take under 1 second).

But on a slow machine, with a cold cache, it apparently can exceed that limit. This gives up to a full minute for the process to complete.


rbeezer:
Merged, thanks!

PR #2418, opened 2025-02-16
CSS: Better handling of long names in navbar buttons

ascholerChemeketa:
Addresses issue with longer titles in navbar buttons:
https://groups.google.com/d/msgid/pretext-dev/20e12ac3-2524-4598-b963-93c3f3e26763n%40googlegroups.com?utm_medium=email&utm_source=footer

### DIFF
--- a/css/components/page-parts/_navbar.scss
+++ b/css/components/page-parts/_navbar.scss
@@ -44,7 +44,11 @@ $center-content: true !default;
       // Disable normal borders - top/bottom provided by container
       // Use extras/_navbar-btn-borders.scss for side borders if desired
       border-width: 0;
+    }
 
+    .name {
+      overflow: hidden;
+      text-overflow: ellipsis;
     }
   }
 
@@ -92,7 +96,7 @@ $center-content: true !default;
   }
 
   .index-button {
-    width: 70px;
+    min-width: 70px;
   }
 
   .runestone-profile {


rbeezer:
Testing with Italian.  Of course.

*  Wider "Index" button is great.
* Prev/Up/Next still get truncated (expected), but there is no ellipsis.  Not in (recent) Chrome nor (recent) Firefox.

ascholerChemeketa:
Those buttons are generated with these localizations:
```xml
    <!-- Keep these short, so buttons are not overly wide, 4 characters maximum -->
    <localization string-id='previous-short'>Prec</localization>
    <localization string-id='up-short'>Su</localization>
    <localization string-id='next-short'>Succ</localization>
```

They are truncated in source, not by HTML/CSS. So there is no ellipsis to display.

rbeezer:
Dang.  Sorry for the noise....

rbeezer:
I made a `previous-short` button text insanely long.  Got an insanely long button.  No ellipsis.  Does the button need a `max-width`?

In addition, we already  have text for full names, and text for short names.  We use the short names as abbreviations on the buttons.  The long names go in tooltips.  The OP's request now strikes me as a bit misguided/problematic.  We already have a mechanism to keep text from spilling out of the buttons.  If we automatically truncate with ellipsis, we are more automatic but maybe lose an option to be more accurate/informative?

rbeezer:
Yes, the Index button changes are good and should be kept.  I'll keep those, and reply to OP about current scheme.

rbeezer:
Merged, with just changes for the Index button.  I think we misunderstood teh OP, maybe this was all he was suggesting be changed.  Thanks!

PR #2422, opened 2025-02-24
New journal styles and frontmatter improvements

oscarlevin:
The six commits here take care of some frontmatter for articles maintenance (style the html, add author-level support elements consistently, improve the guide, and fix the schema).  Additionally, there are three new journal styles implemented in the final commit (which required some refactoring of the original amsart style and latex-classic).

In particular, most of the remaining issues from PR #2391 have been resolved.

Note that the "ejc" style (for the Electronic Journal of Combinatorics) requires the existence of `e-jc.sty`.  For testing, this can be downloaded from [their website](https://www.combinatorics.org/files/e-jc.sty).  We will need to think about how to ensure such files are available if we want to build pdfs, but that's a question for another day.

### DIFF
--- a/css/components/elements/_front-matter.scss
+++ b/css/components/elements/_front-matter.scss
@@ -68,26 +68,32 @@
     margin: 4em 2em;
   }
 
-  .abstract > .title {
-    font-size: 1.125em;
-    font-weight: 600;
-    line-height: 1.125em;
-    display: inline;
+  .keywords, .support {
+    margin-top: 2em;
   }
 
-  .abstract > .title::after {
-    content: ".\2009\2009\2009";
-  }
-  
-  .abstract > .title + .para {
-    display: inline;
+  .abstract, .keywords, .support {
+    & > .title {
+      font-size: 1.125em;
+      font-weight: 600;
+      line-height: 1.125em;
+      display: inline;
+    }
+
+    & > .title::after {
+      content: ".\2009\2009\2009";
+    }
+
+    & > .title + .para {
+      display: inline;
+    }
   }
 
   .colophon {
     .copyright {
       margin-top: 2.5em;
     }
-  
+
     .license {
       margin-top: 2.5em;
     }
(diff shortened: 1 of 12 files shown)


oscarlevin:
And a 7th commit to add bibinfo to the sample article for testing.

rbeezer:
OK, a lot of review completed, but more to do.  Here is a partial list of things to think about in the meantime.

*  We need to deprecate `address` for `location`?  Fix it in -assembly, make warnings, etc?
* I can't tell what the `tf` joiurnal is.  Once we have a good scheme, we'll need to document journals and abbreviated names.
* There are some empty `\textit{}` for cases (with two leading spaces?).  Should be able to condition on when this argument is empty and not write the italic part at all?
* Love the comments for the ends of environments.  Nit: why does the end of a `paragraphs` have two percent signs?

oscarlevin:
Force pushed with fixes for the `\textit{}` for cases and inconsistent end division comments.  

There are two spaces before the case titles to match the style in `pretext-latex-common`.  

As we discussed, `address` was never implemented, just incorrectly in the schema.

The weirdness of `\\&#xa;` starting some cases in the author stuff shows up in `latex-common` as well (also my fault).  I think it makes sense to wait on a fix for that as a separate PR.

rbeezer:
In HTML, with the sample article, the ToC "Abstract" entry makes a pink flash that contains the abstract **and** keywords and more.

Is that the sort of structure (HTML divs, page chunking) we expect?

oscarlevin:
> Is that the sort of structure (HTML divs, page chunking) we expect?

I think so?  I was modeling this off of amsart that puts the keywords and msc inside the abstract.  It would be an easy fix to move them separate, but I don't think they should get their own toc section.

rbeezer:
Merged, as-is.  Tested with base sample article, amsart (PDF build), and ejc (LaTeX source only).  Questions coming at Drop-In.

PR #2425, opened 2025-02-25
Implement journal publisher variable

oscarlevin:
Once you feel caught up on other PRs, I'd appreciate your feedback on this, @rbeezer. 

There will be a many-one mapping from journals to latex-style options.  Example: the Bulletin of the AMS will use `latex-style="amsart"`, but will need to change the `documentclass` .  The "publisher" will be able to set `publication/common/journal/@name="bull-amer-math-soc"` and everything should be set for them.

So far, this PR has the following:
- New publisher variable "journal-name"
- Pretext script for latex gets this name, checks it against a "database" of journals (in `journals/journals.xml`) and sets the `latex-style` accordingly. 
- Start of updates to documentation and publication schema.

Still needed:
- [ ] script that will transform the `journals/journals.xml` file into a pretext table that can be xi:included in the documentation.
- [ ] Decide how to represent additional information about a journal that will be needed for the conversion, including separate documentclass, additional .sty or .cls files that might need to be downloaded, etc.
- [ ] XSL and python code that utilizes that additional information.

### DIFF
--- a/doc/guide/appendices/journals-table.xml
+++ b/doc/guide/appendices/journals-table.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<table>
+  <title>Journals supported by PreTeXt</title>
+  <tabular>
+    <row header="yes">
+      <cell>Full Journal Name</cell>
+      <cell>Code</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Bulletin (New Series) of the American Mathematical Society</cell>
+      <cell>bull-amer-math-soc</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Conformal Geometry and Dynamics</cell>
+      <cell>conform-geom-dyn</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Journal of Algebraic Geometry</cell>
+      <cell>j-algebraic-geom</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Journal of the American Mathematical Society</cell>
+      <cell>j-amer-math-soc</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Mathematics of Computation</cell>
+      <cell>math-comp</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Quarterly of Applied Mathematics</cell>
+      <cell>quart-appl-math</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Proceedings of the American Mathematical Society</cell>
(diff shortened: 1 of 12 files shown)


rbeezer:
I'll give it a survey once I merge the recent related PR.

In the meantime, I have an idea about how to handle the fact that different bits of information will appear in different orders in different journals.  Think Citation Stylesheet Language as a model.  Best discussed at a Drop-In.

oscarlevin:
The more I think about it, the more I'm liking the idea of using a CSL-like structure for describing the LaTeX markup for different journals.  A few questions while I'm thinking about then, which we could discuss at a later time.

1. Am I right that the idea would be to use the `document()` function to read in the journal xml file, and then read parts of it to make decisions about layout?  Or perhaps apply templates templates to the nodes in `document("../journals/journal-name.xml")` and apply templates from the main source inside those templates?  Something like that?  Can the argument of `document()` be stored in a variable?
2. Some of this depends on how bibliographies are going to work.  Specifically, we should figure out how fine-grained we will be with selecting a journal.  Right now, I think there will be a `*-general` for each major publisher, that works with lots of journals that will basically be the same (although you might need to specify the actual journal name in the tex somewhere).  Do we end up with lots of journal-name.xml files that are identical except for the name?  Should we have a single giant xml file with all the journals that provide all the relevant details?  I was thinking that the `journals.xml` database file would also reference the correct information to build bibliographies.  
3. It would be nice if there was a template or enough well-documented examples of these files for contributors to base new journals specifications off of.  Do you have thoughts about where in the repository all these files should go?

That's it for now.  I'll keep thinking and we can discuss next time.

oscarlevin:
I've split this up, rebased it on master, and added some xsl to transform the xml file of journal names into a pretext table that is included in the guide.

As far as I'm concerned, this is ready to go, as a step along the way to getting more journal-specific information into the `latex-classic` stylesheet.

rbeezer:
I like it!  Should be easily expandable as the universe of journals expands.  Maybe this is the Big Bang.

*  There were two 4-argument `get_publisher_variable()` calls in `get_latex_style()`.  I silently added a `get_publisher_variable_report()` and went back to current 2-argument versions.
*  Follow-on commit with some minor reformatting in the Python, mostly to shorten some long lines.  And to lower-case `JOURNAL_XML`.
*  One bit of whitespace.  One bare `latex`.  Little bit of commit message editing.  The usual drill.
*  With a bad journal name, it said "Using defaults instead."  I wasn't sure what that would be.  What is the default journal?.  Once I figured it out, I switched this to be more explicit: "Using the default LaTeX style."

PR #2426, opened 2025-02-27
Log errors from Sage conversion

siwelwerd:
Currently, if a sageplot fails to build, the only error displayed to the user is a "FileNotFound" error resulting from trying to copy the generated image that did not get created.  In particular, there is no way to view the error message produced by Sage, making it difficult to troubleshoot errors.  This tweaks the building of sageplots to capture and log errors from Sage.  As a result, now an author can view the error passed on from Sage by using the `-v debug` flag.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -535,11 +535,28 @@ def sage_conversion(
         stringparams["subtree"] = xmlid_root
     xsltproc(extraction_xslt, xml_source, None, tmp_dir, stringparams)
     with working_directory(tmp_dir):
+        failed_images = []
         for sageplot in os.listdir(tmp_dir):
-            if ext_converter:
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, Drew.  That looks like a great addition.  Working through a **HUGE** backlog of PRs, but will try to get to this real soon.

rbeezer:
Sorry for the delay on this one.  Not only a backlog, but I an slowlly configuring a new machine, and had to get things squared away to run Sage locally.  And then forgot to close this after merging it...

Anyway, this is a great improvement.  Thanks.

Removed one hunk of whitespace, and edited the commit message (as is my usual practice).  So not a true merge, and therefore you'll want to simply delete your own local version of your `sageplot-error-handling` branch.

PR #2427, opened 2025-02-28
Workbooks: fix workspace in exercisegroups

Alex-Jordan:
At least with my 2023 version of texlive, worksheets with an exercisegroup are coming out with the workspace attribute ignored. This was also reported in -support. Hard to say if it is an issue for all LaTeX distributions. The change here to `pretext-latex-common.xsl` fixes that.


rbeezer:
Thanks for the fix.  I duplicated the worksheet into the sample article so tht it will receive regular testing, and split into three commits..  Added some formatting to the `introduction` so pieces look like elements/attributes.

PR #2428, opened 2025-02-28
CSS: darken keyword prism color to meet accessibility requiremenets.

oscarlevin:
A matching fix is already PR'ed on runestone for active code.


rbeezer:
Very good!  Merged as-is.  Thanks.

PR #2429, opened 2025-03-03
Portable html: the css and javascript parts

oscarlevin:
This adds a new publication variable at `/publication/html/platform/@portable`, which when set to "yes" will result in the HTML using a CDN for all css and javascript, and not copy those files to the output.

This is set up to allow the CLI to set a string-param for `cli.version`, which will be added to the URL for the CDN (since we will keep the versions of the css and javascript versioned identically to the cli).

The documentation and schema for the publication file had outdated information about `platform`: they said that `@platform` was an attribute of `/publication/html`, rather than `platform` being a child element of `/publication/html`.  That is fixed in the guide and publication schema, adjacent to the new `@portable` attribute.

Happy to rename things as needed, but this does seem functional now.

### DIFF
--- a/doc/guide/publisher/publication-file.xml
+++ b/doc/guide/publisher/publication-file.xml
@@ -647,14 +647,31 @@
         <subsection xml:id="online-platform-options">
             <title>HTML Platforms</title>
             <idx><h>platform</h><h>HTML</h></idx>
+            <idx><h>host</h><h>HTML</h></idx>
             <idx>online platforms</idx>
 
             <p>The<cd>
-                <cline>/publication/html</cline>
-            </cd>element can have an attribute <attr>platform</attr> with values:<ul>
+                <cline>/publication/html/platform</cline>
+            </cd>element can have an attribute <attr>host</attr> with values:<ul>
                 <li><c>web</c>: the default, meant for self-hosting with no server configuration, features, or assumptions</li>
                 <li><c>runestone</c>: output meant for hosting on a Runestone server (<xref ref="runestone"/>)</li>
             </ul>Here <term>platform</term> refers to the server where the <init>HTML</init> output will eventually be hosted.  The effect is to create minor variations in the output to take advantage of extra features of the indicated platform.</p>
+        </subsection>
+
+        <subsection xml:id="online-portable-options">
+            <title>Portable HTML</title>
+            <idx><h>platform</h><h>HTML</h></idx>
+            <idx><h>portable</h><h>HTML</h></idx>
+            <idx><h>portable html</h></idx>
+
+            <p>To limit the number of files and directories created by the HTML conversion,
+            you can set the
+            <cd>
+                <cline>/publication/html/platform/@portable</cline>
+            </cd>
+            attribute to the value <c>"yes"</c>. 
+            This will result in the HTML using hosted CSS and Javascript files, rather than including them adjacent to the rest of your output.
+            This is especially useful in conjunction with setting the <xref ref="common-chunking-options">chunking variable</xref> to 0 to get a single html file.</p>
         </subsection>
 
         <subsection xml:id="online-style-options">
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -3484,7 +3484,7 @@ def _runestone_services(stringparams, ext_rs_methods):
     # ext_rs_methods - an optional function that can replace the querying
     #                   of the Runestone Services file
     # Result - returns five pieces of discovered or set information,
-    #          see *two* retrn statements, one is intermediate.
+    #          see *two* return statements, one is intermediate.
     #          Failure is an option, if a network request fails
 
     # Canonical location of file of redirections to absolute-latest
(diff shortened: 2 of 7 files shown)


rbeezer:
Appears a `pretext/pretext` build will use `latest`, right?  That's a good thing.  (haven't looked - is that documented?)

rbeezer:
This has a merge conflict.  Probably with #2403?

I thought I fixed it OK, but maybe I need an `ensure` call prior.  Otr is that #2404.  :-(  Anyway, rrebasoing, guessing at a fix, then building HTML gives:

```
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 855, in <module>
    main()
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 731, in main
    ptx.html(
  File "/home/rob/mathbook/mathbook/pretext/pretext.py", line 3864, in html
    if get_publisher_variable(xml, pub_file, stringparams, 'portable-html') != "yes":
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

oscarlevin:
Right, the correct fix would be to use `ensure` but that would require #2404.  Which is why I was waiting to resolve the merge conflict to see which PR would win.

rbeezer:
Bloddy hell.  ;-)  Well, I'm invested in #2404 and this.  So you do want you want and tell me what to do first?  ;-)

oscarlevin:
I just pushed a commit on #2404, but can't figure out the epub thing (I get other errors, so can't figure out how to test correctly).

I think though that this here is easier to fix and less controversial, so perhaps let me resolve the conflict and you can do this one first.

oscarlevin:
That should be correct. And it should be safe to squash just the last two commits

rbeezer:
Your last addition is a merge of master into solo-html.  When I view it online, I'm not sure what I am seeing.

Can you **rebase** your `solo-html` branch and force-push?

oscarlevin:
Done.  I tried using github's online merge conflict resolver, and thought it was just committing that one additional change.  Last time I do that!

rbeezer:
Still getting
```
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 855, in <module>
    main()
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 731, in main
    ptx.html(
  File "/home/rob/mathbook/mathbook/pretext/pretext.py", line 3864, in html
    if get_publisher_variable(xml, pub_file, stringparams, 'portable-html') != "yes":
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: get_publisher_variable() takes 2 positional arguments but 4 were given
```
I don't see `get_publisher_variable()` defined to take four arguments.

oscarlevin:
oooooh.  Okay, I thought it was just the merge conflict.  Went back and modified the functions to align with #2403.  Even tested it in both configurations now.  Third time lucky.

rbeezer:
>  Third time lucky.

`<audible-chuckle/>`

rbeezer:
Sample article has "custom" list markers.  Which go into their own CSS file.  Which I think is always non-empty (and so gets copied along).  Is this file being not-created?  Not sure whay this fails this way.  Will try digging deeper.
```
  File "/usr/lib/python3.11/shutil.py", line 825, in move
    os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/ptx-rwsmpkhn/ol-markers.css' -> '/tmp/ptx-rwsmpkhn/_static/pretext/css/ol-markers.css'

During handling of the above exception, another exception occurred:

  File "/home/rob/mathbook/mathbook/pretext/pretext", line 855, in <module>
    main()
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 731, in main
    ptx.html(
  File "/home/rob/mathbook/mathbook/pretext/pretext.py", line 3898, in html
    move_ol_marker_css(tmp_dir)
  File "/home/rob/mathbook/mathbook/pretext/pretext.py", line 3761, in move_ol_marker_css
    shutil.move(src, dest)
  File "/usr/lib/python3.11/shutil.py", line 845, in move
    copy_function(src, real_dst)
  File "/usr/lib/python3.11/shutil.py", line 436, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.11/shutil.py", line 258, in copyfile
    with open(dst, 'wb') as fdst:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/ptx-rwsmpkhn/_static/pretext/css/ol-markers.css'
```

oscarlevin:
Huh.  I had not anticipated dynamically generated css.  Not sure how to deal with that, unless we write it to the html.

rbeezer:
> Not sure how to deal with that, unless we write it to the html.

I think @ascholerChemeketa had some ideas about how to do that.

oscarlevin:
Until a better solution presents itself, let's just not try to copy that file for non-portable builds.  

This means we are testing for the portable build three times in the html function.  I simplified this by creating a boolean `include_static_files` to make the code look prettier.  I'm not sure how important the order is (place runestone, then copy managed directories, then copy pretext js/css, then xsltproc, then move ol markes), but I'm afraid to mess with it right now.

Tested with the sample article in portable and non-portable modes.  Both build; portable defaults back to non-custom markers, but still looks okay.

rbeezer:
Code comment: `'Add a prefix for the cdn, which is empty unless the portable html variable is true`

Equals "latest", not "empty", no?

The Runestone version is not being read from the Runestone services file, so I see an empty string in `eBookConfig` as part of a diff before/after.  Similarly, I suspect variables like `$rs-js` are empty, so when tokenized, the looping (`for-each`) is empty, so no `link`/`script` are being written into the head.  Not clear to me just which **change** is provoking this.

Shouldn't we be pointing to the Runestone CDN?  Or does the Runestone Services JS/CSS live on `jsdelivr` and it needs to be pointed to?  Certain features, like a multiple choice problem are active without hosting on Runestone.

ascholerChemeketa:
The idea for generated extra CSS was that we can try to include it in the theme building process - have the `cssbuilder` script slurp up the ol-markers file and any future ones and bake them into the theme file. But that doesn't help here.

oscarlevin:
Okay, I think I fixed everything, including a missed xsl diagcess template which wasn't getting the cdn prefix.

Added a new _rs_* function specific for portable builds.  It might make sense to refactor all these functions at some point, but this was the simplest change for now.  Diffs of html come back as expected (although as I expect, I might be missing something again).

rbeezer:
No ill effects on non-portable versions that I can see, so this can move forward.

I would have liked to split out the script stuff from the HTML stuff, but pushed prematurely.

I see a `link` to the `ol-markers.css` file still.  File may be gone, but pointer is still there, I think.

I'd love to see "single-page" HTML as a real possibility for limited documents with all this infrastructure.  I know that is not identical to "portable".

*  Worksheets, `index.html`, and "standalone" pages for `interactive`, `video`, `audio` could all go away in output.  I suspect it could be easy to kill/neuter the templates making those things?
*  Interactives (especially the more custom ones) may need some associated code, similar to external images.
*  It'd be nice to eventually restore some of the documentation about single-page HTML: do this (e.g. chunk level 0), don't do this (e.g. interactives with custom JS), etc.

rbeezer:
I'm going to pick up the first three commits here, so I can attack a couple other PRs.

rbeezer:
Ignore that!  Temporarily lost in my own repo.....

PR #2430, opened 2025-03-04
Portable HTML: svg embedding

oscarlevin:
Adds a template for embedding a latex-image- or prefigure-generated svg directly in the html.  The template is called based on the `$b-portable-html` variable (note, since #2429 is not yet merged, a commented line at 5911 of `pretext-html.xsl` allows for testing; this and the previous line should be removed prior to merging).

I am not sure if I have the namespace stuff right.  I was able to copy the `xmlns:...` "attributes" of the original svg, but had to hard code in the `xmlns="..../svg"` directly (which might be bad, I have no idea).  See lines 6158-6159.

Finally, a small tweak to the css to put a white background behind svg's like images get, needed for dark mode.

I have NOT refrained from copying the generated latex-image folder to output; that would require testing on a pubvar, and I need to think about the right way to do this with the very recent pub-var-report that got moved in.  So it's on my radar.

### DIFF
--- a/css/components/elements/_media.scss
+++ b/css/components/elements/_media.scss
@@ -22,7 +22,7 @@ img.contained {
 
 // ---------------------------------------------
 // images
-.ptx-content img {
+.ptx-content img, .ptx-content svg {
   // for body images in dark mode, we want to be able to force a light colored background
   // as most transparent images will assume that the background is white
   background: var(--ptx-image-bg);
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -3980,6 +3980,9 @@ def html(xml, pub_file, stringparams, xmlid_root, file_format, extra_xsl, out_fi
     if include_static_files:
         # extra css for custom ol markers
         move_ol_marker_css(tmp_dir)
+    if not(include_static_files):
+        # remove latex-image generated directories for portable builds
+        shutil.rmtree(os.path.join(tmp_dir, "generated", "latex-image"), ignore_errors=True)
 
     if file_format  == "html":
         # with multiple files, we need to copy a tree
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -6176,6 +6176,22 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <svg xmlns="http://www.w3.org/2000/svg">
         <xsl:copy-of select="$image-svg-xml/svg:svg/namespace::*"/>
         <xsl:copy-of select="$image-svg-xml/svg:svg/@version"/>
+        <!-- Keep the viewbox, or create one based on the height and width -->
+        <xsl:choose>
+            <xsl:when test="$image-svg-xml/svg:svg/@viewBox">
+                <xsl:attribute name="viewBox">
+                    <xsl:value-of select="$image-svg-xml/svg:svg/@viewBox"/>
+                </xsl:attribute>
+            </xsl:when>
+            <xsl:otherwise>
+                <xsl:attribute name="viewBox">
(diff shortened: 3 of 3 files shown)


oscarlevin:
Just force pushed to remove the two comment lines I left in for "testing" before the publisher variable from #2429.

rbeezer:
Pretty quickly I can see some images that do not render well, with Firefox on Debian.  Also on Chrome.

* Section 10.2 PreFigure, Two are clipped (missing bottom half), two are missing.  Maybe @davidaustinm  can help.
* Figure 10.9, messed up in odd ways.  Figure 10.10 to a lesser degree but in similar ways.
* Figure 10.11, more oddities, perhaps with same root cause as 10.9, 10.11.

No real insights, though I wonder if the images need to be sand-boxed - protected from one another.  Maybe what look like integral signs in 10.11 are reading TeX glyphs that are on the page, but part of a different image using the same names.  Like `id="font_1_1"`, which I see used multiple times.  Rearranging the images might make them behave better if they are first? Or last?

https://stackoverflow.com/questions/45937469/svg-duplicate-ids-on-one-page

If my hunch is right, then maybe our `@unique-id` could prefix an image's ids and references?  Grep/search the XSL for `"svg-edit"` in the EPUB conversion.  Sort of like an assembly pass which could "easily" make adjustments of this sort before embedding.

oscarlevin:
Thanks for the hints, and thanks @davidaustinm for troubleshooting this earlier today.  Both sets of issues are resolved in the latest two commits
1. The prefigure issue was that I removed the height and width attributes in favor of viewBox.  Now I'm testing on the existence of a viewBox, and if not present, creating one based on the height and width.  So no rush on changing anything in PreFigure; this should work whether or not a PreFigure svg has a viewBox.
2. The id's were indeed the issue.  Added a modal template that adds suffixes to id's and their references.  Uses xslt's `generate_id()` function (passed through the templates as a param), which gives each svg its own suffix.

davidaustinm:
Thanks, Oscar.  I just checked this out and the first two PreFigure diagrams in the sample article are sized properly now.  The [annotated diagrams](https://pretextbook.org/examples/sample-article/html/section-graphics.html#prefigure-diagrams-10), however, don't show up.  Those are loaded with the `diagcess` javascript library so I wonder if that library is being included.

rbeezer:
Good - glad the ids explained that.  Two quick comments, we can discuss at Drop-In?
* `generate-id()` can cause a few problems.  Why not use our `uinque-id` from the replaced `image`?
* I think your svg-editing templates can be more economical.

oscarlevin:
Rob: Fixed the templates as we discussed.  Also removed the svg's generated for latex-image so they get copied to the `output/generated` folder.  I did not remove the prefigure ones though, because it appears that the annotated diagrams need them.

David: It appears that those don't show up when you open the html file directly, but they do when you open it inside a web server.  This is true of the usual non-portable builds too.  You would need to do a `pretext view` to see them correctly.

So I think this is ready to go.  I'd love to release this and post of dev to get feedback from people before writing documentation.

davidaustinm:
Yes, you're quite right.  Sorry about that!  The PreFigure stuff is looking good now.

rbeezer:
Merged with all code as-is.   Fixed spelling on `colide` and removed a double-space in some code that seemed like a typo.  Capitalized a bunch of stuff in edited commit messages.

Real good to have some experience with embedding SVG.  Wondering just now if that could save us some headaches with EPUB manifests, etc?

Full-speed ahead on portable builds!

rbeezer:
I always forget to "exclude" result prefixes when introducing a new namespace.  It really makes a mess in the output (for literal elements, if I'm not mistaken).
See 35f45578352f93a4f1254659c75e22561bc839d4

PR #2431, opened 2025-03-04
Portable HTML: xref knowls are links and files not produced

oscarlevin:
A testament to excellent xsl programming previously: this was a super easy change.

Of course, this requires #2429, or you need to define the `$b-portable-html` variable to test.


rbeezer:
Ooh, that was easy.  ;-)

rbeezer:
And easy to merge!  Once I found my place.  Something to note for the eventual write-up on portable output: put `idx` in a `<p>` and then the index just says "Paragraph", which is not very informative at first look.  Link does work, though (good).  Could suggest placing `idx` immediately within block structures if the finer granularity is not critical.

PR #2432, opened 2025-03-06
Added MOM embedded SVG 1st try

mafitch:
Changed URL used so embedded SVG remain. Added code to move SVG into files and update <image> tags. Does not generate PDF yet. Does not handle width (images very large)


rbeezer:
OK, problem `930582` is behaving nicely.  Good work.

Images render well and scale beautifly (the whole point of SVG).  The coin and the right triangle are side-by-side.  Is that the intent or do we need to be more careful about consecutive bare images?

rbeezer:
LaTeX/PDF, of course, needs work.  But it looks like if we make a PDF and put it next to the SVG file it should be just fine.  We will want a PNG for an EPUB destined for Kindle.  So you might as well do that while you are at it.

Look at the `latex-image` conversion routine for hints.  Or perhaps the `asymptote` conversion for hints.  I'm trying to think where we begin with native SVG and convert to other things....

mafitch:
> OK, problem `930582` is behaving nicely. Good work.
> 
> Images render well and scale beautifly (the whole point of SVG). The coin and the right triangle are side-by-side. Is that the intent or do we need to be more careful about consecutive bare images?

In MOM they are side-by-side with no structure, so this is faithful rendering. Some structure could be changed in MOM that would be caried over, but often MOM visual structure does not translate well (discouraged in PreTeXt documentation)

mafitch:
> LaTeX/PDF, of course, needs work. But it looks like if we make a PDF and put it next to the SVG file it should be just fine. We will want a PNG for an EPUB destined for Kindle. So you might as well do that while you are at it.
> 
> Look at the `latex-image` conversion routine for hints. Or perhaps the `asymptote` conversion for hints. I'm trying to think where we begin with native SVG and convert to other things....

That is in a new commit that I just pushed. git isn't showing me the option for a pull request (not sure if it is needed, in same branch).

rbeezer:
Great!  Once you open a PR, the branch is good.  Just push like you did, there is no PR-initation to be done.

mafitch:
The MOM import is now LXML based instead of regular expressions. There may be some additional changes that this enables (e.g., ensuring SVG have namespace listed). Pending but waiting likely for better SVG to PDF conversion.

rbeezer:
Looks good visually.  I'm going to go post on `-dev` about conversion tools.

rbeezer:
Very good!  Nice job with the code.  More substance on `-dev`, procedural stuff here.

*  Added problem 930582 as an example on a follow-on commit.
*  Removed three lines that had `import` statements commented out.
*  One stray hunk of whitespace I had to root out.

Rebuilding website samples right now.

PR #2433, opened 2025-03-13
LaTeX: first pass at texstyle infrastructure

oscarlevin:
There is a LOT going on here, and there will be more to come.  Right now:

- Inside `journals` a folder of `texstyles` contains two different texstyle files describing the latex layout for ams journals and springer nature journals.  A third dependent style is in `dependents`, with only the documentclass changed.
- The xsl that manages all this is `xsl/latex/pretext-latex-texstyle.xsl`.  The goal is that this will replace all the recent journal-specific xsl files.  It reads in the correct texstyle file and does all the processing. 
- A very small change to the python script is merely cosmetic (name of stringparam).

The xsl file really pushed me to understand how xsl works.  There will absolutely be better ways to do some of these things I think.  The merging of a dependent and base texstyle, using techniques from assembly, was particularly fun.  Also the processing of authors: we bounce back and forth between the source pretext and texstyle file a couple times so we can "loop over" the authors and the author data in the correct order.  Fun fun fun.

My plan:
- Fix a few things that are broken (keywords and their separators)
- Implement the other journal styles as texstyles
- Redo most of it when I get feedback and see the more obvious and better way to do the xsl.

So no rush, but it's here if you are currious.

### DIFF
--- a/doc/guide/appendices/journals-table.xml
+++ b/doc/guide/appendices/journals-table.xml
@@ -7,12 +7,40 @@
       <cell>Code</cell>
     </row>
     <row bottom="minor">
+      <cell>AMS journals</cell>
+      <cell>ams</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Elsevier journals</cell>
+      <cell>elsevier</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Springer Nature journals</cell>
+      <cell>springer-nature</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Taylor Francis journals</cell>
+      <cell>taylor-francis</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Annals of Pure and Applied Logic</cell>
+      <cell>ann-pure-appl-logic</cell>
+    </row>
+    <row bottom="minor">
       <cell>Bulletin (New Series) of the American Mathematical Society</cell>
       <cell>bull-amer-math-soc</cell>
     </row>
     <row bottom="minor">
       <cell>Conformal Geometry and Dynamics</cell>
       <cell>conform-geom-dyn</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Electronic Journal of Combinatorics</cell>
+      <cell>electron-j-combin</cell>
+    </row>
+    <row bottom="minor">
+      <cell>Experimental Mathematics</cell>
+      <cell>exp-math</cell>
     </row>
     <row bottom="minor">
       <cell>Journal of Algebraic Geometry</cell>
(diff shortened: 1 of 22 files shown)


rbeezer:
> Fun fun fun.

;-)

More at a more convenient hour.

oscarlevin:
I did quite a bit of refactoring since you last looked at this.  It is in a state that would be ready to merge, but I'm not being held back at this point if it doesn't for a while.  I will keep adding new texstyle files, and with each, will likely need to refactor some anyway.

Three journals: ams-general, taylor-francise-general, and spring-natural-general.  Two dependents of ams-general.  

Should the logic to get the cls/sty files required for a build be inside core or the cli do you think?  I haven't started on that.

rbeezer:
Sounds good.  I'll checkin when I'm ready to take another look.

Can grabbing class/style files be a discrete function that the  pretext/pretext script could execute on-demand?  Could they be fetched once, placed with external files and copied into temporary working directory as needed?

Guess it depends on your vision, which is likely very different from last paragraph.

oscarlevin:
Okay, this is in reasonable shape for review now.  All 17 journals (including 4 "generic publisher" options) are set up to use texstyle files.  This makes the previous xsl for specific journals obsolete!  

The one big piece that is missing is the fetching of required files to build the resulting latex to pdf.  I plan to put that on a separate PR.

rbeezer:
Very good!  This is a great project.  I built a PDF of the sample article, using the EJC style.  Worked great.  Two observations:
* Some (not all) Asymptote images seem to be AWOL.  See Section 10.4.
* Many images/figures seem to be full-width when maybe the source indicates a skinnier width is requested.  Then to get these onto a new page where they fit leaves a lot of vertical whitespace between paragraphs and blocks just before.  It seems.  I've not investigated too carefully.

rbeezer:
OK, almost merged and realized maybe I'm not doing this right.

I did `@latex-style="ejc"`.

If I do `@latex-style="texstyle"`, what else do I do?  In the best spirit of PreTeXt, I guessed:  `@code="electron-j-combin"`.  Then I tried to read the XSL/Python and couldn't see it.

What am I missing about `@latex-style="texstyle"`?

oscarlevin:
No, you need to set the name of a journal ([Subsection 44.1.9: Journal name](https://pretextbook.org/doc/guide/html/publication-file-common.html#common-journal)).  When this is set, the `@latex-style` is ignored (internally automatically set to the right thing based on `journals.xml`).

So it will be interesting to see how the `common/journal/@name="electron-j-combin"` compares to your test with `@latex-style="ejc"`.  The latter can be removed completely after this is working.

rbeezer:
OK, sort of as expected.  Thanks for the clarification, I had not realized this was already documented.

> So it will be interesting to see how the common/journal/@name="electron-j-combin" compares to your test 

Yes!  I'll do that before merging.  To satisfy my curiousity as well.

rbeezer:
For the record - I built the "ejc" texstyle and the "electron-j-combin" journal and compared them with a diff on the resulting `*.tex` files.

@oscarlevin  and I reviewed that during Drop-In.  A couple rough edges, but nothing to prevent this going in now.

rbeezer:
Merged.  Split up the first commit to isolate the new **stylesheet** (nee template).

Great progress - keep it coming!

PR #2434, opened 2025-03-13
Keywords refactor

oscarlevin:
This is a minor refactor of the keywords xsl template.  It puts a semicolon after the last "primary" keyword, otherwise separates keywords with a comma.

However, to support the texstyle work on #2433, the comma separator is now stored in a param that the texstyle xsl template will be able to override.


oscarlevin:
Fixed mess.  Now just a single commit.

oscarlevin:
@rbeezer: I think this will be easy to merge and would be helpful with the texstyle work.

rbeezer:
Merged, as-is.

PR #2435, opened 2025-03-14
CSS: two small denver/greeley css fixes

oscarlevin:
The denver/greeley styles had a bug with non-knowled checkpoints.  This fixes that.  It also makes the highlighting of toc items look better.


rbeezer:
In.  Thanks.

PR #2436, opened 2025-03-18
Schema: program/@interactive can be "activecode"

ascholerChemeketa:
I noticed the schema currently only allows for program/@interactive to be "codelens". It should also allow "activecode".


rbeezer:
Good catch.  Thanks, merged with derivative files updated.

PR #2439, opened 2025-03-20
Program fragment element

ascholerChemeketa:
Adds the program fragment (`<pf>`) element discussed on the pretext-dev.

My latex skills are limited - I would not be surprised to hear there is a better way to insert a `<pf?` in that environment.

### DIFF
--- a/doc/guide/author/overview.xml
+++ b/doc/guide/author/overview.xml
@@ -188,7 +188,7 @@
         <idx>program</idx>
         <idx>console</idx>
 
-        <p>If you are writing about computer science, or more general scientific or engineering topics, you may wish to include sample computer programs, or command-line sessions.  A <tag>program</tag> will contain a complete computer program, or a portion of the code from a program.  Some <tag>program</tag> can execute in an online output format, while others are both editable and executable.  This behavior depends on the language used and the host employed.  A <tag>console</tag> holds a command-line session.  These elements feature monospace fonts, preservation of whitespace, and syntax highlighting.  These may also be placed in a <tag>listing</tag> to be more prominent, or as the target of a cross-reference.  See <xref ref="topic-program"/> for details.</p>
+        <p>If you are writing about computer science, or more general scientific or engineering topics, you may wish to include sample computer programs, or command-line sessions.  A <tag>program</tag> will contain a complete computer program, or a portion of the code from a program, while a <tag>pf</tag> (<q>program fragment</q>) will display a chunk of code inline with the text.  Some <tag>program</tag> can execute in an online output format, while others are both editable and executable.  This behavior depends on the language used and the host employed.    A <tag>console</tag> holds a command-line session.  These elements feature monospace fonts, preservation of whitespace, and syntax highlighting.  These may also be placed in a <tag>listing</tag> to be more prominent, or as the target of a cross-reference.  See <xref ref="topic-program"/> for details.</p>
     </section>
 
     <section xml:id="overview-worksheet">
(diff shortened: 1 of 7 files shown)


rbeezer:
Took this for a quick drive.  **Before** you added schema, etc.  ;-)

Too fast for me, I was going to suggest following on `-dev` thread:  `pdflatex` is sometimes superior to `xelatex` as it has better automatic fine-tuning of typesetting.  But `pdflatex` is stupid about Unicode.  Sure enough, it doesn't like U+E000.

I tried Python `a in ['b']` as an example.  The LaTeX escaping converted quotes to `\textsinglequote` which I guess `listings` did not like.  Is that the escaping we do for "program"?  Maybe something gentler?

And here is an idea that just came to me.  I wonder if whatever escaping **does** happen, perhaps some simple character is guaranteed to not be present?  Otherwise, use a long list of punctuation characters?  Or entities that expand (e.g. `&SIMPLECHAR;`)

Maybe `console` has clues on the right amount of massaging.  And the `text()` of the fragment should not be caught up in the general sanitization (is `c` excluded there?).

rbeezer:
Forgot.  I think I want "Hellow, World" to stay small.  More minimal than the minimal example.  Yes, not real rational.

ascholerChemeketa:
Thanks for the feedback. I was expecting some LaTeX gotchas I was not aware of.

Will take a pass at updating the escaping and delimiters.

Inclusion of hello world was an oops - I usually use it for quick testing of ideas before moving them to SA or SB.

Alex-Jordan:
Why not use regular ASCII? So for example the source
```
Here is an example in C: <pf language="c">printf("The cost is $%f.2", money);</pf>
```
could come out like this in LaTeX:

```
Here is an example in C: \programfragment{c}{Aprintf("The cost is $%f.2", money);A}
```

Something similar comes up in pg extraction (which is why the find-unused-character tool was made) when you want to quote a string for perl, but for some reason you don't want to use an actual quote. So perl lets you do like `q!string!` where the `!` can be whatever. And there, we use:

```
<xsl:with-param name="charset" select="concat($apos, '&quot;:;,./?|`~!@#$%^&amp;*-_=+', &SIMPLECHAR;)"/>
```

(`&SIMPLECHAR;` is all alphanumeric characters except underscore.) For `pf`, you could remove a few things from that which might be problematic for LaTeX, and it seems like it would work out. Except if your program fragment actually uses all of those characters...

ascholerChemeketa:
This likely gets template gets called a lot. And the `find-unused-character` does a brute force linear search of the text one option at a time. That has the potential to get expensive if it keeps getting hits on the characters. So my goal was to maximize the chance that the first character we try as a delimiter is highly unlikley to match anything. But I can go back to ASCII and just start with characters that seem unlikely to appear in code before resorting to alphabet.

I took the sanitization from `<c>`. `<console>` uses the same template.

Via experimentation, it appears that some of the "10 wicked characters" do need escaping in the `\lstlisting`, but we don't want to do the "extras" that are in `escape-text-to-latex`. Should I make a simplified version that just does the necessary escaping?

ascholerChemeketa:
Rebased and force pushed.

Updated the special character processing. There are now multiple similar latex special character filters that likely could be refactored into something that has less repetition. But I am definitely not qualified to do that.

xelatex works fine. pdflatex chokes with `! Improper alphabetic constant.` But, it also chokes with same error on a completely vanilla use of lstinline like `‘\lstinline!import!`.  If I manually delete the big command that looks like:
```
\lstset{literate={ }{{~}}{1}{¡}{{\textexclamdown }}{1}{¢}{{\textcent }}{1}{£}{{\textsterling }}{1}{¤}...
```
pdflatex is happy. So it seems like something in that `\lstset` is causing issues with standard functioning of the `\listings` package in pdflatex.

rbeezer:
I couldn't duplicate the `\lstinline` failure.  So I made an issue where we might track any additional reports.

https://github.com/PreTeXtBook/pretext/issues/2444

PR #2440, opened 2025-03-20
Rename MOM image files and width tag for SVG

mafitch:
This implements the requested re-naming of downloaded graphics files (PNG).
Also implements an <image width=""> for embedded SVG. Like the width tag for other images it makes a wild assumption about how big a 100% should be and calculates a percent.
The author can control how big by adjusting the SVG width tag in the embedded SVG.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -2676,7 +2676,7 @@ def mom_static_problems(xml_source, pub_file, stringparams, xmlid_root, dest_dir
 
                     # Try to read the SVG width and set PreTeXt width based on document width assumptions
                     temp = svg_element.get('width')
-                    svg_width = re.search(r"([0-9]+)([a-zA-Z]*)",temp)
+                    svg_width = re.search(r"([0-9\.]+)([a-zA-Z]*)",temp)
                     if svg_width:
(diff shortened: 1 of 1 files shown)


mafitch:
This is as complete as it can be. All images are downloaded and names are predictable. All SVG are converted using Playwright. Widths are set reasonably.
Testing was completed on a small test document, sample-article, and MiTaL (for a large test). No errors are encountered.

rbeezer:
Finally!  Merged, pretty much as-is.  Sorry for the long wait.  Some notes.

*  I like lots of small commits.  Especially since it is easier to combine than to split.  Which I did here.  All one commit - too hard to selectively edit commented-out code that eventually gets replaced, etc.
*  Removed one line of commented-out code and one blank line with lots of whitespace.
*  Did not understand some stuff with file extensions, maybe we can walk through that at Drop-In sometime.
*  Weird "Download CSV" link in the PDF of the sample article (Checkpoint 33.2).  We should do something sensible with that.
*  Updated all the bits in the sample article - I think I got that right.  Once we have seeds working, that won't be so necessary.

Thanks very much for your work on this one.

PR #2442, opened 2025-03-21
French translations

JeanSebastienTurcotte:
Removed an unecessary hypen in subsubsection translation.


rbeezer:
Got it.  Merci beaucoup!

PR #2446, opened 2025-03-25
Update pretext.py

geoff-cox:
Uses sys.executable to launch http.server for better cross-platform support.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -1124,6 +1124,7 @@ def dynamic_substitutions(xml_source, pub_file, stringparams, xmlid_root, dest_d
     # Spawn a new process running a local html.server
     import subprocess
     import random
+    import sys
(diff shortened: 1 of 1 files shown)


geoff-cox:
When compiling PreText in Windows WSL, I got errors when processing the dynamic substitutions. Looking at the code, I saw that it references "python" to initiate the local server. Since my system uses "python3," I was getting an error. This PR addresses this by using `sys` to detect the Python alias.

rbeezer:
Thanks very much, @geoff-cox!  Sorry for the delay, this came in just after I'd started some extended travel.

This is a great improvement.  And I wonder why we don't go all the way and just use the module itself?  There is a more complicated instance where we use `http.server.SimpleHTTPRequestHandler` to get preview images (in a threaded fashion).

Pinging @dbrianwalton in case there is interest in not relying on the `subprocess` module.

PR #2447, opened 2025-03-25
Generate correct runestone-id for fillin exercises in manifest

ascholerChemeketa:
The new fillins do not currently generate the decorated unique id (including book title) needed within the RS manifest. This fixes that issue.


rbeezer:
Good catch.  Thanks.

PR #2449, opened 2025-03-31
CSS: fix inline headings for tasks in Denver theme

oscarlevin:
When fixing checkpoint exercises so they do NOT have inline headings, I broke tasks.  This fixes the tasks, and should not break the previous change.


rbeezer:
Thanks, now merged in.

PR #2450, opened 2025-04-04
change DoenetML example to be unescaped DoenetML inside a CDATA

dqnykamp:
In the sample article, the DoenetML example is now inside a CDATA so that it is easier to read without all the escaping of the tags.


rbeezer:
Back in the saddle, and this is merged now.  Thanks for cleaning this up!

Would you like to "fix" the version in the sample book?  Which is there to test its use on Runestone.

PR #2453, opened 2025-04-08
Script: add assembly_internal that returns assembled source element tree

oscarlevin:
To let the CLI have access to the (version) assembled source, this adds a new function `assembly_internal` that *returns* the `__XSLTResultTree` element tree produced by xsltproc running on the `pretext-enhanced-source.xsl` template.  

To accomplish this, the `xsltproc` function now also returns this result tree, in addition to writing it as usual (when appropriate).  

Other internal functions could use this, in particular, the publisher variables functions will no longer need to write the result to a file before inspecting the variables (that can be a later PR).

Post of dev: https://groups.google.com/g/pretext-dev/c/134YBSzz-5w

This is what I need to get the Runestone version-produced stuff working for @bnmnetp.

### DIFF
--- a/pretext/pretext.py
+++ b/pretext/pretext.py
@@ -4162,6 +4162,26 @@ def assembly(xml, pub_file, stringparams, out_file, dest_dir, method):
     )
     xsltproc(extraction_xslt, xml, derivedname, None, stringparams)
 
+def assembly_internal(xml, pub_file, stringparams, method):
+    """A version of the assembly function above that just returns the result tree instead of writing it to a file"""
+    # to ensure provided stringparams aren't mutated unintentionally
+    stringparams = stringparams.copy()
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged, with no real changes.  Two follow-on commits under my name:
* Expanded docstrings, code comments, so I can remember what's what.
* My fault - `os.path.join` should make the path-separators, so I fixed those for assembly.  More to do in a few other places, which I might attack later.

Lots of duplication, which I could not really see how to avoid cleanly.  Code comments suggest keeong changes in-sync.

Only tested that `xsltproc()` works as expected.  Further PRs can fix-up anything you discover in use.

Thanks for moving to this.  Only dawned on me the past few weeks that some of our features mean we have authors writing pre-PreTeXt.

PR #2455, opened 2025-04-09
Script: prevent carriage returns from breaking trace generation

ascholerChemeketa:
codelens trace file depends on stripping out newlines in code. If code is authored in Windows, it will have `\r\n` instead of `\n` and the carriage returns will not be stripped. This breaks `pretext.py`'s `tracer` function which thinks the `\r`s are line breaks.

I encountered the issue with code that was xincluded from a text file authored in Windows.
```
  <program label="structs_operators-on-structures-1" interactive="codelens">
<xi:include href="../../programs/structs/structs-operators-on-structures-1.cpp" parse="text"/>
  </program>
```

If I copy/paste the code into the program, everything works as expected (even when authored in Windows). So there is something about the include mechanism that is affecting newline treatment.

This PR strips any carriage returns during trace generation to make sure they don't break. 

Tested in Windows and Linux using `\r\n` and `\n` line endings.

### DIFF
--- a/xsl/extract-trace.xsl
+++ b/xsl/extract-trace.xsl
@@ -58,7 +58,10 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
             <xsl:with-param name="text" select="code" />
         </xsl:call-template>
     </xsl:variable>
-    <xsl:value-of select="str:replace($code-with-newlines, '&#xa;', '\n')"/>
+    <xsl:variable name="removed-carriage-returns">
+        <xsl:value-of select="str:replace($code-with-newlines, '&#xd;', '')" />
+    </xsl:variable>
+    <xsl:value-of select="str:replace($removed-carriage-returns, '&#xa;', '\n')"/>
     <xsl:text>&#xa;</xsl:text>
 </xsl:template>
 


rbeezer:
Windows!  I've received PRs where the entire file affected has its line endings change into the `\r\n` versions!  Always mystifies me for a minute or two.

> there is something about the include mechanism

Well, I wish we had a better understanding of what that was.  I suspect it is just the `parse="text"` case, which is saying to treat the file contents as **characters**, not content with embedded XML elements.  OK, maybe this is it: https://www.w3.org/TR/xml/#sec-line-ends  I guess the `parse="text"` makes the file contents an unparsed entity.

Independent of any of that, one question.  Would it be safer to replace the CR-LF **pair** by a LF and then make the replacement to `\n`?  I wonder if there is an edge-case where a program might use `xD` in some other way?  (Perhaps the same argument applies to `xA` but maybe this is not a concern in either case.)

I'll likely add a follow-on commit to explain the line-ending stuff at link above into the code here and perhaps in The Guide.  Seems like files created on Windows could create surprises lots of other places.

ascholerChemeketa:
Yes, I agree that restricting the fix to CR-LF pairs makes sense. Assume you are just modifying the PR. Alert me if that is not the case.

I thought about a more general text utility, but this is the only place I have ever noticed Windows line ending issues. I usually author in Windows and build in Linux (WSL), so I am kind of a torture test for potential problems with them.

rbeezer:
Thanks for the reply.  Merged as-is, then added a follow-on commit which works on the pairs.  Didn't break anything on Linux (duh!), but maybe a quick test on Windows would be in order?

Also put a bit of a warning into the Guide (which will get updated eventually).

PR #2456, opened 2025-04-09
HTML: provide hints div for fillin exercises

ascholerChemeketa:
Fillin exercises do not generate the solutions-div common to other exercises in HTML builds. This PR adds that generation.


rbeezer:
OK, this works as one would expect.  But I'm not sure it is what we want.  Pinging @dbrianwalton who will know better.

The `answer` element does not exist in these problems, I think.  There is a `solution` element in the markup.  But the knowl containing it does not always render properly.  For example, i see (in 5.18.10) of the sample book:
```
<span class="process-math">\([%= toTeX(theFunction) %]=0\text{.}\)</span> 
```
A `hint` element could be appropriate.

I **think** the `solution` is meant to be under the control of the JS for the problem.  At some point it is shown to the reader?  But that does not really seem to be the case now.

@dbrianwalton: how is the `solution` element meant to behave?  The usual PreTeXt way or under control of the FITB UI?

ascholerChemeketa:
That sounds good for solutions/answers. I'm happy to rip them out of this.

Hints were my major motivator. What do you think about updating this PR to just handle those? I think hints having the same presentation across all exercise types is beneficial.

ascholerChemeketa:
Working on suggested path. Hit an issue with a new sample that includes a hint:
https://github.com/PreTeXtBook/pretext/issues/2460

Once I know the expected behavior, I will author to that spec and push the commit.

ascholerChemeketa:
Force pushed to just add hints and not solution/answer.

Third commit in the latest version removes the old var based fillin samples. Currently it is not super clear that an author should not be using those. Alternatively, they could be more clearly marked, but I can't think of a good reason to leave them in place. But feel free to drop that commit if you want to leave them in for a while longer.

rbeezer:
Great - thanks.

The "old style" FITB are supported, meaning they should run.  But they should be discouraged, and they won't see new work or features.  So they should stick around for quite a while, just to have a test that they are working.

ascholerChemeketa:
Should we add an "Old syntax, do not use" to each of the old ones then? Currently it is not very clear that one should not try to emulate those.

rbeezer:
Right.  I did this for the sample article, so I thought I had all my bases covered.  Clearly not.  See

[Exercises 16.7: Old-Style Fillin-In](https://pretextbook.org/examples/sample-article/html/section-interactive-exercises.html#section-interactive-exercises-9) 
https://pretextbook.org/examples/sample-article/html/section-interactive-exercises.html#section-interactive-exercises-9

bnmnetp:
I think it is pretty confusing that the sample article without an ability to see the source says:
"Do not use this as a model for new exercises. Just for backwards-compatibility."

If I am new to our ecosystem, It says to me, don't make fill in the blank questions.   Maybe we should add another example that shows the right thing to do.

rbeezer:
> without an ability to see the source 

Version with visible source

[Exercises 16.7: Old-Style Fillin-In](https://pretextbook.org/examples/sample-article/annotated/section-interactive-exercises.html#section-interactive-exercises-9) 
https://pretextbook.org/examples/sample-article/annotated/section-interactive-exercises.html#section-interactive-exercises-9

> Maybe we should add another example that shows the right thing to do.

This entire section comes first.

[Section 17: Dynamic Exercises](https://pretextbook.org/examples/sample-article/html/section-dynamic-exercises.html) 
https://pretextbook.org/examples/sample-article/html/section-dynamic-exercises.html

> I think it is pretty confusing

Additional verbiage welcome.

PR #2457, opened 2025-04-09
Add support for iframe <html> tags to dynamically respond to dark mode toggle

geoff-cox:
This PR enables dynamic dark mode support inside iframes by modifying the `pretext-html.xsl` output template to add a `class=""` attribute to the `<html>` element in `iframe` HTML files.

This is especially useful for embedded JavaScript-based interactives that want to mirror the parent document’s dark mode toggle via JavaScript.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -9791,6 +9791,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:call-template name="converter-blurb-html-no-date"/>
         <html>
             <xsl:call-template name="language-attributes"/>
+            <xsl:attribute name="class"/>
             <head>
                 <!-- grab the contents every iframe gets -->
                 <xsl:copy-of select="$file-wrap-iframe-head-cache"/>


rbeezer:
Thanks - sounds like a good idea.  Two comments:
* Pinging @ascholerChemeketa for any comments he might have.
* What do you think of putting an example into the sample article?  It could be as simple as a panel that has colors that change with a move to dark mode.  It'd be a helpful guide for other authors of similar JS interactives that want to have this feature.

ascholerChemeketa:
What is the purpose of creating the empty class attribute? The existing `classList.add()` call in the JS will create a `class` attribute if it does not already exist.

The Sample Article has an iframe with a working demo:
[Figure 14.19](https://pretextbook.org/examples/sample-article/html/section-interactive-authored.html#figure-interactive-numerical-integral) 
https://pretextbook.org/examples/sample-article/html/section-interactive-authored.html#figure-interactive-numerical-integral

As noted in the paragraph above that widget, dark mode for iframes is currently opt in. If you are trying to get a iframe, make sure to add `@dark-mode-enabled="yes"` to the interactive that is generating it:
https://github.com/PreTeXtBook/pretext/blob/7936f3fd1595439fe86adc44ea9a4f8045372801/examples/sample-article/sample-article.xml#L8613

rbeezer:
Thanks, @ascholerChemeketa.

> If you are trying to get a iframe, make sure to add @dark-mode-enabled="yes" to the interactive that is generating it:

I'd forgotten about that!

geoff-cox:
Wow, I didn't realize this was an option, and it accomplishes exactly what I was trying to achieve with this PR. Thanks for letting me know.

geoff-cox:
Closing ....

PR #2459, opened 2025-04-10
WeBWorK: LaTeX radio buttons and checkbox styling

Alex-Jordan:
One small thing peeled off from #2336.

Would you want the checkbox changes here excluded? They are moot until you are using a 2.19 server.


rbeezer:
Very good.  Merged with checkbox changes to patrol later as 2.19 happens.

PR #2461, opened 2025-04-10
WeBWorK: tabular/row/cell bottom rule corrections

Alex-Jordan:
One small thing peeled off from https://github.com/PreTeXtBook/pretext/pull/2336.

In the sample chapter HTML output, you should see an improvement in Checkpoint 1.7.1. The horizontal rule there before this commit only applies to the first cell. But extends all the way (as it should) after this commit.


rbeezer:
Excellent.  Merged as-is.  Thanks.

PR #2462, opened 2025-04-10
WeBWorK: make problem set archive directly from source, not from repr…

Alex-Jordan:
…esentations file

One medium thing peeled off from https://github.com/PreTeXtBook/pretext/pull/2336.

Previously you build the representations file first, and then to make the archive of problem files and set definition files, it's mostly scraped from the representations file. This changes things so that the archive of all that stuff is made directly from source.

### DIFF
--- a/examples/webwork/Makefile
+++ b/examples/webwork/Makefile
@@ -189,23 +189,19 @@ mini-html:
 # by chunk level) with set definition files, set header files, and problem
 # files in a file tree. Separate problem sets for inline and divisional
 # problems. This does not require having the representations file.
-# We turn off dates in file headers, so when we use these for testing we
-# can get more accurate diffs. Dates are useful for single file output
-# (or at least less of an annoyance).  This should not be taken as necessary
-# or best practice for a real PreTeXt project using WeBWorK problems.
 # Adding a "-z" switch will create a "tgz" compressed archive
 
 sample-chapter-problem-sets:
 	install -d $(PGOUT)
 	cd $(PGOUT); \
 	rm -r Integrating_WeBWorK_into_Textbooks; \
-	$(PTX)/pretext/pretext -v -c all -f webwork-sets -x debug.datedfiles no -p $(SMPCPUB) $(SMPC)
+	$(PTX)/pretext/pretext -v -c all -f webwork-sets -p $(SMPCPUB) $(SMPC)
 
 mini-problem-sets:
 	install -d $(PGOUT)
 	cd $(PGOUT); \
 	rm -r WeBWorK_Minimal_Example; \
-	$(PTX)/pretext/pretext -v -c all -f webwork-sets -x debug.datedfiles no -p $(MINIPUB) $(MINI)
+	$(PTX)/pretext/pretext -v -c all -f webwork-sets -p $(MINIPUB) $(MINI)
 
 
 #########################################################################
--- a/xsl/pretext-ww-problem-sets.xsl
+++ b/xsl/pretext-ww-problem-sets.xsl
@@ -42,6 +42,7 @@
 <xsl:import href="./publisher-variables.xsl"/>
 <xsl:import href="./pretext-assembly.xsl"/>
 <xsl:import href="./pretext-common.xsl"/>
+<xsl:import href="./extract-pg.xsl"/>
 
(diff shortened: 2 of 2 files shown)


rbeezer:
This could be a good idea.  But now is not the time, and there is too much going on here for me to consider it in concert with the 2.19 move.

Please leave the generation of the problem sets exactly as they are.  Do not change the stylesheet, do not change the Makefile, do not change the Python.  If the PG itself changes, I'll see it, but it can't be mixed up with other changes.

Now, you want problem source elsewhere.  Fine.  There is a Python function to do that and it will allow you to place them wherever you want them or need them.  If not, then we can discuss whatever needs modification.

Alex-Jordan:
In 2.19, you can process PG locally, good. How does that work? Well, you have to already have the exercise's .pg file sitting there, to apply local PG to. This is unlike using a webwork2 server where the problem file code will be sent as part of a form. So for local PG processing, we have to have already built the .pg files before we build representations. That is why this is necessary and part of the Tacoma PR in the first palce.

I guess you could ask me to do only webwork2 processing stuff with 2.19 first, and bring in local PG processing later. But the code in pretext.py is full of if/then/else statements and other complicated logic where things work depending on the cases of using webwork2 processing, local PG processing, or the renderer processing. And I can't bring myself to smash all that only to have to reconstruct it a few weeks later.

I was under the impression this gave zero diff in output. Did you find that not to be the case?

Alex-Jordan:
A glimpse at the code just now. Because 2.19 needs these .pg files established anyway for local PG processing, the code for webwork2 server processing has actually also been changed to also assume those .pg files already exist and get used. So even to make 2.19 support for only webwork2 server processing, you are asking for a big rewrite. I'm not sure you are aware of how large that rewrite would be and that I just can't get to something like that. Or that this PR is a predecessor to the big 2.19 stuff, not a potential follower. If the Tacoma PR did not exist, is there reason not to consider this one? I did introduce the `debug.datedfiles no` here, so the diff on PG files will show that. Module that I didn't get a diff when I tested this before.

Alex-Jordan:
This will be rebased and conflicts resolved once #2492 through #2499 are processed.

Alex-Jordan:
This is now updated and built off of #2505. If #2505 is processed, this can be rebased off master and the commit here should only show changes in two files.

When I build the problem set archive in #2505 versus here, I find no file diff in the folder tree with all the PG files, etc.

The comments in the Makefile say that you do not need the representations file to make the archive. That was actually false. Before this change, you *do* need the representations file. But now with this change you don't and you can build the problem set archive very fast.

As I've noted elsewhere, this change facilitates local PG processing. It is helpful to have the archive of PG files in a a manged directory, so that local PG can go get that file to render it for making a static version. Once this PR is processed I can return to the big Tacoma one. Big topics remaining there are (1) general PG 2.19 support (2) local PG processing (3) supporting a local .pg file. I may be able to separate them by just rebasing, cleaning up, and making branches where I selectively delete things.

Alex-Jordan:
Rebased off #2505 (which was recently rebased) and a conflict is now resolved.

Alex-Jordan:
OK, this is updated and could be reviewed.

On `master`, one must run `make sample-chapter-representations` and then run `make sample-chapter-problem-sets` to get the folder tree of PG files. On this branch, one only needs to run `make sample-chapter-problem-sets`. (You can clear away the representations file to know for sure it is not relied on.)

This change means there is a fast way to create the .pg files without simultaneously contacting something (potentially a webwork2 server) to render them into static versions. And then these .pg files can optionally be read by local PG for processing following some upcoming PR that whittles away at the Tacoma PR.

Alex-Jordan:
Just a note that the only diffs I find in the folder tree of the pg files are date stamps.

rbeezer:
Passes with flying colors on the first try.  ;-)  Thank-you.  I needed an easy one.

I forgot to make trivial edits.  I'm trying to slowly get rid of spaces at the end of empty tags (`<foo />`).  But **only** when editing anyway.  In other words, don't mix a bunch of whitespace with a bunch of real stuff.  **OR** do a whole file automatically and we will register it as a whitespace commit after merging.  See https://pretextbook.org/doc/guide/html/code-style-whitespace.html

Carry on!

PR #2463, opened 2025-04-10
WeBWorK: remove use and mentions of source/@webwork-problems

Alex-Jordan:
One small thing peeled off from https://github.com/PreTeXtBook/pretext/pull/2336.

We don't use the `webwork-problems` attribute on `source` in a publisher file anymore. Haven't for a while. In `publisher-variables.xsl`, it is recognized and identified as deprecated.

### DIFF
--- a/doc/guide/appendices/cli-transition.xml
+++ b/doc/guide/appendices/cli-transition.xml
@@ -114,7 +114,7 @@
       and the one created by the <c>pretext init</c> command.
       The file created by <c>pretext init</c> will contain some directory management details, such as:
       <cd>
-        <cline>&lt;source webwork-problems="../generated-assets/webwork/webwork-representations.xml"&gt;</cline>
+        <cline>&lt;source&gt;</cline>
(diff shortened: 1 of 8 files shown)


rbeezer:
Merged as-is.  Local rebase of #2336 continues to improve.

rbeezer:
Appears to me that the `*-not-managed` stanzas of the `Makefile` could be pruned, and the two publication files are now empty and could be removed.  I don't think we need any of this for testing anymore - I certainly have not used them in a long time.

I can do the cleanup, just double-checking here before I crank up the chain saw.

Alex-Jordan:
Thanks, and I fully support requiring managed directories for WeBWorK
stuff. Chainsaw away.

I'll rebase this week sometime and address more things you mentioned
recently.

rbeezer:
> Chainsaw away.

Done at 8e9ac2be39a8c32d1802b64bc9cbfe56ed6d4f79

PR #2465, opened 2025-04-11
Fix dark/light mode syncing for interactive iframes

geoff-cox:
This PR attempts to fix an issue where interactive iframes do not correctly toggle between light and dark mode after the first switch. The problem appears to be that the loop propogating the mode only scans through the iFrames when dark mode is true and does not update them when dark mode is turned off.

### DIFF
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -978,17 +978,25 @@ function setDarkMode(isDark) {
     if(document.documentElement.dataset.darkmode === 'disabled')
         return;
 
-    if (isDark) {
-      document.documentElement.classList.add("dark-mode");
+    const parentHtml = document.documentElement;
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, @geoff-cox, for sticking with this one.  I can certainly apply this and test it, but I'd like some input from @ascholerChemeketa as well about the approach.

ascholerChemeketa:
That would be good to fix.

I think it would be better to implement using `element.classList.add(class)` or `element.classList.remove(class)` instead of setting `element.className`. Manually setting `element.className` wipes out any other classes that were authored or that other code might have set.

geoff-cox:
@ascholerChemeketa 
Thanks for the feedback. I updated the code to use e`lement.classList.add` and `.remove` instead of `element.className`.

rbeezer:
Thanks!  Merged as-is.  Slight edits to the commit message.

PR #2466, opened 2025-04-11
CSS: add minimalist 'boulder' theme for standalone pages

oscarlevin:
This adds a new "colorado" theme; not too different from Greeley, except that it hides the masthead and any deactivated buttons.  Intended for use in a chunk level=0 and portable=yes document.

Changes to greeley theme to keep the color setup consistent between the three colorado themes.

Note that I did not try to move anything around after all; there is a title on the first page anyway, and keeping the navbar made sense in case someone wanted to use search.

### DIFF
--- a/css/targets/html/boulder/_customization.scss
+++ b/css/targets/html/boulder/_customization.scss
@@ -1,0 +1,33 @@
+//$color: var(--primary-color) !default;
+//$text-color: white !default;
+
+.hide-type .codenumber:not(:empty)::after {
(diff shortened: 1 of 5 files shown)


rbeezer:
Merged, where, I split out the greely changes.  Thanks.

Does `boulder` need to go into some list in the Guide?  Explaining its purpose as useful for a "standalone" page?  And maybe a suggestion  to use it, where we talk about standalone pages?

PR #2467, opened 2025-04-11
Texstyles: two quick fixes

oscarlevin:
The two commits (which could totally have been one commit) fix the issue with some sections not being included and with the snapshot package not being loaded when requested.

### DIFF
--- a/xsl/latex/pretext-latex-texstyle.xsl
+++ b/xsl/latex/pretext-latex-texstyle.xsl
@@ -120,6 +120,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <xsl:template match="article">
     <!-- Some boiler plate at the top of the file -->
     <xsl:call-template name="converter-blurb-latex"/>
+    <xsl:call-template name="snapshot-package-info"/>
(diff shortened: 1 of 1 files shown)


rbeezer:
Good - better diff with `@latex-satyle="ejc"` and a better diff with `journal/@name="electron-j-combin"` compared to pre-PR.  Merged as-is.

But it seems that the `backmatter/references` is still going missing?  Maybe the `apply-templates` in-question needs a `self::references{not(parent::backmatter)]` or similar?

rbeezer:
Afterthought.  With no real survey of the code.

Would it be cleaner to apply-templates to every division, and have unsupported divisions just do nothing and report that they were being ignored?  (Rather than complicated logic in the match?)

oscarlevin:
There are two things going on here I think.  The `pretext-latex-classic` is doing some exclusion based on what I was and was not able to get working on the first pass at that.  Then, the `pretext-latex-texstyle` is trying to separate "mainmatter" from "backmatter".  Looking at the schema, an article can have lots of children.  If I knew for sure that an article would be structured with sections and backmatter, this would be easier.  

But all that is to say, I will look into this some more today.

PR #2468, opened 2025-04-14
Texstyles: add backmatter references to bibliographies

oscarlevin:
This adds references that are inside `backmatter`.  Previously, only references that were inside an appendix or child of article were included.

I'm sure that bibliographies need a lot more work for texstyle files, since they are not getting any journal-specific styling.  But I'm going to wait for that until we can tackle bibliographies proper anyway.


rbeezer:
Merged.

Worked on **real** bibliographies today.  But not much to report.

PR #2469, opened 2025-04-15
CSS: fix wide rs elements overflowing containers in colorado themes

oscarlevin:
This fixes a bug with the Colorado themes that let rs components that are wide and get extra space spill over the sides of their containing elements.


rbeezer:
Merged, with two whitespace fixes.  ;-)

PR #2470, opened 2025-04-15
Texstyles: fetch and place required files

oscarlevin:
This PR contains three commits.  

First, I refactored the texstyle xsl and python so that the stringparam points directly to the texstyle file instead of just providing a journal name.  I think this is simpler, and we need the texstyle filename for getting extra files anyway.

Commit 2 is just the python.  Added a function that will copy the correct extra files to the tmp build directory, downloading them first if they are missing.   This did introduce a little duplication (the pdf function calls the latex function which gets the journal info, but the pdf function also needs to get the journal info separately), but I don't think this is too time intensive so should be fine.

Finally, the python functions clarified what info we actually need to get the extra files, so I fixed up the texstyle files with the that data formatted correctly.

Interestingly, having these extra files and trying to compile the latex revealed some issues I hadn't thought of (like some environments being already defined).  I'll fix those up on the next pass.

### DIFF
--- a/journals/texstyles/ams.xml
+++ b/journals/texstyles/ams.xml
@@ -5,7 +5,6 @@
         <code>ams</code>
         <name>AMS journals</name>
         <publisher>American Mathematical Society</publisher>
-        <!-- Each file element of required files should have an href to a location where wget can download the file.  @compression describes how the file is compressed (or None).  @path gives the path to the file relative to the extracted resource.  The file at this path will be copied to the build directory before running xelatex/pdflatex -->
         <latex-engine command="pdflatex"/>
         <bibliography-style>spbasic</bibliography-style>
 
--- a/journals/texstyles/dependents/proc-amer-math-soc.xml
+++ b/journals/texstyles/dependents/proc-amer-math-soc.xml
@@ -7,7 +7,7 @@
         <publisher>American Mathematical Society</publisher>
         <extends>ams</extends>
         <required-files>
-            <file href="https://www.ams.org/arc/journals/packages/proc/proc_amslatex/proc-l.cls" compression="none" path="proc-l.cls"/>
+            <file href="https://www.ams.org/arc/journals/packages/proc/proc_amslatex/proc-l.cls" name="proc-l.cls"/>
         </required-files>
         <style-author>Oscar Levin</style-author>
         <date>2025-04-01</date>
--- a/journals/texstyles/dependents/quart-appl-math.xml
+++ b/journals/texstyles/dependents/quart-appl-math.xml
@@ -7,7 +7,7 @@
         <publisher>American Mathematical Society</publisher>
         <extends>ams</extends>
         <required-files>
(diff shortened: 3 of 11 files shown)


rbeezer:
Looks good and works as advertised (modulo environments).  But I'd like to ask for a bit of refactoring and some clean-up.

I'd like `pretext/pretext` to have a feature where you provide the source and publisher file (as usual) and the class file lands in `generated/journal/<journal-name>` as it does now.  But you could have it land in some other directory with the `-d` switch.  (I don't know why you would do this, but I'm suggesting the same behavior as other features.)  This would mean a top-level function in `pretext.py` that does only this.

Then one could go
``` bash
pretext/pretext -d . -c latex-style -p pub_file.xml mybook.xml
```

There is one place where you can say generated stuff goes into default directories for all possible generations of different stuff.  You are setting `generated/journals` when you set `cache_dir` (sort of ad-hoc location for that definition).  Then if you remove `-d .` from above the default landing spot can automatically be the "normal" place.

Now, when you actually build, you can copy or download+save+copy to the right place in the tmp/build location.

Suggestions:
*  `place_required_build_files()` is LaTeX-specific.  Please use names that reflect that.
* Two bits of stray whitespace in each of the first two commits.
* `#copy_required....` stray comment line
* `jounral`

I plan to be at Friday Drop-In if you'd like to walk through some of this.  Love where this is going...

Rob

oscarlevin:
Done.   I think.  I changed the location in generated from "journals" to "latex-packages".  Naming is hard.

Added the component to the `pretext/pretext` script.  

Cleaned up my sloppyness (the extension that was highlighting whitespace got turned off somehow, but I fixed that).

Updated the environments, but now have a hyperref problem that will need to be fixed.  Later.

rbeezer:
Thanks.  Very close.

* `pretext/pretext -d . -f latex-package`  e-jc.sty in default directory (good)
* `pretext/pretext -f latex-package`  e-jc.sty in `generated/latex-packages/` (no directory w/ journal name?)
* `pretext/pretext -f pdf`  e-jc.sty in `generated/latex-packages/electron-j-combin/` (good)

oscarlevin:
That is definitely an oversight, but now that I think about it, I don't understand how to fix it.  Would it be sufficient for the first pattern to always put the file in the journal-name folder relative to the requested directory?

oscarlevin:
I went ahead and made that minor change.  So now `pretext/pretext -d . -c latex-package...` will put `e-jc.sty` in a folder inside the current directory.  Like you said, I don't know why you would want to use the `-d` option here, but if you did, I don't think it is unreasonable to have the required files show up together inside a folder.

rbeezer:
In.  Thanks.  A bit of light editing, no code though.

Are there some old `@latex-style` things to get rid of now?

oscarlevin:
Yes, I think we can get rid of all the older journal xsl styles.

PR #2471, opened 2025-04-18
Implement embed button

oscarlevin:
This adds an optional (but default visible) "Embed this page" button to the navigation bar.  Can be turned off with a publication variable.  The embedded page includes only the `ptx-content`, making it appropriate for Canvas or another LMS.

The guide could use some documentation about this and related features for building "small" documents, or collections of them.  Not sure where to put that, but don't want that decision to hold this feature back.

### DIFF
--- a/css/components/page-parts/_navbar.scss
+++ b/css/components/page-parts/_navbar.scss
@@ -72,6 +72,25 @@ $center-content: true !default;
     height: 34px;
   }
 
+  .embed-popup {
+    padding: 10px;
+    background-color: var(--navbar-background);
+    position: absolute;
+    margin-top: $nav-height;
+    border: 1px solid var(--page-border-color);
+    border-radius: 3px;
+    box-shadow: 0 0 5px var(--primary-color-gray-80);
+    .embed-code-container {
+      width: "100%";
+      padding:5px;
+      .embed-code-textbox {
+        width: 100%;
+        padding: 10px;
+        height: 5em;
+      }
+    }
+  }
+
   .treebuttons > * {
     display: flex;
     align-items: center;
@@ -87,7 +106,7 @@ $center-content: true !default;
     display: none;
   }
 
-  :is(.runestone-profile, .activecode-toggle, .searchbutton, .calculator-toggle, .light-dark-button) .name {
+  :is(.runestone-profile, .activecode-toggle, .searchbutton, .calculator-toggle, .light-dark-button, .embed-button) .name {
     display: none;
   }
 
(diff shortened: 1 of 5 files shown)


rbeezer:
Do we care:  zoom in until you get an on-mobile look (buttons at the bottom).  Now click on embed-button.  I think the dialog is **below** the button and now not really visible out-of-fram, except maybe its top border.

Continuing to review...

bnmnetp:
What is interesting about this is that I think Canvas supports something like the SPLICE protocol, so that if interactve questions were on an embedded page we *might* *maybe* be able to report the correctness for a question back to Canvas.  Someone would need to do some more investigation to see if my vague memory is correct.

oscarlevin:
@rbeezer, yes, we do care, and I had noticed that last night and promptly forgot about it.  I'll work on fixing the css to move it above the button in that case.

oscarlevin:
Force pushed just a change to the CSS commit (and reordered them).  Mobile view works now.

oscarlevin:
@bnmnetp apparently there is an API that Canvas uses. No idea whether that would hook into some sort of assignment, but if so, wow would that be a game-changer.

rbeezer:
Dinner's done.  Back to it.  Thanks for the mobile fix.

Do we want to intertnationalize the `title` of the button?  ("Embed this page")  Maybe simplify it a bit to be easier/shorter to translate?  "Embed Page"?

Maybe we don't do this for other buttons?  Time to start?  Or do we let icons do the speaking?  (This one is a bit less-common.)

Ready to go, other than this question.

rbeezer:
> The guide could use some documentation

Not in the Author's part.  This is a publisher activity.  The Publisher part is sort of setup by conversion type.  "Small Documents" could be a chapter with an overview quality that pointed to other places.  The embedding is definitely an HTML publisher activity.

PS: don't forget that letters and memos are small documents we already have.  LaTeX-only.

oscarlevin:
I'm going to say this is ready to go.  See relevant discussion on -dev: https://groups.google.com/g/pretext-dev/c/TzmIPakQrig.

I'm punting on the localization, especially given that the button might be short-lived.  It looks like "Previous", "Up" and "Next" are localized, but not "Dark Mode/Light Mode" or other "nav" bar buttons.

rbeezer:
> especially given that the button might be short-lived

Agreed.

rbeezer:
Merged, with only one code change:
```xsl
<xsl:variable name="b-has-embed-button" select="$embed-button = 'yes')"/>
```
wouldn't run.  :-(

Broke out the publisher variables bit onto its own commit, which I'd like to be general practice.

Real nice.  I think I'm going to turn this on for the sample article.

oscarlevin:
That's what I get for last minute changing the default and not testing.  Thanks for catching that.

PR #2472, opened 2025-04-23
Program includes updates

ascholerChemeketa:
Adds new mechanisms for building multiple "programs" into one real program on Runestone. Deprecates `@datafile` in favor of broader `@add-files`

Goes with: https://github.com/RunestoneInteractive/rs/pull/699

The new RS code should be safe to merge without any PreTeXt changes. The new PTX code depends on RS code being in place.


rbeezer:
Looks good, but a few nits to clean up, and a couple bigger things.  Not much I can test, I think, so I'm going to rely on your testing, examples in sample book, and release into the wild.

*  Usually I don't like two ways of doing things, but mostly I've allowed comma-separated and space-separated lists in attributes.  And I do like you do to make one or the other internally.  I think we should always allow both and advertise both.  Add space-separated as an option here in documentation?
*  Whitespace in  `rune.xml`, blank line just after "intentionally different".
*  I'd author `.cpp` in a `c` element.
*  Right margin, comment line holding "for the entire question."
*  There's a merge conflict with two recent deprecation messages.  When you rebase you will see what needs to happen.  Please move yours up to the chronological location in list.
* Remind me to build associated schema files!  ;-)

Should the static versions of these multi-part progrms give some indication of where their parts live.  I know a C(++) program might have an `#include` but I think we have discussed having the same program-filename evolve over the course of a book.  How does the braille reader know which `program` is being added into a compilation?  In `pretext-runestone-static.xsl` you could add `xref` that will translate correctly in all static formats.  (Another argument for `@xml:id` on the things we point to!)

And finally, I see two `Unknown error` from `lxml`.  And no better hints.  I hate that.  Maybe you can bisect to get close?  I'm not seeing that output breaks any place obvious.  Maybe `xsltproc` will give a better indication.

ascholerChemeketa:
Force pushed with fixes to identified items.

As discussed in breakout, the identification of fragments is being addressed through author education re: xml:ids on programs vs listings. The SB examples now has a demonstration of this with commentary.

I'm not seeing any `Unknown error`s after rebase and fixes.

Don't forget to build the schema. :)

bnmnetp:
Runestone side is merged and part of 7.5.17

rbeezer:
Checkout commit at `HTML: include filename and RS id even for static programs`, build before and after for Runestone hosting.  Do a "word-diff".

A bunch of `program` (I suspect) are not related to any including, so do not have a `@label`, so produce `id="PTXSB_2_`.  Repeatedly.  Which HTML validators will not like (not unique).

* Warning seems overkill - might be lots of `program` and not much including.
* As-is is a problem.
* Should we not make an `id` when it is not needed and there is not a `@label`?

We could get fancier with a global compendium of `program` that get referenced.  Create or warn only as strictly necessary.

rbeezer:
Last commit (deprecation) should be three commits: schema, sample book, actual deprecation message.

I guess it is too hard to "fix" the old filename associations, by figuring out the `@xml:id` of the included programs and writing an @add-files attribute with the pre-processor?  I guess there may not even be an @xml:id to use.  I hate to make changes that are urgent.  Is the old way being preserved?  I can't tell.

Warning with "Missing filename attribute on".  Where does that go?  Into the manifest?  And then into the database?  Will there always be an @xml:id in this case?

(Sorry for the scattered replies.)

rbeezer:
Not too much digging, but some.  Trying to get this out before I go out the door.

I don't like the complexity of making something and throwing it out (which might be a bit of a simplification).  Before this, I had the following ideas.

Three types of `program`:
*  Those Runestone needs to put into the database (i.e. in the manifest) for including, etc.
*  Those in "programming exercises", already known to Runestone.
*  Those that are just narrative.  Static by choice.

The includ'ed `program` need an `@xml:id` (target) and a `@label` (part of Runestone database identifier).  Others not so much.  We **know** which these are because the `@add-files` attribute points to them.

So make a globally-available node-set (`$runestone-program-id`) from a document-wide traversal:
```xml
<token>hello-word-c</token>
<token>quick-sort</token>
<token>average-from-a-file</token>
...
```
(We could check that targets exist, during construction).  When we hit a `program`, like during manifest creation, we can check if it is in this list.  There is an easy XSL construction to test if a string is a value across a node-set.  I think you just test equality of a string with a node-set.
```xsl
test="$an-xml-id = $runestone-program-id"
```

The schema should make `program/@xml:id` optional, and it could be authored when it is superfluous, but I think we have all the information we need to understand the difference and do the right thing.  Such as making the right Runestone identifier.

Very possible I'm missing some subtlety.  We can chat more Friday morning if that is easier.

ascholerChemeketa:
New force push to accommodate rework based on "generate list of program ids included elsewhere" approach

A few other changes:
* Reworked `runestone-targets` to named template. I needed it to work on string, not just on an attribute containing a string.
* Added `unique-token-set` as a text-util. It essentially is str:tokenize that guarantees unique tokens. I think it could come in handy elsewhere. 
* Made sure we don't emit programs with invalid active language (i.e. '') to the manifest. (e.g. pascal) in [219c113](https://github.com/PreTeXtBook/pretext/pull/2472/commits/219c11351052dfb395b1226351b8e64badd5b717). Those were causing problems with manifest parsing by Runestone when the programs got ids in the commit just after.

rbeezer:
Very good.  Really clean diffs, and I like the granularity of the commits (all 17 of them!).

Still a bit confused about why we make bad ids and then toss them out.  We can discuss at Drop-In.

Would you do an announcment on `pretext-announce`?  Mostly to explain that the deprecation does not blow up the old way via filenames.

Single digits.  ;-)

ascholerChemeketa:
@oscarlevin Will CLI auxillary files need an update wrt this? I'm not sure what is in them, but I see mention in the guide. Something to check before updating the CLI with a version of PreTeXt that includes this PR.

PR #2476, opened 2025-04-28
CSS: limit statement/feedback width in wide hparsons for salem/denver

ascholerChemeketa:
Just submitted a RS PR that removes a hard-coded cap on hparson width: https://github.com/RunestoneInteractive/rs/pull/706

This adds some limiting to the statement and feedback after Denver/Salem stretch out the container.

@oscarlevin should bless change to Denver.

Oscar - it might be worth breaking refactoring `_rs_widen` to be something that we each `@use` after defining different values for the variables. Do you think it is likely that we will continue to want exactly the same modifications just with different numbers?

image

### DIFF
--- a/css/targets/html/denver/_rs-widen.scss
+++ b/css/targets/html/denver/_rs-widen.scss
@@ -62,6 +62,14 @@ $grouping-elements: ".theorem-like, .definition-like, .example-like, .exercise-l
   margin-right: auto;
 }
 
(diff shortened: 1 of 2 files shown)


oscarlevin:
This pr is good for me.  We discussed how to simplify these changes in the future, which can be a future problem.

rbeezer:
Merged, as-is.  Thanks.

PR #2477, opened 2025-04-28
Texstyles: fix appendices and add required files for ams journals

oscarlevin:
This fixes the ams journals' texstyles by adding missing required files and removing the appendices environment (which is not used for these styles).

### DIFF
--- a/journals/texstyles/ams.xml
+++ b/journals/texstyles/ams.xml
@@ -68,15 +68,11 @@
 
     <acknowledgments heading="\section*{Acknowledgments}"/>
 
-
-
-    <appendices env="appendices" />
-
+    <appendices />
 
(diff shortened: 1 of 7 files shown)


rbeezer:
Tested with journal `conform-geom-dyn` on the sample article.  Worked as expected.  Merged as-is.

But the back matter was a real mess.  I guess that is to be expected, since the point here is that this journal does not expect appendices.

Can we see the empty TeXstyle `appendices` element, and skip any PreTeXt `appendix` elements and/or warn that `appendix` will be a mess?  I think three `references` all accumulated at the very end.   When really two of them are intermediate.  Did not investigate much on this.  Maybe mid-article PreTeXt `references` should be skipped?

PR #2479, opened 2025-04-29
Texstyles: tweak Springer Nature to work with its cls file

oscarlevin:
Had to add some new logic in the `pretext-latex-texstyle` file as well, to handle options on documentclass and add some additional preamble code.  Also an option to not use the standard hyperref preamble chunk.

### DIFF
--- a/journals/texstyles/springer-nature.xml
+++ b/journals/texstyles/springer-nature.xml
@@ -15,71 +15,42 @@
         <date>2025-04-01</date>
     </metadata>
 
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks fior the updates.  Tested on `springer-nature` (duh).  Merged as-is.

Noticed that titles on cases do not have any space after them.  Seems to be generally a Texstyles thing (all journals, and only journals, not ths PR).  Did not investigate.

PR #2480, opened 2025-04-29
Texstyles: clean up obsolete latex-style files

oscarlevin:
This removes the various `latex-style` xsl files that predated the improved texstyle infrastructure.  Deleted the xsl and removed those names as options from the publication variables file.


rbeezer:
Love **removing** cruft.  Thanks for the maintenance.  Minimal testing.

PR #2481, opened 2025-04-29
Print css for standalone worksheet 

ascholerChemeketa:
This sets up a separate stylesheet for `media="print"` on printable standalone worksheet pages. 

It could be this is the wrong direction - especially if we end up wanting theme styling (boxes/headings) to "leak through" to print worksheets. But I think this is the best approach if we want a single print style.

Maybe merge as a branch that David/Mitch can work on next week?

They would be doing work in `css/components/_worksheet.scss` and `css/targets/print-worksheet/print-worksheet.scss`.

### DIFF
--- a/css/components/_printing.scss
+++ b/css/components/_printing.scss
@@ -1,14 +1,9 @@
-// TODO - refactor 
+// Basic rules for printing a standard PreTeXt page
 
+// Get rid of UI elements.
+// Leave margin fiddling for system print dialog.
 
-.print-button {
-  position: relative;
-  right: 2px;
-  background-color: LightGreen;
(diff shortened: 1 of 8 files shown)


rbeezer:
Looking good.  Saved a blank page on one worksheet in the sample article, so this is already paying dividends.

*  Five bits of whitespace.  Removed.
*  I broke out the last commit in two.  I like to reserve "Script:" as shorthand for `pretext/pretext{.py}`
*  I'm going to post to -dev, and BCC  @davidfarmer and @mitchkeller so they don't miss it.

Thanks!  Its great to have this all isolated.

PR #2482, opened 2025-04-30
Scorm: implement building scorm archives

oscarlevin:
For the record: a SCORM archive is a zip folder containing a number of files (in our case html) and a manifest file.  This archive can be uploaded to an LMS (or other hosting platform), which will allow an instructor to build a PreTeXt document so that the result can only be seen by students in their course.

While more advanced features might be added in the future, this accomplishes the generation of the archive in the simplest way possible.  One additional template in `pretext-html.xsl` generates the manifest file.  The only customization is to the title of the scorm object (we just include `index.html` as the entry point).  The template is only "called" when the stringparam `html.scorm = "'yes'"`.  

On the python side, this is achieved by letting the `file_format` for an html build be 'scorm'.  Besides setting the stringparam, this then behaves exactly like when `file_format = "zip"`.  Added `html-scorm` as a format for the pretext script.

Happy to discuss other implementations, but did test this and its working.  Assuming we move forward, I will follow up with documentation in a later PR.

### DIFF
--- a/pretext/pretext
+++ b/pretext/pretext
@@ -367,6 +367,7 @@ def get_cli_arguments():
         ("latex", "LaTeX source file (only, an intermediate format for developers)"),
         ("html", "HyperText Markup Language (online/web pages)"),
         ("html-zip", "HyperText Markup Language (single zip file)"),
+        ("html-scorm", "HyperText Markup Language (single zip file including SCORM manifest)"),
         ("revealjs", "PreTeXt slideshow to reveal.js HTML format"),
         ("epub-svg", "EPUB container, math as SVG"),
         ("epub-mml", "EPUB container, math as MathML"),
@@ -752,6 +753,19 @@ def main():
                 stringparams,
                 None,
                 "zip",
+                extra_stylesheet,
+                out_file,
+                dest_dir,
+                None
+            )
+        elif args.format == "html-scorm":
+            # same as html-zip, but with SCORM manifest
+            ptx.html(
+                xml_source,
+                publication_file,
+                stringparams,
+                None,
+                "scorm",
                 extra_stylesheet,
                 out_file,
                 dest_dir,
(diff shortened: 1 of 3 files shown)


rbeezer:
This looks good, but bigger things are afoot.  Following error happens with `html-zip` as well as `htnl-scorm`, so needs fixing, generally.

Build sample article.  Hangs after
```
PTX:INFO    : packaging a zip file temporarily as /tmp/ptx-5dg7febd/html-output.zip`
```
A Ctrl-C gets a traceback that ends with
```
RuntimeError: File size unexpectedly exceeded ZIP64 limit
```
Cursory googling is not giving me anything obvious to try.  `git bisect` might indicate if/when our HTML output got too big?  Or got some new big files?

oscarlevin:
I will investigate.  But apparently zip64 can handle sizes up to 16 EB, and even zip can hold up to 4gb.  So it must be something else.

I didn't test with the full sample article, just my course materials sample.

oscarlevin:
I can't reproduce.  Just build the sample article successfully with both zip and scorm compression.  On linux, but with the CLI.

rbeezer:
Thanks for double-checking.  Mystery solved - the zip file is being deposited into the same directory as the files being zipped.  So a (partial?) version is being zipped into the zip, which trys to zip that, which tries to zip that,....

Maybe because I am on a new version of Linux (Debian) there's been a change in file-writing behavior?  Anyway, it seems a simple conditional will stop it, or maybe we should just build the zip somewhere else.

Fixing this, then back to this PR, but might be later today.

rbeezer:
Very good!  In the end, that was pretty simple.  Thanks for digging into the details on this one.  I'll be interested to see what sort of enhancements can come next.

Merged, as-is, but with some heavy editing on commit messages.  There should be enough `scorm` about to see what is happening.

If you think the features will grow, this could be a whole section in the publisher chapter about HTML.  I'll let you advertise once the CLI is ready and there is some documentation.  Let's play up the small (intoductory) project where an instructor can write their syllabus in PreTeXt, get an accessible version, integrate it into their Canvas (or other LMS), and impress their administration.  ;-)  Or at least get bonus points.

PR #2483, opened 2025-05-01
Portable build improvements

oscarlevin:
My goal for a portable build is to produce a minimal number of files, probably just one.  This gets us two steps closer by embedding the lunr search code inside the html page (only for portable builds, of course) and skips the building of the `index.html` redirect file.  

Next, but requiring further discussion I think, is to deal with worksheets and interactives (both of which produce additional html files).  And after that, look at base64 encoding of non-svg content.


rbeezer:
Merged.  Two steps closer!  Thanks!

Not much testing - only verified that "regular" HTML was not harmed.

PR #2484, opened 2025-05-02
More portable

oscarlevin:
A bunch of small fixes to reduce the number of extra files that get built for a portable build.  Included is an override of chunking level (without warning), removal of standalone pages where possible, removal of worksheet pages and their buttons, and removal of the ol-markers.css (which I understand will soon be obsolete anyway).

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -605,7 +605,8 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <!-- to generate standalone printable and editable   -->
     <!-- versions. $paper becomes HTML class names, e.g. -->
     <!-- LOWER CASE "a4" and "letter"                    -->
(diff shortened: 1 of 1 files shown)


rbeezer:
This a hard one to do a meaningful test on.  But the output (files, and rendering) looks good.  So it must be OK!

Seriously, code looks good, fewer files for sample article, and rendering looks good.

> ol-markers.css (which I understand will soon be obsolete anyway)

I think that is news to me?  Anyway, no reason to hold this up.  But until it goes away, we want to remember we are back in the old situation of custom markers only working in LaTeX.

No low-hanging fruit left?  Ideas on embedding certain other files into the HTML proper?  Base64 for raster images?

oscarlevin:
Right.  I'm going to try some base64 encoding next.  Then I need to think about what to do with the generated and external folders.  Not a big deal of the output is going into a specific output folder for the build, but parallel output would be a mess.

PR #2489, opened 2025-05-05
Retire WW 215

Alex-Jordan:
This should simplify many things for the larger Tacoma PR. Net 165 deletions here.

This removes all of the special case handling if the WW server version is 2.15 (or 2.14). So such old WW servers can no longer be used to generate output. I announced on pretext-announce on June 18, 2024, that this would happen "soon". If this is merged I can announce again that it has actually happened.

In my testing with the webwork minimal and webwork sample-chapter examples, there is no change in either:

- webwork representations file and 
- html output files

between this branch and "master" branch. The only changes are

- timestamps
- binary image files register a diff but are the same size and appear the same to me visually (so maybe a timestamp buried in the binary)

If one does try to use a WW server 2.15 or earlier, there will be an error about that, and a suggestion to use the AIM WW server.

### DIFF
--- a/doc/guide/author/webwork.xml
+++ b/doc/guide/author/webwork.xml
@@ -18,7 +18,7 @@
   <introduction>
     <p>
       With a <webwork/> server
-      (version 2.14 or higher, or <url href="https://webwork-ptx.aimath.org" visual="webwork-ptx.aimath.org"/>)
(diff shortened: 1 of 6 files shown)


rbeezer:
Very good!  That's bye-bye to a lot of cruft.  ;-)

Merged with no code changes, just edits to commit messages.

> this is merged I can announce again that it has actually happened.

Please do!  Bonus points if you can quote the previous message, making it obvious this was planned 11 months ago.

PR #2491, opened 2025-05-08
Reorganize pretext script's library

oscarlevin:
This moves `pretext.py` and `braille_format.py` to a `lib` folder inside `pretext` (the script's folder).  The reason for this is to start using relative imports, as in `from . import braille_format` rather than `import braille_format`.  This is necessary for the CLI (I guess you can't use the latter sort of imports if import the parent as a module).  

The `pretext` script should still work exactly as it has in the past.


rbeezer:
Merged as-is.  Added a follow-on commit getting the right path to the root of the (installed) repository.  I hope this makes life easier on the CLI side.  Thanks.

rbeezer:
Forgot to say.  Merge produced the following report:
```
Merge made by the 'ort' strategy.
 .gitignore                          | 1 +
 pretext/lib/__init__.py             | 0
 pretext/{ => lib}/braille_format.py | 0
 pretext/{ => lib}/pretext.py        | 2 +-
 pretext/module-test.py              | 2 +-
 pretext/pretext                     | 2 +-
 6 files changed, 4 insertions(+), 3 deletions(-)
 create mode 100644 pretext/lib/__init__.py
 rename pretext/{ => lib}/braille_format.py (100%)
 rename pretext/{ => lib}/pretext.py (99%)
 mode change 100755 => 100644
```
The last line is suspicious.  Let's keep an eye out for anybody who has trouble "executing" any of this.

PR #2492, opened 2025-05-09
Sample Chapter: move chunking to publisher file

Alex-Jordan:
This just moves the chunking specification for the sample chapter problem set archive from the Makefile to the publisher file.

### DIFF
--- a/examples/webwork/Makefile
+++ b/examples/webwork/Makefile
@@ -199,13 +199,13 @@ sample-chapter-problem-sets:
 	install -d $(PGOUT)
 	cd $(PGOUT); \
 	rm -r Integrating_WeBWorK_into_Textbooks; \
-	$(PTX)/pretext/pretext -v -c all -f webwork-sets -x chunk.level 1 debug.datedfiles no -p $(SMPCPUB) $(SMPC)
+	$(PTX)/pretext/pretext -v -c all -f webwork-sets -x debug.datedfiles no -p $(SMPCPUB) $(SMPC)
 
 mini-problem-sets:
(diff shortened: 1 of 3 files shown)


rbeezer:
Putting that value in the publication file now means that the HTML version of the sample chapter (at the website) wold be chunked different.

Let's have two publication files, one at level 1 and one at level 2, and adjust the makefile to use the right one in each case.

rbeezer:
Putting that value in the publication file now means that the HTML version of the sample chapter (at the website) wold be chunked different.

Let's have two publication files, one at level 1 and one at level 2, and adjust the makefile to use the right one in each case.

Alex-Jordan:
Oh, I hadn't realized the default chunking for HTML was different. I haven't internalized when chunk level is 1, 2, etc what that actually means.

Before I do what you suggested, any reason to not just go with chunk level 2 for everything, including the set definition files? I know that will mean many more set definition files, some with very few exercises in them. So that will be a big diff for this PR. But I'm wondering if we really care since this is just a sample.

rbeezer:
> just go with chunk level 2 for everything

Yes, that'd be fine.  Nobody is **using** those problems...(I hope).

Alex-Jordan:
OK, now chunking set to level 2.

rbeezer:
Good.  Thanks.  You probably saw that the diffs on the problem sets did a nice job of tracking the renames on a few files.  (Not important, of course, just of interest.)

Old stringparam came off the minimal example in the makefile, so I added it into the publication file for that project.

PR #2493, opened 2025-05-09
Sample Chapter: don't date representations for ease of examining diffs

Alex-Jordan:
When I make some change where I need to examine a diff on the sample chapter representations file, the diff is always full of unimportant changes where it's just the run time timestamp. The first thing I do is scrub those from the diff output. But this change will just stop printing timestamps into the extraction file in the first place. Future diffs on sample chapter output will have a lot less clutter.

### DIFF
--- a/examples/webwork/Makefile
+++ b/examples/webwork/Makefile
@@ -118,9 +118,13 @@ sample-chapter-macros:
 #  Extract webwork problems into a single XML file.
 #  Location is within generated folder (named in the publisher file),
 #  within "webwork" subfolder
+#  We turn off dates in file headers, so when we use these for testing we
+#  can get more accurate diffs. Dates are useful for single file output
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
OK, I will close this. I have something else in mind though that will have the same effect (but leave one date at the top of the representations file). I'll lightly describe that here in case you see a red flag and stop me. The representations file would just have the converter blurb at the top in some element. So the representations file would no longer have the same blurb copied 100 times (in the case of the sample chapter). For now I would also rig the archive routine to prepend that one blurb from the representations file to each .pg file so that .pg files end up with converter blurbs. (That last bit becomes unnecessary when #2462 is resolved.)

I understand you don't find them in the way when reviewing my PRs. For me while developing, and doing a dozen diffs for each one you might do one during a review, it adds up. And I'd like to remove them.

PR #2494, opened 2025-05-09
WeBWorK Minimal: include an OPL problem

Alex-Jordan:
My plan is that the WW minimal example end up with one example each of:

- PTX authored exercise (that it already has)
- Problem from a host course on a webwork2 server (that this PR adds)
- Local problem file (that will come later as the larger Tacoma PR is chipped away at)

Adding this one now will make some future diff smaller.


rbeezer:
Got it.  Thanks.

PR #2495, opened 2025-05-09
WeBWorK Minimal: put webwork element in pub file

Alex-Jordan:
This adds the `webwork` element to the WW Minimal example's publisher file. All the settings are just the default settings. But having this here helps with development by making it easier to change a setting and build.


rbeezer:
Merged!  Thanks.

PR #2496, opened 2025-05-09
Sample Chapter: project file maintenance

Alex-Jordan:
This is just maintenance for the project.ptx file for the sample chapter, to help it be built using the CLI. Also to use a folder name ("html") consistent with what the Makefile build creates.


rbeezer:
Merged, but with no testing, since I'm not yet proficient with the CLI yet.

PR #2497, opened 2025-05-09
Sample Chapter: don't use special characters in description

Alex-Jordan:
Special perl characters are used in a description. Under the new code in the Tacoma project branch, this causes problems. Eventually those problems should be worked on, so that it can be OK to use characters like this in a description. But it's a project for some other day. For now I just want to sidestep this.


rbeezer:
> For now I just want to sidestep this.

I don't want to remove features that some projects may be using.  Whatever is going on with 2.19 this needs to be understood.  There is always the chance that it is a symptom of some larger problem.

Alex-Jordan:
OK, that's fair.

PR #2498, opened 2025-05-09
Guide: remove webwork from publihser files

Alex-Jordan:
The Guide's publisher file has webwork settings, which are unnecessary because there are no live WeBWorK exercises in the guide. (IIRC there was one once, but because it complicated building the guide, we removed it.)

So this removes all that fro the publisher files. It helps reduce clutter when searching for webwork things that need updating.


rbeezer:
Done!

PR #2499, opened 2025-05-09
WeBWorK: consolidate course-password and user-password

Alex-Jordan:
Once upon a time, I think that Mike Gage had a scheme in mind for authenticating into a WeBWorK course for a remote procedure call where two passwords would be necessary. Their names are misnomers, I think, but were called `coursePassword` and `userPassword`. We set things up so that PreTeXt would be ready to work with both. But all that was ever needed was `coursePassword`. The other one has never been needed or had an effect.

So this PR:

- mostly scrubs away both of them, and puts a simple `password` in place instead
- except out in the wild, publisher files and stringparams might still use `coursePassword`, and that is still supported for backwards compatibility (but with a deprecation warning)
- also versions of WW 2.18 and earlier are expecting to hear a param named `coursePassword`, so that is still used in two places by `pretext.py` when it needs to talk to a webwork2 server
- and also I do not want to edit the `pretext-webwork.js` files here that go with old versions of WeBWorK. Copies of those are hosted in places like the utmost server and even with individual projects that are hosted up on the internet somewhere, since we started using local js files. So those files still look for `coursePassword` as a data attribute on a certain div in PTX HTML output. And so that div keeps a data attribute with name `coursePassword` too.

All of this should set things up to simplify the big Tacoma PR with 2.19 support.

### DIFF
--- a/doc/guide/publisher/publication-file.xml
+++ b/doc/guide/publisher/publication-file.xml
@@ -1001,16 +1001,6 @@
             </p>
         </subsection>
 
-        <subsection xml:id="publication-file-webwork-course-password">
-            <title>Course Password</title>
-            <p>
-                A course password (distinct from a user password) is in the attribute
-                <cd>
-                    <cline>/publication/webwork/@coursepassword</cline>
-                </cd>
-            </p>
-        </subsection>
-
         <subsection xml:id="publication-file-webwork-user">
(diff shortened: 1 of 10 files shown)


rbeezer:
Good clean diffs.  Merged.  Thanks.

PR #2500, opened 2025-05-10
Prefig cache prep

oscarlevin:
The three commits will allow the CLI to cache prefigure assets, the same as we do for latex, sage, and asymptote.

Note that the middle commit does change the behavior of the script when just tactile outformat is selected.  With the change tactile output will always be put in the `tactile` subfolder.  This way, if someone builds tactile and then pdf, the files won't be overwritten.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ doc/guide/webwork-representations.ptx
 script/mjsre/node_modules/
 script/mjsre/package-lock.json
 pretext/__pycache__/**
+pretext/lib/__pycache__/**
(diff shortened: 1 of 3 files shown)


rbeezer:
Includes a previously merged commit 1c89ac6d0d7311af7dda57c722491e649cbfc73d

Don't do anything, it got skipped.

rbeezer:
Merged, with just light editing of commit messages.

Only tested SVG geberation, so didn't explore visual/tactile PDFs.

Net +2 lines of code.  ;-)

Issue #2478, opened 2025-04-29
worksheet dimensions and scrolling in html vs print


Tanaquil18:
I was playing around with a worksheet, in a branch of a project using Denver theme.  As of CLI 2.17, there is now a scroll bar which makes html look nice, but the printing from html shows the scroll bar and obviously can't scroll.  Before CLI 2.17, I was using CLI 2.15, and the overflow looked very bad in the html, but printing from the html made a good-looking pdf where everything fit nicely on one line. @mitchkeller @davidfarmer 

Source is here: https://github.com/davidaustinm/fcca/blob/optional-lab-worksheet/source/sec_optionallab.ptx

Tanaquil18:
Because it's on a branch, it's not deployed anywhere live, at least not yet.  If that changes before next week, I'll update with a link.  

I do have some worksheets in another project that is deployed, if you discover you want more live examples.  I haven't built in a few weeks though and there are no scrollbars there right now.

rbeezer:
Where are we on this one?  #2501 might have had some impact?  In different words - does this need work and we now know what to do, or should it be closed?

davidfarmer:
Closing.  There may be problems depending on the theme.  If so, new issue.

PR #2501, opened 2025-05-10
worksheets have appropriate width before printing

davidfarmer:
Worksheets before printing were inappropriately narrow, because `max-width` of `.ptx-content`
was too small.

This over-rides that for .worksheet.

Now it looks okay both before and after hitting "print".

This takes care of at lease some cases of #2478


rbeezer:
Merged.  Thanks.

As-is, other than one stray bit of whitespace - a newline at the bottom of the file.  ;-)

PR #2502, opened 2025-05-10
LaTeX: Fix missing $ when deciding if should load multicol package

mitchkeller:


rbeezer:
Oops... introduced at ae20daf85c9d658e0319446b917e2f7277c426b0

rbeezer:
Forgot to close this up.  **Great catch!**  Package loading is pretty forgiving in the sample article, so these are hard to notice.  Thanks.

PR #2503, opened 2025-05-11
HTML: Hide toc-toggle and ptx-sidebar when toc-level is 0

mitchkeller:
Working to fulfill the promise in the Guide's page on Single Page HTML that if you set `tableofcontents/@level` to 0, then "there will be no sidebar Table of Contents at all".

### DIFF
--- a/css/components/helpers/_buttons-default.scss
+++ b/css/components/helpers/_buttons-default.scss
@@ -39,6 +39,10 @@ $border-radius: 0 !default;
     cursor: not-allowed;
   }
 
+  &.hidden {
+    display: none;
+  }
+
   &.open {
     color: var(--button-hover-text-color);
     background-color: var(--button-hover-background);
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -11759,6 +11759,11 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
 <xsl:template match="*" mode="primary-navigation-toc">
     <button class="toc-toggle button">
+        <xsl:if test="not($toc-level)">
+            <xsl:attribute name="class">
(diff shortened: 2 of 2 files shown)


ascholerChemeketa:
Why render them to HTML at all if they are going to be hidden?

But CSS wise, it looks fine. 

I like that the CSS is a compound style. `&.hidden` as opposed to `.hidden`.

For brand new things, I would encourage something BEM style: `.toctoggle__hidden` so there is no possible intersection with other  stylesheets that might be injecting a rule to target the generic class name `.hidden`. But the legacy button code doesn't do that. And lots of other bits of PTX depend on a generic class `.hidden`. I think Mitch's approch was the right one there.

ascholerChemeketa:
I tested with completely ripped out html elements instead of hidden via css. `default-modern` and `salem` look fine. `denver/greeley` have an issue, but it is likely fixable without too much pain.

rbeezer:
Thanks, @ascholerChemeketa!

I'm going to merge this now for @mitchkeller (after testing).  You two can sort out if it should be upgraded to be more modern.

rbeezer:
I've merged this so @mitchkeller  can use it tomorrow from the nightly.

HTML changes look good, some CSS I'm unsure about.  Some items are not horizontally aligned as nicely as before, but not critical for this use, I think.  Maybe a "centered" theme would do better?

Slightly rewritten (silently) as we discussed in-person.  Definitely needed some `xsl:text` action.

Seems **all** buttons could now react to a `hidden` attribute.  See consequences at be3c49b7141f06b1d440d48247699a7db6703376

In consultation with @ascholerChemeketa I'd like some follow-up (since this came with a deadline) if necessary.  Limit scope of the change.  Do not create certain HTML items (instead of hiding them)?

PR #2504, opened 2025-05-12
Sample Book: Add multiple choice with math content in feedback

mitchkeller:
Also fixes a `@label` that should be `@xml:id`.


rbeezer:
Very nice example, that's a great addition.  Math renders fine for me (on Firefox), maybe @bnmnetp has addressed this.

And thanks for the drive-by fix.

PR #2505, opened 2025-05-12
WeBWorK: better control over blank lines

Alex-Jordan:
Over in #2462 I want to refactor so that problem set archives (pg files, set definition files, etc) are generated directly from source. This means `pretext.py` would no longer be part of the process.

Currently, either through laziness or technical limitations, `extract-pg.xsl` produces PG content with excess blank lines in places. And `pretext.py` condenses all such instances into a single blank line. But that will not happen once archiving bypasses `pretext.py`.

So this commit:

- preemptively removes that trimming code from `pretext.py`
- refines `extract-pg.xsl` to produce fewer blank lines in the first place

The net effect is that in some places, the PG output has a blank line where there wasn't one before.

- Sometimes this is actually an improvement: now there is a blank line preceding a block of latex-image code.
- Sometimes this is turning what was a single blank line into a double or triple blank line. This is OK. It happens following lists or list items. I don't think I'll ever work out how to actually suppress the excess blank lines, because of the complications from list items being structured or unstructured. And other complications like if the `p` surrounding the list has test nodes following the list. And is the list the last element in the `p`. And more.

The diff in the problem set archive for the sample chapter only shows these things I just mentioned. I see no diff at all in the latex file output. For HTML, the only diff is in the base64 encoding of the problems. This is the same base64 encoding used to fetch the static versions, which are then written into that latex file. So I am confident that all the problems render in the live HTML.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -1883,10 +1883,6 @@ def webwork_to_xml(
                 )
             else:
                 formatted_pg = pghuman[problem]
-            # opportunity to cut out extra blank lines
-            formatted_pg = re.sub(
-                re.compile(r"(\n *\n)( *\n)*", re.MULTILINE), r"\n\n", formatted_pg
-            )
             pg.text = ET.CDATA("\n" + formatted_pg)
         elif origin[problem] == "server":
             pg.set("source", source[problem])
--- a/xsl/extract-pg.xsl
+++ b/xsl/extract-pg.xsl
@@ -172,9 +172,14 @@
         </xsl:attribute>
         <!-- 4. human readable PG (for PTX-authored)                               -->
         <pghuman>
-            <xsl:apply-templates select=".">
-                <xsl:with-param name="b-human-readable" select="true()" />
-            </xsl:apply-templates>
+            <xsl:variable name="pg">
+                <xsl:apply-templates select=".">
+                    <xsl:with-param name="b-human-readable" select="true()" />
+                </xsl:apply-templates>
+            </xsl:variable>
+            <xsl:call-template name="consolidate-empty-lines">
+                <xsl:with-param name="text" select="$pg"/>
+            </xsl:call-template>
         </pghuman>
         <!-- 5. PG optimized (and less human-readable) for use in PTX output modes -->
         <pgdense>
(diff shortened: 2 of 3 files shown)


Alex-Jordan:
Rebased now that #2492 is processed. Shouldn't make a difference though.

rbeezer:
Despite appearances, I've spent a lot of time with this one.  Mostly because I have a real hard time understanding expressions like
```xsl
test="*[last()]/self::ol|*[last()]/self::ul|*[last()]/self::dl and (not(*[last()]//p) or normalize-space(node()[last()]/self::text()))"
```
I am always suspicious of `first()`, `last()` and `position()` on **authored source**, though I can't replicate the times it has bit me.  Just gun-shy.  And I don't thunk I have had so much cause to write many tests that are so complicated.

But maybe I don't understand PG well enough to understand the desired effects here.

I guess I'd be willing to have a few too many blank lines in return for simpler code.

Anyway you can give this a thorough once-over and maybe simplify it some?  Even if the PG is not quite as clean as we would like?

Alex-Jordan:
I added a utility to `pretext-text-utilities.xsl` that consolidates adjacent blank lines into one blank line. So now this is better described as moving that consolidation out of the python and into the XSLT.

Then I removed much of the changes that were previously here (before the force push). I left a few places where this adds new lines (if only to often, but not always, have them removed later by the consolidation utility).

Alex-Jordan:
Now split into two commits.

Alex-Jordan:
> Did always using sanitize-text get you an extra newline from running the template on a null string?

Yes, that's right.

There are many exercises in the sample chapter without a `pg-code` section. They are mostly just there to test something. So now with the "if" wrapper, a newline is removed from the PG for those exercises.

Another newline gets added in those places by something else. So for human readable PG, you see no change because the multiple blank lines get condensed now as they did before. But for non-human-readable PG (the kind that gets converted to base64 and used as a `server-data` attribute) this removal of an unnecessary line break shows up in the diff.

It's not going to make an efficiency impact, but it was a remnant from when I meticulously reviewed each and every line break and scrubbed away what was not needed. Ultimately in the non-human-readable PG, that removes a blank line entirely (as opposed to condensing consecutive newlines into one). So it's a bit different than other newline removal in this PR.

rbeezer:
> But for non-human-readable PG (the kind that gets converted to base64

That's what I was forgetting!  Thanks.

PR #2506, opened 2025-05-13
Runestone: remove multiple choice exercise statement from ul

ascholerChemeketa:
As wondered at in the existing comment, multiplechoice questions are generating invalid HTML by placing the statement inside the `<ul>`. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ul#technical_summary

I noticed because an attempt to put a mermaid diagram in an exercise did not work correctly.

This moves the statement outside of the `<ul>` and wraps it in a div. The `exercise-statement` and `exercise-interactives` classes can hopefully be slowly added to RS components - currently each component has a largely unique set of styling rules.

This does result in the `Choose one` text being too close to the question. Addressing that in a RS PR to also fix dark mode font color for that text.


rbeezer:
Merged as-is.  Looks great off-Runestone.  The "Choose one" rule is new?  I like it.  We could mimic it in static (maybe).

It'd be good if @bnmnetp and I had a break from new features and went back to modernize our coordination.  After we get general matching sorthed out.

PR #2509, opened 2025-05-14
CSS: ensure .hidden doesn't display

oscarlevin:
@mitchkeller discovered that using the `?embed` query string for embedding a ptx page doesn't always hide the outer page elements on all themes.  For example, in `greeley`, other css display commands were overriding the `.hidden {display: none}`.  

This adds `!important` to the `.hidden` selector's display.  @ascholerChemeketa, any reason you can think of that this would be a bad idea?  It seems like if we have put `hidden` in the class, that should always take precedence.

Tested with all the themes and I didn't notice any problems (and the `?embed` is working as expected now).

### DIFF
--- a/css/components/elements/_misc-content.scss
+++ b/css/components/elements/_misc-content.scss
@@ -133,7 +133,7 @@ article.theorem-like .emphasis {
 }
 
 .hidden {
-  display: none;
+  display: none !important;
 }
 
 /* genus and species in italics */


ascholerChemeketa:
Seems reasonable for `.hidden` given the way we use it.

Pedantic add on:

`.hidden` is super convenient - the CSS doesn't need to know exactly what is hideable. JS or HTML can make anything hidden by slapping on that class. But easy has a cost.

What happens when some library that we import directly or indirectly has a `.hidden` that makes something grayed out instead of `display: none`? What happens when some HTML/CSS generated by PreTeXt is dropped into an LMS page?

The good news about `.hidden` is that in most places it probably does mean `display: none`. Something like `.active` or `.visible` is another story. Every library likely has a different idea what those means and they may even mean different things for different components within the same library. As soon as one library `!important`s its properties, it is likely going to blow up everything else.

So for similar issues with other decorator classes, I would recommend a more declarative approach. SCSS makes it easier than plain CSS to mix in modifiers so that they only affect things that the SCSS explicitly. This demo shows a `hideable` mixin used to make some but not all of the other classes respond to `.hidden`:
https://sass-lang.com/playground/#eJx9jsFuwjAQRO/+ihUgDogEodySS7jxGxt7o65w7Mi7QUGIf6/lqqjtoceZnXk753O/CMFGUKT1LLrpzA/HRh9TtsxujEErUbS3Fq7k76Rs8QiXxOg7s5sTT5geVSm0sG2aJrdMP/HKAT7YEQ6e4GkA9nWWjkIRAI5l9vhoIcRAXbZe5mVMrdGWwPucyKPyvUR6DtYvjt7grnRmXasJ879asj9g+k3AQaJf9B/C6QDf41wkyZMUcBzJKhxOX/wBQ6A/4JFXcgXwCdifano=

rbeezer:
Great!  Merged as-is.

@ascholerChemeketa: feel free to issue a ground stop (on -dev?) with anymore work on `.hidden` until it gets refactored with modifiers (or other similar work).

PR #2511, opened 2025-05-14
Sample book: Replace broken Java examples with working ones

ascholerChemeketa:
Two of the examples of Java program in the sample book are broken. One has no hope of ever working, the other has multiple issues. This replaces both with working examples.


rbeezer:
Thanks!  Not really tested, since I don't have a back-end handy.  Website update eventuallly.

`tests`  for one had loads of whitespace in some lines, and then I just got rid of the (clean) blank lines.  Closed up three import statements.  I think that is still legal syntax.

bnmnetp:
pulled and tested on my end.  Both new examples work great.

PR #2512, opened 2025-05-16
Runestone: add exercise-statement class to ac_question

ascholerChemeketa:
Adds a class `exercise-statement` to activecodes. Trying to use that style across all RS items that suck in exercise statements.


rbeezer:
Thanks for taking on this project.  Merged as-is.

Do you think `ac_question` will go away?  Only curious.  I'd bet it may be deeply entrenched?  ;-)

PR #2513, opened 2025-05-16
Codelens upgrades - questions, starting-step, and cross build target traces

ascholerChemeketa:
Adds:

* Markup for the existing question feature in Codelens. See mini demo here:
https://computerscience.chemeketa.edu/ascholer/pretext-test/codelens/code-lens.html
* `starting-step` attribute for codelens to skip past program setup.
* Cross build target traces.

Pairs with RS https://github.com/RunestoneInteractive/rs/pull/718
The RS PR should be backwards compatible with old PTX code/tracefiles.
This also should be backwards compatible. "New" tracefiles require the new `pretext.py` but will work with old RS code (using the old JS for asking questions and they won't honor `starting-step`). "Old" tracefiles parse correctly with new `pretext.py` code.

~~Last two~~ First two commits add code to the trace files so that they will attempt to find the id of the codelens they are in instead of always using a baked in id to identify the right codelens. This makes it so that a codelens trace built for a plain web target can be used in a RS build and vice-verse. Helpful while testing other parts of PR as well as for future maintenance.


rbeezer:
A quick comment after a superficial visual scan, so this can move forward.  Do Lenz' `xml-to-string` templates function just like our modal `serialize` templates?  (Are they better?)

If so, might you experiment with dropping it in as a replacement and see how it behaves?  The annotated source procedure maybe would be a really comprehensive test?  (I think.)

ascholerChemeketa:
Force push with:
* typo fix in sample
* intended first two commits rotated all the way to the front
* simplified trace file parsing in `pretext.py` to not handle old and new format from `extract-trace.xsl`. Since that py file is what runs extract-trace, there is no need for it to handle two flavors of output from xsl.

I tested the Lenz vs homegrown templates.

Differences:
* space vs no space before end of self closing tags
* homegrown escapes `>` which Lenz does not
* Lenz preserves XSL comments
* homegrown strips out `include-source` attribute
* Lenz appears to have much more nuanced handling of namespaces and other tricky bits

So I could redo this PR to use the existing serialize. I thought I was having an issue with it, but it appears whatever issue I thought I was having was maybe something else.

Or I could leave it in and we could look to replace existing `serialize` with the Lenz version. Perhaps with a parameter that controls comments?

ascholerChemeketa:
BTW - a little conflicted on how to deal with trace files. I generated them in a couple spots. But they don't play super well with git since they have giant lines. Might be better to leave update for them to one commit at end.

rbeezer:
Thanks for chasing down "XML to strings".  I think I would like just one routine at a time, never a mix.  ;-)

*  Separate PR to integrate Lenz, then this.
* Do this with existing, then swap in Lenz.

ascholerChemeketa:
Converted to draft - will rebase once [Lenz PR](https://github.com/PreTeXtBook/pretext/pull/2519) is merged.

rbeezer:
#2519 is now in.

rbeezer:
When you rebase, can we consider using something beside "question", which will conflict with the block of the same name?

ascholerChemeketa:
@rbeezer Rebase looks good, but have not done rename yet.

Just about anything reasonable ends up with some kind of name collision.

I could do `<program-question>`. But `<program><program-question>` feels really redundant. 

I don't think it is that big an ask to recognize that as a direct descendant of `<program>`, `<question>`, means something different than a `<question>` not nested under a `<program>`.

But give me a name and I'll make a swap.

ascholerChemeketa:
Also... can't believe I am asking the question, but are `codelens` and `activecode` really attribute values on `<program @interactive>`? Now that they have a fuller compliment of features, it feels kind of weird. There are a host of other attributes on program, and even subelements (!?!?!), that only make sense given a particular value of `@interactive`.

Aside from having to basically give up on the schema being correct for different combinations, just writing clear documentation is getting harder and harder.

Maybe `<activecode>` and `<codelens>` are containers for a `<program>`.  They are first class components that are designed for rendering a program in a highly interactive fashion. (Or to some listing + link/qr code/whatever in a static version.)

They get the attributes/elements specific to them. Program keeps what makes sense there. Compare this to what we have currently:

```
<codelens label="foo" starting-step="10">
   <program language="cpp">
   ...
   </program>
   <question line=10 answer="x">
        <feedback>...</feedback>
   </question>
</codelens>
```

```
<activecode label="bar" add-files="addh-v1" compile-also="addcpp-v1" compiler-args="-std=c++20, -Wall">
   <program language="cpp">
   ...
   </program>
   <tests>
      ....
   </tests>
</activecode >
```

There is a clear list of things that make sense only for activecodes. And same for codelens. This syntax would make those semantics clearer.

This would be a big change. Even if everyone thought it did make sense, I'm not proposing it for now. I think there would be a riot if we told all the RS people working on books right now to change all the code they are half done with.

Maybe it is something to aspire to and keep as an eventual goal? Maybe before next summer we could have a backwards compatible version and some script to help with conversion.

Happy to move this one to dev... using this as a "safe space" to ask the question without freaking anyone out there.

rbeezer:
> Maybe checkpoint instead of question?

Sure!  I think that captures the intent.  There will be some confusion with how we name "inline exercises" as "Checkpoint M.N" but that's no reason not to use it.

ascholerChemeketa:
To draft.

Waiting on `-dev` feedback on `<xsl:variable name="exercise-style" select="'dynamic'"/>` in extract-trace.

Would like to fix that issue as part of this PR.

ascholerChemeketa:
Question for @rbeezer and @bnmnetp:

Is it better to require `@xml:id` on a codelens in an exercise to guarantee consistent trace file name? Or stick with convention for other programs and warn that trace file name is dependent on where `<program>` is in `<exercise>`?

I assume the first, but I also hate violating the logic around `@label` and `@xml:id` wrt to programs/exercises. For other programs, it is `@label` to make it work in RS and `@xml:id` if you want to link to the code somewhere else (another program).

Problem:

Convention is `@label` for `<program>`s. Unless they are in `<exercise>` in which case `@label` is on that. 

Codelens uses `@visible-id` for filename, which will be derived from `@label` if there is no `@xml:id`. In the case of:
```
        <exercise label="program-codelens-cpp-questions">
            <title>A C++ program as CodeLens exercise</title>
            <statement>Run the codelens and answer the questions it asks.</statement>
            <program interactive="codelens" language="cpp" include-source="yes">
```
The `@visible-id` will be `program-codelens-cpp-questions-3` as the program is the 3rd element in the exercise. This will change (and necessitate rebuilding trace file) if the author add/removed the title, put a hint before the program, etc...

Specifying an `@xml-id` on the program locks down the `@visible-id` to that value. The version below will always have `program-codelens-cpp-questions-program` as its trace filename.
```
        <exercise label="program-codelens-cpp-questions">
            <title>A C++ program as CodeLens exercise</title>
            <statement>Run the codelens and answer the questions it asks.</statement>
            <program xml:id="program-codelens-cpp-questions-program" interactive="codelens" language="cpp" include-source="yes">
```

ascholerChemeketa:
Updated to use `@unique-id` for tracefile names. Rebased to resolve CSS merge conflict. Retested - looks good.

rbeezer:
> Do you prefer the explicit list without &PROJECT-LIKE;

Yes, definitely.  And I loved your code comment: `This is an implicit PROJECT-LIKE...` in another spot.  But I share your concerns.

I'm pretty careful in my searching for places that need a change.  I think this was a **contributing factor** to the recent RS  whack-a-mole, but perhaps not the only thing in play.

The two entities are of limited utility when you use more complicated XPath paths.  An alternative looks like
```
test="parent::*[&PROJECT-LIKE-FILTER;]"
```
but experience says wildcards and (complicated) filters can have **HUGE** impacts on run-time.  This might be place where it is not so bad.  Perhaps another chance to profile???  ;-)

ascholerChemeketa:
New force push. Added error reporting in the case of no valid `@label`.

rbeezer:
Thanks!  Rearranged commits look great, and I can now do better testing.

Two bits of whitespace: near the self-closing tag bit, and in `pretext.py`.

I think maybe the CSS is stale (and keeps me from building HTML):

```
PTX:INFO    : Building custom css theme: theme-salem
PTX:DEBUG   : Theme options:{"options": {}, "contrast-checks": {"color-main": "#fff", "color-do": "#fff", "color-fact": "#fff", "color-meta": "#fff", "primary-color-dark": "#23241f"}}
PTX:ERROR   : ✘ [ERROR] Undefined variable.
   ╷
78 │     max-width: $content-width;
   │                ^^^^^^^^^^^^^^
   ╵
  /home/rob/mathbook/mathbook/css/targets/html/salem/_rs-widen.scss 78:16  @use
  - 84:1                                                                   root stylesheet [plugin sass-plugin]

1 error
/home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:1477
  let error = new Error(text);
              ^

Error: Build failed with 1 error:
error: Undefined variable.
   ╷
78 │     max-width: $content-width;
   │                ^^^^^^^^^^^^^^
   ╵
  /home/rob/mathbook/mathbook/css/targets/html/salem/_rs-widen.scss 78:16  @use
  - 84:1                                                                   root stylesheet
    at failureErrorWithLog (/home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:1477:15)
    at /home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:946:25
    at runOnEndCallbacks (/home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:1317:45)
    at buildResponseToResult (/home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:944:7)
    at /home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:956:9
    at new Promise (<anonymous>)
    at requestCallbacks.on-end (/home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:955:54)
    at handleRequest (/home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:648:17)
    at handleIncomingPacket (/home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:673:7)
    at Socket.readFromStdout (/home/rob/mathbook/mathbook/script/cssbuilder/node_modules/esbuild/lib/main.js:601:7) {
  errors: [Getter/Setter],
  warnings: [Getter/Setter]
}

Node.js v24.3.0

  File "/home/rob/mathbook/mathbook/pretext/pretext", line 879, in <module>
    main()
  File "/home/rob/mathbook/mathbook/pretext/pretext", line 740, in main
    ptx.html(
  File "/home/rob/mathbook/mathbook/pretext/lib/pretext.py", line 4237, in html
    build_or_copy_theme(xml, pub_vars, tmp_dir)
  File "/home/rob/mathbook/mathbook/pretext/lib/pretext.py", line 4135, in build_or_copy_theme
    _build_custom_theme(xml, theme_name, theme_opts, tmp_dir)
  File "/home/rob/mathbook/mathbook/pretext/lib/pretext.py", line 4096, in _build_custom_theme
    raise e
  File "/home/rob/mathbook/mathbook/pretext/lib/pretext.py", line 4093, in _build_custom_theme
    raise Exception("Failed to build custom theme")
Exception: Failed to build custom theme
```

rbeezer:
> current conversion needs an implementation of the "code-wrapper" template

I'd noticed that, and figured it was my problem,

> maybe a passthrough version

Well, that would be sweeping the problem under the carpet.  That message is intended exactly for developers, and I now realize is doing its job very well.  Took me a while to understand.

The new `checkpoint` and its `prompt` and `feedback` are "literary" content, and more specifically have `c` elements within your one example.  So back we go again.  You do need `pretext-html.xsl` I believe to accurately process the `checkpoint` into HTML as part of the trace file (untested, but hightly confident).  That is what the error is telling you, and I will guess that it goes away if you make this change (and your `c` content will re-emerge, wasn't it going missing?).

But more importantly, and I guess you won't be thrilled, we have "new" content - questions in the `checkpoint`.  I'm not seeing how this is being handled for static output formats.

ascholerChemeketa:
The checkpoint questions are not about the program. They are about the running state of the program at a particular instance in time. *"What is the value of `i`?"* where `i` appears in multiple scopes, or is is inside a loop, or is dependent on the parameters to a function call, etc... is almost a kaon. Without the context of a particular execution step, it is meaningless. They are a feature of the interactive and make no sense outside the interactive presentation.



Maybe CodeLens and ActiveCodes should get the QR code treatment? I think the current presentation (to basically pretend they don't exist in static) is likely to lead to confusing static texts. 

In books that have been designed with interaction in mind, ActiveCodes are not just a "here is the program, oh and you can try running it if you like". Instead, they are a tool to make the presentation of the material fundamentally interactive: *"Run this program and examine the compiler error..."*, *"Now change  the function call's parameter to `(int& x)` and run the program again"*, ... At best any construct like that feels weird in a static version, at worst they cease to make much sense.

A really determined author could use versions to make "interactive" and "static" presentations of part of the text. But absent that effort, I feel like we should be letting the reader know that they are missing out on something that is designed to be interacted with.

ascholerChemeketa:
You are correct about `pretext-html`. About to do a force push with it back in and trace file update.

rbeezer:
> They are a feature of the interactive 

Yes, I understand 100% about the purpose and interactive vs. static.  (More than you will ever know!  ;-)  )

> Maybe CodeLens and ActiveCodes should get the QR code treatment?

Yes, this thought crossed my mind.  It'd be a big change, I guess.  But perhaps not?

PR #2514, opened 2025-05-16
FITB: input width changed to size

ascholerChemeketa:
The attribute `width` on an HTML text input does not do anything. It should be `size`.


rbeezer:
Thanks!  Took the opportunity to silently spread the attribute construction across three lines.

Pinging @dbrianwalton so he is aware.

PR #2517, opened 2025-05-19
FITB: Sanitize static auto-generated solution for fill-in-the-blank

### LINKED ISSUE #2516: FITB: Automatic solution is not sanitized
When a fill-in-the-blank exercise is authored with a `statement` containing labeled/xml:id'd children, the automatically generated `solution` for that exercise duplicates the entire statement resulting in duplicate xml:id errors.

See this conversation:
https://groups.google.com/g/pretext-support/c/34E4PpRSlhA/m/EHGQ1rFLAAAJ

The error was first identified when images were included in a problem.

I believe that the fault occurs in xsl/pretext-runestone-static.xsl lines 1040-1060. This block of code duplicates the original `statement` replacing all of the `fillin` elements with their correct answers. It was implemented with a recursive copy of the statement xml hierarchy, and only the `fillin` is overridden with a change.

To preserve the original intent of the automatic solution, what is the best alternative?
- Copy only text-based elements (e.g., `p` and `ul` and `ol`) and skip any other children
- Copy only paragraphs that *contain* fill-in-blanks that require substitution, skipping all other elements (including paragraphs) that involve no substitution

dbrianwalton:
Clear xml:id and label from copied exercise/statement children being used in automatic static solutions.

Resolves #2516

### DIFF
--- a/xsl/pretext-runestone-static.xsl
+++ b/xsl/pretext-runestone-static.xsl
@@ -983,6 +983,18 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     </xsl:copy>
 </xsl:template>
 
+<!-- Don't duplicate xml:id. Nobody should point into the solution. -->
+<!-- Note that any internal link references will point to the original not the copy -->
+<xsl:template match="@xml:id" mode="fillin-solution"/>
+
+<!-- Append to a label -->
+<xsl:template match="@label" mode="fillin-solution">
+    <xsl:attribute name="label">
+        <xsl:value-of select="."/>
+        <xsl:text>-fitb-solution</xsl:text>
+    </xsl:attribute>
+</xsl:template>
+
 <xsl:template match="statement" mode="fillin-solution">
     <xsl:variable name="exercise" select=".."/>
     <solution>


rbeezer:
Oh man, this is complicated!  Your PR is great, but building an example to stress it opened up a can of worms.  See new thread I opened on `-dev`.  (Sorry!)

rbeezer:
Where are we on this one?  I went back and read the 2025-05-19 posts on `pretext-dev`.  Can't say I have internalized everything, but it looked like it had progress towards some changes here.

rbeezer:
I asked Cluse to review this, which as not necessarily helped **me** that much.  But perhaps, @dbrianwalton, there will be something to chew on.  Comments welcome!

rbeezer:
I reviewed this PR on Rob's behalf, alongside #2516 and the recent [pretext-support thread](https://groups.google.com/g/pretext-support/c/34E4PpRSlhA).

**As a fix for the literal #2516 report** — duplicated `statement` → duplicate `xml:id` → hard error — this is correct and minimal. Dropping `@xml:id` on the copy removes the error, and the `@label` suffix keeps a duplicated `latex-image` filename distinct, which is why the TikZ case extracts cleanly.

**Two problems remain that I think are the actual issue:**

1. *Silent numbering drift.* Removing `@xml:id` eliminates the error, but the xeroxed statement still carries numbered content (figure-like elements, numbered `md`/`men`). That content still gets numbered, so a static build renders one more numbered item than the dynamic build, and every number downstream diverges between formats — now silently rather than as a crash. That's arguably a worse failure mode than the one being fixed.

2. *The `@label` half is order-dependent.* In assembly, `representations` (where `fillin-solution` runs) is at `pretext-assembly.xsl:663`, but `@label`-from-`@xml:id` derivation is the later `labels` pass (~line 693). So the suffix only fires for an *authored* `@label`. An image carrying only `@xml:id` ends up on the copy with neither attribute, and its extracted filename falls back to an internal id — distinct, but not the stable `-fitb-solution` behavior the comment describes.

**Where this seems to point.** Brian's own suggestion in the thread — don't reproduce the whole statement; let the author write a `solution`/`hint` with `<eval>`/`fillin` placeholders next to the blanks, or auto-generate a plain `answer` for the simple case — avoids all of the above: no duplication, no numbering drift, no label collision. On the question of whether a FITB ever *generates* a numbered item: it doesn't. A bare substitution is only ever a number or short expression; the numbered items appear solely because the entire statement is copied. Remove the copy and the question is moot.

So, two paths:
- **Stopgap:** keep the `@xml:id` drop to stop the crash, but add an `xsl:message` when the duplicated statement contains numbered content so the drift isn't silent — and drop or rework the `@label` half given the ordering above.
- **Preferred:** redirect toward the `<eval>` / generated-`answer` design. `<eval>` isn't in the schema yet, so that's a prerequisite.

Brian — does the no-duplication direction match what you had in mind, and is there a real case where an author needs the full statement reproduced rather than just the blanks-with-answers? Worth settling that before more preprocessor code gets written.

dbrianwalton:
Rob, yes this does match what I had in mind.

If we can avoid duplicating the statement, we make things cleaner. Auto-generated `answer` that is not just filling in the blanks would be useful. Early decisions were concerned that just listing the answers without the context were problematic, and that is where the duplication came from. I don't think it is worth the cost we see now.

Note, I'm not entirely sure what Claude's comment about `<eval>/fillin` placeholders next to the blanks refers to without it somehow implicitly considering still duplicating the statement. The `solution` wouldn't have blanks.

rbeezer:
Thanks for taking a fresh look.

Well, if Claude seems confused, maybe it is.  I see this from time to time, and once questioned, I sometimes get a really different take.  Sorry for the noise.  The great insights usually far outweigh the junk.

Should I merge this for now, and wait on a further PR to go to a more basic #answer without a recycling of the #statement?  Or go in another direction?

dbrianwalton:
Note that it is the #solution that is being generated which creates the issue, not answers.

The PR only kills any xml:id's in the copy and appends something to any labels. I haven't been able to pay close enough attention to the last year to know if there are other issues, such as all the work you've been doing with numbering. The best practice would definitely be for the author to write a custom solution.

But I think merging now would be good. There is no obvious path to a general strategy for a solution other than just listing the answers — but that wouldn't be a solution.

rbeezer:
> Note that it is the #solution that is being generated which creates the issue, not answers.

Right.  Too many balls in the air right now.  ;-)  Thanks for the attention to this one.

Long-term, let's think about making sure "numbered stuff" doesn't leak into duplicates and throw off overall numbering.  Allow an authored #solution (for which we will eventually control content via the wschema), and let #answer be answers.

Merged as-is, with just a shorter commit message.  I'll close #2516 if it is not automatic.  Thanks!

PR #2519, opened 2025-05-19
Switch xml serializing to Lenz templates

ascholerChemeketa:
Requested separate PR with Lenz serialize routines.

Looks to give a clean diffs other than some spaces at end of tags and the fact that his templates do not escape `>`

Not sure about the `@href` seralize in pretext-jupyter. No idea how to test that.


rbeezer:
Been musing on this one.  I temporarily added a step/variable (right near the end) to escape greater-than.  I want to believe that is not necessary, but I'd also like to finish tests with minimal changes (so I don't miss anything).  I'd get rid of some spaces in closing tags in source to help with that.

I can't see any harm, except perhaps for efficiency.  We could try removing it later, but I'm not sure I would be any more comfortable then.  :-(

rbeezer:
OK, switch is good now.  Lots of testing: HTML, annotated HTML, sample book, sample article (and for Runestone), Jupyter.  Thanks for breaking out this big change, it needed some attention.

*  Added Lenz' verbatim as first commit, with no changes, so we can track **our** changes.
*  Introduced escaped "greater-than" addition, under my name.
*  No need to import several times, once in "text utilities" it is in "-common" then widely available to conversions.
*  Ditched `as-authored-source` since suspect attributes are no longer coming through, since we operate on an earlier tree from the pre-processor for View Source versions.
*  Namespaces behaving once template is initiated at depth 1.
*  Jupyter `@href` is doing no harm, so I left it alone.
*  Now have source comments in annotations!
*  For annotated source: a few bits of extra insignificant space, some reorderings, a few quote marks got escaped.  Nothing major or concerning, and likely improvements.

One big change: `@include-source` is coming through when used on a `program`, where the source is displayed.  I don't think the `xml-to-string` mode is the right place to deal with it.  (I like that we have made only one change to Lenz' routines.)  And I guess I'm not clear on the exact purpose of `@include-source`.

ascholerChemeketa:
All sounds great. Thanks for picking up the ball and running with it.

Leaving `@include-source` is fine. That is the attribute we added to programs to allow author indicated annotated output. Rational was getting XML source for program samples into Sample Book (or other reference works) so authors focused on Runestone can see the source of what they are looking at without having to grep in rune.xml for a likely string.

PR #2520, opened 2025-05-20
Title colon cleanup

oscarlevin:
As discussed on -dev (https://groups.google.com/g/pretext-dev/c/oUMo42G3Vq0), this removes the colon separating a title from a subtitle (in the three places in the `pretext-html`, and puts it back as CSS just in the masthead.  Expected results are that nothing looks any different in the masthead, but the titlepage, which has the subtitle on the next line, does not have a colon following it.  

Tested successfully both with and without a subtitle.


rbeezer:
Even some whitespace corrections!  ;-)  Looks good.  Merged.

Issue #2524, opened 2025-05-21
asyncio.get_event_loop is deprecated - needs to be relaced


ascholerChemeketa:
`asyncio.get_event_loop` is deprecated, so at some point we need to figure out a replacement.

It also chokes if run from an async context. RS builds books in an async context - fortunately, all uses of `get_event_loop` are in static build pathways RS does not need. But if use expands to dynamic builds, there will be an issue.
 
Official advice appears to be "rewrite anything that uses async to be async all the way up:
https://discuss.python.org/t/calling-coroutines-from-sync-code-2/24093/5

ascholerChemeketa:
`nest_asyncio` can be used to patch this. We did so temporarily in RS:
https://github.com/RunestoneInteractive/rs/pull/723

BUT... that package is archived.

rbeezer:
Closing as done. The deprecated `asyncio.get_event_loop().run_until_complete(...)` calls were replaced with `asyncio.run(...)` — in the MyOpenMath PDF conversion (92088444) and the interactive-preview image generation (e51cca6b), both on 2026-07-13. No `get_event_loop()` calls remain in the build pathways. (A single cosmetic `# get_event_loop()` comment lingers on one import; it will be tidied in routine housekeeping.) Thanks @ascholerChemeketa for the `nest_asyncio` pointer along the way.

*Claude Opus 4.8 (1M context), acting as a review assistant for Rob Beezer*

PR #2525, opened 2025-05-21
Script: add warning comment about possible future asyncio issue

ascholerChemeketa:
@bnmnetp and I are working through an issue with some RS python code that gets called from both synchronous and asynchronous contexts. Running in an async context causes `asyncio.get_event_loop` to blow up. (By design apparently.)

This adds a note to pretext.py that `asyncio.get_event_loop` is a potential timebomb. If `asyncio.get_event_loop` ever gets into dynamic build pathway, it is going to blow up builds done through RS. It is also deprecated, but the suggested replacements do not solve the core issue that Python maintainers really do not support the idea of functions that try to use asyncio from a context that is already async.

@oscarlevin No need for action, but you should be aware of this one. If CLI ever tries to call pretext.py functions from an async context it is not going to work. 

Added issue:
https://github.com/PreTeXtBook/pretext/issues/2524


rbeezer:
We use `asyncio.get_event_loop()` **three** places.  Should they all get a warning?  If so, I can silently duplicate the one provided here.

ascholerChemeketa:
Maybe? I will update and force push based on that and lessons learned in RS in the last 12 hours.

Incidentally, when you use runestone to do a build, we are hitting these. At least some times. It appears default behavior is build all the possible things, even if they aren't needed for a web build. I'm seeing messages that 1+ of these async calls aren't being awaited. Which is OK since we don't care about their output. Not ideal though.

The officially recommended fix would be to make everything in the stack that uses these async. We tried using `nested_asyncio` in RS but it broke something else. That said it is used indirectly in libraries RS uses and seems OK there. Which is why I suppose it isn't the officially sanctioned fix... it does some brittle monkeypatching of ascyncio.

ascholerChemeketa:
Force pushed with changes. Now each call has a note about what to search for if the line blows up.

rbeezer:
Thanks.  I guess I understand the general nature of the problem here.  But maybe I'll pick your brain at Drop-In at some point.  Thanks for anticipating the future!

PR #2527, opened 2025-05-22
TOC customizable icons

gigamonkey:
This is a variation of my earlier PR that allows some customization of the TOC icons via CSS. It is limited to the icons that can be specified in the content of a `<span class="icon material-symbols-outlined" aria-hidden="true"></span>` but that at least accounts for the chevrons of before and the +/- that I'd like to see.

Note. I haven't been able to figure out how to do a full end-to-end test of building the whole pretext CLI, etc. to test these changes but I've tested the JS and CSS by hacking it in my built book. So this should work but apologies if I've misunderstood something about how the scss works.

### DIFF
--- a/css/components/page-parts/_toc-basics.scss
+++ b/css/components/page-parts/_toc-basics.scss
@@ -101,6 +101,16 @@ $nav-height: 36px !default;
     text-align: left;
     flex-grow: 0;
   }
+
+
+  /* ---------------icon customization----------------- */
+  .toc-item > .toc-title-box > .toc-expander > .icon:before {
+    content: "add";
+  }
+
+  .toc-item.expanded > .toc-title-box > .toc-expander > .icon:before {
+    content: "remove";
+  }
 }
 
 
--- a/css/components/page-parts/_toc-basics.scss
+++ b/css/components/page-parts/_toc-basics.scss
@@ -259,7 +259,7 @@ and then clear indent if there is a codenumber */
     align-items: center;
 
     .icon {
-      font-size: 30px;
+      font-size: 1.25em;
       line-height: 18px;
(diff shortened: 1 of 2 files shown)


ascholerChemeketa:
Works great on a full build.

The `toc-expander` class is only used within a `focused` toc. So the style rules might make more sense nested under the `.ptx-toc.focused {...` block instead of the more general `.ptx-toc {...` block. But they work correctly where they are.

@gigamonkey My personal recommendation is to develop without the CLI when doing stuff like this. Just use the pretext script directly:
https://pretextbook.org/doc/guide/html/pretext-executing.html

If you have the pretext repository and are in `pretext/examples/sample-book` you can do something like:
```
python ../../pretext/pretext -vvv -f html -c all -d out -p publication-structural.xml sample-book.xml
```

Including the CLI in the process is necessary if you are testing what it does to manage builds. But for builds themselves you don't need it. Going directly to the script requires a few more command line args, but gives you more control and one less piece to worry about.

rbeezer:
Thanks, @ascholerChemeketa for the review.

@gigamonkey: I need to update my preferences for contributions.  I thhink what you saw was to discourage people from updating on their own after I'd already startd a review.  ;-)

Looks to me like this changes **all** themes.  Maybe we/stylists want a mix?  Give me a 'sec to post on `-dev` and lets discuss there.

gigamonkey:
As I mentioned on -dev, I think we need to have some icons by default and then themes should be able to override them? I don't presume to have an opinion about what would be best for all PreText themes but would like to get +/- for my book.

rbeezer:
Merged, notes on #2530.  Thanks for initiating this and staying on top of it.

PR #2529, opened 2025-05-22
Documentation update relating to Interactive Exercises

dbrianwalton:
Three small updates to documentation:

- sample-article: Add an example for Dynamic Fill-In Exercises illustrating the use of a javascript test using jscmp to facilitate checking multiple blanks so that the order a student answers does not matter.
- sample-article: Correct grammar in a different fill-in exercise that was missing a verb
- guide: Spelling error correction from "thatn" to "than"


rbeezer:
Thanks for the maintenance, and the new example.  Rebuilding website now and will advertise new example.

PR #2530, opened 2025-05-23
ToC expander styling differentiation

ascholerChemeketa:
Builds on https://github.com/PreTeXtBook/pretext/pull/2527 to restore chevrons to non-Salem themes.

Adds notes to the CSS README about options for projects like this.

First commit should just be squashed into Peter's and credited to him. The expander rules are only needed in the "focused" section.


gigamonkey:
Thanks @ascholerChemeketa !

gigamonkey (commented):
I wonder if this is why Rob isn't seeing the chevrons in the other themes? Because without the `.focused` these rules are less specific than the ones that set the +/-

rbeezer:
Thanks, @gigamonkey.

> Should this be .ptx-toc.focused ?

That seems to be a fix! 

@ascholerChemeketa: I can silently meld this in if you just want to check-off on it.

ascholerChemeketa:
Sorry about that. I swear it was working... 

Yes, @gigamonkey's fix looks good, please swap it in.

PR #2532, opened 2025-05-26
Runestone: change placement of True or False in static

sean-fitzpatrick:
A simple swap of two lines of code so that the "True or False?" prompt comes first in static formats.

In PDF, here is the "before" picture:
image

And here is the "after" picture:
image


rbeezer:
Merged, as-is.  Thanks!

PR #2534, opened 2025-05-27
HTML: Refactor permalinks

oscarlevin:
This cleans up permalinks by moving their production to build time rather than using javascript to insert them.  Should make page loads faster, and we have finer control over what gets a permalink.  Took the opportunity to also clean up the CSS a bit (in particular, I want some bonus points since permalinks for figures now show up just to the left of the figure caption).

To test this, I recommend first checking out the first commit and building.  You will see two sets of permalinks: the old ones with the slanted link chain, and new ones with google's material UI horizontal link chain.  That way you can verify that everything that was getting a permalink still is.

In the second commit, I revert the link icon for the new permalinks to the unicode sideways link.  I suggest squashing this commit in with the first one.

Next two commits are for the JS and CSS respectively.  Javascript is much simpler, and we use event listeners rather than hard coding `onclick` into the html.  

As for the XSL, I had to do some Inclusion/Exclusion to get permalinks on what I want but not on the first paragraph of named blocks (which get their own permalink).  Added an entity to handle this.  Also added to pretext-common to explicitly make the `<page>` element non-numbered since this was throwing errors.

### DIFF
--- a/css/components/elements/_description-lists.scss
+++ b/css/components/elements/_description-lists.scss
@@ -18,7 +18,7 @@ dl:is(.description-list, .glossary) {
   margin-top: 1em;
   margin-left: 0;
   margin-bottom: 0;
-  overflow: hidden;
+  //overflow: hidden; // Removed to show permalinks
 
   dt {
     font-weight: bold;
--- a/css/components/elements/_permalinks.scss
+++ b/css/components/elements/_permalinks.scss
@@ -1,11 +1,20 @@
 // TODO - refactor
 $opacity: 0.0 !default;
 
-/* so that we can position things (like .autopermalink) absolutely wrt these items */
+/*
+So that we can position things (like .autopermalink) absolutely wrt these items.
+Any element whose permalink should be positioned directly to the left of it
+should be in this list.
+*/
 section,
 article,
+li,
+blockquote,
+figcaption,
+dt,
 .exercisegroup,
 .discussion-like,
+.solution-like,
 .para {
   position: relative;
 }
@@ -22,9 +31,8 @@ article,
   margin-top: 0 !important;
 }
 
-li > .para > .autopermalink {
-  left: -3.4em;
-  top: 0;
+li > .autopermalink {
+  left: -3.5em;
 }
 
 .autopermalink > * {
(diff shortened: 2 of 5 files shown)


rbeezer:
Trial run looks good, but for one thing.  Build times are doubling for me.  :-(  And of course, it is the whole first paragraph thing.

The problem is "run-in headings" that combine a title/heading into the first paragraph.  No place to put the paragraph's permalink clickable.

First thought was that CSS is controlling the run-in look, so maybe similar rules could hide the permalink?

Which made me think.  Could you kill permalinks more directly for run-in situations?

```xsl
<xsl:template match="theorem/p[1]|lemma/p[1]|...|remark/p[1]|..." mode="permalink"/>
```
Laborious, but I suspect it would be much faster, and you still are expressing the parent relationship.

I swear there is some way to make this more compact, like
```xsl
(theorem|lemma|...|remark|...)/p[1]
```
but I've nver been able to get it right.

oscarlevin:
Interesting.  So `<xsl:template match="theorem/p[1]"/>` is significantly faster than `<xsl:template match="p[1] [parent::theorem]" />`?  Just trying to understand the logic.  But I'm happy to try one of these fixes.

Alex-Jordan:
> Just trying to understand the logic.

I think the full story about things like this is XSLT-processor dependent. But it seems to be a good rule of thumb that predicates are slower than paths. The path is incorporated in how the looping through children is executed, letting lots of nodes be skipped entirely from consideration. But a predicate is considered after the fact, and rather than be skipped, lots of nodes will be tested (and the predicate will evaluate to false for many of them).

It's secondary to the above consideration, but when working with multiple predicates, consider if `[A][B]`, `[B][A]`, or `[A and B]` might be best.  For example I wonder if `match="p[parent::theorem][1]"` would be faster than `match="p[1][parent::theorem]"` by having a more discriminatory filter applied earlier. 

> (theorem|lemma|...|remark|...)/p[1]

A match pattern is like `theorem[...]/p[...]`. It's a path but can have predicates at any level in the path. Note with this definition, no union symbols. Then you can union multiple match patterns. But you can't union match patterns and then use that to create a conceptually more complex match pattern. At least for XSLT 1.0, I think we're stuck with `|` being the last operator in the order of operations.

Alex-Jordan:
> At least for XSLT 1.0, I think we're stuck with | being the last operator in the order of operations.

Like, you can't even do `match="(theorem|remark)"`. The parentheses exterior to a union kill it.

Alex-Jordan:
This might go in a different direction, but I was looking over the XSLT here and I think you could leave out `INNER-STRUCTURAL-FILTER`. And instead use:

```
match="*[not(
                &STRUCTURAL-FILTER;
                or self::sidebyside
                or self::stack
                or self::headnote
                or self::introduction[&INNER-STRUCTURAL-FILTER;]
                or self::conclusion or self::page
             )]/p[1]"
```
Apologies if I'm too far into beer time to get that right. I did not test, nor do I want to make a prediction about speed.

oscarlevin:
After trying a few different directions, I think the best bet is to generate all these "extra" permalinks and then remove them with CSS.  That would allow a theme that doesn't have run-in titles to re-enable them if needed.

Cleaned up the commits to do this.  Should be ready to go.  And speed is back down to "normal".

rbeezer:
> And speed is back down to "normal".

You mean speed is back to **wicked fast**.  ;-)

More in just a bit.

rbeezer:
*  Working well!
*  Yes, run-in headings obscuring a margin-permalink is a style choice, so I think it makes a lot of sense to have the **logical** permalink to the first paragraph remain, and then the style needs to deal with it (hide it).
*  CSS is bleeding into the "worksheet printing" CSS.  Know anything about that?  ;-)
*  Love all the JS going away.
*  Maximum (additional!) bonus points for getting the link onto the caption of a figure.  Use them wisely.

> More in just a bit.

Well, not so fast, it seems.  Down a deep rabbit-hole, but I think I found the White Rabbit at the bottom.

I'm getting multiple warnings:
```
PTX:WARNING : * PTX:WARNING: could not translate string with id "p" into language for code ""
```
Did you not get these also?

A `p` is not able to see its way clear to know what language applies so it can for "Paragraph" for tooltip-text.  Its ancestors are
```
p/feedback/test/evaluate
```
Then the trail mysteriously goes cold.  Next should be `evaluation`.  FITB, which I will take up with @dbrianwalton  on -dev right now.

But maybe we should kill permalink for "p" with an "evaluation" ancestor for now?  Or sit tight for -dev and experiment, not in that order.

rbeezer:
Sounds like the same problem.  Can you put some of the above on the -dev thread I just started.

rbeezer:
Merged.  But prematurely.

Forgot about worksheet-print CSS.  Well, forgot to update themes at all, then I saw it again.

Plus - I don't think we want these in EPUB, Kindle, or Jupyter (no, or limited, Javascript) so they are just going to be sitting there?  Should be easy to kill in those two templates.  Extra commit has my name on it, so we know who to **blame**.

Thanks - this is great.  I'm guessing a bit of fine-tuning, like in the past, as to location, but it should not be as widespread.

I'm going to do a consolidated "clickable" announcement with #2535, once that is done.

rbeezer:
>  Is that something that still needs to be done? 

Yes, didn't get to it.  (Bibliographies instead!)

Check that they come through unwanted.  Then I think a simple killing of any use of `mode="permalink"` will do it.  Check that they are gone?  Of course Kindle doesn't need it, just EPUB and Jupyter.  Oh, slideshows!  How could I forget?  Thaks.

oscarlevin:
That is a much better use of your time, IHO. 

I'm on it.  Oh, is epub really only available for books?  I got loads of chunking errors when I tried to build the sample article to epub.

PR #2535, opened 2025-05-28
Add copy-element functionality for program code chunks

skiadas:
This PR adds a copy-to-clipboard button to the top right of program code chunks, similar to the functionality on GitHub Markdown code chunks. Still to do:

- [x] Check how this behaves with activecode and codelens. They should not be affected at all as they are not using class="code-box"
- [x] Should this also work for `<cd>` elements? They only produce a pre element without a surrounding tag, so adding a button would be challenging at the moment, unless we modify it.
- [x] Test different themes
- [x] Test side-by-sides (looks like button not present at all)
- [x] Should this be toggleable somehow at various levels? Or just always on?
- [x] While this sort-of works for console, it would need some adjustment. And there is also the question of what text should be copied in the console case. Currently everything, including prompts etc as well as all of the input and output, is copied.
- [x] The button visually disappears when `@line-numbers` is used

### DIFF
--- a/css/components/chunks/_codelike.scss
+++ b/css/components/chunks/_codelike.scss
@@ -50,5 +50,58 @@
   width: 3.0em;
 }
 
+.clipboardable {
+  position: relative;  // To anchor the code-copy button
+  display: flow-root;
 
+  .console + .code-copy {
+    top: 1em;
+  }
 
+  .program + .code-copy {
+    top: 0.5em;
+  }
+
+  .code-copy {
+    position: absolute;
+    right: 0;
+    top: 0;
+    opacity: 0.25;
+    border-width: 0;
+    background: none;
+    cursor: pointer;
+    z-index: 1;
+    white-space-collapse: collapse;
+    padding: 1px;
+    scale: 0.85;
+
+    &:hover {
+      opacity: 0.7;
+    }
+
+    .checkmark {
+      display: none;
+    }
+
+    &.copied {
+      opacity: 1;
+      color: mediumseagreen;
+      cursor: not-allowed;
+      pointer-events: none;
+
+      .copyicon {
+        display: none;
+      }
+
+      .checkmark {
+        display: inline;
+      }
+    }
+  }
+}
+
+@media print {
+  .code-copy {
+    display: none;
+  }
+}
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -1087,3 +1087,33 @@ window.addEventListener("DOMContentLoaded", function(event) {
         }
     }
 });
+
+// START Support for code-copy button functionality
+document.addEventListener("click", (ev) => {
+    const codeBox = ev.target.closest(".clipboardable");
+    if (!navigator.clipboard || !codeBox) return;
+    const button = ev.target.closest(".code-copy");
+    const preContent = codeBox.querySelector("pre").textContent;
+    navigator.clipboard.writeText(preContent);
+    button.classList.toggle("copied")
+    setTimeout(() => button.classList.toggle("copied"), 1000);
+});
+
+document.addEventListener("DOMContentLoaded", () => {
+    const elements = document.querySelectorAll(".clipboardable");
+    for (el of elements) {
+        const div = document.createElement("div");
+        div.classList.add("clipboardable");
+        el.classList.remove("clipboardable");
+        el.replaceWith(div);
+        div.insertAdjacentElement("afterbegin", el);
+        div.insertAdjacentHTML("beforeend", `
+    <button class="code-copy" title="Copy code" role="button" aria-label="Copy code" >
+        <span class="copyicon material-symbols-outlined">content_copy</span>
+        <span class="checkmark material-symbols-outlined">check</span>
+    </button>
+            `.trim());
+    }
+});
+// END Support for code-copy button functionality
+
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -8709,7 +8709,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:param name="b-original" select="true()" />
     <xsl:element name="pre">
         <xsl:attribute name="class">
-            <xsl:text>code-display tex2jax_ignore</xsl:text>
+            <xsl:text>code-display tex2jax_ignore clipboardable</xsl:text>
         </xsl:attribute>
         <xsl:choose>
             <xsl:when test="not(@showspaces) or (@showspaces = 'none')">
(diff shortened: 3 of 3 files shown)


skiadas:
Added support for cd elements. @ascholerChemeketa I wrapped cd in a div and moved the code-display tag to the div instead. Thoughts? Where might this run into problems style-wise?

rbeezer:
> I'll stop messing with it and wait to hear from you.

OK, neck-deep in bibliographies, but will have a look real soon.  Maybe even now.

skiadas:
I scaled it to 0.85, seemed to work ok for the oneliner on my screen.

ascholerChemeketa:
Good call re: disable on print.

New position has added benefit of tabbing through content hitting the code container before advancing to the copy button. Feels like a more logical progression.

I see an issue with scrolling - the icon remains fixed relative to the content, not the scrolling container.

I think position sticky and a calculated left (something like calc(100% - 30px);) might do the trick.

Testing sample for issue:
```
    <figure>
      <caption>A <tag>program</tag> and a <tag>console</tag></caption>
      <sidebyside margins="5%"
                  widths="40% 45%">
        <program language="r">
                <code>
                n_loops &lt;- 10
                x.means &lt;- numeric(n_loops)
                for (i in 1:n_loops){
                    x &lt;- as.integer(runif(100, 1, 7))
                    x.means[i] &lt;- mean(x)
                }
                x.means
                </code>
            </program>
        <console>
          <input prompt="pi@rpi ~$ ">gcc -o intAndFloat intAndFloat.c</input>
          <input prompt="pi@rpi ~$ ">./intAndFloat</input>
          <output>
                19088743 (integer) and 19088.742188 (float)
                </output>
          <input prompt="pi@rpi ~$ "/>
        </console>
      </sidebyside>
    </figure>

    <figure>
      <caption>A <tag>program</tag> and a <tag>cd</tag></caption>
      <sidebyside margins="5%"
                  widths="40% 45%">
        <program language="r">
              <code>
              n_loops &lt;- 10
              x.means &lt;- numeric(n_loops)
              for (i in 1:n_loops){
                  x &lt;- as.integer(runif(100, 1, 7))
                  x.means[i] &lt;- mean(x)
              }
              x.means
              </code>
          </program>
        <p>
          <cd> n_loops &lt;- 10 x.means &lt;- numeric(n_loops) for (i in 1:n_loops){ x &lt;-
            as.integer(runif(100, 1, 7)) x.means[i] &lt;- mean(x) } x.means </cd>
        </p>
      </sidebyside>
    </figure>

    <program>
        Ipsum dolor sit amet, consectetur adipiscing elit. Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
        Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
        Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
    </program>

    <console>
      <input>
      Ipsum dolor sit amet, consectetur adipiscing elit. Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
      Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
      Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
    </input>
    </console>

    <p>
      <cd> Ipsum dolor sit amet, consectetur adipiscing elit. Sed do eiusmod tempor incididunt ut
        labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco
        laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in
        voluptate velit esse cillum dolore eu fugiat nulla pariatur. </cd>
    </p>
```

skiadas:
I could not get it to work with the button within the pre element, so I switched to introducing a div wrapper.  But I made it into a separate commit so we can revert it more easily. 

The main problem I had with the wrapper approach was the inconsistent amount of top margin applied to code elements, and I am not sure there is a good reason for this:

- no margin on cd
- 0.5em margin on programs from _prism
- 1em margin on console from user stylesheet, which I am very uncomfortable about.
- on top of that, somewhat inconsistent behavior on how the relative placing happens based on whether the parent element is displayed flex or block:  sbspanels are flex, code-box is implicitly block. But maybe there is something else I am misunderstanding there.

So one proposal, implemented in the new commit I added, is to set the code-box to also display flex, to echo what sbspanels do. I didn't see any immediate problems with that but maybe I am missing something. Then I had to simply adjust the top based on whether there is a simbling program or sibling console.

The alternative, if we don't want to mess with turning code-box into flex, would be to simply special-case the spacing within sbspanels.

Or maybe there is some magical setting of flex behavior that I am missing, which can help fix the behavior in a simpler way.

skiadas:
Here is a bit of the situation I was running in, sounds like there is either a rendering bug or something I am not understanding very well, first the image:

image

So what i have here is:
- A div (with the red border). It is set to display flex column in some of the cases, and display block otherwise.
- In it there is a paragraph and just below it another div that is relatively positioned (no amounts, just to act as a basis for the absolute positioning of elements inside it). This div has a green border on some of the pictures, and no border otherwise. It is also given a infinitesimal padding of 0.001em on some of the pictures, and no padding otherwise.
- Inside this relatively positioned div there are two elements: a paragraph (purple border), with a margin-top, and a button (with the A in it).
- The button is set to absolute positioning,  with a top of 0 and a right of 1em. Changing the top doesn't change the fundamental disagreement shown in the picture.

Now the weird thing: In almost all cases the button is positioned relative to the green-bordered div, which is its closest relatively positioned ancestor, which is the expected behavior I believe. With one exception: If that ancestor has no border, and has no padding, and is inside a container that is set to display block and not flex. This is the 2nd image on the top row.

How this relates to our case: The 2nd image at the top row is the behavior inside a code-box, while the 1st image at the top row is the behavior inside a sbs.

ascholerChemeketa:
I'd like to check out your demo before opining on the best path. See -dev for request

skiadas:
https://codepen.io/skiadas/pen/raVxxrE

ascholerChemeketa:
SASS and final behavior look good to me!

skiadas:
> SASS and final behavior look good to me!

@rbeezer Let me know if you need anything else for this PR.

rbeezer:
OK, this has been merged as-is.  Back from three weeks of travel over the past six weeks, most of it related to supporting PreTeXt.  I **always** have an eye on the list of PRs and if it is clearly indicated as ready-to-go I **will** eventually get to it.

* The `print-worksheet` CSS seems to be affected, not sure if that is undesirable or not.
* Would we want to give the PreTeXt `pre` element the same treatment?  I think it just means adding a class.
* I think it would be good if we didn't have this new class come through in derived HTML (e.g. EPUB, Jupyter, maybe RevealJS).  There could be a template that adds the class, then it could be neurted in any "non-JS" version of HTML output.  @skiadas - might you want to give that a go and increase your XSL-fu?  I can give more hints during a Drop-In.

skiadas:
Sure, I'll try to stop by a drop-in, probably next week.

Issue #2533, opened 2025-05-27
Fillin exercises no longer show in Runestone assignment page (not in manifest)


bhoffman0:
Fillin exercises in an activity are not showing up in the Runestone assignment page. Brad says that fillins in an exercise do show up in the Runestone manifest but ones in an activity do not. They did use to show a few weeks or months ago though. Here's the markup for one: 
```<activity label="fillDecVar2">
      <statement>
        <p>What type should you use for a shoe size like 8.5?</p>
        <p>
          <fillin answer="double" mode="string" />
        </p>
      </statement>
      <evaluation>
        <evaluate>
          <test>
            <strcmp use-answer="yes" />
            <feedback>
              Correct.  Any variable that needs to values after the decimal point should be declared as a double.
            </feedback>
          </test>
          <test>
            <!-- else -->
            <feedback>What type allows for a decimal value?</feedback>
          </test>
        </evaluate>
      </evaluation>
    </activity>```
This shows up fine on the lesson page, but is not listed in the assignment page as a possible problem because Brad says they did not make it to the runestone-manifest.xml.
For example, 1.2.10 - 1.2.12 are missing here:

![Image](https://github.com/user-attachments/assets/e954fc99-d033-46d1-b2b5-0f3ebdd19a1a)

Is it also possible to show the chapter/lesson numbering in the assignment page?

Thanks, 
Beryl

rbeezer:
Thanks, Beryl, for chasing this down.  @ascholerChemeketa put in a fix, and a bunch of work to make this more reliable in the future.  Over at #2536.  And thanks to @bnmnetp, too.

bnmnetp:
Oscar released a new cli  this morning.  I did a manual build of csawesome2 and I now see the problem above in the runestone-manifest.xml file.

I have not yet rebuilt the author server and will not get to that for a while, hopefully later today.

rbeezer:
Thanks for tne ACK.

> I now see the problem

Took me a while to realize that the "problem" addressed on this PR has been fixed and now you see the `exercise` in the manifest.  ;-)

PR #2536, opened 2025-05-28
Runestone: simplify exercise matching logic for manifest; bugfix for fillin

ascholerChemeketa:
Fillins are not being written to manifest in activities and other non-exercise locations. 

This replaces the matching code for what to write to manifest with something much simpler. We've had to do multiple bugfixes in the previous logic... that tells me it is too brittle.

xsltproc timing shows no real runtime difference between original approach and this less efficient one. The template isn't matched enough to matter.

Verified against sample book manifest, as well as two other programming heavy books. Only diff I got was the expected inclusion of fillins from more locations with new code.

### DIFF
--- a/xsl/pretext-runestone.xsl
+++ b/xsl/pretext-runestone.xsl
@@ -665,137 +665,64 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     </label>
 </xsl:template>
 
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
Fix for https://github.com/PreTeXtBook/pretext/issues/2533

PR #2537, opened 2025-05-28
Add @randomize for cardsort

bnmnetp:
At Andrew's request, I added @randomize to cardsort.  The default is to randomize so if you don't want the premises shuffled set it to 'no'

### DIFF
--- a/examples/sample-book/rune.xml
+++ b/examples/sample-book/rune.xml
@@ -3285,12 +3285,12 @@ TEST_CASE( "Test the add function" ) {
             <title>Matching Problem, Dates</title>
             <idx>matching US dates</idx>
             <statement>
-                <p>Match each event in United States history with the year it happened.</p>
+                <p>Match each event in United States history with the year it happened. Hint: this one isn't randomized.</p>
             </statement>
             <feedback>
                 <p>Review <url href="https://www.britannica.com/list/25-decade-defining-events-in-us-history" visual="www.britannica.com/list/25-decade-defining-events-in-us-history">Encyclopedia Brittania, 25 Decade-Defining Events in U.S. History</url>url.</p>
             </feedback>
-            <cardsort>
+            <cardsort randomize="no">
                 <match>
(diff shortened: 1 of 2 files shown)


bnmnetp:
Hmmm, that is a good question that I hope @ascholerChemeketa will answer since he asked for this.   And I took it as a personal challenge to write XSL. 😉

rbeezer:
I agree that many-many might benefit.  But thus PR is just about `cardsort` and they are marked-up quite differently.

I've implemented specific orders for `premise` and for `response` when making a static version of many-many.  Which might be irrelevant.  But I think some sort of order-preservation (authored order?) for interactive would be a useful feature.  I think static will just go with authored order absent any `@order` attributes.

rbeezer:
`@randomize="premise response" (default)

`@randomize="premise" 

`@randomize="response" 

`@randomize="none"

Or complementary with an attribute like `authored-order` of `fixed-order`?

rbeezer:
Can we restart this discussion on -dev?  With matching questions on a separate thread?

rbeezer:
Oops I mean multiple choice. See I'm  already confused.

rbeezer:
I'm closing this for now.  Re-open if necessary.  We can try again after some discussion on `-dev`.

ascholerChemeketa:
@rbeezer Can I do a quick PR before -dev discussion to clean up the terminology in the sample book examples?

Not helping that both types are labeled "Matching" in the SB. And that one of the examples of a matching problem (not cardsort!) uses playing cards as its domain.

bnmnetp:
I thought @rbeezer was going to make the cards on the left images. :-)

ascholerChemeketa:
Is ask for permission and then not waiting for it better/worse than asking forgiveness instead?

https://github.com/PreTeXtBook/pretext/pull/2538

rbeezer:
> not waiting for it

I've been chasing the White Rabbit.

PR #2541, opened 2025-06-01
WeBWorK: escape dollar signs

Alex-Jordan:
Dollar signs do not need to be escaped when written in PGML. But they can be, and it's harmless.

Meanwhile, dollar signs an author writes will soon (with 2.19 support) show up in a PG file, inside the string used for an image description. They do need to be escaped there as this usage is beyond the scope of PGML. So this tiny PR escapes them now.

There is a diff in PG output from the backslashes that this adds. (But not in LaTeX or PDF.) It will be nice a few PRs later, when 2.19 support really comes in, for the dollar-sign-escaping to already be done and not part of that later large diff.

### DIFF
--- a/xsl/extract-pg.xsl
+++ b/xsl/extract-pg.xsl
@@ -2145,7 +2145,11 @@
 <!--   think this is a bug and white space should not prevent        -->
 <!--   the action.                                                   -->
 <!--                                                                 -->
-<!-- * &, %, $, ^, ~  No need to escape in PGML                      -->
+<!-- * &, %, ^, ~  No need to escape in PGML                         -->
+<!--                                                                 -->
+<!-- * $  No need to escape dollar in PGML, however this needs to be -->
+<!--      escaped when in image descriptions. And harmless to escape -->
+<!--      in PGML.                                                   -->
 <!--                                                                 -->
 <!-- * _  If there are two of these in a line, the parts in          -->
 <!--      between are italicized.                                    -->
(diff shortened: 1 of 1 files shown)


rbeezer:
OK, what do I know.  This looks to me like PERL variables are getting their leading dollar-sign escaped.  Is that what you want?

Before: `\[$NDASH\]\*, \[\_\_\_\]\{$answer\}`
After: `\[\$NDASH\]\*, \[\_\_\_\]\{\$answer\}`

Perefaced with `Brackets, paired, in PGML constructions;` so you can locate it.

rbeezer:
Answering my own question; I see no changes in LaTeX, so it must not be too disruptive.

I do see quite a few images changing (three or four) - which was more than I expected.

Alex-Jordan:
> Before: `\[$NDASH\]\*, \[\_\_\_\]\{$answer\}`
> After:` \[\$NDASH\]\*, \[\_\_\_\]\{\$answer\}`

This is a place where PGML code is being demonstrated as part of the text for the reader to see. So it's not actually a perl variable `$answer`, it's the test string "$answer". And the backslash here is as expected.

I see something I missed before, an actual fix. This is in the diff (unimportant part represented with dots):

```
< <image ...><description>this image has pictures of text with special characters like  %, and @</description></image>
---
> <image ...><description>this image has pictures of text with special characters like $, %, and @</description></image>
```

Before this change, there is a place where perl is seeing `$,` and reading that as a string variable. Of course it is empty. Now it is `\$,` and comes out as it should.

I see 6 images files have changes (6 png images, and the 6 corresponding pdf images). However the tex files that are used to make these are also present and unchanged. And I can't see actual differences when I look in at the images. I think they must all be metadata. These are the full extent of the latex-image images in the sample chapter. The other images are built by PGgraphmacros. So that is consistent with the hypothesis that the differences are just unimportant metadata changes.

Alex-Jordan:
I just rebased this to keep things current. No actual changes.

PR #2542, opened 2025-06-01
WeBWorK: multiple choice repair for 2.19

Alex-Jordan:
After this PR and #2541 are processed, I have a branch ready to submit that covers local PG processing for making representations. (It speeds up representation-making by a lot.) That only works when the local PG repository is 2.19 (or later) though. PG 2.19 sends static PTX back where a few things are slightly differently compared with 2.17, most notably in the multiple choice family of questions. Multiple choice were redesigned somewhere along the way with 2.19.

So for the sake of the local PG pull request that is coming up, this PR adds a bit of the 2.19 support needed so the local PG 2.19 can work. (There will be larger changes on the `webwork2` side when I get to supporting 2.19 all the way.)

I could include these changes with the local PG processing PR, but this feels logically separate. It happens to be needed for local PG processing, but it is really about supporting PG 2.19 in general.

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -2697,12 +2697,12 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <!-- This may only be needed as support for older servers' generated PreTeXt. -->
 <xsl:template match="statement//var[@form = 'checkboxes']/li[(p[. = '?']) or (normalize-space(.) = '?')]" mode="webwork-rep-to-static"/>
 
-<!-- "var/@form" come back from the server as a result of authored -->
+<!-- @form comes back from the server as a result of authored      -->
 <!-- "answer forms" and should be rendered as lists in static      -->
 <!-- representations.                                              -->
 <!-- NB: this does not preclude the match below (scrubbing default -->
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I just rebased this off #2541. Accidentally. Meant to rebase of `master`, but it should be OK if #2541 gets is processed first.

rbeezer:
Working this now...

Alex-Jordan:
Thanks for #2541. I rebased this now off `master` and will lift the draft status.

rbeezer:
Just merged the "draft" version.  Looked ready to go.  Wanna go undraft the other one that **maybe** is ready?  ;-)

rbeezer:
Merged as-is.  No changes I could see in the diffs.  Do we need examples in the sample chapter?  Or is this entirely pre-emptive?

Alex-Jordan:
This is entirely preemptive. The current WW servers are not going to send back `ul[@form]`, `ol[@form]`, or `dl[@form]`. But with the next PR, it'll be using local PG processing, with PG 2.19. So without this, the sample chapter will break (or look incomplete) when you test that PR. I could have included this with that PR, but this has a distinct purpose so I separated it.

Issue #2458, opened 2025-04-10
FITB: unspecified mode silently produces broken exercise


ascholerChemeketa:
From mailing list. Tagging @dbrianwalton 

Currently, if you have a <fillin> that lacks an @mode, a broken JSON blob is generated (preventing loading of the problem by Runestone) and there is no warning.

The code in <xsl:template match="fillin" mode="dynamic-feedback"> only produces the first part of its output if the mode is explicitly set.

Would it make sense to have a default mode (presumably string)? Or to try to infer a type?
If not, some kind of error message should probably be logged.

This is what is produced currently:
...
"feedbackArray": [
  [
    {, "feedback": "Correct!"},
    ...

Where the leading comma is invalid. What should be produced (assuming string was the default and ????? the answer):
...
"feedbackArray": [
  [
    {"regex":"?????", "feedback": "Correct!"},
    ...

PR #2543, opened 2025-06-01
FITB: Refactoring dynamic-feedback and defensive JSON creation

dbrianwalton:
Refactored `mode='dynamic-feedback'` from just matching `#fillin` to reference `#evaluate` so that the feedback evaluation is directly accessed instead of using exsl:node-tree. This allows localization within feedback creation since now the template has access to the language attribute of the original node-tree.

Removed `mode='body'` in template call on feedback element (was missed on an earlier clean-up)

Went through all of the `<xsl:choose>` and ensure that either `<xsl:otherwise>` was present, or not necessary, or flagged an error. This is the underlying cause of #2458. A problem that lacks a mode will now trigger an error, as will using undefined modes or missing or mismatching `#evaluate` elements.

### DIFF
--- a/xsl/pretext-runestone-fitb.xsl
+++ b/xsl/pretext-runestone-fitb.xsl
@@ -1177,7 +1177,7 @@
 
 <xsl:template match="feedback" mode="serialize-feedback">
     <xsl:variable name="feedback-rtf">
-        <xsl:apply-templates select="*" mode="body"/>
+        <xsl:apply-templates select="*" />
     </xsl:variable>
     <!-- serialize HTML as text, then escape as JSON -->
     <xsl:call-template name="escape-json-string">
--- a/xsl/pretext-runestone-fitb.xsl
+++ b/xsl/pretext-runestone-fitb.xsl
@@ -15,17 +15,6 @@
 <!-- Actual rules for substution when generating HTML     -->
 <!-- ==================================================== -->
 
-<!-- These need to be replaced by localization calls.      -->
-<!-- Or maybe push localization strings to Runestone,      -->
-<!-- since this feedback is only useful in an interactive. -->
-<xsl:variable name="defaultCorrectFeedback">
-    <p>Correct!</p>
-</xsl:variable>
-
-<xsl:variable name="defaultIncorrectFeedback">
-    <p>Incorrect.</p>
-</xsl:variable>
-
 <!-- Convert fillin tag to an input element on the page -->
 <xsl:template match="exercise[@exercise-interactive='fillin']//fillin
                      | project[@exercise-interactive='fillin']//fillin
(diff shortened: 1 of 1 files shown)


rbeezer:
Looks good, but I think there is one oversight someplace.  Sample book, HTML output, diff before/after PR. Seems there are two "incorrect" strings, rather than a "correct" and an "incorrect".  Likely I could find it, but I bet you'll go right there.  Please add a **new** commit here, and I'll get in merged into the right place.  Long version for content, then short version.

```json
"problemHtml": "<div class=\"para\" id=\"ex-demo-jscmp-primes-2-1\">What is an example of a prime number less than 20? <input type=\"text
\" id=\"ex-demo-jscmp-primes-blank1\" size=\"5\"\/><div class=\"autopermalink\" data-description=\"Paragraph\"><a href=\"#ex-demo-jscmp-p
rimes-2-1\" title=\"Copy heading and permalink for Paragraph\">🔗<\/a><\/div><\/div>",
"solutionHtml": "",
"blankNames": {"blank1": 0},
"feedbackArray": [[{"solution_code": "function() {\n    var testResults = new Array();\n    testResults[0] = \n                        [2
, 3, 5, 7, 11, 13, 17, 19].includes(Number(ans))\n                        ;\n    return (testResults[0]);\n}()", "feedback": "<div class=
\"para\" id=\"ex-demo-jscmp-primes-3-1-1-2-1\">Any number from the list <span class=\"process-math\">\\(\\{2, 3, 5, 7, 11, 13, 17, 19\\}\
\)<\/span> is a prime number less than 20.<\/div>"}, {"solution_code": "function() {\n    var testResults = new Array();\n    testResults
[0] = \n                        function(){\n                        const val=Number(ans);\n                        if (val <= 0) { retu
rn \"You need to give a positive integer.\" }\n                        if (val >= 20) { return \"The integer must be less than 20.\" }\n 
                       for (let i=2; i<=3; i++) {\n                            if (val % i == 0) { return `Your answer is composite; for 
example, it is divisible by ${i}.` }\n                        }\n                        return false;\n                        }()\n    
                    ;\n    return (testResults[0]);\n}()", "feedback": [-"Correct!"},-]{+"Incorrect."},+} {"feedback": "Incorrect."}]]
}</script></div>
</div></div>
```

See near the end, with diff markers `([-,-]; {+,+}'
```json
"feedback": [-"Correct!"},-]{+"Incorrect."},+} {"feedback": "Incorrect."}
```

rbeezer:
Two comments, while I'm here.  These can wait, do not address them here - I'm ready but for change above.

* Those are very long commit messages!  I will edit.
* Nearby to some changes I saw several:  `xsl:text`, `xsl:value-of`, `xsl:text` all on one line.  I think the code is more readable if they were split onto three lines.  You won't get intervening whitespace into your output, it is ignored.

rbeezer:
One more thing!  Sorry!

* There is a `mode="location-report"` template which you can use after error messages.  Would save you reporting `visible-id`, and may generally be more helpful to authors.  Again, address that later on another PR.

dbrianwalton:
The new version is actually correct here. There was an error that I had fixed but did not realize was present in any of the examples. The `#test` that is being evaluated at that stage for feedback is actually the 2nd test and the javascript function is determining which specific feedback should be given for incorrect responses. If the function is unable to select a specific response, then the `feedback` should be used as generic feedback and should have been `Incorrect`. 

The error was that there had been a test on the variable `$b-correct` that only checked `test="$b-correct"` instead of looking at the actual value `yes` or `no`. It was fixed in commit 6410c7eb when I switched to using localized response strings.

rbeezer:
> The new version is actually correct here. 

Aah!  Good I didn't go looking to fix it myself.  ;-)  Will merge just now.

PR #2545, opened 2025-06-01
PDF: add option to save tex source and asset folders to output directory

oscarlevin:
Adds a flag, `save_tex`, in the `pdf` function that when true will copy the `.tex` and asset folders to the destination directory.  I did not do anything with the `pretext/pretext` script other than fix the signature and apply the default `False` value to the argument.

Possible concerns:
1. To copy the asset directories, I utilized the `copy_managed_directories()` function, which was previously set up to error if the asset directories existed in the target directory.  Since these directories would likely exist for the purpose here, I added `dirs_exist_ok=True`.
2. The tex and asset folders are always copied to the `dest_dir.`  Should this be adjusted to the parent directory of `out_file` if that is not None?
3. Should we always copy the tex source if the latex fails to build?

Oh, I also greatly expanded the docstring for the `pdf` function so I could document the new argument.  Thanks to copilot for the helpful suggestions.  In VS Code, the doc string has very nice formatting.


rbeezer:
Another thought.  Do everything to get ready for a LaTeX compilation, but don't actually do it.  Then use `copy_build_directory()` to place the goods.

oscarlevin:
Something I realized just this morning is that the CLI's LaTeX build doesn't actually give you all the build files you need, in case you are building for a journal that has external .sty files.  I feel like there are too many options for me to wrap my head around now.

rbeezer:
OK, taking a fresh look at ths.  Following is all from the `pretext/pretext` perspective.

Now:
* `-f latex`: just the LaTeX file in output directory, mostly fro development debugging
* `-f pdf`: just the PDF file in output directory

Potentially::
* `-f latex-plus`: LaTeX file and all necessary image files. etc  in output directory, pre-compilation
* `-f pdf-plus`: PDF file, LaTeX file, necessary image files, log files, etc in output directory, build fails or not

Latter two would be a matter of copying the build directory at the right moment.

oscarlevin:
I think those are reasonable options.  I've force pushed that implementation, and now I think this is ready to go as far as core is concerned.  When merged, I'll add options to the CLI.  I will probably default to the `pdf-plus` behavior whenever the CLI uses debug mode, but also allow other options.

rbeezer:
Thanks for the rebase.   Likely at the same moment you thought "I should ping that PR", I was thinking, "Didn't we add that?"

Not sure why I haven't keyed in on this anyway, given the list is getting shorter.  Soon.

PR #2546, opened 2025-06-02
Add support for configuring a doenetml interactive

siefkenj:
A doenetml <interactive> can now be provided with a specific version of deoenetml in <docinfo> or on an individual interactive level.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -9982,10 +9982,47 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
 <!-- DoenetML header libraries -->
 <xsl:template match="interactive[@platform = 'doenetml']" mode="header-libraries">
-    <link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/doenet-standalone-test@0.6.0/dist/style.css" />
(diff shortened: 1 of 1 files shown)


rbeezer:
What is the purpose of the "override" URLs?

siefkenj:
> What is the purpose of the "override" URLs?

When debugging along with DoenetML, those get set to `http://localhost:...` to load the local versions.

rbeezer:
I thought so.  ;-)  We don't want authors doing that.  We don't want source to be pointing off at some unauthorized (crappy) version of Doenet that somebody forked.  And so on.  In other words, we don't want to entertain the possibility of weird stuff in source.  We want quality source, here meaning you can only vary the version number of **real* Doenet.

Look at the `debug.*` series of string parameters.    Make two (CSS, JS).  Then you can provide those in the standard way and you can set things up to prefer those if they are non-empty.  Don't provide `localhost` as a default, you will later want to see attempted uses based on empty/non-empty.  With this setup, the non-standard URLs will not be distributed in source.

How about I merge this, after I remove the override stuff, and you can add that on a new PR later.  Otherwise this runs the risk of sitting for a couple weeks (not much time on my end).  :-(

siefkenj:
Sounds good!

rbeezer:
OK, silently slimmed down.

Not sure what the best strategy will be for using string parameters, but I'm sure you will think of something workable.  Might be nice to have an example (in the sample article) using an old version, just to make sure it is semi-documented and frequently tested.

Likely a website update tomorrow.

PR #2548, opened 2025-06-03
Normalize and fix rendering of authored whitespace in programs for HTML

ascholerChemeketa:
The HTML static and dynamic (activecode vs now) handle authored whitespace slightly differently and eat whitespace in some places it might be desirable to add it. This fixes overzealous whitespace pruning and normalizes the behavior between the two outputs.

Clean diffs on SB. None of the issues existed in it or are not apparent in diffs (different handling between static and activecode programs).

Notes:
TESTING ONLY commit illustrates problem using hello-world. Assuming you will drop that commit after using it for before/after. I could move some/all of those into a random place in SA if you want.

2nd commit has the string-utils testbed I mused about before and some tests for existing (but not all) utils.

3rd commit adds a `substring-after-preserve` util and tests for it.

4th is the actual fix for authored whitespace in programs

5th drops an unused string util. It does the same job as `trim-end` with `$preserve-intentional = true()`

### DIFF
--- a/examples/hello-world/hello-world.xml
+++ b/examples/hello-world/hello-world.xml
@@ -38,7 +38,119 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- 6.  Experiment with other available conversions                                    -->
 
 <pretext>
-    <article xml:id="hello-world">
-       <p>Hello, World!</p>
-    </article>
+  <article xml:id="hello-world">
+    
+    <p>No whitespace at all - inline elements/content</p>
+    <program>
+      <preamble>AAAAAAAAAAAAAAAAA</preamble>
+      <code>BBBBBBBBBBBBBBB</code>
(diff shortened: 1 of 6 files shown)


ascholerChemeketa:
Working on related issue raised by Haris on -dev. Draft for now...

ascholerChemeketa:
Looks like it resolves what he asked. Back to ready status.

rbeezer:
Merged as-is (sans demo commit).  Thanks for the maintenance work, and I like the new tests!  Thanks for the demo commit.

A couple of follow-on suggestions:
* Can you craft an example like the demo commit that would stress any changes to the newer bits of code (for the sample article)?  (Yes, I know that is what the tests are for!)
* How about a simple/short `README.md` in `tests/` just saying what each file is for (including the `xsltproc pretext-text-utilities-test.xsl null.xml` instruction)?
* I don't think `--xinclude` is needed to execute the tests?  And for me, I think I only need one dash these days anyway.

PR #2552, opened 2025-06-05
add cardsort; premise has order attr

wrigjl:
Add cardsort as an option for a matching problem. Also, allow premise to have an order attribute.

(Be gentle with me please, this is the first time I've meddled in the affairs of XML validation).

### DIFF
--- a/schema/pretext-dev.rnc
+++ b/schema/pretext-dev.rnc
@@ -469,14 +469,20 @@
                 attribute number {text}?,
                 StatementExercise,
                 Feedback?,
+                Matches,
+                Hint*, Answer*, Solution*
+            Matches = (
                 element matches {
                     Match+
-                },
-                Hint*, Answer*, Solution*
+                } |
+                element cardsort {
+                    Match+
+                })
             Match =
                 element match {
                     attribute order {xsd:integer}?,
                     element premise {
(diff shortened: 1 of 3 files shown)


rbeezer:
Two commits merged to address this PR and extend the schema.

**Commit 1** (amended from your original): Adds `<cardsort>` to the `Matches` production, removes the deprecated `<matches>` element and `@order` on `<match>`, allows multiple `<premise>` per `<match>`, and switches the content model for premise and response in cardsort from `mixed {BlockText?}` to `TextLong` so that inline elements like `<m>` validate.

**Commit 2**: Adds `<matching>` as an alternative to `<cardsort>` in the `Matches` production, with `<premise>` (optional `@ref`) and `<response>` (`@xml:id`, optional `@order`), also using `TextLong` content.

**About derived files.** The schema source of truth is `schema/pretext.xml`, which is a literate program. The `.rnc` and `.rng` files are derived from it by running `xsltproc` with `xsl/pretext-litprog.xsl` to extract the compact RELAX-NG (`.rnc`), then `trang` to convert to RELAX-NG XML (`.rng`). The script `schema/build.sh` automates this. PRs that modify the schema should edit only `pretext.xml` and include the regenerated derived files.

PR #2553, opened 2025-06-06
Sample book: add an image to a fillin

bnmnetp:
Add an example of a fillin with an image in the statement.  This is good for testing, especially Runestone assignments where we need to rewrite the url to the image.


bnmnetp:
I know what I did...

Will work on a fix and get it out early this week.

bnmnetp:
Ok, 7.7.7 fixes it.  Or actualy leaves the URL alone instead of rewriting it.

rbeezer:
Working!  So merged.  Thanks for getting this in place.  Website update sometime later today...

PR #2555, opened 2025-06-10
Guide: change xml:id to label for generated images

sean-fitzpatrick:
The section on images in the guide does not reflect the change that was made from using `xml:id` on the `<image>` element to using a `label` on the tag indicating the type of generated image.

- I was not 100% sure that this change affected `sageplot` and `mermaid` as well, but I guessed it was unlikely that we'd be inconsistent on this
- The use of `<description>` (without a `<p>`) to provide the alt text is supported but doesn't match the instructions at the top of the page. I wondered if these should be changed to `<shortdescription>`.

### DIFF
--- a/doc/guide/author/topics.xml
+++ b/doc/guide/author/topics.xml
@@ -2546,9 +2546,9 @@ displayed line, and there are no <c>\\</c>s.  Use <c>\amp</c> to mark the alignm
                 <p>Authoring is straight-forward.  Inside an <tag>image</tag> include a child <tag>asymptote</tag> to hold the code.  For example:</p>
 
                 <pre>
-                &lt;image xml:id="gaussian-histogram"&gt;
+                &lt;image&gt;
                     &lt;description&gt;A histogram of Gaussian data.&lt;/description&gt;
-                    &lt;asymptote&gt;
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, @sean-fitzpatrick!  Just back from being off-duty.

*  "top of the page"  - where exactly?
* Mermaid images are not being handled correctly.  I may post on `-dev`.  So this part will have to wait, I guess.
* There is an `xref` to `rna-codons-figure`, which has moved off an `@xml:id`.  Likely the `@xml:id` needs to be on `figure` (and should use a different string/value) and then the label can be used on the `latex-image` and the existing image files will be OK.

sean-fitzpatrick:
In anticipation of #2560 I've left my changes in place for the mermaid instructions, but I can adjust those if it turns out we can't make mermaid work the same as the other formats.

sean-fitzpatrick:
I'm closing this, because I forgot to create a new branch for it, which means I can't do a clean pull request for a different issue. I'll make a new PR on a branch, once I get my copy of master reset.

sean-fitzpatrick:
I'm closing this, because I forgot to create a new branch for it, which means I can't do a clean pull request for a different issue. I'll make a new PR on a branch, once I get my copy of master reset.

PR #2557, opened 2025-06-12
FITB: fix host-platform for extracting dynamic substitutions 

dbrianwalton:
When trying to build a complete target (e.g. sample-book) for runestone, the CLI calls the extract dynamic substitutions routine with the target platform as runestone. However, the stand-alone pages that are produced for extraction are actually being hosted locally and so need to use host-platform as `web`.

Modified `pretext.py` to override the stringparam for host-platform for this call.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -1099,6 +1099,8 @@ def dynamic_substitutions(xml_source, pub_file, stringparams, xmlid_root, dest_d
         stringparams["publisher"] = pub_file
     if xmlid_root:
         stringparams["subtree"] = xmlid_root
+    # Always act as though web is the target
+    stringparams["host-platform"] = "web"
 
     tmp_dir = get_temporary_directory()
 


dbrianwalton:
Question: Should I have made a copy of the stringparams and modified the entry in a copy? I just realized that the CLI isn't calling the pretext/pretext script for each resource — it is sequentially calling internal functions with the stringparams as it works through the resource list. So modifying the stringparams internally in the dynamic_substitutions function may have a downstream side-effect.

@oscarlevin might have insight on this

dbrianwalton:
Thanks @bnmnetp for confirming.

oscarlevin:
Yes, almost(?) all the functions that manipulate the stringparams make a copy first

dbrianwalton:
I just tried to build sample-book with pretext/pretext/pretext (not CLI) with the substitutions were not available. I used the noparts publication file with format html and platform runestone and I didn't see that it ever looked for the substitutions. Maybe I was remembering something else. [When I tried using the runestone publication file, my installation choked on something related to a custom theme, which is why I tried noparts]

oscarlevin:
What would be helpful in terms of the CLI is to know which formats really
require the dynamic_substitution function to run.  Currently this is set up
to be required for all formats.  But if we can remove it from HTML, that
would be great.

dbrianwalton:
The dynamic formats, html and runestone, do not need the substitutions because they run the actual javascript.

All static formats (everything else?) will need the substitutions.

oscarlevin:
Perfect.  Then I'll make the change for the CLI so there will be no need
for playwright on Runestone builds.

rbeezer:
Yes, @oscarlevin, this file should only be consulted for output formats with "dynamic" exercise styles, which means broadly interactive, not Brian's use of the word dynamic.  And only consulted if the source has any of Brian's FITB exercises.

In other words, JS-enabled HTML output formats only.

dbrianwalton:
Sorry, @rbeezer, I did misunderstand. I had copied what was being done in the cli. I'll shortly push another commit that matches the pretext.py model.

rbeezer:
Thanks, @dbrianwalton.  Sorry to be picky - it just helps to make maintenance easier and thus makes the code less buggy.

dbrianwalton:
It should look much cleaner now.

rbeezer:
Yes, thanks for the quick work.  Should be able to merge soon after I am back home.  Landing right now

PR #2558, opened 2025-06-14
Footnotes for URLs

oscarlevin:
There is no good reason to include the footnote containing the URL for a link in HTML (like there is for print where the link might be text and not itself a URL).  This adds a class to such footnote elements and sets the CSS to not display them by default (a style author could override this).

Note that this does not interrupt numbering which is good, since numbers will still match other output formats.

### DIFF
--- a/css/components/elements/_footnotes.scss
+++ b/css/components/elements/_footnotes.scss
@@ -5,6 +5,10 @@ $border-radius: 0px !default;
   position: relative;
 }
 
+.ptx-footnote.url-fn {
+  display: none;
+}
+
 .ptx-footnote[open] {
   display: contents;
 }
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -1188,7 +1188,6 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <!-- NB: interior of "website" is a "url" in author's -->
         <!-- source, but the pre-processor adds a footnote    -->
         <!-- Only one presumed, and thus enforced here        -->
-         <!-- Kill this? -->
         <xsl:apply-templates select="url[1]|fn[1]" />
     </div>
 </xsl:template>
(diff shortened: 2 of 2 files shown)


sean-fitzpatrick:
This makes sense to me, but I'll play devil's advocate for a moment: 
Say you have 5 URLs and then a footnote.
Could a user of the HTML get confused if the number on the first footnote is 6?

I think they probably won't care, but this was the first thing that came to mind for me.

oscarlevin:
Perhaps a longer-term fix would be to use a different sort of counter for url footnotes (in print) than for regular footnotes.  I think it is better for the numbers to match between output formats than for numbers to not be skipped.

sean-fitzpatrick:
I agree that you want the different formats to match.

mitchkeller:
> Perhaps a longer-term fix would be to use a different sort of counter for url footnotes (in print) than for regular footnotes. I think it is better for the numbers to match between output formats than for numbers to not be skipped.

I likely agree that a long-term fix of using something like letters for URL footnotes in print to avoid inconsistencies between the footnote numbers in different formats is a good idea. However, my top priority is ending the excessive footnote production for links in HTML for online use. If this PR is going to sit for a while, I likely will just implement something else for the projects I need to build new copies of before August.

mitchkeller:
What I've found is that setting `@visual=""` is what can suppress an automatic footnote. (If there's something I've missed, please point me toward it.) I don't find this to be a good solution, as HTML output looks best when the `url` has content (not the URL). However, since the projects I work on also want to produce print, I *like* the automated footnotes for formats other than online HTML. (I'm even willing to agree that EPUB/Kindle might benefit from the automated footnotes.) In HTML, however, they take up space and attention. This is worse when on the same counter as authored footnotes. This requires readers to guess if each footnote is a URL footnote that provides no information they wouldn't have if it weren't there or if it's worth opening. To me,  the automated footnotes *in HTML* are a violation of PreTeXt principles 4 ("PreTeXt respects the good design practices which have been developed over the past centuries.") and 6 ("PreTeXt supports online documents which make use of the full capabilities of the Web.").

rbeezer:
> setting @visual="" is what can suppress an automatic footnote

Right.

>  I like the automated footnotes for formats other than online HTML.

So you want two versions.  Use version support and put two URLS where you have one now.  One is for HTML with no footnote, the other is for print with a footnote.  And your numbers don't match up at all.  :-(

Anyway, I like this reasoned discussion better than negative assertions.  Would you like to revive the `pretext-dev` thread, including much of this, and highlighting the dichotomy between print and online?

PR #2560, opened 2025-06-16
LaTeX: Use mermaid@label for extraction filename

ascholerChemeketa:
Fix for issue noted here:
https://github.com/PreTeXtBook/pretext/pull/2555
Bases Mermaid filename on label applied to `<mermaid>` element itself.

This makes mermaid extraction work with what is in that PR, but it would break existing instances. I don't see any deprecation warning or repair for the change. It feels like one both of those should be put in place before this gets merged. I'm on the road starting tomorrow. Some access to code, but limited time.


ascholerChemeketa:
It looks like this now is related to #https://github.com/PreTeXtBook/pretext/pull/2568

PR #2562, opened 2025-06-19
add multiple choice question with checkboxes even with only 1 correct…

Tanaquil18:
… answer


rbeezer:
Merged as-is.  Thanks!

Will go live on website soon.

PR #2565, opened 2025-06-24
HTML: Use javascript to create pages for worksheets

oscarlevin:
This is the second of a series of PRs to improve worksheets (the first was #2540).  It implements a page-breaking algorithm based on the Knuth-Plass algorithm used for line-breaking in TeX that tries to add roughly uniform increases to workspace across pages of a worksheet (similar to the algorithm suggested by @Alex-Jordan).   This allows a worksheet to still get reasonable page breaks even if the author does not specify pages using the `<page>` element (which could be deprecated in favor of this).

Prior to this PR, a `<worksheet>` that did not contain `<page>` elements would not render correctly when trying to print from html.  So even if this is not the best way to manage `<page>`-less worksheets, it is still an improvement.  If a worksheet does contain `<page>` elements, then the XSL will insert `<section>`s with `class="onepage"` in the HTML, and this javascript will treat those as they have always been treated.  So this PR should be 100% backward compatible.

The bulk of the PR is in the second commit; the first commit just fixes an oversight with the CSS for worksheet printing (should have been part of #2540).  The third commit updates the sample article to include versions of the first two worksheets that have `<page>` still and removes the unnecessary `<page>` elements from the remaining worksheets (the page breaks result in the same pages as the old version for these latter worksheets; the only worksheet that is different is the first and I did not adjust the workspace on the `<page>`-less version so a developer might be able to test adjusting those to get their preferred page breaks).


oscarlevin:
I would like to do some more work on this, unless you have already done work on testing.  I'm going to overhaul the CSS and javascript for authored pages to make it consistent; I'm afraid if I tried to do that on a separate PR, we would get too many conflicts.

Alex-Jordan:
I am noting here that in a few weeks, I would like to explore changes to the actual HTML structure that would support simple CSS and JS to reach the desired end goal. I believe the JS here assumes keeping the current HTML structure, and I think that we could have a really clean JS and CSS situation if we are willing to change the HTML.

Of course, I understand not wanting to have worksheet HTML diverge much, if at all, from other HTML. I just want to explore that track at some point not too far out now.

oscarlevin:
Just force-pushed a few tweaks to the CSS (now tables will have their rules shown, for example).

oscarlevin:
For the printing to the edges thing, your print dialog should have an option for margins.  If you select the default margins, then you will get the correct thing.  At least that's how it works in Chrome.

PR #2566, opened 2025-06-25
HTML: Internal links only using anchors, not filename, for portable builds

oscarlevin:
In a portable build, it is reasonable to want to rename the single html file that results (for sending to admin in case you are building a syllabus, for example).  For this to work, internal URLs need to only contain the anchor tag, not the filename.  This PR inserts a test for not being a portable build before the "containing filename" is inserted as part of the url.


oscarlevin:
Okay, I fixed both issues with a single commit.  It makes more changes to the xsl (splitting out a modal `url-anchor` template`, but I think the logic is better.  Tests on `$chunk-level = 0` rather than `$b-is-portable` so this also fixes the aforementioned bug (which also applied to frontmatter, it turned out).

PR #2569, opened 2025-06-26
HTML: add descriptions to Asymptote images

sean-fitzpatrick:
Two changes to the Asymptote template:

1. If `<shortdescription>` is present, write it to the `title` attribute of the `iframe` containing the Asymptote image
2. When `<description>` is structured by `<p>`, add the "info" button containing the long description below the Asymptote image.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -6153,7 +6153,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <iframe src="{$html-filename}" class="asymptote">
             <xsl:if test="shortdescription">
                 <xsl:attribute name="title">
-                    <xsl:apply-templates select="shortdescription"/>
+                    <xsl:value-of select="shortdescription"/>
                 </xsl:attribute>
             </xsl:if>
         </iframe>


rbeezer:
Putting content (from an author) into an attribute is a **Very Dangerous Thing**.  Does not take much to mess-up the HTML with a rogue attribute.

*  Schema says `shortdescription` is only text.  Good.
*  Somebody **will** try markup.  Meet them at the pass by using `xsl:value-of` and do not entertain their markup with `xsl:apply-templates`.
*  I only see one extant example in the sample book, and it is too simple to be a stressor.
*  Make an example where the text has a less-than character (escape it in source), and an ampersand for good measure.
*  Make another example with (nonsensical) single and double quotes and see if you can break the output routines.

Convince me that my fears are unfounded.  ;-)

sean-fitzpatrick:
That's a good point! 
The reason I didn't worry (or thought that similar stress testing may have already occurred) is that for other image elements we write `shortdescription` to the `alt` attribute. 
(In fact I'm sure I modelled the code here after what is done when writing `img` tags.)

sean-fitzpatrick:
I had time to test this today. Using `xsl:apply-templates`:

- If I put `<shortdescription>This is a "test"</shortdescription>`, the HTML tag for the iframe gets `title='this is a "test"'` (with single quotes)
- Switching to single quotes gets me `title="this is a 'test'"`
- Putting `<m>x^2</m>` into a short description gives me `\(x^2\)` in the title. The screen reader will not like this, but it doesn't break the html
- Both `&lt;` and `&amp;` get written verbatim (they remain escaped). I could put them in without escaping, but that will give invalid XML, and things will break (as they should)

Issue #2570, opened 2025-06-26
@provide-dark-mode="no" doesn't disable dark mode


mitchkeller:
As discussed in [pretext-dev](https://groups.google.com/g/pretext-dev/c/sU6PMykpWK8/m/0AQ_oFIfAgAJ), `@provide-dark-mode="no"` is not disabling dark mode. It appears that this is across all themes. (Tested with CLI 2.20.0.) A post from @ascholerChemeketa in that thread from January suggested that a fix would be forthcoming, but it doesn't appear that a PR to fix it has been submitted.

rbeezer:
Resolved at #2571

PR #2571, opened 2025-06-26
HTML: fix detection of theme options in publication file

ascholerChemeketa:
Addresses https://github.com/PreTeXtBook/pretext/issues/2570

The location of options on a theme in the CSS changed from attributes on sub-element to attributes on `html/css/`. That change did not get reflected to the code that looks up options like `@provide-dark-mode`. This fixes that.


rbeezer:
Thanks, @ascholerChemeketa!  Merged, though I suspect I did not give this any real testing.

PR #2572, opened 2025-06-26
Guide: add options for local installation with updated cli install notes

oscarlevin:
Now that the docker image is working better for codespaces, @mitchkeller pointed out that if one installs docker desktop on their machine, it is super easy to get the codespace experience locally (without needing an internet connection).  This provides a very quick way for a new pretext user to install all the software they will need.

This PR attempts to document this, and provides a section of the guide dedicated to the three local installation options.  

There are also a few tweaks to the CLI instructions (including bumping the minimum python version up to 3.10).

### DIFF
--- a/doc/guide/author/processing.xml
+++ b/doc/guide/author/processing.xml
@@ -22,7 +22,7 @@
             <ol>
                 <li>
                     <p>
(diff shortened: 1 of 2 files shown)


rbeezer:
Merged as-is.  Thanks.

PR #2573, opened 2025-06-27
Add event handler to close TOC when clicking outside it.

gigamonkey:
Small quality-of-life improvement: add some Javascript so if the TOC is open and the reader clicks somewhere outside the TOC it closes.

### DIFF
--- a/css/dist/theme-salem.css
+++ b/css/dist/theme-salem.css
@@ -7,7 +7,6 @@
 :root {
   --base-content-width: 725px;
   --content-padding: 48px;
-  --auto-collapse-toc: yes;
 }
 body {
   margin: 0;
@@ -3740,6 +3739,7 @@ body.pretext {
 }
 :root {
   --content-font-size: 1.1rem;
+  --auto-collapse-toc: yes;
 }
 @media screen and (width < 800px) {
   body.pretext {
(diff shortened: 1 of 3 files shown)


ascholerChemeketa:
I like the concept. A lingering ToC dropdown does violate expectations.

The CSS change needs to happen to the SCSS file that is used to generate `theme-salem.css`.

See the `css/README.md` for some tips on files to edit. But quick version is you want to make the edit to something in `css/targets/html/salem` - likely to `theme-salem.scss` in the existing rule targeting `:root`.

ascholerChemeketa:
BTW, thanks Peter!

rbeezer:
Great!

@gigamonkey - make a clear indication when you have this ready to go, and maybe @ascholerChemeketa can give it a last fleeting once-over.

Thanks, both.

gigamonkey:
@ascholerChemeketa I think my latest commit moves it to the right place? (Though I don't really know my way around scss so took a stab at what seemed like the right spot.)

ascholerChemeketa:
@gigamonkey Yes, looks great. There are fancy things you can do with SCSS, but the basic syntax is just CSS.

@rbeezer Looks good from here.

rbeezer:
Thanks, @gigamonkey.  Working well for me.

Repackaged as a JS commit and a CSS commit.

I routinely create the stuff in `dist/`, so you don't need to include that in PRs.

Thanks, @ascholerChemeketa for the help with this one.

PR #2574, opened 2025-06-28
added PreFigure diagram to sample book

davidaustinm:
Adds a PreFigure diagram for testing

### DIFF
--- a/examples/sample-book/cyclic.xml
+++ b/examples/sample-book/cyclic.xml
@@ -368,6 +368,37 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 <latex-image><xi:include href="tikz/cyclic-roots-unity.tex" parse="text" /></latex-image>
             </image>
         </figure>
+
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, David!

Since we don't complicate development with the CLI, I generated the necessary images, and then tested a couple output formats to make sure I'd done things right, and added them into the PR.

No annotations?  Would it be good to try those from within Runestone?  Make a new PR iif you want to add those.

rbeezer:
I can't find where Brad requested this.  I wanted to post a link to Figure 4.3.14.

bnmnetp:
It was on the RunestoneInteractive/rs repo.  But I see this!  Thanks @davidaustinm @rbeezer

davidaustinm:
> No annotations? Would it be good to try those from within Runestone? Make a new PR iif you want to add those.

I'm expecting a new version of `diagcess.js` relatively soon, which will require a small change on my part.  I'll test on Runestone and then send a PR.  Thanks for taking care of this!

PR #2576, opened 2025-06-29
Towards integrating STACK into PreText

geoo89:
For context, see [this discussion](https://groups.google.com/g/pretext-dev/c/gkZ4sSj8nVk/m/rRBPEoe-AQAJ) on `pretext-dev`. 

I've been working on generating static representations for STACK questions. I've written a function as a starting point (to be extended with error handling, etc), and you can test it on the example STACK question by running the [`stack_to_pretext.py`](https://github.com/IDEMSInternational/pretext/commit/3c92a36d72aab8001d30a319c54e5d7f64a8b756#diff-069b78ae2d106f911fdf64ef19129b0269aa34eaf0329b532cb82f2f6f620072) script I added. (Note: The public STACK API server used in this example is a bit temperamental, so if you get an error the first time around, try again.)

I'm trying to figure out how to integrate this into the larger asset generation process. I assume this should happen akin to the `webwork_to_xml` function which generates the `webwork_representations.xml` with static and PG versions of the webwork questions. In the STACK case we'll only have the static version (at this point anyway), as for the dynamic version a STACK question XML file will simply be referenced in the HTML. I guess those STACK question XML files would then be assets, e.g. like image files referenced in the text.

The part I don't have a good understanding of is the work done by the XSL templates. For example, in `webwork_to_xml`, we apply `extract-pg.xsl` to the input XML and then pick out the server parameters from the resulting XML file (e.g. to get the API URL, which must originate from the publication PTX file), so that we can call the server to retrieve the questions and/or generate the static versions. Obviously, a lot more is going on in that XSL template, most of which isn't applicable to STACK, but I don't understand what I'd need to mirror from there and what to ignore.

Similarly, I don't really understand the work the XSL templates are doing  when building a target, such as HTML or pdf. For STACK, all that'd need to happen is either take the static version from the representations file that we generated (e.g. PDF target), or simply insert a single `div` referencing the STACK question XML file (HTML target), something like
```
<div class="container-fluid que stack" data-qfile="questions/integrate.xml">
```
from the [example](https://github.com/IDEMSInternational/pretext-stack-example) where I embedded an interactive STACK question into HTML generated by pretext, as well as including the javascript headers. (I'll work on improving the actual Javascript logic to make it behave more closely to how webwork questions behave.)

Some pointers how to proceed from here would be really helpful, especially around the XSL templating. (I can make some sense of the `pretext` and `pretext-cli` python code if I dig into it a bit, but high-level guidance would still be helpful.)


rbeezer:
Thanks, @geoo89!  I was thinking about this recently.

Let's do most of the discussion on `pretext-dev`.  Can you cut/past the above onto the thread there?  And I'll reply.  Thanks.

geoo89:
There are only 4 new commits (the last 4), I can rebase once I'm back home
( Saturday). Meanwhile you could look at the individual diffs if you want

rbeezer:
Thanks for the rebase.  Yes, `dev` is a long-ago fiction.  I should probably remove it now.

I've taken a look through this, and it looks like there will be a lot to go on.  I have two big PRs in front of this one, both are on-pause (as of a few minutes ago), so we will need to see how those go before I can put a third ball up in the air.  Sit tight for a bit.

rbeezer:
A few follow-on commits to slightly adjust the flow here.  You'll want to look at them closely.  Lots of discussion on -dev thread, coming up in a minute or two..

I've done a fair amount of reorganization, should make more sense later.  Three instances of stray whitespace - I'm pretty religious about keeping those out of source - just makes for headaches later.  Hopefully credits on commits (Chris v. Georg) are accurate enough.

I'm going to close this, let's do further modifications on new PRs.  I like short, to-the-point PRs better than a bunch of unrelated stuff all at once.  (This one was fine, we had to get started!)

Please do not delete or modify anything here.  I have skipped the Python/static commit, and saved a copy locally, but this will be the backup if I need it again.

Thanks for the good work here, it was easy to work with.

PR #2577, opened 2025-06-30
FITB: dynamic-substitutions using node

dbrianwalton:
Change creation of the dynamic-substitutions.xml file that is used for FITB problems when a static format is being generated to use Node in place of the playwright/local server approach.

Note: Prior to use, it is necessary to run `npm install` inside the pretext/script/dynsub folder in order to be sure that the btm-expressions package is available when node is running the script. This will need to be added to the CLI installation process when we are ready to proceed.

I also took advantage in the later commits to reformat the resulting XML file so that it is more easily human readable and comparable, with separate lines for each substitution. The original script had redundant substitution entries when the same object was evaluated multiple times. This cleaned version only lists each object once.

This made it apparent that the sample-article appears to evaluate two math objects without being in math-environment, as some substitutions appeared as entry text and others appeared in LaTeX format. The new version *forces* math objects into their LaTeX representation for the substitution. [Neither old or new had the ability to look for different substitutions, depending on whether the author wanted a raw representation or a latex representation -- this needs to be added as a future update, perhaps using an attribute in the `<eval-subst>` elements?]

I also did some preliminary testing to see what would be needed with a possible future shift from `node` to `deno`, and left some commented code where the changes will be needed.


rbeezer:
*  I had to resolve one merge conflict when I rebased, it was about the `stringparams.copy()` change.
* Do you think the `script/dynsub/README.md` could use a hint to do `npm install`?
* Showstopper below.  Is my `node` a bit old? (Note my root directory is still called `mathbook`!)
```
PTX:INFO    : Generating substitutions.
PTX:ERROR   : Dynamic substitution process failed: file:///home/rob/mathbook/mathbook/script/dynsub/dynamic_extract.mjs:2
import { BTM } from 'btm-expressions/src/BTM_root.js';
         ^^^
SyntaxError: Named export 'BTM' not found. The requested module 'btm-expressions/src/BTM_root.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'btm-expressions/src/BTM_root.js';
const { BTM } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:171:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:254:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:474:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:109:5)

Node.js v22.3.0

PTX:DEBUG   : Temporary directories left behind for inspection: ['/tmp/ptx-g1zb53hf']
```

rbeezer:
Updated `node` using `nvm`, or at least I think so:
```
$ which node
/home/rob/.nvm/versions/node/v24.3.0/bin/node
$ node --version
v24.3.0
```
Get the same error report citing `Node.js v22.3.0`

rbeezer:
Thanks, @oscarlevin.

> Did you npm install?

Yes, I did.  I'm just trying to update my `node` executable on a hunch.

dbrianwalton:
Because the error is when running node and not npm, I assume that your node_modules package does have the btm-expressions module successfully installed. The reported error looks like it has to do with how node is expecting imports to work.

I copied the error message and the Google search AI summary suggested that the error is related to CommonJS vs ES Modules and sent me to a documentation page for node 24.3.0.  (https://nodejs.org/api/esm.html#commonjs-namespaces)

Let me try a different syntax for importing and see if we can get it to work on both versions of node.

dbrianwalton:
Looking closer at the original error message, you were using node v22.3.0. If I change to that version, I also get the error. A change was implement in node v22.10.0 that makes Node able to deal with the import in a way that works consistently.

I believe that to support a version below that, I'm going to need to find how to redo the npm packaging that is capable of supporting the different styles of JS required for Runestone and Node. Do we know what version of node we need to support?

oscarlevin:
What will happen when a user of the CLI tries to build if they don't have node v22.10.0?  Do they need to run an `npm install` from somewhere?  

The CLI checks the version and runs `npm install` in the right place for themes.  I assume this will be needed here too?  Please submit a PR or give me instruction to take care of this.

Issue #2579, opened 2025-07-01
GeoGebra javascript error for variable name error


dbrianwalton:
GeoGebra material ids that start with numbers fail to load.

The error is that the automatic javascript uses bare material id to start a variable name, so when the first character is not a valid symbol for a variable name, the javascript fails to load the content.

PR #2580, opened 2025-07-01
GeoGebra: Fix JS variables to have valid characters

dbrianwalton:
GeoGebra template for loading content requires forming several variables that contain key information about how to load the applet that runs GGB content. Issue #2579 showed that if a user chose an id for a figure that started with a number, that created an error.

This prepends the string 'ggb_' before the id to avoid this issue.

Are there any characters that could be in `visible-id-no-dash` other than A-Z, a-z, 0-9 and _ to worry about? If so, we'll need to add an additional filter layer.


rbeezer:
Thanks, @dbrianwalton!  That's perfect.  I added a couple comments to the code so the necessity of `ggb_` was obvious.

PR #2581, opened 2025-07-02
HTML: add aria-label to permalinks

ascholerChemeketa:
When using a screen reader (NVDA at least - others might vary), permalinks are pronounced link buttons are pronounced as "Link, link". This adds the same text the tooltip uses so they get pronounced as something like "Link, Copy heading and permalink for Paragraph".

Heads up @oscarlevin

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -3023,7 +3023,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- A figure wrapping the sidebyside could be knowled -->
 <!-- if they need to be hidden.                        -->
 <xsl:template match="sidebyside/figure|sidebyside/table|sidebyside/listing|sidebyside/list" mode="is-hidden">
-    <xsl:value-of select="false()" />
+    <xsl:value-of select="false()" /> 
 </xsl:template>
 
 <!-- Overall enclosing element -->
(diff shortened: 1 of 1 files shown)


xsl/pretext-html.xsl
```diff
@@ -3023,7 +3023,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- A figure wrapping the sidebyside could be knowled -->
 <!-- if they need to be hidden.                        -->
 <xsl:template match="sidebyside/figure|sidebyside/table|sidebyside/listing|sidebyside/list" mode="is-hidden">
-    <xsl:value-of select="false()" />
+    <xsl:value-of select="false()" />
```
oscarlevin, line 3026: Extra white-space
```suggestion
    <xsl:value-of select="false()" />
```

xsl/pretext-html.xsl
```diff
@@ -11471,21 +11471,30 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
...
 <xsl:template match="*" mode="permalink">
+    <xsl:variable name="tooltip-text">
+        <xsl:apply-templates select="." mode="tooltip-text"/>
+    </xsl:variable>
+    <xsl:variable name="permalink-description">
+        <xsl:apply-templates select="." mode="type-name">
+            <xsl:with-param name="string-id" select="'permalink-tooltip'"/>
+        </xsl:apply-templates>
+        <xsl:text> </xsl:text>
+        <xsl:apply-templates select="." mode="tooltip-text"/>
```
oscarlevin, line 11482: Not `<xsl:value-of select="$tooltip-text"/>`?

oscarlevin:
I definitely think this is an improvement.  Still curious whether we should hide these completely from screen readers.  

Out of curiosity, is there a difference in setting a variable and then doing a `value-of` instead of applying the template in its place?  Or is it that doing that is better if you need it in more than once place?

xsl/pretext-html.xsl
rbeezer: The pot calling the kettle black!  ;-)

rbeezer:
> is there a difference in setting a variable and ...

Since `$permalink-description` is used twice, is moderately involved to construct, and is "local" to this template, it makes sense to build it once in a variable.  However, as submitted, `$tooltip-text` is placed in a variable, used once, and a second opportunity to use it was skipped.  Since this is already a widely-used "global" template, i think it is cleaner to just use it twice.  I do not think the efficiency penalty is too great (though @ascholerChemeketa  might make me eat those words).  Any penalty would improve if numbering was done more sensibly.

PR #2582, opened 2025-07-04
CSS: improve printing from html; unified across themes

oscarlevin:
Largely because the Colorado-based themes do different things from the other themes with page margins, if a user tried to print from html (in general, not a worksheet), these were getting really extreme margins.  This PR fixes that, and improves printing for all themes by using the `@page` selector to set 0.75 inch margins around each page.  Page breaks right after headings are avoided, as are page breaks inside sage cells.

Tested on all themes.

### DIFF
--- a/css/components/_printing.scss
+++ b/css/components/_printing.scss
@@ -15,10 +15,23 @@
     display:none;
   }
 
-  .pretext .ptx-main .ptx-content {
-    margin-left: auto;
-    margin-right: auto;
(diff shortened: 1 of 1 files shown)


rbeezer:
Printed pages look great!  Did not see an improvement to my experience on #2565.

(Can `c` elements be made to print as just monospace heavy-weight font, without a surrounding box?)

Merged as-is.

PR #2583, opened 2025-07-05
CSS: Better printing of code and figures/tables/exercises

oscarlevin:
As requested in #2582, inline code will now appear without a background or border (just monospace font) when printing from HTML.  Also added some rules to avoid page-breaking inside a table and figure and exercise (mostly to avoid separating the title/caption from the rest of the content).

### DIFF
--- a/css/components/_printing.scss
+++ b/css/components/_printing.scss
@@ -23,7 +23,7 @@
     margin: 0;
   }
 
-  .ptx-sagecell {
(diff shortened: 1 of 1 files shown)


oscarlevin:
I was wondering about that.  I know how to do it with JavaScript, but not sure if js can detect when the browser's print command is pressed.  Or maybe there is a css way to do that.

PR #2584, opened 2025-07-06
HTML: add toggle to highlight workspace in a worksheet print preview

oscarlevin:
If you author a worksheet, you should specify a minimum height for your workspace (right, @Alex-Jordan?).  When you view the worksheet in HTML, we make the rendered workspace fit onto pages as best we can.  Sometimes this increases or decreases workspace.  But now there will be a checkbox in the print preview that when checked will highlight the rendered workspace, together with a column in the right margin of the workspace that stretches the distance you requested.  Tool-tip alerts you to this purpose, and green/orange highlights whether the the rendered amount respects the author's wishes.  

If this looks good, then I think we should add a publisher variable that would hide this toggle (the idea being that this is really just an author-tool; instructors will likely keep it on, since it isn't too distracting, but for mature projects, it could be turned off).

### DIFF
--- a/css/targets/print-worksheet/print-worksheet.scss
+++ b/css/targets/print-worksheet/print-worksheet.scss
@@ -179,58 +179,11 @@ section.worksheet > .heading + .para {
   }
 }
 
-
-// Note: leaving the following rules in place in anticipation of implementing a
-// workspace preview feature in the future.
- .onepage .workspace.preview {
-  border: 1px dashed grey;
-  background: rgb(240, 255, 255);
-/*  Sally suggests light and dark blue
-  background: linear-gradient(
-      #eef 0px, #eef 200px,
-      #eef 200px, #99f 205px,
-      #99f 205px, #99f 100%)
-*/
-}
-.onepage .workspace.squashed.preview {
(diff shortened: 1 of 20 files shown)


oscarlevin:
Oh, and I also sneaked in a commit that adds localization support for the "Print preview" title on its page, since I did the localization work for the "highlight workspace".  Probably should have also added localization support for the tool-tip, but that idea came later and I'm tired now.

rbeezer:
Very nice, I like the colored bars.  These are just suggestion/comments.  All your call, and any changes can go on a new PR.  Merged as-is.

*  Checkbox stays lit-up from one worksheet to the next.  Is that the desired behavior?
*  Could the "author-specified height" actually be reported in the tooltip?  E.g., "author-specified height of 1.5in".
*  New checkbox might be even less obtrusive if it were on the same line as "Paper size"?
*  Would colored bar work well on the left side of the highlight?  On the right I want to drag it like it is a scrollbar.  ;-)

Announce when you are ready and where you think best.  I'm of two minds about the necessity of a publisher variable for this (especially if it moves up a line).

Thanks for your persistence on this one - its looking good.

PR #2585, opened 2025-07-06
Add CSS styling for .booktitle.

gigamonkey:
I noticed that the modern, component-based CSS styles didn't have any styling for .booktitle. Added one copied from `legacy/pretext_add_on.css`. (I think I added it the right place so it will apply to all the themes?)


rbeezer:
Merged!  Thanks for the quick fix!  See follow-on commit, affected all themes, it seems.

PR #2587, opened 2025-07-07
HTML: only link ol-marker.css when it exists

jjrsylvestre:
See title. Avoids 404 in case project does not contain any `ol|ul` with `@marker`.

Also snuck in a little simplification that was discussed in the comments of #2402 (separate commit).

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -252,6 +252,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:apply-templates select="$document-root//ol[@marker and count(. | key('marker-key', @marker)[1]) = 1]" mode="ol-markers"/>
     </ol-markers>
 </xsl:variable>
+<xsl:variable name="b-needs-custom-marker-css" select="boolean($document-root//ol[@marker])"/>
 
 
 <!-- ######## -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks for moving on this one @jjrsylvestre!

Can we define a global variable that determines if `ol-markers.css` will be empty?  Something like:
```
<xsl:variable  name="b-needs-custom-marker-css" select="boolean(ol[@marker....]|ul[@marker...])"/>
```

Its not particulary expensive (unless the logic gets tortuous) and then we can **not** make the file and **not** make the `<link` and every single one of bazillions of page loads is that much cleaner?

jjrsylvestre:
I've pushed another commit on top that tests the boolean variable without searching the whole document tree, I assume that will be more efficient. You can either squash that new commit into the previous one, or drop it.

rbeezer:
Thanks for the changes, good to get this sorted.  A couple of silent changes.

* Used the more efficient boolean (thanks for that!) and added alternative in a code comment.
* Wrapped the file creation in the boolean.  Since we know if it is needed or not, we might as well bail out, rather than relying on the file to be empty and not built.

Note to self: we can place this in the right directory straight away and remove some Python that moves it around.

jjrsylvestre:
>     * Wrapped the file creation in the boolean.  Since we know if it is needed or not, we might as well bail out, rather than relying on the file to be empty and not built.

Sounds good. It was wrapped in a boolean before this pull request, but I removed it because in my testing I found that the file wasn't created if the `select` in the enclosed `apply-templates` was null. But I guess better not rely on implicit behaviour of the xslt implementation.

PR #2588, opened 2025-07-07
CSS: fix bug preventing most palettes from being specified

oscarlevin:
The recent worksheet printing PRs introduced a bug which prevented non `default-modern` CSS themes from specifying a palette.  This fixes that.

Sorry for that bug.


rbeezer:
Bugs always get my attention!  Merged.  Thanks for the quick fix.

PR #2591, opened 2025-07-07
CSS: tabulars scroll when too wide

ascholerChemeketa:
Wide tabulars that are wider than the viewport currently stretch the container they are in. This can end up looking awkward (check SA Graphics section in denver), causes issues  for mobile width in any theme, and 
makes it difficult to view the right side of the table and any context for the table at the same time (caption, surrounding text).

This makes tabular-boxes that are too wide scroll instead. It also makes them "expanded" in default-modern and salem so they continue to occupy a bit of extra space if possible.

@oscarlevin What do you want for denver/greeley? I tried just copy/pasting the expand logic from default-modern, but it ended up having issues at various widths. Given the limited extra space in the virtual page, I think not trying to "expand" the table might be the best approach.

### DIFF
--- a/css/targets/html/denver/_parts-paper.scss
+++ b/css/targets/html/denver/_parts-paper.scss
@@ -58,9 +58,7 @@ $page-width: $sidebar-total-width + $main-width + $main-right-margin;
   $navbar-breakpoint: $navbar-breakpoint,
 );
 
-@use 'rs-widen' with (
-  $content-width: 600px,
-);
+@use 'rs-widen';
 
 .ptx-main {
   margin-right: $sidebar-total-width;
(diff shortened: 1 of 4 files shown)


oscarlevin:
Okay, maybe it is just a artifact of chrome's mobile simulation; it seems to happen or not happen at random.  So I don't think this should hold back the PR.

oscarlevin:
I noticed an unwanted side effect of this change: exercises that contain a table automatically get widened (on salem/denver).  It happens even when the table is small enough not to need it.    For example, check out exercise 4 of Worksheet 35.3 using a theme that uses rs-widen.

oscarlevin:
Thanks.  With runestone exercises there is a clear visual reason why the container needs to be wider.  But with tables, that is less obvious.  It's tricky.

PR #2592, opened 2025-07-07
Close toc on subsection selection

gigamonkey:
As I discussed with @ascholerChemeketa during the drop-in today, it seems that an auto-hiding TOC should also auto-hide when a selection within the same page is made which I overlooked before. This PR fixes that.

### DIFF
--- a/js/pretext.js
+++ b/js/pretext.js
@@ -81,9 +81,6 @@ window.addEventListener("DOMContentLoaded",function(event) {
         sidebar.addEventListener("click", function (event) {
             if (event.target.closest('a') && event.target.closest(".toc-subsection")) {
                 toggletoc();
-            } else {
-                console.log('not a subsection');
-                console.log(event.target);
             }
         });
     }
--- a/js/pretext.js
+++ b/js/pretext.js
@@ -60,14 +60,30 @@ window.addEventListener("DOMContentLoaded",function(event) {
         e.stopPropagation(); // keep global click handler from immediately toggling it back
     });
 
-    // For themes that want it, install a click handler to auto close the toc.
+    // For themes that want it, install click handlers to auto close the toc
+    // when the reader clicks anywhere outside it or selects a subsection.
+    // (Selecting other sections or chapters navigates away from the page so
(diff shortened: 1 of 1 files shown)


gigamonkey:
Oh, whoops. Just noticed this doesn't clear out the highlight of the selected item. Let me fix that.

gigamonkey:
Ugh. This is a bit trickier than I have time to deal with right now. Feel free to ignore this PR until I get a chance to come back to it. Or close it if you'd rather and I'll open a new one once I've got it working properly.

gigamonkey:
Okay. I think it is good to go now.

rbeezer:
> Selecting other sections or chapters navigates away from the page so effectively closes the TOC.

For a Runestone-hosted book, the organization is chapters and sections ("subchapter" in Runestone-speak).  But a publisher could chunk a book where each `subsection` is its own HTML page.  And you can request greater depth on the ToC.  (I have not done the experiment to see what that looks like now in the new themes.)  Now does a `subsubsection` need to be considered?

That may not be 100% clear or accurate - but my point is: we have great flexibility in chunking and in the detail level on the ToC.  Do you need to be even more general with this feature?

gigamonkey:
I wondered about that but convinced myself that this would probably work in that case since things would still be within a .toc-subsection. But it occurs to me there's perhaps a more direct way to test the thing we actually care about that so let me see if I can do that instead.

gigamonkey:
Okay, this should directly test for the thing we care about: if the click in the TOC is a click on a link that takes us to the same page, it should hide the TOC. Seems like that should work for any level of TOC.

gigamonkey:
@ascholerChemeketa does this seem right to you now?

ascholerChemeketa:
Gave it a quick test dive. Seems to work as advertised with arbitrary ToC depths.

oscarlevin:
This is really cool.  Consolidated commits and merged as is.

oscarlevin (commented):
This is great.  I think something similar for all themes for mobile view would be nice (since that has the toc overlay

PR #2593, opened 2025-07-07
CSS: workaround for Firefox bug not respecting variable margins in @page

oscarlevin:
Not sure if this is enough of a bug to warrant your immediate attention, although this should be a quick one.  

The issue with Firefox not respecting the margins when printing worksheets seems to be a bug with Firefox (although it is also present in Safari I think).  While you can absolutely set margins for printing in the `@page` rule, these browsers to not evaluate the CSS variables.  This adds default fallbacks, which do get recognized correctly.  

For now, this is an easy fix.  When custom margins get implemented, we will need to be careful and override this rule.

### DIFF
--- a/css/targets/print-worksheet/print-worksheet.scss
+++ b/css/targets/print-worksheet/print-worksheet.scss
@@ -299,6 +299,8 @@ section.worksheet > .heading + .para {
   }
 
   @page {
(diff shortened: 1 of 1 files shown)


rbeezer:
Et voila!  Printing looks real good - used "Default" for margins in the Firefox print dialog.  Thanks!

PR #2594, opened 2025-07-08
WeBWorK: don't call pg_macros needlessly

Alex-Jordan:
Without this change, when `webwork_to_xml` is called, both `webwork_sets` and `pg_macros` are called. But `webwork_sets` also calls `pg_macros`. So there's a redundancy here, and it's wasteful that `pg_macros` is called twice.

At first glance, we would just remove the `pg_macros` call from `webwork_to_xml`. But when `webwork_to_xml` calls `pg_macros`, it requires the macros file to be built in a certain predictable place.  And when `pg_macros` is called directly by a user, it needs to place the macros file in the "destination directory", which is a different place.

So the changes here are meant to get around this puzzle. `webwork_to_xml` will still call `pg_macros` and tell it where to build the macros folder. But when `webwork_to_xml` calls `webwork_sets`, it now uses a new argument telling `webwork_sets` to skip the `pg_macros` call. We get rid of a redundant call of `pg_macros` without changing the behavior of a user calling `pg_macros` directly.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -1292,7 +1292,7 @@ def webwork_to_xml(
         os.mkdir(ww_macros_dir)
 
     # construct the generated pg files, etc, which may need to be read later for rendering problems
-    webwork_sets(xml_source, pub_file, stringparams, ww_pg_dir, False)
+    webwork_sets(xml_source, pub_file, stringparams, ww_pg_dir, False, False)
     pg_macros(xml_source, pub_file, stringparams, ww_macros_dir)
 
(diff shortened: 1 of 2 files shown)


rbeezer:
I'd prefer to not have named arguments with defaults (yes, others have slid them in).  I want consumers of the library to get errors rather than silent changes in behavior.

Please revise the definition and call once with `True` and once with `False`.

Not sure if @oscarlevin at the CLI needs to be informed.

Alex-Jordan:
I made the changes you requested.

PR #2597, opened 2025-07-09
Guide: edits to WeBWorK author guide

Alex-Jordan:
These are edits to the WeBWorK portion of the author guide. They cover a lot of what is in the Tacoma branch, and merging them now just helps whittle things down more and more. All of the changes here are still correct for the code as it currently stands.


rbeezer:
Thanks for the updates~  Merged as-is.

PR #2598, opened 2025-07-09
Reveal CSS structural update

ascholerChemeketa:
Removes baked in CSS from reveal template and uses version built by cssbuilder script.

Adds pubvar to specify custom css files for revealjs builds.


rbeezer:
Dang.  Forgot to put the PR number onto the merge commit.

So here goes, it is at: d6a94f812102828122673c9524021653201c5a9b

PR #2600, opened 2025-07-09
Javascript: improve display of authored workspace height for worksheet print preview

oscarlevin:
As suggested in #2584, this moves the authored workspace bar to the left of the workspace (looks less like a scroll bar) and puts the value of the authored workspace into the tool tip.


rbeezer:
Very nice!  (I can almost read the Javascript!)  Merged as-is.

PR #2602, opened 2025-07-12
Worksheet margins via publisher variables

oscarlevin:
This adds publisher variables for worksheet margins (but still allows specifying margins on a per-worksheet basis) and then uses these for both LaTeX and HTML printing.  The worksheet margins specified in `docinfo` are now simply ignored (no warning yet).  

Happy to tweak this as discussion happens on -dev here: https://groups.google.com/g/pretext-dev/c/FNRWpbAHWbw

### DIFF
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -621,19 +621,6 @@ function urlattribute() {
 
 // The new method for creating pages and adjusting workspace //
 
-// Assumptions: needs to work for both letter (8.5in x 11in) and a4 (210mm x 297mm) paper sizes.  We will work in pixels (96/in): those are 816px x 1056px and 794px x 1122.5px respectively (1 inch = 96 px, 1 cm = 37.8 px).  We assume that the printing interface of the browser will do the right thing with these.
-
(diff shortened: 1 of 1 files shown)


rbeezer:
Looks good.  Some comments 

> The worksheet margins specified in docinfo are now simply ignored (no warning yet)

Why not test on presence of `docinfo/latex-output/worksheet`, say it is being ignored, and direct to the new `common` publisher apparatus?  I'm assuming you think it is too hard to "honor the intent" there.

For the `mode="worksheet-margin"` template.  
*  How about a `match` on `worksheet`?  Always that way (good).  It'll add a bit of type-safety.
*  How about naming the two parameters `author-direction` and `publisher-direction`?  Wordy, yes.  But it would have helped me understand it a bit sooner.

oscarlevin:
Good ideas.  It wouldn't be difficult to continue to respect docinfo margins, but I don't know if it is worth continuing to have that available.  I can do whichever you prefer.  It wouldn't be a repair step, just extra logic inside the worksheet-margins template.

rbeezer:
Was ready to merge this, and got waylaid for hours with https://github.com/PreTeXtBook/pretext/issues/2619.

I can come back to this soon, but not now.

PR #2603, opened 2025-07-12
CSS: fix worksheet printing bug by removing top margin on pages

oscarlevin:
This fixes a bug with worksheet printing from HTML that @mitchkeller identified.  Added a `margin-top: 0` to the `.onepage` rule, and nested the print rules to ensure high enough specificity to override the conflicting rules in `_spacing.scss`.


rbeezer:
Merged as-is.  Concocted a less-harsh commit message.  ;-)  Thanks!

PR #2604, opened 2025-07-13
CSS: refactor widening code for Salem/Denver

ascholerChemeketa:
Refactors widening code for Salem/Denver to build on generic expandable mixin. Also simplifies the structure of sharing/customizing rules for the two themes.

Fixes undesired widening of all containers with tables mentioned here: https://github.com/PreTeXtBook/pretext/pull/2591#issuecomment-3054405586

@oscarlevin Fixing the overly aggressive widening motivated a refactor. Partly because I learned a new trick for passing information around in SCSS. 

Say `module-x` has a `$foo: 200px !default;`. If you `@use 'module-x' with ($foo: 100px);` to use that code and override the `foo` variable, and then later on (even in some nested location), `@use 'module-x';`, you get the overriden version. You can see this at work in Denver/Salem both `@use`ing `'sizin-globals'` but passing in different overrides for the variables in it. Then, when they import `rs-widen`, which does `@use 'sizing-globals';` it gets whatever was already established in that module.

This reduces the need to pass something like `$content-width` around through multiple levels of import. I think there is potential to simplify some other logic, especially with regard to color palates. But leaving that for another day.


ascholerChemeketa:
@oscarlevin Table widening now only very conservatively targets what can be identified as "top level" tables that are not in any grouping/lists/etc... To make centering work more reliably, the side margin for figures that are tables is removed.

rbeezer:
When I tried to rationalize tables, there was a big problem: LaTeX tables are rigid, and HTML tables can be squishy.

I know that squishing an HTML table is not always a great look.  How much of that behavior was retained in the big CSS change, and how will it be different with this?

ascholerChemeketa:
There aren't any rendering changes to tables that "fit", or to the squishing of content when it can be made to fit. e.g. this table appears the same pre and post on a narrowish screen width:
image

The only intended change is to tables that can't be made to fit as they have unbreakable contents or just too much contents. They now scroll individually instead of making the whole page scroll horizontally.
image
image

ascholerChemeketa:
@rbeezer The image formats table is not using `<p>` in cells and so all of its content is rendered into cells with `.lines`. That rule forces `white-space: nowrap`. Presumably to make them behave like latex tables?

oscarlevin:
@ascholerChemeketa, now when I try to build on salem/denver, the table in section 10 does not scroll or widen, just overhangs.

oscarlevin:
Thanks!  Merged with just a few whitespace deletions.

rbeezer:
@oscarlevin:  I find  `git commit --author "foo <foo@bar.com>"  ...` necessary at times.

oscarlevin:
Apologies @ascholerChemeketa.  Definitely not intentional; I must have done a `git reset` instead of a usual `git rebase` at some point.  I'll be more careful in the future.

PR #2605, opened 2025-07-14
Disable code-copy on derived HTML

skiadas:
This picks up the questions raised at #2535 I'll try to drop in tomorrow to discuss any issues

- Uses a variable to disable adding the clipboardable class name in derived HTML documents  (epub, revealjs, jupyter)
- Another question was regarding the print-worksheet stylesheet receiving the `.clipboardable` styles from `_codelike.scss`. I think that is not an issue in practice, as the key affected element of `.code-copy` is disabled in print media, but if @oscarlevin has some concerns we can figure out how to disable it.
- Another question was about applying this to the `pre` element. I'm not sure that's a good idea in general, I believe the `pre` element is used on some cases like sage-cells, or activecode, where we would not want this to apply, or at least I am not sure how to apply it given the overwrites happening. Is there some part that generates the "pre elements not affected by other things"? Maybe that's a place to add it?

Question: Would it be better if the variable started out as false, then overwritten as true in the file that produces the html output? Would that be `pretext-basic-html`?

### DIFF
--- a/xsl/pretext-epub.xsl
+++ b/xsl/pretext-epub.xsl
@@ -186,6 +186,9 @@
 
 <xsl:variable name="b-endnotes-have-math" select="$endnotes-have-math = 'true'"/>
 
+<!-- Disable clipboardable -->
+<xsl:variable name="b-add-clipboardable" select="false()"/>
+
 <!-- ############## -->
 <!-- Entry Template -->
 <!-- ############## -->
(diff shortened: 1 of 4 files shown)


rbeezer:
Thanks, @skiadas!

*  Consolidated into a refactor and an expansion of use.
* Edited `<xsl:template name="insert-clipboardable-class"></xsl:template>` to an empty element `<xsl:template name="insert-clipboardable-class"/>`
* Empty elements do not need a space towards the end (`..." />`, see example in previous bullet) - that was bad advice I got in the early days.  I didn't fix them here, and you don't need to go back and do it.

Thanks again!

PR #2607, opened 2025-07-14
CSS: long code-inline blocks scroll if they overflow container

ascholerChemeketa:
.code-inline that are too wide scroll instead of overflowing and possibly causing page to scroll.

This does not try to make anything wider, but is compatible with the `expandable` mixin. If that is applied to code-inline you get a wider div that scrolls when it runs out of space

See: sample-article/out/section-pre-formatted.html#section-pre-formatted-10 for good sample.


oscarlevin:
Looks great.  Thanks.

Issue #2606, opened 2025-07-14
link solutions/answers back to original in html


Tanaquil18:
When answers/solutions are not visible in context but instead <solutions> are added to the backmatter or somewhere else, I would like to be able to quickly get from an answer in the back to the problem in-context.  

This is already implemented in pdf where the type/number of the activity or exercise is a link back to the problem in context.  Another option could be to mimic what happens when automatic lists of theorems or definitions or interactive problems are created: there is an "in-context" link.

rbeezer:
Closing as implemented. #2610 ("Add backlinks to elements in solutions sections") adds exactly this — an in-context backlink from an answer/solution collected in the backmatter (or elsewhere) to the original problem, mirroring the behavior HTML already had for generated lists and that PDF has for exercise numbers. It's on master. Thanks.

*Claude Opus 4.8 (1M context), acting as a review assistant for Rob Beezer*

PR #2610, opened 2025-07-17
Add backlinks to elements in solutions sections

skiadas:
This should be addressing #2606 but also includes a fair amount of refactoring:

- Created a heading-generic template to uniformly handle the common elements to all heading generations (determining tag from heading, adding heading class). All other heading-generating templates now call on this one.
- Add boolean parameter to this heading-generic template and any others that needed it to indicate that the headings should actually become links back to the original location of those headings.
- Added some test cases to the sample article for heading cases that appeared to not be getting tested.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -12919,6 +12919,28 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
         </section>
 
+        <section xml:id="literate-programming" label="literate-programming">
+            <title>Literate Programming</title>
+            <p>There is support for literate programming using the <tag>fragment</tag> tag. It should always contain an <attr>@xml:id</attr> or a <attr>@filename</attr> attribute, and may contain <tag>code</tag> segments as well as <tag>fragref</tag> segments referencing other fragments. The final code is to be assembled starting from a fragment with a filename attribute and traversing the tree of fragment references.</p>
+            <p>For example here is the gcd algorithm in Python. Note that indentation is hard to get right at the moment.</p>
+            <fragment filename="gcd.py">
+                <title>The GCD algorithm</title>
+                <code>
+                    def gcd(a, b):
+                        while(b):
+                </code>
+                <fragref ref="gcd-loop" />
+                <code>                        return a</code>
+            </fragment>
+            <p>The key part is the inner part of the loop</p>:
+            <fragment xml:id="gcd-loop">
+                <title>The key loop part</title>
+                <code>
+                            a, b = b, a % b
+                </code>
+            </fragment>
+            <p>This double assignment changes both a and b.</p>
+        </section>
         <section xml:id="atomic-items" label="section-atomic-items">
             <title>Atomic Objects</title>
 
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -926,7 +926,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     <case direction="cycle">
                         <p>Does the first statement imply the second?</p>
                         </case>
-                        
+
                     <case direction="cycle">
                         <title>The <q>middle</q> case</title>
                         <p>Does the second statement imply the third?</p>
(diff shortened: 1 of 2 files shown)


Alex-Jordan:
I haven't looked closely at this, but I note that it touches on a lot of delciate code that ensures heading levels elevate correctly. Like, not following an `h2` with an `h4` without using an `h3` in between. Are you confident that this behavior is preserved?

I used the Firefox headingsMap extension a lot when we put all that together. It highlights when headings are out of line like that. There are still thinks that it won't catch, that need manual eyes. For example, if there are adjacent headings that should be `h2`, `h3`, but this change makes them `h2`, `h2`, then only manual oversight can see that.

One of the complications is that the heading level of something where it is born might be (for example) 3. But then when that thing appears somewhere else, like in a "solutions" page, it may need a different heading level.

skiadas:
> I haven't looked closely at this, but I note that it touches on a lot of delicate code that ensures heading levels elevate correctly. Like, not following an `h2` with an `h4` without using an `h3` in between. Are you confident that this behavior is preserved?

I'm as confident as I can be I suppose. So the first two commits add new "tests", then the moving around is done on the 3rd and 4th commit, and my guide throughout that process was a before- and after- diff of the sample-article build with the publication-crc settings, and observing that the versions are identical (except for the obviously different compile-time timestamp). So at least on the sample-article source code the results of the refactoring should be identical to the before state.

Then the last commit, which adds the backlinks, simply surrounds the title part with an anchor tag (does this need any accessibility extras btw?). 

The algorithm for determining what heading level to use has remained unaffected as far as I can tell, in the mode="hN" template. But yes it's a biggish change, which is why I wanted as many eyes on it as possible.

Are there other projects that I should use as test cases, beyond the sample article?

Alex-Jordan:
If you were doing diffs on the output HTML from the sample article, that is great. It has all of the complicated cases from when this was tested before, and no other projects were used in testing at that time. If you didn't see any "hN" changing their level, I'd be confident things are good in that front.

If it's not much trouble, if you can post the sample article build somewhere public, I'd look at the accessibility question you asked.

rbeezer:
Back to this one, since it seems to have gotten hung up.

Like @alex-jordan, the refactoring of hN values looks pretty involved and will take more study.  But Claude Code likes it (based on before/after testing), so I have confidence.  I may also take a close look at the hN levels, perhaps after this goes in.

Why is there a new section about literate programming?  What is it testing?  Commits with new source material should be rotated to be last when making a PR, so I can do before/after testing easily without them.

Next comment will be from Claude Code.  aria-label could be added now, but I won't let this sit just for that.  And CSS could be an add-on, especially if some `-dev` discussion is needed.

This rebases nicely on master, so you could do that, make any edits and move the new source material commit to the tip, and force-push.  The review is lengthy, but it looks like it'll be a good guide for me once I hop in to look more carefully.

rbeezer:
I reviewed this PR after rebasing on current master (clean, no conflicts). I built the sample article HTML before and after, with the literate-programming section and exercisegroup introduction changes removed from the source so the diff would isolate XSL-driven changes only.

**The refactoring is behavior-preserving.** With identical source, only solutions pages and pages containing inline `<solutions>` divisions differ. All other pages are identical (except timestamps). Heading levels (`h3`/`h4`/`h5`/`h6`) are preserved throughout.

**The backlinks work correctly.** Exercise, project, and task headings in all solutions sections get `<a>` links back to their birthplace. Inline solutions divisions on born pages (e.g., "Solutions for This Subsection" on the `section-interesting-corollary` page) also get backlinks, which is consistent behavior.

Some concerns:

1. **Exercisegroup introduction change (commit 2)** -- The change wraps the exercisegroup introduction content in `<title>` and `<statement>`. I believe `<introduction>` does not allow these as children -- it takes block-level content directly. This needs to be checked against the schema.

2. **Accessibility (following up on @Alex-Jordan's question)** -- The `<a>` elements wrapping heading content have no `aria-label`, `title` attribute, or visually-hidden text indicating they are backlinks. A screen reader would encounter a link but have no indication that it navigates back to the exercise's origin rather than being the heading itself. Something like `aria-label="Back to Checkpoint 4.9"` could help.

3. **No CSS for heading links** -- The `<a>` tags inside headings will pick up default browser link styling (blue, underline). Is there existing CSS that handles links inside `.heading` elements, or does this need a companion CSS change?

4. **Minor: same-page backlinks use full paths** -- When a `<solutions>` division appears on the same page as its exercises, the links are `href="page.html#id"` rather than just `#id`. Not broken, but slightly redundant.

The refactoring itself is solid -- funneling all heading generation through `heading-generic` is a nice simplification and the `b-make-link` parameter is a clean way to opt in to backlinks.

rbeezer:
Not so long after all - I got a much longer tour through the changes, which will help me, and not add much here.  Also, the "full URLS" is not a real problem.

rbeezer:
Thanks, @skiadas, for the changes (and so quickly!).  That was a lot of detailed work to put all of this together.

I've run tests by hand, and they look good.  Claude Code confirms that it looks good, and found a stray colon, which I have silently fixed.  That report is next.

PR #2611, opened 2025-07-18
Auto close TOC when navigating back with back button.

gigamonkey:
I noticed one more subtlety with the auto-closing TOC: if you open the TOC and then click a link that navigates to a different page and then hit the back button the TOC is still open when you come back. Which seems wrong to me since it should go away once you make a selection. This PR fixes that.

### DIFF
--- a/js/pretext.js
+++ b/js/pretext.js
@@ -108,6 +108,13 @@ window.addEventListener("DOMContentLoaded",function(event) {
                 toggletoc();
             }
         });
+
+        window.addEventListener('pageshow', (e) => {
+            if (e.persisted) {
+                sidebar.classList.remove('visible');
+                sidebar.classList.add('hidden');
+            }
+        });
     }
 });
 


rbeezer:
I can't duplicate the back-button behavior.  Maybe it is browser-dependent (I'm on Firefox).

@oscarlevin - I thought this might be a good one for you anyway.  Can you take it for a drive?  If it looks good, go ahead and merge.

oscarlevin:
I also wasn't able to reproduce the issue this solves and am curious which browsers load cached versions via forward/back button navigation.  Anyway, this doesn't hurt anything if it never gets applied.  Added a comment and rephrased the commit message.

Thanks.

gigamonkey:
Huh. Happens to me on both Chrome and Safari on MacOS. But yeah, Firefox does a page refresh on the back button so the TOC goes away without this.

PR #2612, opened 2025-07-18
WeBWorK: use custom names for generated and external folders

Alex-Jordan:
WeBWorK PG files need to refer to things in the generated and external folders, including when they are rendered for static representations. To date, these have just been the default "generated" and "external", so it went unnoticed that certain things broke if custom names were used for these folders. That particular issue was fixed as a side effect of one of the recent PRs. But moving forward, I'd like to move the sample article to use custom names for these folders. So that testing on new features is done with custom folder names.


rbeezer:
> Newline thing was not intentional.

I didn't really think so.  The point was it never should have made it into the PR.

PR #2613, opened 2025-07-18
WeBWorK: better handling of LaTeX macros in interactive exercises

Alex-Jordan:
Before this commit, when a WW exercise uses a custom LaTeX macro, and when that exercise is going to be rendered live, we open each chunk of math with whatever custom LaTeX definitions are needed for that chunk of math. This is pretty inefficient, since you might be redefining macros over and over again in the same exercise. Also it clutters up the MathJax tool that gives a user the tex code that was used to generate that chunk of math.

This changes all that. The PG being used for rendering the live version of the exercise now has no latex macro definitions at all. Instead, all of the project's custom LaTeX macros are loaded into MathJax once at the start of the problem (which, I should remind everyone, is within an iframe and not subject to the MathJax happenings in the ambient page.)

To accomplish that, the LaTeX macros text is kept in a visually hidden place in the ambient page, so that javascript can grab it and insert it at the start of the problem. This needs to be a separate instance of writing the LaTeX macros into the ambient page from what is already there. Because what is already there gets ingestedby MathJax, and becomes something completely different in the DOM to how it was born.

I can't think of an alternate way to reach the objective, except maybe to put the latex macros into a data attribute instead of into an element. I'll make that change if it seems better.

Either way, the second copy of the latex macros text will only show up in the HTML if the project is using WeBWorK. So no unnecessary clutter for other projects.

### DIFF
--- a/js/pretext-webwork/2.16/pretext-webwork.js
+++ b/js/pretext-webwork/2.16/pretext-webwork.js
@@ -476,7 +476,11 @@ function handleWW(ww_id, action) {
                  max-width:100%;
             }
             </style>` +
-            '</head><body><main class="pretext-content">' + form.outerHTML + '</main></body>' +
+            '</head><body>' +
+            '<div id="latex-macros" class="hidden-content process-math" style="display:none"><span class="process-math">\\(' +
+            document.getElementById('latex-macros-text').textContent +
+            '\\)</span></div>' +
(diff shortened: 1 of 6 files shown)


rbeezer:
> or maybe I add something even more distinctive to lock in on

I know I am on standby.  We could add a LaTeX comment with a magic string.  There is one floating around that might be available for reuse...

PR #2615, opened 2025-07-19
Wrap text datafile/pre/@source in CDATA

gigamonkey:
This seems like the Right Thing™ insofar as a data file should not need to be mangled to be legal XML text. However this change could possibly break existing books that have run into the problem this patch is fixing by adding escaping or CDATA into their datafiles.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -855,7 +855,7 @@ def datafiles_to_xml(xml_source, pub_file, stringparams, xmlid_root, dest_dir):
     # Each file receives a single element as its root
     # element. These are templates for that entry
     image_info = '<pi:image-b64 xmlns:pi="http://pretextbook.org/2020/pretext/internal" pi:mime-type="{}" pi:base64="{}"/>'
-    text_info  = '<pi:text-file xmlns:pi="http://pretextbook.org/2020/pretext/internal">{}</pi:text-file>'
+    text_info  = '<pi:text-file xmlns:pi="http://pretextbook.org/2020/pretext/internal"><![CDATA[{}]]></pi:text-file>'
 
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
Initially posed on wrong PR. @rbeezer's post on -dev and this crossed in flight.



Our conclusion is that `@source` on datafile should probably be deprecated.  (Correct me if that wasn't *our* conclusion Rob).


Usually `@source` usually means that the referenced asset is being loaded at run time. This `@source` is confusing because the asset is being loaded at build time (as text) into the XML before it is built. (Which is probably why you wanted to put the program into `external`, which is really for things that are going to be loaded at run time.)

It also is potentially confusing because you can use `<pre>` outside of a datafile, but `@source` is not going to work in those locations.

Anywhere else you wanted to include code from a file (like in the body of a `<program>`) you would have to use `<xi:include parse="text" href="filepath"/>` to inject the contents. That handles escaping the text to make it valid XML.

So `pre@source` is just a special case alternative to writing that.

The recommended authoring pattern for putting the contents into a separate file would be:
```
<datafile>
  <pre>
    <xi:include parse="text" href="filepath"/>
  </pre>
 </datafile>
 ```

rbeezer:
> kind of weird

If you mean the `pre` element ("the element" is not clear to me), it is necessary to distinguish the case where the element is an `image`.  Though maybe we coulda, shoulda, done that with an attribute.

gigamonkey:
Lol. I forgot Markdown also lets you use arbitrary HTML. What I said, some of which got eaten by Markdown:

> Makes sense to me. At that point it's kind of weird to me to have the `<pre>` element at all, but I guess that's to contrast to the possibility of an `<image>` child.

PR #2617, opened 2025-07-19
Clarifying the dl docs a bit.

gigamonkey:
I was confused about how to use dl and had to got digging through example PreTeXt to figure it out. Hopefully these few sentences will save the next person a few minutes.

### DIFF
--- a/doc/guide/author/topics.xml
+++ b/doc/guide/author/topics.xml
@@ -3542,7 +3542,7 @@ displayed line, and there are no <c>\\</c>s.  Use <c>\amp</c> to mark the alignm
                   <ul>
                     <li>
                       <p>
-                        <attr>xml:id</attr>s are used to link to a program from other parts of the book source code. Anny program that you want to use as the target of an <attr>add-files</attr> or <attr>include</attr> attribute in some other program must have a unique (within the book) <attr>xml:id</attr>.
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
@rbeezer and I discussed this at drop in this morning.

Our conclusion is that `@source` on datafile should probably be deprecated.  (Correct me if that wasn't *our* conclusion Rob).


Usually `@source` usually means that the referenced asset is being loaded at run time. This `@source` is confusing because the asset is being loaded at build time (as text) into the XML before it is built. (Which is probably why you wanted to put the program into `external`, which is really for things that are going to be loaded at run time.)

It also is potentially confusing because you can use `<pre>` outside of a datafile, but `@source` is not going to work in those locations.

Anywhere else you wanted to include code from a file (like in the body of a `<program>`) you would have to use `<xi:include parse="text" href="filepath"/>` to inject the contents. That handles escaping the text to make it valid XML.

So `pre@source` is just a special case alternative to writing that.

The recommended authoring pattern for putting the contents into a separate file would be:
```
<datafile>
  <pre>
    <xi:include parse="text" href="filepath"/>
  </pre>
 </datafile>
 ```

rbeezer:
Thanks for taking the time to expand this.

We really shouldn't assume authors know anything about HTML, but I guess I started it with "We borrow from HTML..."

I'd prefer unrelated commits not be added when a PR is outstanding.  If you want to develop a PR without involving me (yet), please mark it as a draft.  The typo fix caused me to make a mess I then had to clean up (I assumed it was part of the `dl` discussion....)

rbeezer:
Shoot.  Thought I had merged this yesterday when I commented.  Now merged.

PR #2618, opened 2025-07-19
Publisher pagebreaks

Alex-Jordan:
Draft PR.

This lets publishers specify xml:ids for where they want additional page breaking to occur. There are two issues to sort out yet, and I feel that the discussion has reached the point where it may help to see some code.

1.  Do we like the name "insertions"? Both for the publisher file element, and for the additional assembly phase.
2. I get some sort of RTF-related alert when I use this. `xsltFreeStackElem: Unexpected RVT flag 3`. I can't sort out what it is telling me, or how my usage of things here differs from elsewhere in the code base. Alert notwithstanding, everything "works" in the sense that page breaks come out where specified.


Alex-Jordan:
> I get some sort of RTF-related alert...

I made a change that skirts the alert while leaving the feature functioning.  There is a comment about this change now in `pretext-assembly.xsl`. It's surely an inefficiency to repeatedly use `str:tokenize()` like I'm doing, rather than store it in a variable, but my testing shows no noticeable change in timing.

I'm going to lift draft mode now. Although perhaps we still are not in love with "insertions".

Alex-Jordan:
This should apply to `exercise` too. That should be part of `&NUMBERED-BLOCK;`.

rbeezer:
This looks real good, but I think there is one big problem.

Adding a `pi:page-break` element during assembly totally throws off the ids that are manufactured automatically.  Like "fundamental-theorem-5-2".  We make an "assembly-id" at one point, hoping that it will remain the same when we make a "visible-id" on a second pass.  I mean to try to sort this out, but have no great ideas right now.  The insertions are happening in-between.  Which I think will lead to a big disconnect.

Yes, authors should be using labels.  But imagine a `latex-image` without.  Add an insertion, file name changes, LaTeX won't compile until you re-generate images.  That sounds like a frustrating process as you iterate through introducing page breaks (say the affected image was late in a chapter).

I have not manufactured an example.  If you think this fear is unfounded, we can dig deeper.

Approach I was thinking of **before** seeing above.  Introduce an attribute on the element with the `@xml:id`, say `@pi:page-break="yes"`.  The cost: every supported block needs an adjustment to every template for that block which early on applies a template to conditionally insert the line-break.  But an attribute, in a way, feels more natural to me - "this block is special and will start on a new page".

Also, if I was to merge this right now, I'd need to explode the second commit into four: publisher variable, assembly, LaTeX, Guide.  There is a small merge conflict with the entities, it is easy to fix, look for an addition of `handout` to the previous last definition since this PR was last touched.

rbeezer:
> Would it make sense to just do these insertions all at the end

That might work, but I really want to make the pre-processor simpler, not more complicated.  It is really making me nervous these days.

Now I'm wondering - why use -assembly at all?  The publisher variable is great, and is globally available.  Once into a block, test its `@xml:id` against the variable, and maybe drop a `\newpage`.  A template can do this, it just needs to be applied in each block.  This is strictly a LaTeX thing - so why not keep it all in the LaTeX conversion?  Conversely, why test for these page breaks every single time we apply a stylesheet, such as the `extract-*.xsl` suite?

Alex-Jordan:
Yes, the way it is now was in an effort to add minimal code. Just the one
template to -latex. That necessitated using an assembly pass. But if we let
that go, assembly is not needed.

Alex Jordan
Mathematics Instructor
Portland Community College

rbeezer:
> Yes, the way it is now was in an effort to add minimal code.

Yes, that was a good idea.  A clever idea.  Maybe too clever.  ;-)  -assembly is too popular and getting too complicated.  I've got plans to give it a good once over...

Alex-Jordan:
Same question for `audio` and `interactive`, which -assembly also converts to `sidebysde`.

Alex-Jordan:
A different question while I was working on this, much less important than the first question. In -latex, there are a few templates that match on `figure|table|listing|list`. This is identical to the `&FIGURE-LIKE;` entity. Is there any reason I should not change these to match on `&FIGURE-LIKE;`?

rbeezer:
Nicely done.  You got the order right.

One edit:  "audo" to "audio".  Otherwise as-is.

"Copied verbatim from pretext-common.xsl" is a red flag, especially without a code comment by the original to indicate the copy might need subsequent editing.  Can you mimic the "post-sbsgroup" template?
*  make a "pre-sbsgroup" template
*  define it to be a no-op in -common
*  implement it in -latex-common to make a possible newpage
*  add an example to the sample article, if it is not there already

I've merged this, but would really like to see the above hook happen.

Would you like to do the honors on -announce?  Be sure to call out the "pagefitters".  That's a term of endearment, so try to use it that way.  ;-)

PR #2621, opened 2025-07-21
LaTeX: fix bug affecting worksheet exercises

### LINKED ISSUE #2619: Sample article worksheet causes LaTeX to fail
The sample article is not surviving LaTeX compilation.  Not code, but rather the introduction of page-less worksheets for testing.  I shoulda caught it sooner.

`git bisect` to locate, so I don't think new code is to blame.  An existing bug coming to the surface.  

```
! LaTeX Error: There's no line here to end.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.13070 \end{divisionexercise}
                              %
```

Then a whole lot of "commenting out" to see what the problem was.

Commit 32262dda4922a4535bcfb10fe1c66b50c51337ac was the best I could do to remove just enough to make compilation go.

@oscarlevin  - not suggesting this is your fault, but I could use another set of eyes on this.  I'm out of gas at the moment.  Revert the commit (on a branch) to test?

oscarlevin:
Using `\par` instead of `\newline` to surround the workspace strut appears to fix the problem.

Will close #2619.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -14933,7 +14933,88 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                     <statement>
                         <p> Now generalize these ideas to a context outside of electrical circuits. Consider the network of streets given in the diagram (with one-way directions as indicated).</p>
 
-                        <p>2025-07-20: this exercise has been gutted so LaTeXt compilation will succeed.  It has likely lost some of its use as a test case.</p>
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks very much @oscarlevin!  I should be back onto this one later today.

rbeezer:
Thanks again for the fix, especially while you are busy with other things.

Added a follow-on commit to move some existing margins off of points and into inches.  A bit over-eager, should have followed #2602.

PR #2624, opened 2025-07-22
Aside rework as born hidden knowls

ascholerChemeketa:
See discussion:
https://groups.google.com/g/pretext-dev/c/WcgbtDJbFm8/m/OKNtHsuNBAAJ

This does not implement some of the bigger ideas (allowing publisher to select what aside/remarks to "asidify"). Just about getting HTML implementation to a solid resting place.

This has the "hover to reaveal" in the commits with a commit at the end to rip that out. It looks like the consensus is to not use that behavior.  If you really want, I can do a force push that pulls that code out of the earlier commits. But the logic is non-trivial, I wouldn't mind having a record of how to preview on hover a knowl in the repo.

@oscarlevin If you want to modify what Denver is doing, let me know. I can rip out denver stuff and let you commit later. Or you can make a PR on my fork and I'll merge to this.


oscarlevin:
I like what you have done with this, both for `denver` and in general

rbeezer:
Merged, mostly as-is.  Rotated sample article addition to be last (removing removed hover logic would have been easy no matter where it was).  Left in the commit with all the hard work to make the hover-reveal function (which was an exception!  ;-) )

Testing and all would have been easier if there had been a "heading divider" PR, then an "aside" PR.  As it was, I worked around it OK.  And doing two merges was easier than I'd expected.

Thanks for this - its great to get this on a better footing.

PR #2625, opened 2025-07-23
documentation edits

rhinopotamus:
Two small edits to the documentation:
- I am always looking up Table 4.5.2 about xref text options; I thought it would be helpful to have the actual xref tags displayed in a listing afterward.
- Oscar recently wrote a document explaining asset generation through pretext generate, which will be very helpful for people debugging weird generation things.

### DIFF
--- a/doc/guide/author/processing.xml
+++ b/doc/guide/author/processing.xml
@@ -390,6 +390,10 @@
                     </li>
                 </ul>
             </p>
+
+            <p>
+                Further information on <c>pretext generate</c> is available on <url href="https://github.com/PreTeXtBook/pretext-cli/blob/main/docs/asset-generation.md">the <c>pretext-cli</c> github repository.</url>
+            </p>
         </subsection>
 
         <subsection xml:id="cli-project-manifest">
--- a/doc/guide/author/topics.xml
+++ b/doc/guide/author/topics.xml
@@ -1140,7 +1140,108 @@
                         <cell><xref ref="topic-divisions" text="custom">Extra</xref></cell>
                     </row>
                 </tabular>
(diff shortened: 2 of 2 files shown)


rbeezer:
@rhinopotamus - these are good suggestions.  Did you try building HTML or LaTeX/PDF before submitting this.  Neither really works.  A `tabular` does not belong in `code`, though it **is** a creative idea.

A `table` with cells wrapped in `c` might work, though I fear it will be quite wide.  A big `pre` is another option, maye using some identation to organize/group.

rhinopotamus:
I tried to build but had an issue since I can't install sage on my work laptop.

(Relatedly: more modular source of these docs would make it a lot easier to contribute / test / build!)

I think you're ultimately right that a larger reorg of this table is a better way to go. Will ponder.

rhinopotamus:
> There's no real good way to modularize XML. "It is what it is" - one monolithic tree.

Oh, no, for sure, I mean like [this](https://pretextbook.org/doc/guide/html/processing-modular.html).

rhinopotamus:
Ok, here's a new version that I think is better in several ways:
- the link to the pretext-cli github doesn't have `<c>` in it and therefore looks nicer;
- I put the code for `<xref>` examples in `<hint>` tags in a second row so they automatically knowl.

rbeezer:
@rhinopotamus: a `hint` only goes as a child of an `exercise` (or PROJECT-LIKE).   It might behave well here in this table, in HTML output, but that is just a happy accident.  LaTeX is likely a mess (did you try building that?).

I love the creativity, but we need to stick to the syntax, especially for the Guide.

PR #2627, opened 2025-07-24
WeBWorK: pretext-webwork js file for a 2.20 server

Alex-Jordan:
This adds a js file for interacting with a webwork 2.20 server. It is an exact copy of the 2.19 file, since at this time, there is no know difference in behavior. But if you presently try to use a 2.20 server, the PTX HTML will go looking for a 2.20 js file and not find it.

So to test, on master build the sample chapter using `webwork-dev.aimath.org` as the server (specified in the publisher file). This is now a 2.20 server. When you try to load the HTML, it will fail to render live problems. (Of course, it will fail anyway if you try to view without using a proper web server to view.)

Following this PR, it should work.

### DIFF
--- a/js/pretext-webwork/2.20/pretext-webwork.js
+++ b/js/pretext-webwork/2.20/pretext-webwork.js
@@ -1,0 +1,670 @@
+//Critical:
+// TODO: Think about what should be done for an essay question.
+
+//Accessibility:
+
+//Enhancement:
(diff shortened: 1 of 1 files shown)


Alex-Jordan:
I view the AIM WeBWorK dev server as not a public resource. I might change radically something there for PTX testing, so no one should be relying on it. Only recent exception is that I suggested to Brad that he use it for testing the sample chapter inside Runestone. I'll check in with him.

PR #2631, opened 2025-07-24
Bare pg

Alex-Jordan:
This is a replacement for #2620 and #1696 to let the full text/code of a PG problem be used directly in a `webwork` element.

- First commit has documentation and the necessary changes to xslt and schema.
- Second commit adds one example each to the webwork minimal example and the sample chapter. This is a fundamental thing that I felt should also be in the minimal example.

### DIFF
--- a/doc/guide/author/overview.xml
+++ b/doc/guide/author/overview.xml
@@ -534,14 +534,14 @@
         <idx><h><webwork/></h><h><webwork/> exercise</h></idx>
 
         <p>
-            It is possible to include <webwork/>/PG problems in your source. These can either be
-            authored using <pretext/> or referenced from a host
-            <webwork/> course. A static version of the exercise will be rendered in your static
-            output formats such as <latex/>/PDF. A <q>live</q> version will be rendered in HTML
-            output. If the HTML is hosted on a Runestone server, the student's progress completing
-            the live exercises can be tracked. You can also extract the problems into a single
-            archive suitable for uploading into a <webwork/> course. This is a big topic, so see the
-            dedicated <xref ref="webwork-author"/> for details.
+            It is possible to include <webwork/>/PG problems in your source. These can be authored
+            using <pretext/>, included from local PG files, or referenced from a host <webwork/>
+            course. A static version of the exercise will be rendered in your static output formats
+            such as <latex/>/PDF. A <q>live</q> version will be rendered in HTML output. If the HTML
+            is hosted on a Runestone server, the student's progress completing the live exercises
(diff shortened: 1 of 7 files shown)


oscarlevin:
I think this looks really good, and it seems to work exactly as I would expect in my testing.  Happy to take care of merging unless @rbeezer wants to take over now (I would split up the first commit to pull out the guide stuff).

Alex-Jordan:
> Please split out the changes to the Guide on their own commit (penultimate) and changes to the schema (last).

OK, done.

> I want to de-emphasize (and abolish, if possible) pointing to problems on host courses. Can you remove those suggestions from the Guide?

I did not review everything. There was one place nearby where this current feature is being described, and I rewrote it to emphasize OPL usage only.

> I want to demonstrate having a PG problem right in the source. Please describe that in the Guide rather than making it look like xi:include is the only possibility. And include an example of that in the sample chapter. Outdent it to the left margin if you feel that is warranted.

OK, done. Both are outdented.

Alex-Jordan:
I am aware that we could use sanitize-text here, but then I'm confused about what we mean by "source". If you have an indented PG problem as text content of a `webwork`, that text is not valid PG. It maybe (if the author aligns things well) becomes valid PG after sanitize-text does its thing. So what do you want "source" to mean when it's an external coding language like this? Is it "source" that could directly go into the external code interpreter? Or is it PTX source that will be processed and potentially altered before going to that external interpreter?

rbeezer:
Tested first commit (before and after) with usual WW server at AIM (v2.18?).  Tested all commits (before and after) with WW dev server at AIM (v2.20?).  Just sample chapter.

*  Actual file for new "Local PG File" was not included.  I have fixed that, but it was quite a speed bump.
*  Second scenario above.  `my-generated/pg` does not seem to change.  I'd expect two new files for the two new "Local" examples.
*  Second scenario above.   Same two problems have expected changes in LaTeX.
*  Second scenario above.   Same two problems seem fine initially in HTML, but do not activate.
*  Edited all but one commit message to conform to customs/expectations.

Perhaps second and fourth bullets are related?  I forget how  `my-generated/pg`  is used exactly.

Given that 2.18 seems OK without any local files, and 2.20 builds OK, I am going to go ahead and merge this.  But perhaps there is some fixing still needed?

Alex-Jordan:
I'm sorry for that. Some of my workflow produces all these new files in the example/webwork folder, and the new .pg file was drowned in the mix of noise from those showing up in my git diff. I should have been aware though. Also for the minimal example. 

I'm working on a PR that will clean things up.

PR #2634, opened 2025-07-26
Guide: clarify target selection in getting started tutorial

### LINKED ISSUE #2626: Tutorial suggests building unavailable target
I'm following the [Your First PreTeXt Document](https://pretextbook.org/doc/guide/html/tutorial-first-document.html) tutorial in a GitHub Codespace. The “Building for print” subsection suggests that I can select ”PreTeXt > Build another target ... > print-latex” to build a LaTeX version of the project. I tried that, but no such target is available (only `print` and `web`), and manually typing it in does nothing.

Digging into `project.ptx`, I found that there are only two `<target>` elements as children of `<project>/<targets>`: `print` and `web`. In the commented-out block below that, I found a snippet `<target name="latex" format="latex"/>`. Once I copied that from the commented-out block to `<project>/<targets>`, I was able to select `latex` as a target and everything worked as the tutorial said.

So either the project-generating tool needs to include an active `print-latex` target originally, or the tutorial should be updated to include this extra step of editing `project.ptx`. I think the latter is better.

oscarlevin:
Will close #2626

### DIFF
--- a/doc/guide/introduction/tutorial.xml
+++ b/doc/guide/introduction/tutorial.xml
@@ -411,7 +411,7 @@
 
     <introduction>
       <p>
-        The browser-based GitHub Codespaces workflow described in <xref ref="quickstart-example"/> is a quick and easy way to use <pretext /> without needing to install any additional software.  If you prefer to have a local setup of PreTeXt, we recommend one of the following three setups.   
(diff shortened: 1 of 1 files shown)


rbeezer:
Merged this and forgot to close it.  Done now, will go close #2626.

PR #2636, opened 2025-07-28
JS: fix TOC auto-scroll

jjrsylvestre:
From what I understand, `offsetTop` only returns the offset from the element's parent, not from top of page.

@ascholerChemeketa can you take a look to see if this is the correct fix?


jjrsylvestre:
Huh? What browser are you using? Are you actually using the edited version of `pretext.js` from this PR?

Here is my scroll position when I navigate to `section-video.html#section-video-4` in the sample article: [image](https://sites.ualberta.ca/~jsylvest/pretext-dev/scroll-toc/scroll-toc.png)

jjrsylvestre:
@ascholerChemeketa OK, made some small adjustments based on your suggestions, but honestly it's not that big of a difference in the final scroll position, so not really sure why it's not working for you. Works for me.

ascholerChemeketa:
I am using Chrome. Currently only have 768 vertical resolution.

I realize that I only get that page on a reload (F5). When I open some other page, then navigate to the YouTube section of the ToC and navigating to it, I end up with the right part scrolled into view. But then if I F5 the page, I end up scrolled to the bottom. The `tocEntryTop` value on the fresh load is 747 and on refresh is 4394. I also got the incorrect position if I used history to navigate back to that location after navigating somewhere else. Which suggests a timing issue with calculating the position.

But, the new version appears to behave consistently on fresh load, reload, or history navigation.

Feel free to add a `hashchanged` event listener to update the active subsection on intra-page navigation. Or issue it and I'll check it out once I'm off the road.

PR #2641, opened 2025-08-06
Add `handout` division with expanded workspace options

oscarlevin:
This PR implements a new division, `handout` which mimics `worksheet` in that it gets a print preview for HTML and custom margins in PDF, but does not have the special treatment of exercises that a `worksheet` does.  Since a common use case for a handout is "guided notes" in which the instructor provides printouts of the day's notes with room for students to fill in pieces that will be done in class, we expand the elements that can accept `@workspace` to include almost all blocks, `p` and `li`.  Technically, this additional workspace opportunity is available in `worksheet` as well, but we could lock that down if we wanted.

Diffs after the first 7 commits show only that two "bugs" were fixed: previously, workspace was added to the copy of elements in (backmatter) `solutions` in HTML, and project-like were not getting workspace in LaTeX.  The refactoring of workspace fixed this.

Added examples to the sample article, which revealed some bugs with the print-worksheet css, which are now also fixed.

A later PR can do some cosmetic cleanup of the code (e.g., "worksheet" is used where now "worksheet/handout" or "printable" would be more appropriate), add info to the guide, and amend the schema (although worksheet is still only in the dev schema).

### DIFF
--- a/css/targets/print-worksheet/_chunks-worksheet.scss
+++ b/css/targets/print-worksheet/_chunks-worksheet.scss
@@ -23,7 +23,8 @@ $border-radius: 8px !default;
 .project-like,
 .remark-like,
 .openproblem-like,
-.computation-like {
+.computation-like,
+.assemblage-like {
   > .heading {
     display: block;
     font-size: 1.1em;
--- a/css/targets/print-worksheet/print-worksheet.scss
+++ b/css/targets/print-worksheet/print-worksheet.scss
@@ -77,7 +77,7 @@
 
 
 // Set font appropriate for printing
-section.worksheet {
+section.worksheet, section.handout {
   font-size: 11pt;
   font-family: var(--font-print);
 }
@@ -103,7 +103,7 @@ summary {
   display: inline;
 }
 
-details > div.knowl__content {
+.knowl__content {
   display: inline;
   margin: unset;
   border: none;
@@ -113,6 +113,11 @@ details > div.knowl__content {
 
 .knowl__link {
   all: unset
+}
+
+.code-inline {
+  border: unset;
+  background: unset;
 }
 
 .onepage article {
@@ -139,12 +144,12 @@ details > div.knowl__content {
   margin-left: 0;
 }
 
-section.worksheet > .heading {
+section.worksheet > .heading, section.handout > .heading {
   display: inline;
   font-size: 1.1em;
 }
 
-section.worksheet > .heading + .para {
+section.worksheet > .heading + .para, section.handout > .heading + .para {
   display: inline;
 }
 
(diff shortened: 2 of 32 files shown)


rbeezer:
Bad accident on nearby highway caused total closure leading to Tacoma Narrows Bridge.  Massive detours right through our (dead-end) neighborhood.  What better to do than abandon car travel to a bike trail and instead stay trapped at home and review this?  ;-)

In good shape.  Adding a new division is a major undertaking!

### Changes to existing conversions
* HTML: `hint` for `exercise-vector-addition-2` (and "pages" version) is getting workspace in backmatter solutions.  Should a hint have workspace? (Ever?)  Certainly we don't want workspace in solutions?
* LaTeX: some extra workspace on a couple actvities.  I think this is intentional (and not in solutions, but did not check, saw this first).
* Otherwise, I think this looks good in terms of not making any regressions.

### New Handouts
* HTML printing seems to look real good.
* PDF, Subsection 36.1, Derivative Rules: something wonky is going on, last bullets off in a second column?  36.2 looks good.

### Code
* Lots of code (variable names, comments) says "worksheet" when it now means "worksheet or handout".  Can we give this pair a name?  Like "printable" but not that.  "classroom-able"?  No, not that.  Something - just for the code.  Here's a non-exhaustive list of places I saw this:
* `worksheet/handout` looks like XPath to me.  `"worksheet" or "handout"` (3x)
* comment: "at least one worksheet and no other specialized divisions."
* `$only-worksheets`, `$all-worksheets`
* `mode="worksheet-margin"`
* Probably more - these were just in diffs (sometimes the context).  Let's clean these up - think about folks reading the code later and wondering what is being said versus what is really happening.

oscarlevin:
> HTML: hint for exercise-vector-addition-2 (and "pages" version) is getting workspace in backmatter solutions. Should a hint have workspace? (Ever?) Certainly we don't want workspace in solutions?

I'm hoping that you just switched the new/old when you were looking at diffs.  The current conversion to HTML left workspace on exercises, even in backmatter solutions.  This PR removed them (as a side effect of the refactor).

> LaTeX: some extra workspace on a couple actvities. I think this is intentional (and not in solutions, but did not check, saw this first).

This is the other bug that the PR fixed by accident.  Previously project-like in latex worksheets failed to have workspace respected.

So given this, except for the strangeness in pdf (which was already broken) and the cosmetic cleanup (which we need a word for and can be done later), I think this is good to go.

oscarlevin:
(By the way, for the weirdness in the first handout: #2665)

rbeezer:
> Double dog swear! Is that a thing?

Good enough for me.  ;-)

> just don't know what to call "worksheet/handout"

It'll come to you at some odd moment.  Or to me.  Or if not, you could risk appealling to the wisdom of -dev?

rbeezer:
Merged as-is.  Thanks.

I'll let you decide on limited announcement on -dev, or full rollout on -announce.  Now or a bit later.

PR #2642, opened 2025-08-06
XSL: fix bug preventing workspace in HTML when LaTeX pubvar set to no

oscarlevin:
The `sanitize-workspace` modal template lives in `pretext-common.xsl` as was using the value of the publisher variable `latex-worksheet-formatted` in deciding whether to allow workspace on an item (if the value was "no" then no workspace should appear in LaTeX).  This was bleeding through to HTML though, which also uses the template.

To fix, I removed the test on that publisher variable in `-common` and add a copy of the template in `-latex-common`.  The new template checks if the publisher wants latex worksheets formatted, and if so, applies the imported version of the template (otherwise it does nothing, which is the desired behavior).

This might cause a merge conflict with #2641, which I can resolve when appropriate.

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -6534,10 +6534,10 @@ Book (with parts), "section" at level 3
 <xsl:template match="&PROJECT-LIKE;|exercise|task" mode="sanitize-workspace">
     <!-- bail out quickly and empty if not on a worksheet    -->
     <!-- bail out if at a "task" that is not a terminal task -->
-    <!-- bail out if publisher file says to not format worksheets        -->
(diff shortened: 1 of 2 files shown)


rbeezer:
Nicely done with `apply-imports`.  Built 8 scenarios with sample article

*  LaTeX v. HTML
*  `@formated`: yes v. no
*  PR v. no PR

No changes in 7 scenarios, suspect situation restored the "missing" workspaces.  Merged as-is, except some light editing to whitespace (added for clarity/consistency, not deleting stray whitespace).

Thanks for your work on this while you had this all in your head.

PR #2644, opened 2025-08-08
Update documentation to reflect change in use of visual URLs

mitchkeller:
Commits here are likely more granular than you'll ultimately want, @rbeezer, but I figured giving you a commit history where you can squash was better than requiring you to disentangle commits if some things aren't getting merged.

There's a bit of a side commit here (b573dcb) to the sample article since my systematic review of all uses of `url` turned up an outdated URL to @davidaustinm's *Understanding Linear Algebra* that probably should not be squashed with the other sample article commit.


rbeezer:
Thanks - that was a lot of changes!  (I had suspected so.)

Greatly appreciate the granular commits, always easier to combine.   Today - just too spent to think about it, so left it as-is.  ;-)

PR #2645, opened 2025-08-08
Fix bug: ensure runestone profile menu opens above navbar in mobile view

oscarlevin:
@DavidCooperWCU noticed a bug with the CSS for the Runestone Profile (person) menu; it was opening downward in mobile view, which since the button was already at the bottom of the page was not particularly helpful.

We already have a selector for this case, but because the original selector is being included inside `.runestone-profile`, it's specificity was too low to override properly.  "Fixed" by adding `!important`, which I know isn't ideal, but seemed reasonable in this case.

Also fixed an invalid display property for the original selector.

@ascholer?

### DIFF
--- a/css/components/helpers/_buttons-default.scss
+++ b/css/components/helpers/_buttons-default.scss
@@ -53,7 +53,7 @@ $border-radius: 0 !default;
   position: relative;
 
   .dropdown-content {
-    display: hidden;
+    display: none;
     position: absolute;
     background-color: var(--dropdown-background);
     min-width: 160px;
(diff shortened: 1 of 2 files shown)


oscarlevin:
Thanks that's a really nice description of the options.  I'll update the PR with the suggested change.

oscarlevin:
Force pushed with the changes suggested by @ascholerChemeketa.

rbeezer:
My oops now.  Forgot to tag with PR number.  But merged as-is.

See 2c42ba4ae8a6ef2df4b2b49d4440aec0e445150e

PR #2646, opened 2025-08-09
Bare pg fixes

Alex-Jordan:
This is a follow up to #2631.

- Thank you for taking the time to fill in the gap of the missing PG file in #2631.
- This also adds the file that was missing for the minimal example.
- Now "bare PG" problems get new PG files written into the generated folder.
- The live version worked for me in testing, but it could have been that I had stale files in my generated folder from one of the previous iterations of this feature.  I just built [this one](https://spot.pcc.edu/~ajordan/temp/WWSC/sample-ww-chapter-4-2.html) cleanly and both problems work there.


rbeezer:
OK, that checks out nicely and as expected.  Thanks.

Please place new content **after** new code, it makes my testing harder (I know it makes your testing easier the other way around).   Just rote them after you are done as part of formin the PR.

rbeezer:
s/rote/rotate/

PR #2648, opened 2025-08-09
WeBWorK: process PROJECT-LIKE correctly

Alex-Jordan:
Note: This will have a small clash with #2646. Not sure which to process first, as I think the Active Calculus project is interested in both so that their WW server can transition to webwork.runestone.academy.

Up until this PR, `webwork` within a PROJECT-LIKE did not work out well in many scenarios unless the PROJECT-LIKE was specifically a `project`. Also the word "Project" was used in a hard coded way, rather than allow for localization or a custom name.

### DIFF
--- a/examples/webwork/sample-chapter/sample-chapter.ptx
+++ b/examples/webwork/sample-chapter/sample-chapter.ptx
@@ -343,17 +343,31 @@ ENDDOCUMENT();
                     <webwork><xi:include parse="text" href="pg/Adding_SingleDigit_Integers.pg"/></webwork>
                 </exercise>
 
-                <project label="copy-webwork-add-numbers">
-                    <title>Inside a <tag>project</tag></title>
-                    <introduction>
-                        <p>
-                            If you like, you can have a <webwork/> inside a PROJECT-LIKE block.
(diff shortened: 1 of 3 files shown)


rbeezer:
*  Diffs look good, though with new `subsection` and `exploration` there's a lot of noise with new numbers, so I could have missed something.
*  I added a `title` for the new `subsection`.
*  Diff for "problem sets" seemed lacking?  I didn't see the new `exploration`.  The only change I saw was
```diff
+problem_start
+source_file = Integrating_WeBWorK_into_Textbooks/1-Instructive_Examples/1_1-Arithmetic/1_1_1-Adding_SingleDigit_Integers.pg
+problem_id = 
+problem_end
```

oscarlevin:
Hi @Alex-Jordan, this still doesn't seem to be working.  @Tanaquil18 and I are trying to build with the CLI (which now has this PR), but we are still not getting any project-like workwork pg files being generated.   Not sure how you were able to get those in your demo.  

It does look like the contents of these projects are going into "extracted-pg.xml", but they are not being made into their own .pg files.

rbeezer:
Please make a PR!  Thanks for sorthing this one out.

PR #2649, opened 2025-08-10
Css fixes for colorado themes

oscarlevin:
This fixes the tombstone placement for ends of proofs in theme-denver and theme-greeley, fixes an issue with wide elements not being placed correctly in theme-greeley and theme-boulder, and tweaks the navbar and page shadow to give a cleaner look.

Will need a follow-on theme generation commit.

### DIFF
--- a/css/colors/_palette-dark.scss
+++ b/css/colors/_palette-dark.scss
@@ -91,7 +91,7 @@ $colors: map.merge(
   // Buttons & Widgets
   "button-background": var(--background-color-gray-15),
   "button-border-color": var(--background-color-white-25),
-  "button-hover-background": var(--primary-color),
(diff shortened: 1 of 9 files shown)


rbeezer:
No real testing, and just two minor tweaks to commit messages.  @ascholerChemeketa likes his "Chemeketa" handle?  Might not make a difference.

I was in the mood to clear this, so @ascholerChemeketa can revert and modify if he feels strongly...

rbeezer:
Forgot to say: I'll let you clear out the "end-of-proof signs" thread that was addressed here.

ascholerChemeketa:
Yeah, @ascholer goes to a different email. @ascholerChemeketa is the one that I associate with PreTeXt/Runestone stuff.

I like the dark mode updates.

PR #2654, opened 2025-08-10
CSS: always center image-archive links

ascholerChemeketa:
The only place these really show up in examples is in SA SBSs. There they have always been centered because of the old sbspanel code. New sbs css does not cause that centering.

I think we do want that centering, and it should be applied universally, not just in sbs.


rbeezer:
Yes, those archive links don't seem to have been too popul;ar.  Thanks for fixing them up all the same.

PR #2657, opened 2025-08-11
HTML: override pubfile specified legacy css themes

ascholerChemeketa:
Substitutes out any specified legacy CSS theme with closest modern one. Emits warning that change is being made.

This code runs in the `html-theme-name` variable setting where the selection logic already is. There still is a "standard" deprecation message that comes from the pubfile variable itself. 

I don't think two messages hurt... Eventually the whole "legacy" selection block in `html-theme-name` maybe goes away and any unrecognized theme including legacy ones becomes `default-modern`. At that point we just leave the standard pubfile deprecation message. 

I think this code is just for the interim period where the deprecation is going from voluntary compliance to enforced compliance.

Needs release coordination with @oscarlevin to make freeze point clear for anyone who needs a legacy style for longer. And possibly an announcement at that point.

### DIFF
--- a/xsl/publisher-variables.xsl
+++ b/xsl/publisher-variables.xsl
@@ -2329,26 +2329,39 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <!--                              -->
 
 <xsl:variable name="html-theme-name">
+    <xsl:variable name="warning-message">PTX:WARNING: The "FROMSTYLE" style requested in publication/html/css is deprecated. Your book will be built with theme="TOSTYLE". See the "Styling" section in Ch 29 of the Guide for other options.</xsl:variable>
     <xsl:choose>
         <xsl:when test="$publication/html/css/@theme">
             <xsl:value-of select="$publication/html/css/@theme"/>
         </xsl:when>
         <xsl:otherwise>
-            <!-- legacy style detection -->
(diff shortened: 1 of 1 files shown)


rbeezer:
@oscarlevin.  How about?
*  We clear the whole WW activity bug.
*  I push this.
*  You make a release.

rbeezer:
Merged!  Thanks, @ascholerChemeketa!

Can you do the `-announce` drill?  Perhaps cast this as "the new themes have seen some iteration and are now in good shape.  So the transitional phase keepin the old ones is over."  In addition to the jankification that was bound to happen and is now happening.  Point to The Guide for where folks can leanr their options and what to do.  Thanks.

rbeezer:
@oscarlevin - I think you are clear to make 2.26.0 at this juncture.  I picked up your WW CSS for Denver.

oscarlevin:
@ascholerChemeketa , you can say that this is part of 2.27.0 of the CLI

ascholerChemeketa:
@oscarlevin Yep. Typing it right now.

FYI - I think the changelog for this has a few typos:
`Lagacy themes for HTML will not be automatically replaced with their new equivalents.`

Lagacy -> Legacy
will not -> will

oscarlevin:
Oops, I was working too fast.  I'll fix those eventually.

PR #2658, opened 2025-08-11
WeBWorK: restore PROJECT-LIKE processing

oscarlevin:
This restores an edit that was part of #2648 but somewhere got reverted when fixing a merge conflict.  This should now correctly create the `.pg` files for source-authored webwork problems inside project-like.


Alex-Jordan:
The change from this should only be new files appearing in the `my-generated/webwork/pg/` subfolder tree. And the sample chapter `my-generated` folder is in `.gitignore`. So you won't see any differences using git. My own routine is I make copies of the `my-generated` folder after each build and use regular `diff` to see any changes.

rbeezer:
Aah, moving too fast.  May have forgotten "--no-index" on my use of "git diff", which acts as if you aren't inside a repo.  Will retry in the AM.  Thanks.

PR #2659, opened 2025-08-12
Script: add longer interactive preview timeout "method"

oscarlevin:
Sometimes an `interactive` needs to load external resources (such as Doenet) which takes longer than the 5 seconds we give for loading the page before taking the screenshot.  This adds an option as `-M --method` on the `preview` component, with values "fast" (default) or "slow" (10 second timeout).

The CLI will also get a flag that has the same effect.


rbeezer:
Behaves as advertised, with no ill-effects.

We can add new speeds if necessary, with new adjectives.  Numerical values could be a future option.

Adjusted a couple of code comments that were now a bit stale, plus added one extra space in the message about using a bad value for the method.  Otherwise, merged as-is.

Sorry for the delay getting to this one.  Thanks for working up this one.

(You are responsible for the signature change on the one routine, so you know what to do!)

PR #2662, opened 2025-08-12
CSS: don't apply extra styling to activated webwork exercises

oscarlevin:
The upgrade to the 2.20 webwork server means the structure of an activated webwork problem includes classes `exercise exercise-like`.  In the `denver` theme, these are selected to display the "C-box" styling for inline exercises.  This was bleeding through to this case, so I added another parent to the "don't apply this to that" list.

If this can get merged before the next release, I'd appreciate it.


oscarlevin:
Thanks for squeezing this one in.

PR #2663, opened 2025-08-13
Use the code-wrapper template to process tag/tage/attr

skiadas:
tag/tage/attr were using a "wrap in c and reevaluate" approach that resulted in double-escaping and needless processing especially in latex. This attempts to simplify the call by simply using `value-of` on the text content of the tag rather than apply-templates, and using the code-wrapper template instead of "wrapping in c and reevaluating".

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -10521,37 +10521,34 @@ http://andrewmccarthy.ie/2014/11/06/swung-dash-in-latex/
 
 <!-- A tag, with angle brackets and monospace font -->
 <xsl:template match="tag">
-    <xsl:variable name="the-element">
-        <c>
+    <xsl:call-template name="code-wrapper">
+        <xsl:with-param name="content">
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, @skiadas!

I have almost exactly this on a branch!  But I fixed up some bad uses in the sample article and in the Guide.  And I've been dragging my feet on some additions to the documentation.  Tell you what - I'll merge mine, but make you the author on the part you've done here.

Rather than
```xsl
<xsl:value-of select="text()"/>
```
I've done
```xsl
<xsl:value-of select="."/>
```
Not sure which is superior.  Latter has definitely been my style consistently.

Alex-Jordan:
XML element and attribute names can have underscores. Will using value-of escape them as needed in LaTeX output? (I haven't checked, just thinking about the one potentially problematic character that is legal in an XML name.)

rbeezer:
Documentation has long said:

> the content of these elements should only be the 26 lower-case letters and a dash/hyphen.

PR #2664, opened 2025-08-14
Script: use log messages instead of print commands

oscarlevin:
In a recent webwork generation, I saw some errors about problems being empty.  But these were not showing up in the saved the logs.  This fixes that particular print statements, as well as one in the citations function.


rbeezer:
Thanks for catching that TODO!  Merged as-is.

Tried to provoke that error for CiteProc.  No luck.  Perhaps our processing makes it impossible?  Anyway, can't hurt and is a required argument to one of the methods.

PR #2667, opened 2025-08-19
Common: fix backmatter reading solution selection

oscarlevin:
While preparing a PR to cosmetically clean up the worksheet|handout comments, I ran into what appears to be a copy/paste error for the solutions-generator.  Only noticed it since I was searching "worksheet" and it looked out of place.


rbeezer:
Yep.  Amazing that has been there for so long.  ;-)  Not stressed by the sample article, but I think I will punt on adding an example, since the mistake was just cit-and-paste gone bad.  Thanks!

PR #2669, opened 2025-08-19
Improve floating aside handling in lists and multi-col environments

ascholerChemeketa:
Addresses floating aside issues in trickier locations as identified by @jjrsylvestre on -dev.
* Disable floating behavior in multi-col environments
* Avoid list title bleed through to asides (or anything else with a `.title` in `li`s


rbeezer:
Merged as-is.  (I'm still not wild about `asside` down inside lists and exercises...)

I'll let you announce on the thread with Jeremy.

PR #2670, opened 2025-08-19
CSS: unify inline code spacing

ascholerChemeketa:
The prism styling currently overrides some of the spacing that `.code-inline` tries to set. This unifies the padding of `.code-inline` that is or is not getting language syntax highlighting.

You should currently notice a slight difference between highlighted and unhighlighted fragments here:
https://pretextbook.org/examples/sample-article/html/section-programs.html#section-programs-21

After applying those all have the same visual padding.

### DIFF
--- a/css/components/chunks/_codelike.scss
+++ b/css/components/chunks/_codelike.scss
@@ -4,6 +4,11 @@
   line-height: 1.2;
 }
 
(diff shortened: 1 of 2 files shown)


rbeezer:
Very good!  merged as-is.  Thanks.

PR #2671, opened 2025-08-20
Add <veil>, an inline reveal/hide element for guided notes

geoff-cox:
## Summary
This PR introduces a new PreTeXt element, <veil>, for hiding inline and block content until the button labelled "reveal" is clicked. This version focuses more on the HTML side of the element and only offers minimal support for print.

## HTML
 - Renders a compact “reveal” button.
 - On click/Enter/Space, hidden content is revealed inline with a thin accent border; clicking inside the revealed area hides it again.

## LaTeX/Print
 - Inline veils print as their content underlined via \uline{…}.
 - Multiline/block veils are printed plainly (no underline).
 - Veils inside math are not supported yet; content is printed as-is with a warning.
 - Hiding content in print is not supported yet. Further dev discussion is necessary.

## Testing
 - Added a section in the `sample-article.ptx` containing a comprehensive list of test samples.

No schema or publisher-variable changes are included in this PR; those can follow in a subsequent change.


geoff-cox:
@rbeezer 
Thanks for getting back to me. I wasn't able to make it into the drop-in this week. I'll try for next week. I am thinking the easiest approach would be to start with a fresh clone of PreText and resubmit the PR.

rbeezer:
@geoff-cox - will you be in Ann Arbor again this June?  We could restart this together then.  I think it really just complications with `git` that are holding this up.

geoff-cox:
@rbeezer I totally forgot about this. Yes, I will be there. 

This shouldn't be hard to fix. Maybe, I can take a look at it sooner.

PR #2673, opened 2025-08-21
CSS: codelike blocks always get background color

ascholerChemeketa:
Add some kind of background color to all codelike elements.

Absence of a background color sometimes makes it hard to see the extent of the block (and what the copy to clipboard will apply to):
image

And is sub-optimal when they are used in a container with a background color:
image


oscarlevin:
Thanks!  Swapped order of commits, but kept content as-is.

PR #2676, opened 2025-08-23
QR codes for interactive programs

ascholerChemeketa:
This adds QR code generation for interactive programs (activecode/codelens).

Because programs already may have width issues in print, I don't want to do the link side by side with the program itself. So instead, an assemblage with a link/qr is added after the program. I think it is good to have that block visually separated as a "here is something not part of the core text". But I feel no real ownership of that layout.

image

I figure you will want to do your own test of generation, so no commit with the updated QR codes.

First commit fixes a breaking bug if version 8+ of qrcode is installed:
https://pypi.org/project/qrcode/
If the rest of this ends up languishing, it might be worth breaking that commit out.


ascholerChemeketa:
Draft while working on:
https://github.com/PreTeXtBook/pretext/pull/2682

ascholerChemeketa:
Updated based on results of `<pi:localize>` work.

Content since last review:
* Individual xml sidecar files for urls
* Use pi:localize
* Dropped commit moving static-url templates from -assembly to -html (see below)

I kept the `extraction-wrapper` override in `extract-qrcode` as it is the best way to have a single decision based on `$baseurl` to either generates error message or do work.

I believe `dataurl` is the now the only thing that still relies on `"static-url"` templates during static build. So moving the `"static-url"` templates needs to wait on a solution for it.

ascholerChemeketa:
Force pushed to rebase.
Two new commits at end. One to change structure of the SBS. Once we pick a final format I will merge it into other commits. Other has hello world testing setup. It will get dropped.

Here is a:
```
sbs
   paragraphs
      p
         em
            title text
      p
          url
```

A `title` instead of the first paragraph would be more semantic, but looks ugly when rendered with an inline table.

I think this is going to be impossible to make pretty in assembly given the current tools.

ascholerChemeketa:
Sounds good. Force pushed that structure with `@pi:indent="no"` on the `<p>`s.

rbeezer:
Working on this one.  I'm leery of the import of `pretext-html.xsl` into the focused stylesheet for extraction.  And we need a more general fix, since I would like links to "in situ" versions of `interactive`, etc.  See #2819.

I've got more to look at, so do not add anything here right now, might be a day or two...

ascholerChemeketa:
Roger.

Not so sure about the standalone page. It would break programs that use other elements on the page as data files of code to include. And assigned exercises already end up on an out of context page.

For purposes of the QR destination, I personally would want people to end up on the relevant book page, not in a one-off for that program. Both on principle and because the standalone pages don't currently make it easy to get back to in-context - we expect someone navigates to them from the book, not starts there.

oscarlevin:
I believe due to this PR, the cli is throwing errors when it tries to build html that contains interactive elements.  

```
error: processing with /home/oscar/.ptx/2.38.3/core/xsl/pretext-html.xsl has failed
critical: Cannot resolve URI /files/GitHub/dev/pretext-cli/new-pretext-project/interactive/generated-assets/qrcode/interactive-infinity-url.xml
critical: Failed to build without errors.  Exiting...
```

Is it true that the html build now requires QR codes to be generated?  Of course the CLI builds QR codes for static formats, but until now we haven't done so for html.

Anyway, adding qrcodes to the list of assets to build for HTML seems to fix the issue.  So I'm happy to add that, but wanted to make sure this was the intended behavior.

Also, do I need to check the source for `.//program[@interactive]` in case the source doesn't have other interactives?

rbeezer:
The `qrcode` format is now also making files of URLs based on the base URL.  I can't think of why we are trying to read those files for an HTML build.  Must be a new `exsl:document()`.  I might be able to investigate in a few hours.  So maybe CLI should generate as a short-term measure?

oscarlevin:
That's what I've done, so no rush on this.

ascholerChemeketa:
I had noticed that requirement. It is not directly related to QR codes. The requirement comes from reading url's for `audio|video|interactive[not(static)]` from the new sidecar files that was added in
6672af74419ecbecaf8a50ff35ea63c397ad5d77

I had meant to bring it up. It would be nice not to have HTML depend on those sidecars. If you are in a tight edit loop with something like Codechat or nodemon doing partial rebuilds and you add an audio/video/interactive, you now have to break out of your normal rebuild and go trigger a qr code extration.

rbeezer:
I think I can write those URLs as attributes, but since the computer chunk-level variable will be overridden by a conversion, it will mean parameterizung the containg-fikename template.

Though maybe the real problem is that the standalone-filename is in the problematic file, that does not need to know anything about chunking.  I believe this is the heart of the bug.

rbeezer:
>  I can't think of why we are trying to read those files for an HTML build.

The read/opening of the sidecar files was placed well before it was needed, and should have been buried in part of a `xsl:choose`.  Fixed at 05f38f327fb198ab27da5c9b49353c2b70793e0e.  So, @oscarlevin, you should be able to let your guard down on the HTML builds.  Thanks for catching this one.

For the first time, I'll blame Claude.  Or maybe Claude's supervisor.

rbeezer:
And in the FWIW category, it seems the `qrcode` package, which I must have updated to v8.0 for this PR, is making images about a quarter of the previous size.

PR #2677, opened 2025-08-24
Improve `dl @width` options

oscarlevin:
This started as a PR to just fix a bug in which `dl` with `@width="narrow"` didn't display correctly in HTML.  That is accomplished here.

The guide also claims that `@width` can have value "wide", but this never seems to have been implemented.  So I did that for LaTeX and HTML.  And then also added an example to the sample article.

There is a small inconsistency between LaTeX and HTML for narrow though: in LaTeX, a narrow `dl` that has a `dt` too big to fit in the allocated `11ex` will make the `dd` start on the next line.  I think this is how the narrow option was supposed to work for HTML, but the CSS that was there was broken so I simplified it to always keep the `dd` on the same line.  In HTML, the narrow `dt` is also flush right (same as the other widths), but this is not true for LaTeX.

There are other things to do for `dl`.  We have talked about a slightly different style for the nested terms.  And the nested terms in LaTeX have strange spacing (or at least, not consistent with HTML).  There is also still an issue with `dl` as the first child of a `<paragraphs>` that  needs work in LaTeX.  All this for another time.

### DIFF
--- a/css/components/elements/_description-lists.scss
+++ b/css/components/elements/_description-lists.scss
@@ -69,33 +69,23 @@ dl.description-list {
     margin-left: 22ex;
   }
 
-  .narrow {
+  &.narrow {
     dt {
-      margin-top: 0;
-      width: unset;
-      max-width: 55ex;
-      text-align: left;
+      width: 11ex;
     }
 
     dd {
-      margin-left: 12ex;
-      margin-left: 0;
-      margin-top: 0;
-      width: 31em;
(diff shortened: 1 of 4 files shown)


rbeezer:
CSS: 11ex, 15ex, 25ex
HTML: 11ex, 18ex, 25ex

Conscious, or no?  I may procede with review and merge.  So just let me know and I can hand-edit, if necessary now or later. (I.e, don't do a force-push or folllow-on commit).

PR #2680, opened 2025-08-25
Favor publisher value for HTML toc focused mode over theme value

ascholerChemeketa:
Second commit addresses:
https://groups.google.com/g/pretext-dev/c/PqQfGGJXKYA/m/1JuWVrw1AQAJ?utm_medium=email&utm_source=footer

While testing noticed that not specifying any theme/style gives a warning that theme `""` is not supported and that default modern will be used. First commit fixes that.


oscarlevin:
I think these both look good, and am happy to fully test and merge later today.  But note that these are changes to `publisher-variables.xsl`, so let me know @rbeezer if you want to handle this instead.

PR #2681, opened 2025-08-26
Add appendix contents to runestone manifest

ascholerChemeketa:
This adds `appendix` items to the RS manifest. They can contain:
* datafiles
* source elements for programs that are linked to from elsewhere (same items that are written as `<source>` in chapters)

They do not contain any `<questions>` as those require a chapter with an integer chapter number. An appendix with exercises etc... would require reworking data assumptions in RS.

### DIFF
--- a/examples/sample-book/backmatter.xml
+++ b/examples/sample-book/backmatter.xml
@@ -81,6 +81,20 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         </section>
     </appendix>
 
+    <appendix>
+        <title>Some Runestone Elements</title>
+        <p>These elements should end up in the Runestone manifest.</p>
+        <datafile filename="datafile-appendix.txt" xml:id="datafile-appendix" label="datafile-appendix">
+            <pre>
+            Ipsum lorem...
+            </pre>
+        </datafile>
+
+        <program language="py" xml:id="program-appendix" label="program-appendix">
+            print("Hello from the appendix!")
+        </program>
+    </appendix>
+
     <xi:include href="./gfdl-mathbook.xml" />           <!-- GFDL License text -->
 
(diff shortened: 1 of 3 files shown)


rbeezer:
There are (of course!), three empty `<appendix/>` elements in the manifest.  At a minimum, maybe RS should be warned - this sort of thing has been overlooked before.  Please checkoff that this is what you really want to have happen and I can merge.  I know this material really won't be seen as it enters and leaves the manifest on the RS side.

rbeezer:
I'd love to see just a bit of context in the new appendix:  "Here is a data file", "Here is a bit of a program."  "You might want to put library code back here like this."  That sort of thing.

And perhaps the Guide can suggest things like `datafile` get collected in an `appendix`, maybe even structured with `section` if there is a lot of stuff?

These documentation additions can go on a new commit, don't hold up this one.

rbeezer:
> I'm not sure I agree in general 

I tend to operate with rules-of-thumb that include "fewer parameters is better."  Necessary for recursion, sometimes necessary for switching behavior.

Differentiating variance in template behavior by a finer set of `match` is one avenue.  You can imagine that here:
```xsl
match=datafile[ancestor::appendix]
```
which be a bad idea, since the companion template would be nearly identical.

Or differentiating variance by location/surroundings.  The suggestion here: look up the tree to see where you are, because that makes a difference.  And then I think templates are "less connected", more standalone, independent - and hence more reliable.

rbeezer:
> Let me know if you like that for similar situations in the future.

That is certainly fine, but also not always necessary.  Always easier to combine commits rather than explode one.

Merged as two commits.

Thanks for this - it'll be a nice way to move "unimportant" stuff off into a corner.  Advertise as you see fit - I'll put this PR on the thread in -dev.

PR #2684, opened 2025-08-28
CSS: Add custom styling/theming project to examples

ascholerChemeketa:
Two purposes:

1) We don't encourage customizing CSS, but just about everyone "in the know" does it. And many not in the know yet ask about it. So this has simplified examples of how to add custom CSS and how to do it in a way that interacts well with the theme (e.g. making use of the "primary-color" defined by the theme). No more needing to point to @oscarlevin's book as a sample. 😄 

2) There probably should be an in-house example of a fully custom theme. @trshemanske did a great exploration of creating a `tacoma-crc` (https://github.com/trshemanske/tacoma-crc) that I have pointed to in the past. But this gives us a sample of how to do so that can be kept up to date with any future changes.

### DIFF
--- a/examples/custom-theming/README.md
+++ b/examples/custom-theming/README.md
@@ -1,0 +1,24 @@
+#  Styling/Theming Samples for PreTeXt
+
+This project has samples to illustrate how the HTML output of PreTeXt can be customized. The `projects.ptx` file defines the following build targets:
+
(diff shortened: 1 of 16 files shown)


oscarlevin:
This is very helpful and well thought out.  Merged with a slightly reworded commit message.

PR #2686, opened 2025-08-31
CSS: Fix inconsistent worksheet inline headings

oscarlevin:
I was noticing some strange heading/number displays for exercises in a worksheet.  Sometimes the paragraph started inline, and other times not.  Two things were at play that are fixed here.

1. in denver and related themes, I was not including the inline-header-mixin.

In c29101d636f279412f3f7771f3d372c9c3fedf41, part of the changes removed a `display: inline` rule for 
```
.para.logical > .para:first-child
```
This meant that the first line of exercises that was in a `para logical` would be put on the next line after the exercise number.  

@ascholerChemeketa, could the removal of that rule (in `_spacing.scss`) have been unintentional? If not, then I don't know what else it was trying to fix, so I'm not sure if my adding it back in will break something else.

### DIFF
--- a/css/components/chunks/helpers/_inline-heading-mixin.scss
+++ b/css/components/chunks/helpers/_inline-heading-mixin.scss
@@ -9,16 +9,20 @@
       content: "\2009";
     }
 
-    & + .para {
+    // Following sibling also needs to be inline for heading to line up correctly
+    // For now limit to logical paragraphs and introductions as they seem to cover
+    // the plain text cases. Not clear we would want same treatment for other
+    // kinds of content
+    & + :is(.para, .para.logical, .introduction) {
       display: inline;
-    }
-
-    & + .introduction {
-      display: inline;
-    }
-
-    & + .introduction > .para:first-child {
-      display: inline;
+      // possible first child that is the actual text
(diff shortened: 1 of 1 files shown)


ascholerChemeketa:
The comment in the original line (`// make sure first para child of logical paragraphs doesn't get extra space`) indicates that rule was about preventing extra space. That effect is no longer needed based on the other changes in c29101d636f279412f3.

Normally `.para`s are blocks. I think it is best to leave `.para.logical > .para:first-child` the same unless there is a clear reason to change it and then try to limit where that is changed.

What about adding something like:
```scss
    & + .para.logical > .para:first-child {
      display: inline;
    }
 ```

To the `_inline-heading-mixin`. There is a similar construct for introductions there. We could combine the two into something like:

```scss
    // Following sibling also needs to be inline for heading to line up correctly
    // For now limit to logical paragraphs and introductions as they seem to cover
    // the plain text cases. Not clear we would want same treatment for other
    // kinds of content
    & + :is(.para, .para.logical, .introduction) {
      display: inline;
      // possible first child that is the actual text
      & > .para:first-child {
        display: inline;
        }
    }
 ```

PR #2688, opened 2025-09-04
HTML banner customization variables

ascholerChemeketa:
Subtitle and byline may not always make sense to render in banner.

Sample reasons someone may want them off:
* Long subtitle
* Lots of authors (physical space issues)
* Book is more a collection of content than an authored work

This adds publisher variables to disable rendering of both items in the banner. And reduces by 2 the number of reasons for someone to ask about how to add custom CSS to their project. 😄

### DIFF
--- a/examples/sample-article/publication.xml
+++ b/examples/sample-article/publication.xml
@@ -144,6 +144,9 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
         <feedback href="http://example.com/not-implemented.html"><!-- My Button Text --></feedback>
         <!-- A "portable" build is meant to be a single file that is transportable -->
         <!-- <platform portable="yes"/> -->
+        <!-- Banner options, these default to yes but can be set to "no" to -->
+        <!-- remove the indicated element from the top banner               -->
+        <banner byline="no" subtitle="no"/>
     </html>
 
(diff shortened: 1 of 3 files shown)


rbeezer:
This looks really good, but no publisher variable goes in without documentation.  That has served us well.

General blurb into

[Section 29.1: HTML Publisher Options](https://pretextbook.org/doc/guide/html/html-publisher-options.html#html-publisher-options) 
https://pretextbook.org/doc/guide/html/html-publisher-options.html#html-publisher-options

Terse mechanical specifics into

[Section 44.4: Online (HTML) Options](https://pretextbook.org/doc/guide/html/publication-file-online.html#publication-file-online) 
https://pretextbook.org/doc/guide/html/publication-file-online.html#publication-file-online

Then incude two `xref` pointing from one to the other.

ascholerChemeketa:
Updated with documentation.

rbeezer:
Thanks very much for the documentation.  Very nice work on this whole PR.  Merged as-is, but for
* Switched sample article settings to `yes`
* Added one sentence to the general write-up in the Guide, to the effect that these items will still show up on the title page, etc.  When I first saw this I initially forgot why anyone would bother having them in their source if they were not ever visible.  :-(

PR #2692, opened 2025-09-06
Guide: how to spell check HTML output

Alex-Jordan:
While I could have spell checkers in my editors, I haven't put in the time to configure them for XML. Let alone for chunks of other code like latex-image or pg-code. Instead I use what's described here in Firefox, which makes a quick-and-dirty spell checker on the output pages. If I'm reviewing something in final stages, I run this on each page as I review. I thought this might be of some general interest for the guide.

### DIFF
--- a/doc/guide/appendices/editors.xml
+++ b/doc/guide/appendices/editors.xml
@@ -919,5 +919,26 @@
             </subsubsection>
         </subsection>
     </section>
-
+    <section xml:id="subsection-spellchecking-output">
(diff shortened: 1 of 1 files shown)


rbeezer:
Interesting.  ;-)  Thanks for the addition.

I split the incantation across three lines at the semi-colons.  So it would render properly in PDF output (which is stalled at the website).  I know that breaks a nice copy out of HTML output with the new clipboard function.  And while I was there I capitalized JavaScript properly.

PR #2696, opened 2025-09-09
Program formatting logic consolidation followup

ascholerChemeketa:
Consolidation and WS fix requested here: https://github.com/PreTeXtBook/pretext/pull/2630

Program formatting logic tested against SA and SB. Clean on both.


rbeezer:
Took a quick peek.  I like the XSL-thonic creation of the boolean parameter!

PR #2701, opened 2025-09-18
STACK: Fix relative links returned by API in JS library

geoo89:
The STACK API sometimes returns links to assets that should be displayed in the question.
These links are relative to the URL of the STACK API; however, when directly rendering them on a website, they will be interpreted as relative to the website, and thus not work.
This fixes it but prefixing the URL of the STACK API.

This PR build on top of #2693 in order to avoid merge conflicts (both PRs touch the same part of the STACK JS file). Only the last commit is new. Merging this PR will also merge #2693


rbeezer:
Merged as-is.  Thanks.

PR #2708, opened 2025-09-26
WeBWorK in Runestone: let books use generated ww from other books

oscarlevin:
This adds a "baseCourse" data attribute to each webwork problem and then sets the javascript to use that if it can to specify the path to the .pg file for the problem when activating.  

The javascript should only have any effect if viewed in runestone.  The `data-baseCourse` is only added when `data-problemSource` is, which should only happen when there is local .pg files to be loaded.

Draft for now since I am not in a position to test this fully yet, but @Alex-Jordan, I'd love to get your thoughts when you have a chance.

### DIFF
--- a/js/pretext-webwork/2.20/pretext-webwork.js
+++ b/js/pretext-webwork/2.20/pretext-webwork.js
@@ -16,6 +16,7 @@ async function handleWW(ww_id, action) {
     const ww_processing = 'webwork2';
     const ww_origin = ww_container.dataset.origin;
     const ww_problemSource = ww_container.dataset.problemsource;
+    const ww_baseCourse = ww_container.dataset.basecourse;
     const ww_sourceFilePath = ww_container.dataset.sourcefilepath;
     const ww_course_id = ww_container.dataset.courseid;
     const ww_user_id = ww_container.dataset.userid;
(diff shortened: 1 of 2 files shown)


bnmnetp:
The Javascript looks good to me.

oscarlevin:
Tested outside of Runestone and it looks like it is working fine there.  So as soon as @Alex-Jordan signs off, I think this is ready to go.  

The sooner the better for getting this feature available, although I don't think it raises to the level of emergency or anything like that.  Books that want to make their source-authored webwork available for other courses will need to rebuild anyway.

Alex-Jordan:
Sorry for not checking on this. As you may have heard, I now live in a war-ravaged city, so there's just a lot going on :)

I haven't tested. Thanks for seeking my input/review, and I'm comfortable with things if everyone else here is. My only thought while browsing the code changes are, what happens to a project that has not declared a document-id?

bnmnetp:
A project without a document-id will be categorically rejected by the Runestone BDFL.

Alex-Jordan:
More thoughts/questions. Which maybe have simple answers.

1. This adds `data-baseCourse` to the HTML, even if this is not an HTML build for Runestone. Would it be better to add this as `data-documentID` (or something like that)? It's coincidentally (in a sense) the same as a Runestone basecourse, but that is something the js can deal with within the block where the js is only active if it's a Runestone build.
2. Any reason not to make the regex more specific, to match on something the ends with `-ww-rs`?

oscarlevin:
Currently underscore is NOT banned, but @bnmnetp and I think we should do this.  Brad checked and there are currently no books on Runestone that use WeBWorK and contain an underscore in the document-id.  

Any book built after this PR is merged will never need to parse the id to get the basecourse for Runestone.  So we can even deprecate this check after all the current Runestone books are updated.

Alex-Jordan:
Officially, xml:id~s should only have alphanumeric, including underscore, and hyphen. It would be nice to have one fewer set of rules like that to remember, and put the same limitations on document ID. Just a slight preference, but I'm not following why it's a concern.

Note that document ID is used for at least one thing outside of Runestone. (The neglected catalog of PreTeXt projects.)

PR #2709, opened 2025-10-11
Runestone: add exercise hints to runestone-manifest

ascholerChemeketa:
Currently, if an exercise has a hint, that is not visible on assignment pages. Although a student can hit "Show this question in the book", there is no indication they should do that.

This allows hints to be added to the RS manifest where they will be displayed on the assignment page (see screenshot). They also will appear on other pages that use the manifest (grading).

It does not affect rendering in the normal book pages.

Note that changing this flag only makes it possible for an exercise with a `<hint>` to add a "solutions" block containing the hint to the manifest version of the exercise. Exercises without hints will not get any extra content.

image

image

Heads up @bnmnetp - no changes needed to RS

### DIFF
--- a/examples/hello-world/hello-world.xml
+++ b/examples/hello-world/hello-world.xml
@@ -38,7 +38,41 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- 6.  Experiment with other available conversions                                    -->
 
 <pretext>
-    <article xml:id="hello-world">
-       <p>Hello, World!</p>
-    </article>
+    <book xml:id="hello-world">
+
+      <chapter>
+        <title>Hello World</title>
+
+        <section>
+          <title>First Section</title>
+          <p>This is a bare minimum example of MathBook XML.  It has a chapter, a section, and a paragraph.</p>
+          <exercise label="inline-ex">
+            <statement>
+              <p>
+                Do something.
+              </p>
+            </statement>
+            <hint><p>Try harder.</p></hint>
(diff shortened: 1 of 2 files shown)


ascholerChemeketa:
Not ready to merge, but could use some feedback...

As I dug in deeper on determining pub values for exercise components, I started getting nervous about the very similar but not identical constructs in use for determining what kind of exercise something is and what components it should display.

I started attempting to consolidate that logic. Did so for RS and latex.

I'm a bit scared to touch Webwork. And not 100% sure how this interacts with dry-run.

So, when you get a chance to just take a quick glance, please let me know if I should:

1) Backtrack and just put new logic in pretext-runestone for the manifest. Don't try to consolidate the logic.
2) Looks fine, clean it up and submit for real. Add comments to Webwork and any other spots using similar logic that next time they are touched, they might want to leverage the same logic.
3) Looks like a good idea, but we should udpate dry-run and/or webwork to rely on same logic now.

ascholerChemeketa:
OK, cleaned up the changes.

Existing commits have a no-change refactor, a commit that should only cause changes in RS manifest, a commit that should result in no changes in Latex, but consolidate logic, and a testing commit with hello world changes that you can drop before merging.

rbeezer:
> I'm going to go with uncovered an existing bug.

Good call.

Yes, it'd be nice if we could deprecate this (with a limited-time warning?), that'd allow us to remove these instances from the sample article (and move this improvement forward).

PR #2710, opened 2025-10-19
HTML: no permalinks in multiple choice statements

ascholerChemeketa:
Trying to render permalinks in multiple-choice problems ends up obscuring and preventing interaction with the label that should be a clickable way to select the desired option.

2025-10-19 09_46_18-doctest-1b cpp - welcomeprogramming - Visual Studio Code

Rather than try to do some special rendering logic, it seems better to just not render permalinks in that context.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -11666,6 +11666,10 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- appear at  https://github.com/PreTeXtBook/pretext/pull/2534    -->
 <xsl:template match="feedback/p" mode="permalink"/>
 
+<!-- 2025-10-19: "p" in the label vor a multiple choice option overlaps -->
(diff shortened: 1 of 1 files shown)


bnmnetp:
Where are you seeing these?  I though we had eliminated them with CSS both in context and on the assignment page.

ascholerChemeketa:
Ahhh, it looks like I was on an out of date branch of RS. They are indeed hidden in current builds. PTX still renders them into the HTML, but RS CSS successfully hides them.

Although if they are always hidden, should they even be in the generated HTML?

rbeezer:
The defining use-case for a permalink is a late-night email from an instructor to a student.  "Here, look at this to answer your question."  I don't see the individual choices of a multiple-choice being too useful in that scenario.  And don't they appear in random order sometimes?  Will my "permalink" be your permalink?  I guess so, since identifiers are based on teh order in the source.  Anyway, I don't see much use in keeping these.

PR #2712, opened 2025-10-26
Promote worksheet to stable schema; add handout to dev

oscarlevin:
This makes the "worksheet" element official.  To facilitate adding additional worksheet-like divisions, we call such things "Printout" as we do elsewhere internally.  This makes adding `handout` much easier than when we originally added `worksheet` to the dev schema.  

The only difference in the rules for worksheet is that now an article and book can contain worksheets at their root.  

A few considerations: 

1. Prior to this commit, a division could be structured either as a sequence of `BlockDivision` or `Paragraphs` followed by exactly one of the specialized divisions (`ReadingQuestions`, `Exercises`, `Solutions`, `References`, `Glossary`, and indeed `Worksheet`) in any order, or it could be structured with a subdivision or `worksheet` followed by any combination of subdivisions or specialized divisions in any order.  Now that the `Printout` division has multiple variants (in the dev schema at least), does it still make sense to allow a single worksheet *or* handout among the specialized divisions?  
2. I have not tried to specify where `@workspace` can go beyond exercises in Printouts.  The way that is implemented is to create a new version of `Exercise` called `PrintoutExercise`, as well as `PrintoutTask` that is identical to `Exercise` and `Task` except that it has an additional `@workspace` attribute.  I suppose I could do this for every block, but this seems like a lot of additional work.  Perhaps there is a way to extend these more simply?  Or maybe we say that `@workspace` is legal on all these elements regardless of whether they are descendants of a Printout, just that the attribute is ignored outside of a Printout.  Interested in hearing your thoughts on that.

Happy to fix this PR to address the above points, or any other issues you see.

### DIFF
--- a/schema/pretext-dev.rnc
+++ b/schema/pretext-dev.rnc
@@ -157,6 +157,19 @@
                 }
 
             
+            # Main handout definition
+            Handout =
+                element handout {
+                    PrintoutAttributes,
+                    MetaDataAltTitleOptional,
+                    (Objectives? & IntroductionDivision?),
+                    (Page+ | PrintoutBlock+),
+                    (Outcomes? & ConclusionDivision?)
+                }
+            # Add Handout to Printout options
+            Printout |=
+                Handout
+            
             Solutions |=
                 element solutions {
                     MetaDataAltTitleOptional,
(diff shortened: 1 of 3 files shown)


rbeezer:
Thanks!  And doubly for building the derived products.  Merged as-is.

Schema browser is not being rebuilt - I must have the wrong Java since I moved to Debian.  Perhaps I will make that my first website task.

PR #2713, opened 2025-10-26
CSS: fix fillin styling for non-math, non-interactive fillin blanks in HTML

ascholerChemeketa:
I noticed that the style options for non-interactive, non-math filllin blanks are broken in HTML. Pretty sure this happened in the great CSS retooling and has gone unnoticed until now. Rendering is currently always underlined. This restores box/shaded renderings.

Look for it at 4.7 in SA for examples. Will have to change pub file `common/fillin/@textstyle` to see the issue.

@oscarlevin This is probably coming your way

### DIFF
--- a/css/components/elements/_fillin.scss
+++ b/css/components/elements/_fillin.scss
@@ -1,36 +1,29 @@
 
 .fillin {
   display: inline-block;
-  border-bottom-style: solid;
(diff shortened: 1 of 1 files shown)


rbeezer:
Check that `CSS: updste themes` for a title.

PR #2714, opened 2025-11-01
HTML: no permalinks in Parsons problem blocks

ascholerChemeketa:
Currently `<p>` tags in Parsons blocks get permalinks inserted. I don't see any good that can come of having them there. The Parsons rendering code prevents the permalinks from even appearing, so it is just cruft HTML.

Although I don't know of any issues currently, given all the downstream processing the blocks go through, adding a bunch of extra HTML is asking for trouble.


rbeezer:
Agreed!

Any reason not match on `exercise/blocks/block/*` or similar.

And now that I think about it, shouldn't we protect PROJECT-LIKE and `task`?

I merged this as-is.  Improvements, if desired, can come on a new PR.  Thanks!

rbeezer:
> I'll do a new PR

That all sounds great.  Thanks.

PR #2715, opened 2025-11-01
Use stack-moodle element

geoo89:
Uses `stack-moodle` instead of `stack` elements to include stack questions as exercises.

Moves the namespace attribute out of the question file and into the parent `stack-moodle` element.

I've managed to skip the id-generation for the STACK questions.

But if I include my question file via `xi:include`, the warnings "There is an apparent XML element (<options>) inside an <input> element" are back and the numbering is increasing in steps of 2 as well. It doesn't happen if I include the content of the question file directly in the .ptx source.


rbeezer:
See 4 recent commits starting at a125b96a9a890e8e304ca8938ba67193a7407556, which accomplish the aims of this PR.  More discussion on `pretext-dev`.

PR #2716, opened 2025-11-05
Fix bug causing undesirable pagebreaks between first and later tasks in worksheets

oscarlevin:
A little while back I fixed another problem with tasks in worksheet exercises by splitting off all tasks except the first one of an exercise/project-like into their own row for computation of  workspace heights (just for the HTML print preview).  Unfortunately, this kept all the tasks inside the exercise for when the height of the exercise was computed.

This fixes that bug by actually removing the subsequent tasks (using javascript) outside of their parent exercise (in the print preview html) so heights are calculated correctly.

The first commit adds another test worksheet to the sample article that illustrates the original problem.  The second commit applies the fix.


rbeezer:
Looking good.  Works as advertied.  Merged as-is.

PR #2718, opened 2025-11-10
More rs permalink quash

ascholerChemeketa:
Follow up to #2714

Expands permalink quashing from exercises to exercises/tasks/project-likes.
Second commit is for testing that it works on non-exercises.

Third commit adds to paragraphs to other interactive types to create unwanted/needed permalinks for demonstration purposes.
Fourth gets rid of them.


rbeezer:
Curious why
```
p[ancestor::choices[parent::*[&PROJECT-FILTER; or self::task or self::exercise]]]
```
cannot just be
```
p[ancestor::choices]
```
Won't `choices` (and similar) only ever have those parents?  (Latter can also be written as `choices//p`.)

PR #2719, opened 2025-11-11
Add Kotlin capability

dmusican:
This PR adds capability to PreTeXt to handle Kotlin code. I've added an example in the sample book to show its use. If the book is built to static html, this seems to be sufficient to make everything work. For interactive code that runs on a Runestone server, there are changes that need to come to the Runestone code base... but this needs to be in place first. (I've got those changes ready to go, but I'm starting here.)

Thanks! If I can clean this up or help in some way, please let me know.


dmusican:
Thanks for the help, @ascholerChemeketa  and @rbeezer. I've updated the PR; let me know if I need to make more changes or if all is good.

PR #2731, opened 2025-12-02
HTML: add debug stringparam to override theme name

ascholerChemeketa:
Put it under the debug namespace to make it clear this is not the right way to set theme and avoid confusion with `$html-theme-name`.

Use: `-x debug.html.theme-name salem`, `-x debug.html.theme-name default-modern`, etc...


rbeezer:
Thanks very much - works as advertised.  I did have to cleanse my existing publication file to not set a `palette` since the one I had for default-modern was not supported in `denver`.  As to be expected.

PR #2733, opened 2025-12-05
HTML: Fully expand ToC to active item when subpage is navigated to

ascholerChemeketa:
Currently, navigating to a ToC link that is a sub page element (subsubsection when chunks are sections) does not expand the ToC to all the way. (Since XSL only knows what chunk the page corresponds to, not what href within the page will be selected.)

This makes sure the path to the active ToC item is fully expanded.

Testing:
Use the first commit that changes SA ToC depth to 4. 
Navigate to 4.6.2 and then refresh page.
ToC should look like:
image

Then apply second commit, rebuild to get fresh JS, and navigate to same place, then refresh.
Now you will get:
image

### DIFF
--- a/js/pretext.js
+++ b/js/pretext.js
@@ -246,7 +246,30 @@ window.addEventListener("DOMContentLoaded", function(event) {
                 expander.title = "Expand" + (itemType !== "" ? " " + itemType : "");
             }
         }
-      }
+    }
+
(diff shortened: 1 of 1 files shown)


rbeezer:
Very good!  Thanks for the tune-up.  Merged as-is.

PR #2734, opened 2025-12-06
CSS: Salem. Suppress conclusion styling inside RS exercises

ascholerChemeketa:
Removes styling for conclusion in project-like inside RS exercises when using Salem theme. (See https://groups.google.com/g/pretext-dev/c/K96i-x24aAE/m/X_bHOW7NCQAJ?utm_medium=email&utm_source=footer)

@bnmnetp

We likely need some follow up work to differentiate the conclusion from contents of any particular task. But that cuts across themes and likely belongs in RS codebase as it will be styling the HTML RS injects.

I'm planning on doing a spacing pass over break for PTX content dropped into RS exercises. There are many small instances of too much or too little margin/padding when the styles are blended. Will look at follow up as part of that.


oscarlevin:
Excellent.  Merged.

PR #2735, opened 2025-12-06
Warning vs error vs bug alignment

ascholerChemeketa:
Tagging people not named Rob Beezer who appear to have last touched given items. Please reply with "Go ahead and change to warning" or "That should stay an error". If "stay an error", I'll revert that commit as part of a rebase.

- [x]  Commit cb9fdbc
Multiple closely related. Some of these are @Alex-Jordan, other @rbeezer. Think it should be all or nothing on them.
- [X] Commit 7e40b3b748ac9a4a7068b80f88bd061ae11d3668
Alex already reviewed.

Other ones "belong" to Rob. Could be code he moved from elsewhere. @rbeezer feel free to point me to others who should be pinged about individual items.

These seemed like cases where changing WARNING/ERROR/BUG might better communicate the nature of the issue. This potentially simplifies tools (CLI) reporting on the severity/nature of issues and users understanding how broken their build is in the "I just rebuilt this 5 minutes before class, can I use it" case.

The boundary between ERROR/WARNING is admittedly fuzzy, but I'm using:

* BUG : Issue that is likely not related to author's source. e.g. missing template in a particular output mode.
* ERROR : Something is clearly missing/broken in output. Little to no chance that part of the output is useable.
* WARNING : Something is deprecated. Or, something was built and will likely be a valid rendering, but may not match author intent (default value used).

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -11202,7 +11202,7 @@ http://andrewmccarthy.ie/2014/11/06/swung-dash-in-latex/
 <xsl:template match="mathbook|pretext" mode="table-paragraph-cells-warning">
     <xsl:for-each select=".//tabular">
         <xsl:if test="row/cell/p and not(col/@width)">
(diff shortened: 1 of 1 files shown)


rbeezer:
cb9fdbc163f9959d6aae3c72f5bdffbcd11aa9e4 looks good, I ticked the box.  Will now cruise remainder.

PR #2736, opened 2025-12-07
Runestone: add id to activecode div in runnable Parsons

ascholerChemeketa:
The activecode that is created when a "runnable" Parsons is completed does not get an  `@id` on the div that becomes the activecode. This does not prevent the activecode from working, but it does mean its code gets persisted/restored from the DB using an id of `""`. Since all runnables share that same id, if the user modifies the code of any of them, it shows as a change to all.

Smaller change than it looks, really just adds an `@id` to the div that becomes the activecode when the Parsons is complete. The rest of the changes are just indentation cleanup.

Heads up for @bnmnetp - this requires no RS side changes.


bnmnetp:
Seems like a bad idea for me to be storing anything with an identifier of `""`

PR #2738, opened 2025-12-10
Make Mathjax rendering optional in shortanswers

ascholerChemeketa:
If an answer is not supposed to involve mathematics, rendering it with MathJax and displaying the result is at best confusing and at worst will actually display something that makes no sense.

This adds a switch to the `<response>` to disable mathjax rendering and a book level switch to change the default behavior. Otherwise, the default is to render with MathJax in order to maintain existing behavior.

Note that you should see the change in the HTML output, but RS will still render MathJax until this PR is in place:
https://github.com/RunestoneInteractive/rs/pull/1082

To test without a RS server, make sure to use stringparam `short-answer-responses always`.

Tagging @bnmnetp


ascholerChemeketa:
This can be used for experimentation:
https://runestone.academy/ns/books/published/PTXSB/short-answer-exercises.html

While `An $8\times 8$ checkerboard costs $3.50.` does what you expect, `A checkerboard costs $3.50. That is $2.00 too much.` does not behave as you would expect/desire.

We could use a regex to guess at whether there is something that could be rendered as math. But that does not mean it should be rendered as math.

At the extreme, given a piece of text, it is fundamentally unknowable if the intent was to embed math content. IMO, anything that can't be guaranteed to correctly guess intent should have a way to specify that intent.

rbeezer:
I think that parsing a reader's response to a short-answer question should be a Runestone responsibility.  Or conversely, it should not be an author's responsibility to predict (or restrict/enable) what sort of syntax gets used.  So I don't think the proposed extension to PreTeXt belongs.

With asymmetric delimiters, say `\(` and `\)`, I would think they could be paired and marked-off for MathJax to interpret.  Any leftovers could be dealt with as straight text.

We can discuss further on `pretext-dev` if desirable, either as PreTeXt (the language) or aspects of parsing math within free-form text.

rbeezer:
> the intent was to embed math content

That intent lies with the reader, no?  Maybe the text-box should have a checkbox ("Use math syntax").

You are correct that $ is a very bad choice for delimiters.  That is part of why LaTeX introduced \(, \).

rbeezer:
This was discussed in Drop-In.  Some work will be done on the Runestone side to improve things such as previewing an answer.  Closing for now.

PR #2740, opened 2025-12-10
Assembly: carry an @xml:id from audio|video|interactive[not(static)] …

Alex-Jordan:
…over to its static sidebyside

This is discussed in #2618.

Among the PRs I expect to open and update this week, first to evaluate can be this one or #2739, in either order.


rbeezer:
Thanks for the pointer to `biblio`.  A `biblio` is largely useless without an `@xml:id` since you should reference it via an `@ref`..  But I guess people will make `biblio` in a `references` just for show.

rbeezer:
And I fixed up the `biblio` version at 2f2c5c24e82d6419ba89f37dff45060c23c4103d.  Thanks for pointing that one out.

PR #2745, opened 2025-12-19
Printout options: hide solution-like elements for printing worksheets and handouts

oscarlevin:
Currently if you enable hints/answers/solutions for worksheets, these will show up when you go to print them via the print preview.  This adds some checkboxes on the print preview page that let the user hide these elements (toggling the checkbox will reload the page to recompute workspace, but the choices persist in local storage).  Useful if an instructor wants to include solutions on the main page but not have them clutter the printout.

Final commit includes modifications to the sample article in the 35.5 worksheet show you can test the print preview options.


rbeezer:
Very nice!  Merged.  Minor comments to come in a little while.

PR #2751, opened 2025-12-27
Same page printout previews

oscarlevin:
The print preview for worksheets and handouts is currently handled by creating a separate page with just that "printout" and some different CSS.  This eliminates those additional files.  The main reason this is needed is for portable builds to still get printable worksheets and handouts, but this also simplifies the code and reduces the number of files that need to be built.

How it works: now on any page that contains worksheets or handouts, a header for print previews is included (but hidden by CSS).  The links to the print preview now open the current page with a `?print-preview=[id-of-printout]` parameter.  This reloads the page (so the back button still works the same way as it always has) and tells the javascript to load just that printout and use the css for print previews.

While getting the javascript to work for this (second commit) I also refactored to make the javascript more robust.  In particular, switching page sizes while the highlight workspace is on no longer requires a full page reload, and this eliminates a minor bug that showed up if a user highlighted workspace, then unhighlighted it and then switch pages (doing so would set workspace incorrectly).

As an added bonus, this cleans up the pretext-html templates around the print-preview controls, as suggested in #2745

### DIFF
--- a/css/components/page-parts/_body.scss
+++ b/css/components/page-parts/_body.scss
@@ -79,6 +79,10 @@ body.standalone {
   background: var(--page-color, white);
 }
 
+// Hide print preview header div in normal view
+.ptx-content .print-preview-header {
+  display: none;
+}
 
 body.pretext > a.assistive {
   padding:6px;


rbeezer:
Cursory visual review looks good.  Do you know the `descendant-or-self::` axis?  Perhaps for use in

```xsl
<xsl:with-param name="b-has-printout" select="self::worksheet or self::handout or .//worksheet or .//handout"/>
```

oscarlevin:
Ah!  I did not know about that.  Agree that's better.  Will update soon, unless you beat me to it.

PR #2752, opened 2025-12-27
WeBWorK: reorganize PG macro file inclusion

Alex-Jordan:
I'm getting started on some cleanup (in multiple meanings of the word) to WeBWorK/PG code. For starters, I'd like to reorganize and consolidate how macro files are detected and loaded, and that is happening here.

* Some comment information was outdated or just wrong.
* The little comments that explain what each macro file is used for are out of place here. Someone should just follow the link to where PG macro files live on GitHub if they need to investigate what something is used for. But also, this file is not a place for curious authors to be looking in the first place.
* Replace multiple uses of `.//pg-code` with storing it once as `$pg-code`. Might have some effect on efficiency, although testing suggests it's negligible at the scale of the webwork sample chapter.
* Use of str:tokenize allows for consolidating code into a for-each loop. Together with removing the small comments mentioned above, the net effect is 225 fewer lines of code.
* The reorganization will help me know better where to add new macro files in the future.

I looked at before/after with the webwork representations file for the sample chapter. The only changes are with a few macros being added in a different order, owing to the reorganization.

### DIFF
--- a/xsl/extract-pg.xsl
+++ b/xsl/extract-pg.xsl
@@ -841,7 +841,7 @@
         <xsl:with-param name="b-human-readable" select="$b-human-readable" />
     </xsl:call-template>
 
-    <!-- three standard macros always, order and placement is critical -->
+    <!-- two standard macros always used -->
     <xsl:variable name="standard-macros">
         <xsl:call-template name="macro-padding">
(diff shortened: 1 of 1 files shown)


rbeezer:
Very good!  Love all the code consolidation.  ;-)  Merged as-is.

Issue #2622, opened 2025-07-21
Code-copy elements for code blocks inside tables are not properly formatted


skiadas:
A code like the following:

```xml
          <row header="no" left="minor">
            <cell bottom="minor" right="minor">
              <p>
                <cd>11</cd>
              </p>
            </cell>
            <cell bottom="minor" right="minor">
              <p>
                <cd>the square root of 11 is about 3.32</cd>
              </p>
            </cell>
          </row>
```
Ends up producing some very visible copy-element links.

<img width="514" height="288" alt="Image" src="https://github.com/user-attachments/assets/e95d7a17-0e57-452c-a4dc-6fccfa03f9d1" />

rbeezer:
Resolved by not offering the button there: code inside a `tabular` cell no longer receives a copy button.  The button serves a block of code; inside a table cell the content is short and the floating button crowds the cell, as your screenshot shows, so there we decline to provide one.  Merged to master in PR #3123.

*Claude Fable 5, acting as a coding assistant for Rob Beezer*

PR #2756, opened 2025-12-29
Deprecate the use of pre@source in datafiles

skiadas:
See #2755 

This PR:

- Adjusts the example that used dataset[pre@source] to use xi:include instead.
- Amends the guide to suggest this approach instead.
- Adds a deprecation warning.


rbeezer:
I will need to re-familiarize myself with this one, but there are comments and questions from my latest post (two months ago) that maybe need a response?  Moving to a draft until this restarts.

skiadas:
Yes this likely needs some more thinking, I'll put some time on it this week. For reference here is the discussion on -dev about it:  https://groups.google.com/g/pretext-dev/c/qQm2PY_Os78/m/h3dyrxtcBgAJ

skiadas:
I'm back to working on this, along with an overall transition to a #file tag. Hoping to have some more concrete proposals soon. Trying to work out the path for supporting binary files more (zips/jars etc). That support would require runestone and activecode changes. Anyway for now just reporting work is happening.

rbeezer:
Thanks for checking in.  I have a new assistant (Claude Code!) so am a bit more efficient thatn usual.  Mostly around the next few weeks with one short exception

I hope you saw the code-copy button removal just now on https://github.com/PreTeXtBook/pretext/issues/2622 - I hope that is not too draconian

skiadas:
Superseded by #3139: the new `file` element replaces `datafile` (including `pre@source`) with first-class binary support.

PR #2758, opened 2025-12-31
Diagcess:  update library and PreFigure diagrams in examples

davidaustinm:
There is a new version of the `diagcess` library, which led to some changes in PreFigure.  When annotations are present in a PreFigure diagram with label `foo`, then PreFigure will write out two SVGs:  `foo.svg`, which has `@height`, `@width`, and `@viewBox` attributes, and `foo-diagcess.svg`. which only has `@viewBox`.  In HTML, we feed `foo-diagcess.svg` into the `diagcess` library.  This resulted in a small change to `pretext-html.xsl` so that we load `foo-diagcess.svg` when there are annotations.  In all other instances, we simply load `foo.svg` as before.

This PR has the new `diagcess.js`, makes the change to `pretext-html.xsl`, and updates the PreFigure diagrams in the sample article, slideshow, and epub sampler.  The diagram in the epub sampler has also been resized.

I rebased before pushing this branch with --force-with-lease.  I'm not totally confident in my git skills so just let me know if I can change something to make it easier.


rbeezer:
Looking good.  Readers, authors, publishers will never know there need to be two versions of the SVG.  ;-)

Your git work looks just great on my end.  Merged as-is.  Thanks!

PR #2760, opened 2026-01-05
Publisher pagebreaks

Alex-Jordan:
This follows up from #2618, making a `pre-sbsgroup` template that can be used as the mechanism to make a pagebreak right before an sbsgroup in PDF.

Also I saw a nearby bad space in the -common file.  I removed it on its own commit, but right before the other commit involving that file. So you could squash those two if that's better. Or if you just don't want that let me know and I'll force push without that commit.

### DIFF
--- a/examples/sample-article/publication.xml
+++ b/examples/sample-article/publication.xml
@@ -168,7 +168,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
         <worksheet formatted="yes"/>
         <asymptote links="yes"/>
         <insertions pagebreaks="quotations exercises-section-multiple pre-roll-countdown-8"/>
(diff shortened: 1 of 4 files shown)


Alex-Jordan:
Just saw that you requested an example in the sample article. Will check on that now. Moved to draft status until that is done and I hear about if you want that space-removing commit here too.

rbeezer:
I'd forgotten about `pre-sidebyside`!  Thanks for picking that up.

Looks good visually - will test (and merge) soon when I can.  @davidaustinm is generally interested in these hooks for use with PreFigure annotated diagrams, so pinging him here.

oscarlevin:
I was about to publish a new version of the Cli, so that should be out in time for the announcement.

PR #2808, opened 2026-03-28
Update localizations

jirilebl:
Update localizations for Czech


rbeezer:
Thanks very much!  Great to have some more quotation marks going...

PR #2809, opened 2026-03-30
WeBWorK assembly: replace ww-id pass with per-exercise representation files

rbeezer:
This PR removes the dedicated WeBWorK identification pass and replaces the single monolithic `webwork-representations.xml` file with one XML file per exercise, named by the exercise's `@assembly-id`.

**Motivation.** The old pipeline stamped a `@ww-id` on every `webwork` element in a separate assembly pass, then used those IDs to look up entries in a single representations file. The `@assembly-id`, already computed in the `assembly-label` pass, serves the same purpose — and is the exercise's stable, meaningful identifier. Per-exercise files make incremental regeneration natural: only changed problems need to be re-sent to the server.

**Commits follow the pipeline order:**

1. `Assembly: remove WeBWorK identification (ww-id) pass` — eliminates the `webwork` mode pass, the `$all-webwork`/`$webworked` variable chain, and the `@ww-id` stamping template.
2. `WeBWorK: use parent exercise @assembly-id as problem identifier` — `extract-pg.xsl` now reads `../assembly-id` from the parent exercise.
3. `WeBWorK: use assembly-id of original exercise for copied-from reference` — for `@copy` problems, the `copied-from` attribute now carries the assembly-id of the source exercise rather than the xml:id of the source `webwork` element.
4. `Script: write @assembly-id on webwork-reps elements` — replaces `ww-id` attribute with `assembly-id` on each `webwork-reps` element written by the server communication loop.
5. `Script: write per-exercise WeBWorK representation files` — each problem is written to `{assembly-id}.xml` in the representations directory instead of appending to a single file.
6. `Publisher variables: replace webwork-representations-file with webwork-representations-dir` — the publication file variable switches from pointing to a single file to pointing to the directory.
7. `Assembly: use @assembly-id with per-exercise representation files` — the representations pass loads `{dir}/{assembly-id}.xml` per exercise; error messages note the presence of an old-format `webwork-representations.xml` as a sign that regeneration is needed.
8. `HTML: read @assembly-id from webwork-reps for div and iframe` — updates the two locations in HTML output that used `@ww-id`.

**Breaking change.** Existing generated WeBWorK files must be regenerated. The error messages in the assembly pass will identify the problem and point to the old file if present.

**Testing.** Tested against `examples/webwork/sample-chapter/` using `pretext/pretext` on both master and this branch. WeBWorK representations were regenerated from `https://webwork-ptx.aimath.org` using the new per-exercise code (76 files produced, including all `@copy` problems). Before/after builds were run for HTML, LaTeX, and PG problem sets (`webwork-sets`); all three produced the same file counts with no warnings or errors. The only differences in output are the intended ones: `ww-id`-based identifiers replaced by `assembly-id`-based identifiers, image filenames now use the exercise label rather than an opaque counter, and PG file timestamps. The four `assembly-*` developer formats were also compared before/after on the sample-chapter; `assembly-version` output is identical, and the differences in the other three reflect only the identifier renaming.

🤖 Authored with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -1168,8 +1168,7 @@ def webwork_to_xml(
     ww_reps_dir = dest_dir
     ww_images_dir = os.path.join(dest_dir, "images")
 
-    # file path for the representations file
-    ww_reps_file = os.path.join(ww_reps_dir, "webwork-representations.xml")
+    # per-exercise representation files live in this directory, named {assembly-id}.xml
 
     # where generated pg problem files will live (each .pg file will usually be deeper in a folder
     # tree based on document structure and chunking level)
@@ -1423,11 +1422,8 @@ def webwork_to_xml(
         if count > 0:
             raise ValueError("PTX:ERROR: unable to establish connection to local socket")
 
-    # begin XML tree
-    # then we loop through all problems, appending children
     NSMAP = {"xml": "http://www.w3.org/XML/1998/namespace"}
     XML = "http://www.w3.org/XML/1998/namespace"
-    webwork_representations = ET.Element("webwork-representations", nsmap=NSMAP)
     # Choose one of the dictionaries to take its keys as what to loop through
     for problem in origin:
         if origin[problem] == "webwork2":
@@ -1484,7 +1480,7 @@ def webwork_to_xml(
                 socket_params["sourceFilePath"] = os.path.join(external_dir, path[problem])
 
             msg = "sending {} to socket to save in {}: origin is '{}'"
-            log.info(msg.format(problem, ww_reps_file, origin[problem]))
+            log.info(msg.format(problem, ww_reps_dir, origin[problem]))
             clientsocket.send(json.dumps(socket_params).encode('utf-8'))
 
             buffer = bytearray()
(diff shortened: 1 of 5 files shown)


rbeezer:
I plan to re-do the testing, but have already reviewed (and refined) this a lot.

Comments from @Alex-Jordan ?  

@oscarlevin : see point 7 above about transitioning.  If there is more that can be done to help the CLI, just say so,

rbeezer:
Mostly for @oscarlevin.

> Maybe a more dire concern: what happens if you try to build and even one of the individual files is missing?

Well, if a file is missing, the attempt to open it will fail.  XSL does not have any kind of condtional to check on the existence of a file before opening it (that I know of).  We already do this sort of one-XML-file-per-problem for MyOpenMath and for STACK.  The WW processing does some checks for "bad" XML and the like.  It could write a dummy file when teh processing fails, no matter what, and maybe it does that already.

I'm ready to merge this (after a few checks that it has not bit-rot).  How should we coordinte with the CLI and checks related to generation and re-generation?

rbeezer:
For the record, @oscarlevin suggests merging this, and the CLI wil react to how best to recognize the need for re-generation, both relative to this change, and long-term.

rbeezer:
There was an additional commit to resolve some mis-matches on `id`.  That got fixed in.  Also a message about a deprecated approach to specifying the (old) file of representations was restored.  Otherwise, merged as it has been here for several weeks.  More to come.

PR #2812, opened 2026-03-31
Numbering: replace multi-number with pre-computed block-struct attribute

rbeezer:
## Summary

- Replace the `multi-number` recursive template (80 lines of ancestor-walking logic) with a `@block-struct` attribute stamped on every division during the assembly pass
- New `block-structure-number` and `truncate-pad-struct` templates in `pretext-numbers.xsl` read the pre-computed attribute and truncate/pad to the requested number of levels
- Fix decorative specialized divisions producing incorrect structure chains (the documented `multi-number` bug)
- Fix `figure/sbsgroup/sidebyside/figure` (and table, listing, list) missing the `structure-number` template match, so subfigures inside an sbsgroup now correctly inherit their parent figure's full number
- Move `is-structured-division` and `is-specialized-own-number` to `pretext-numbers.xsl` to consolidate numbering logic
- Rename misnamed `$equation-levels` variable to `$exercise-levels` in the inline exercise structure-number template

## Testing

24 HTML builds: sample article + sample book (no parts, structural parts, decorative parts) × 3 numbering level settings (existing defaults, all-zero, all-three).

**Sample book (9 configs):** Zero non-timestamp differences across all part configurations and numbering levels.

**Sample article (3 configs):** Exactly 12 changed lines per config, all in `section-side-by-side.html` — the expected sbsgroup bug fix. 12 subfigures (a–l) inside `figure/sbsgroup/sidebyside/figure` now correctly show their parent figure's full number in permalinks:
- Level 0: `"Figure (a)"` → `"Figure 268(a)"` (was missing parent number entirely)
- Level 1 (default): `"Figure 26(a)"` → `"Figure 26.42(a)"` (was using bare serial instead of full number)
- Level 3: same correction with deeper structure numbers

*Claude Opus 4.6, acting as a coding assistant for Rob Beezer*


rbeezer:
Force-pushing to fix a bug found during additional testing.

The original 24 HTML builds used `sample-book.xml` for all book configurations, including structural and decorative parts. This was wrong — those configurations need `sample-book-parts.xml`, which wraps chapters in parts. With `sample-book.xml` there are no parts in the source, so the structural and decorative publication files had no real effect on the division hierarchy, and the parts code path was never exercised.

When rebuilt with the correct source file, every block structure number in a book with parts gained a spurious `.0` component (e.g., `3.2.0.1` instead of `3.2.1`). The cause: `@block-struct` already excludes parts (they are squelched during assembly), but the `$levels` parameter passed by the callers (e.g., `$numbering-blocks = 3` for a book with parts) still counts from part depth. The old `multi-number` template handled this by decrementing `$levels` when it encountered a part or backmatter ancestor. The new `block-structure-number` template now does the same — it subtracts one from `$levels` when parts are present and the block is inside a part or backmatter. Blocks in frontmatter (which have no part ancestor) use `$levels` unmodified. The subtraction is clamped to zero for the edge case of level 0 with parts.

The fix is folded into the commit that introduced `block-structure-number` and `truncate-pad-struct`.

Same testing regime as before — 24 HTML builds (4 configs × 3 level settings), now using `sample-book-parts.xml` for the 6 structural/decorative parts builds. All 9 book configs show zero non-timestamp differences. All 3 article configs show only the expected 12-line sbsgroup bug fix.

*Claude Opus 4.6, acting as a coding assistant for Rob Beezer*

PR #2815, opened 2026-04-02
Create modular javascript for bundling

oscarlevin:
With the help of copilot I have a possible way forward on making the javascript modular and easier to maintain.  I'm marking this as draft, since it definitely needs more eyes and testing, but from what I've seen, there are no functional changes to the sample article or sample book.  

The idea is that we will maintain a js/src directory with modular files that get bundled by a esbuild function (see `script/jsbuilder`)  This creates `js/dist/pretext-core.js` and `js/dist/pretext-search.js` (since we don't always provide the search javascript).  

There was an initial attempt to remove jquery (and I think we removed the reliance on jquery from pretext_add_on's replacements), but I think maybe some runestone js requires us to have a copy of jquery?  I don't really understand that.

So @rbeezer and @ascholerChemeketa, do you think this is the sort of thing we want?  Any suggestions?  When we are mostly happy with the new file structure I will remove the old files we no longer need and clean up all the commits.


bnmnetp:
Removing jQuery will break the interactive components in static html builds.
This would be project on the runestone side as the Sphinx build system always include jQuery in a runestone build.

I'm sure it can be done, but I think I would rather put my efforts into the continued removal of jQuery from the components themselves and once that is complete then you would not need to include it anywhere.

Brad Miller, acting like he knows what he's talking about.

rbeezer:
There is some knowl/JS work at #2845, and MathJax 4 work at #2818, so that may require some adjustments here.

ascholerChemeketa:
More work is hitting the JS as part of the accessibility PRs I am currently working on.

Another useful feature of getting a build system in place will be to make internationalization of JS strings easier. Stuff I am touching I am wrapping with shimmed calls to `i18next.t( )`. I don't think it makes sense to include the real thing from each JS file. But once we have build in place, we can include `i18next` (or a similar solution) in one place and then have one string table that editors can use to add languages. I am picturing one file, or folder of files, like the xsl localizations that make it a little bit easier to make UI components work in languages other than English.

oscarlevin:
This pr has grown stale.  I don't think it will make sense to try to fix this up, but next week I can try to replicate the ideas in it to have one that is worth acting on.  Perhaps before then we can think about what the modularization structure should be.

rbeezer:
And maybe one of our assistants can seed the UI-only strings by moving some out of the XSL localizations?!

oscarlevin:
Closing this in favor of #2919

PR #2180, opened 2024-06-20
Mj4 knowl fix

ascholerChemeketa:

Issue #1841, opened 2022-10-02
MathJax 4


Alex-Jordan:
The first alpha for MathJax 4 was just announced: https://github.com/mathjax/MathJax-src/releases/tag/4.0.0-alpha.1

This issue is a reminder to monitor the features and progress of MathJax 4, informing decisions about when to eventually move to using it.

rbeezer:
Inserted the new Javascript `script` tag for first alpha release, under influence of `debug.mathjax4` parameter.  No configuration adjustments implemented (or even considered).

Seems to be working OK.  Commutative diagram in Subsection 7.6 of sample article has short arrows.  Broken arrow in Subsection 7.1 seems to be unrelated (present with MJv3).

rbeezer:
Just thought to check.  Arrows (both examples) are good in Chrome.  Bad versions are in Firefox.

Alex-Jordan:
We've generally (not completely) ruled out XML children of a PTX math element.  But something new with MJ4 is you can have HTML inside the math content. I haven't thought hard about it but that may open up interesting possibilities for PTX elements inside math.

rbeezer:
> @rbeezer, can you provide me the example diagram? I will look into the issue.

Thanks!  I was going to wait a bit for confirmation.  ;-)  Here is the `CD` which is authored as a PreTeXt `<me>` element, resulting in HTML and screenshot.

Oooh, that's weird.  Screenshot has full arrows.  And now I see that if I zoom to just 110% I get full arrows.  And full at 90%.   (At 100% arrows are all short and pushed to the right, which I can't show you!)


![Screenshot 2022-10-03 at 10-46-12 SA Mathematics](https://user-images.githubusercontent.com/1881714/193644715-8f47f953-028a-49ac-86ff-35ffd8f8c6da.png)


```tex
\begin{CD}
 E[x]/\langle p(x) \rangle    @>\psi>>  F[x]/\langle q(x) \rangle\\
@VV{\sigma}V        @VV{\tau}V\\
 E(\alpha)     @>\overline{\phi}>>  F(\beta)\\
@VVV        @VVV\\
 E     @>\phi>>  F
 \end{CD}
```

```html
<div class="displaymath process-math" data-contains-math-knowls="">
\begin{equation*}
\begin{CD}
E[x]/\langle p(x) \rangle    @&gt;\psi&gt;&gt;  F[x]/\langle q(x) \rangle\\
@VV{\sigma}V        @VV{\tau}V\\
[...]

PR #2818, opened 2026-04-04
HTML: upgrade to MathJax 4 for browser rendering

rbeezer:
Upgrade HTML output from MathJax 3 to MathJax 4 (CDN stable release 4.1.1).

- Incorporates Andrew Scholer's `mathjax_startup.js` module from #2180, which inlines the knowl extension and exports a configuration function
- Replaces the inline MathJax 3 configuration block in `pretext-html.xsl` with a module import calling `startMathJax()`, plus the MathJax 4 CDN script with `defer`
- Adds `defer` to Runestone service script tags for correct load ordering
- Removes `mathjaxknowl3.js` (superseded by the startup module)
- Removes `debug.mathjax4` parameter and `$mathjax4-testing` variable (no longer needed since MJ4 is now the only path)

The `textmacros` extension, added to the MJ3 configuration in #2817, is loaded by default in MathJax 4 and does not need to be listed explicitly.

This is Phase 1 of the MJ4 migration (issue #1841), covering browser-side rendering only. The offline Node.js processing in `mj-sre-page.js` (used for EPUB and braille) is independent and unchanged.

*Claude Opus 4.6, acting as a coding assistant for Rob Beezer*


js/mathjax_startup.js
```diff
@@ -0,0 +1,147 @@
...
+        }
+        )
+      },
+    }
+  }
+
+  if(opts.htmlPresentation) {
+    mathJaxOpts['options']['menuOptions'] = {
+      "settings": {
+        "zoom": "Click",
```
dpvc: I would not recommend using `Click` for this unless you also include one of the modifier keys, as clicking (and double clicking) is used with the expression explorer, and this will cause it to both start the explorer and show the zoom.

dpvc (commented):
A few recommendation for streamlining the configuration.

rbeezer:
Thanks very much @dpvc!  I'll see about getting those in place.

You are right about "Click".  We also support knowls inside display math, which are activated/opened by a mouse click.  We will need to think more carefully about how we want to separate that from the Explorer functions.

rbeezer:
Latest commit applies David Cervone's configuration recommendations from the PR discussion:

- Remove `base`, `ams`, `newcommand` from `[+]` packages and `[tex]/newcommand` from loader (already bundled in `tex-mml-chtml.js`)
- Move `NodeUtil` lookup outside `GetArgumentMML` to avoid repeated lookups
- Use MJ4 compact `CommandMap` format with inline function
- Simplify `pageReady` callback

Did not change `defer` attributes (under separate evaluation) or `Click` zoom (conflicts with expression explorer, needs more thought).

*Claude Opus 4.6, acting as a coding assistant for Rob Beezer*

rbeezer:
Merged, with some modifications discussed above.  In particular, I dropped `@defer="true"` from the loading of Runestone scripts, but left it in place for MathJax.  Broader announcment coming just now on `-dev`.

PR #2822, opened 2026-04-07
Expand meta tags for better social media card support

StevenClontz:
Re: https://groups.google.com/g/pretext-dev/c/pTX_B_EDLWc/m/m7vTCxulBQAJ

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -197,6 +197,11 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <!-- use of the translation, "external/"                 -->
         <directories data="numerical"/>
 
+        <!-- A blurb for describing  -->
+        <blurb>
+            This is a simple string (no markup is allowed) that describes
+            this document, a sample article for the PreTeXt markup language.
+        </blurb>
     </docinfo>
 
     <!--
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -197,7 +197,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <!-- use of the translation, "external/"                 -->
         <directories data="numerical"/>
 
-        <!-- A blurb for describing  -->
+        <!-- A blurb that briefly describes the document -->
         <blurb>
             This is a simple string (no markup is allowed) that describes
             this document, a sample article for the PreTeXt markup language.
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -2689,6 +2689,17 @@ Book (with parts), "section" at level 3
     </xsl:choose>
 </xsl:template>
 
+<!-- Plain subtitle: used for HTML <meta> tags -->
+<xsl:template match="*" mode="subtitle-plain">
+    <xsl:choose>
+        <xsl:when test="subtitle">
+            <xsl:apply-templates select="subtitle/node()[not(self::fn)]" mode="plain-title-edit"/>
+        </xsl:when>
+        <!-- just empty if there is no subtitles, no default -->
+        <xsl:otherwise/>
+    </xsl:choose>
+</xsl:template>
+
 <!-- We do not wrap an "m" element as part of a plain title -->
 <!-- This will misbehave for m/xref and m/fillin, etc, but  -->
 <!-- these devices should not be in a title anyway          -->
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -11614,14 +11614,14 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:param name="name"/>
     <xsl:param name="content"/>
     <meta>
-        <xsl:if select="$property">
+        <xsl:if test="$property">
             <xsl:attribute name="property">
                 <xsl:value-of select="$namespace"/>
                 <xsl:text>:</xsl:text>
                 <xsl:value-of select="$property"/>
             </xsl:attribute>
         </xsl:if>
-        <xsl:if select="$name">
+        <xsl:if test="$name">
             <xsl:attribute name="name">
                 <xsl:value-of select="$namespace"/>
                 <xsl:text>:</xsl:text>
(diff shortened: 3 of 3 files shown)


rbeezer (commented):
**Summary**: Expands the existing OpenGraph `<meta>` tags to also emit Twitter/X card meta tags, adds `og:title`, `og:description`, and `twitter:card`, and uses `docinfo/blurb` (already in the dev schema for Runestone) as the description source with subtitle as fallback.

**Issues:**

1. **`blurb` is dev-schema only.** The `blurb` element is defined in `pretext-dev.rnc` (for Runestone's shelf categorization), not the main schema. This PR repurposes it for social metadata. That's fine conceptually, but it means authors using the stable schema will get validation errors if they add `<blurb>` to their `docinfo`. Should `blurb` move to the main schema, or should the PR note this dependency?

2. **Significant code duplication.** The `og` and `twitter` calls for title, image, image:alt, and description are nearly identical — differing only in the attribute name (`property` vs `name`). The `social-meta-element` template already handles both via separate `$property` and `$name` params, but each caller duplicates the full block. A loop or a wrapper template that emits both at once would halve the code.

3. **Title logic is inconsistent.** The subtitle is appended to the title only when `$docinfo/blurb` exists (`<xsl:if test="$docinfo/blurb and $document-root/subtitle">`). This means: blurb + subtitle → title is "Title: Subtitle", description is blurb. No blurb + subtitle → title is just "Title", description is subtitle. The subtitle disappears from both when there's no blurb. Should the subtitle always appear in the title regardless of blurb?

4. **`summary_large_image` without guaranteed image.** The Twitter card type is unconditionally set to `summary_large_image`, but the image tag is only emitted when `$b-has-baseurl and $docinfo/brandlogo`. Without an image, `summary_large_image` will render poorly — Twitter shows a blank image area. Should fall back to `summary` when there's no image.

5. **Hard-coded "Document Logo" alt text.** The `og:image:alt` and `twitter:image:alt` use the literal string "Document Logo". This isn't localizable and isn't descriptive. The `brandlogo` element doesn't currently carry alt text, so there may not be a better source, but worth noting.

6. **Empty comment.** Line with just `<!--  -->` (the one before "image") — looks like an editing artifact.

7. **`article:author` may not work as expected.** The [OGP article spec](https://ogp.me/#type_article) defines `article:author` as a URL (profile page), not a name string. Passing `personname` as the content won't produce useful results for Facebook's parser. The `book:author` property has the same issue but was pre-existing.

**Minor:** The seven incremental commits ("first pass", "fix xsl:if", etc.) could be squashed before merge for a cleaner history.

rbeezer:
(1) is not important, and of course, I would squash everyting just as part of my routine.

StevenClontz:
> 3\. **Title logic is inconsistent.** The subtitle is appended to the title only when `$docinfo/blurb` exists (`<xsl:if test="$docinfo/blurb and $document-root/subtitle">`). This means: blurb + subtitle → title is "Title: Subtitle", description is blurb. No blurb + subtitle → title is just "Title", description is subtitle. The subtitle disappears from both when there's no blurb. Should the subtitle always appear in the title regardless of blurb?

😶

> No blurb + subtitle → title is just "Title", description is **subtitle**.

> The subtitle disappears from both when there's no blurb. 

C'mon Claude, get your act together. 😅

StevenClontz:
b867072 fixes up 2&4&6, the others are hallucinations or wontfix.

StevenClontz:
Poking this -- did I miss anything I needed to do to move this forward?

rbeezer:
> Poking this

Been away for several weeks.  Just catching up now...

rbeezer:
The use of

```xsl
<xsl:apply-templates select="$document-root" mode="subtitle" />
```
will process children of `subtitle`, such as `m` and `q` and `em` and introduce markup that we really do not want inside an attribute value (for example nested quotation marks).

See `mode="title-plain"` and construct a `mode="subtitle-plain"`.  Likely not much to do other than to apply the `mode="plain-title-edit"` template to nodews, much as in the same way as is done for `title`.

rbeezer:
I guess authors need to be URLs, and there is no good way to broadcast this information?

rbeezer:
I'm marking this off as a draft while we wait on a plain subtitle template.

StevenClontz:
Thanks for the poke -- at a Doenet workshop right now and am in dev mode, so good timing. Does 5b29b07 wrap this up?

rbeezer:
Thanks, @StevenClontz, the follow-up and sticking with this one.  Should help folks advertise thir projects better.  All for that!

Fixed-up into three commits, but no actual changes.  Will likely add some follow-up commits in just a second

rbeezer:
Just added four commits, the final three are worth reviewing.  Thanks again!

StevenClontz:
It could use a higher-quality image, but we're looking good for the Sample Article:

image

A minor note: it was slightly disappointing to see that Claude got a shout-out but I was not added as a co-author for any of the relevant commits to master. My ego doesn't particularly need the stroking, but it shouldn't be overlooked for a newer contributor that they recieve credit for contributing to a major project like PreTeXt.

rbeezer:
> but we're looking good for the Sample Article:

Great, thanks for the test.

> Claude got a shout-out

I only see an acknowledgement of a code review, here in the comments at GitHub for teh PR, which has become my practice when work is not my own.  I'm not seeing Claude acknowledged in the actual commits anywhere?  Maybe I'm not finding it?

StevenClontz:
image

Maybe those Claude coauthor commits are follow-up work beyond what I did (I didn't dig in). It's just that my community-maintainer brain raises a flag when a PR is marked as closed without merging on GitHub (an unfortunate artifact from the workflow you use for managaing pretext contributions), but it actually is implemented (great!), but the contributor isn't marked as author or co-author on the relevant commits.

rbeezer:
An unfortunate artifact of my workflow is that I strive for a readable history.  Clear commit messages and focused commits.  That means I rewrite a lot of people's work since I accept a lot of poorly-formed stuff.  I'm not going to apologize for that.

Commits bd51540aee21753f0aa0a9a097ceab2661afd818 and 221d737622f8a86cea7d45dea3ddbe16228b0471 are your work.  I can, and do, replace the author's name and email routinely when I need to consolidate a long string of commits.  In this case, that did not happen and was just an oversight in the process of doing the extra work to get this in.  I do apologize for that.  The merge commit does reference #2822 so anyone who goes there would see your role clearly.

The subsequent commits are joint work that Claude and I did, which surfaced in the course of your work, so those are accurate.  `git log --graph`

I do go to great lengths to make sure everyone's contributions are recorded.  But I do make mistakes.  Please take the time to get your facts in order before critiquing that work.

PR #2829, opened 2026-04-12
Script: add profiling checkpoints to html()

ascholerChemeketa:
Pulled out from: https://github.com/PreTeXtBook/pretext/pull/2329

-------------------------

This pull request adds a lightweight timing and profiling utility to the `pretext/lib/pretext.py` module, enabling optional logging of elapsed time at key stages of the HTML build process. The main change is the introduction of a `Stopwatch` class and its integration into the `html()` function, allowing developers to profile and monitor build performance when needed.

**Profiling and Logging Enhancements:**

* Introduced a `Stopwatch` class for measuring and logging elapsed time between key events, with optional logging controlled by the `profile.py` string parameter. (`pretext/lib/pretext.py`)
* Integrated the `Stopwatch` into the `html()` function, adding time logs at major build steps such as loading publisher variables, placing Runestone services, copying managed directories, copying CSS/JS, running XSLT, and completing the build. (`pretext/lib/pretext.py`) [[1]](diffhunk://#diff-38443e70f3814e0da63a79b48ef5349fe85c5e598bf82a068c415c0190753b5cR4861-R4863) [[2]](diffhunk://#diff-38443e70f3814e0da63a79b48ef5349fe85c5e598bf82a068c415c0190753b5cR4874) [[3]](diffhunk://#diff-38443e70f3814e0da63a79b48ef5349fe85c5e598bf82a068c415c0190753b5cR4884) [[4]](diffhunk://#diff-38443e70f3814e0da63a79b48ef5349fe85c5e598bf82a068c415c0190753b5cR4903-R4915) [[5]](diffhunk://#diff-38443e70f3814e0da63a79b48ef5349fe85c5e598bf82a068c415c0190753b5cR4951-R4952)

**Dependency Update:**

* Added an import for the standard `time` module to support the new timing functionality. (`pretext/lib/pretext.py`)

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -91,6 +91,8 @@ import re
 # contextmanager tools
 import contextlib
 
+import time
+
 # cleanup multiline strings used as source code
 import textwrap
 
@@ -4856,6 +4858,9 @@ def html(xml, pub_file, stringparams, xmlid_root, file_format, extra_xsl, out_fi
     # to ensure provided stringparams aren't mutated unintentionally
     stringparams = stringparams.copy()
 
+    log_time_info = stringparams.get("debug.profile", False) == "yes"
+    time_logger = Stopwatch("html()", log_time_info)
+
     # Consult publisher file for locations of images
     generated_abs, external_abs = get_managed_directories(xml, pub_file)
     # Consult source for additional files
@@ -4866,6 +4871,7 @@ def html(xml, pub_file, stringparams, xmlid_root, file_format, extra_xsl, out_fi
 
     pub_vars = get_publisher_variable_report(xml, pub_file, stringparams)
     include_static_files = get_publisher_variable(pub_vars, 'portable-html') != "yes"
+    time_logger.log("pubvars loaded")
 
     if include_static_files:
         # interrogate Runestone server (or debugging switches) and populate
@@ -4875,6 +4881,7 @@ def html(xml, pub_file, stringparams, xmlid_root, file_format, extra_xsl, out_fi
         # even if we don't need static files, we need to set stringparams for
         # Runestone Services information.
         _cdn_runestone_services(stringparams, ext_rs_methods)
+    time_logger.log("runestone placed")
 
     # support publisher file, and subtree argument
     if pub_file:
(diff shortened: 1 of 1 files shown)


rbeezer:
Looking good visually.  You didn't like the comment that equaled the function name?  ;-)

rbeezer:
And shoulda said - thanks for breaking this out.

rbeezer:
I was confused by "profile.py" when I did a visual review.  Do we have a module named "profile.py"?  Or is it from some associated library?  And why does it matter?  I'll need to chase that down, I thought.  How about "html.profile"?  Yes, it is all in the Python, but it is also the HTML build being profiled.  More consistent with names in use otherwise.

Claude flagged this as well.  That review coming next.  Mostly nitpicking stuff, but maybe something you would like to react to.

We have a developer's section in the Guide.  Let's record the existence of this switch there somewhere with just a single sentence.  Likely no great/obvious place, so don't sweat it.  It'll get reorganized eventually.

You can pile on new commits, if you keep documentation on the last one by itself, and I'll clean up.

rbeezer:
Thanks for breaking this out from #2329, much easier to evaluate independently.

Clean and useful. A few observations:

**Stringparam naming:** The param is `profile.py` here but was `profile-py` in #2329. The dot-py reads like a filename rather than a parameter name. Is `profile.py` intentional, or should it be `profile-py` (or something else entirely)? Just want to make sure the two PRs agree.

**Minor:**
- The diff removes a blank line and the `# build or copy theme` comment between `copy_html_js` and `build_or_copy_theme`. Not a problem, but unrelated to the profiling work — could be a separate cleanup or just left alone.
- `reset()` is only called internally — could be `_reset()`, but the class is small enough that it doesn't matter much.

Otherwise looks good to merge once the naming question is settled.

*Claude Opus 4.6 acting as a review assistant for Rob Beezer*

ascholerChemeketa:
How about `debug.profile` or `debug.profile-py` ?

I could see adding timestamp logging to other pathways and wanting to use the same flag to say "show me timing" while debugging.

rbeezer:
I like `debug.profile`.  Thanks.

ascholerChemeketa:
Force pushed with stringparam renamed to ` debug.profile`

rbeezer:
Merged as-s, but for some whitespace.  Questions coming in Drop-In.

PR #2832, opened 2026-04-12
Sample article: add PNG of annotated PreFigure diagram

davidaustinm:
This adds the PNG of the recently added annotated PreFigure diagram in the sample article and addresses the issue raised in #2824


rbeezer:
Thanks for the clean-up, and the retorts.

PR #2834, opened 2026-04-13
Python: refactor LaTeX multi-pass compilation into shared helper

rbeezer:
## Summary

- Two places in `pretext.py` implemented the same "compile LaTeX, rerun until settled" logic with different idioms: a `while` loop for latex-image compilation and a `for`/`break` loop for PDF compilation.
- Extracted a single `_latex_compile()` helper that uses the `while` approach, with an `os.path.isfile` guard (from the PDF version) and a `capture_output` flag to preserve the behavioral difference in stdout handling.
- Both call sites now delegate to the helper, reducing duplication and ensuring consistent rerun logic.

## Test plan

- [ ] Build a document with latex-images (e.g. sample article PDF) and verify images compile correctly
- [ ] Build a full PDF document and verify multi-pass compilation works

*Claude Opus 4.6 acting as a coding assistant for Rob Beezer*


rbeezer:
Merged with an adjusted commit message.

PR #2836, opened 2026-04-29
Fix for Epub CSS compatibility with older epubcheck 

ascholerChemeketa:
Makes a new file to break out web specific styling that is not used for epub/kindle (e.g. interactive widgets).

Uses that to remove a few rules that fail epubcheck < 5.3.

Motivation:
https://groups.google.com/g/pretext-support/c/8qvWd5Tdulk/m/QFymmNSfCQAJ?utm_medium=email&utm_source=footer


rbeezer:
Claude suggests a `has():` is still present.  Comment coming next.  I'm prepared to merge this, but will wait for your verdict on this one.  (Sounds to me like the motivating discussion **is** about use on selectors.)

rbeezer:
Thanks for tackling this, @ascholerChemeketa — the split is clean, and the reduction in `epub.css`/`kindle.css` is exactly what was needed.

One question on completeness before merge. After this PR, the EPUB build still contains a `:has()` selector, emitted from `css/components/interactives/_other.scss:36`:

```scss
.interactive-iframe-container__opener {
  …
  &:has(+ figcaption, + .instructions) {
    float: left;
    …
  }
}
```

Because `_pretext.scss` still does `@use 'interactives/other'` unconditionally, this rule lands in `dist/epub.css` (line 1688 in the rebuild) and `dist/kindle.css` (also line 1688).

Was the original epubcheck < 5.3 failure on a *property* (`scale`, `white-space-collapse`, `pointer-events`, etc.) rather than on `:has()` at the selector level? If so, this rule is harmless and nothing more to do. But if epubcheck rejects `:has()` itself, the simplest fix would be to move `@use 'interactives/other'` from `_pretext.scss` into `_pretext-web.scss` — interactive iframes don't render in EPUB anyway, so the rule is dead weight there.

ascholerChemeketa:
Yes, there is a much broader issue. I didn't notice it when I did a quick test of this because of https://github.com/PreTeXtBook/pretext/pull/2841 - I tested using `epub-mml`, which it turns out has no css file baked in. Thus epubcheck happily passed it.

Other pseudoclasses than `:has()` (e.g. `:is()`) also are going to throw off the broken parser in the old epubcheck. And those are used in many places that I think we do need/want in epub's css.

Recommendations:

1) This should no longer be considered a fix for the google books issue. It is a good thing to do (and extend), but it just solves the smaller issue of not throwing cruft css in the epub files.

If you want, we can back burner this until I get time to factor out more things as part of this one PR.

2) If passing old epubcheck is mission critical, we need to figure out something else. A short term fix might be serve up "legacy" css. A long term fix might be to add a transpilation step to the CSS build that turns modern css into older constructs. That sounds like a lot of headaches to support use of an outdated piece of buggy software.

rbeezer:
> to support use of an outdated piece of buggy software.

Right.  We are not doing that.

Thanks for #2841 - I'll get there soon, but not right away,

Do you think we can finish this one off by doing the move for `interactive` and then draw a line in the sand?  Maybe that will let Humanities Clinic (OP) move forward, and anything more will mean waiting on Google to upgrade?  I'd like to give them the chance to experiment with putting our EPUB on Google Play.

Is that a decent compromise?  It'll let this start with some small isolations, and then it can be improved while on the back burner.

ascholerChemeketa:
Forced pushed with a pass of moving additional low hanging fruit from `_pretext.scss` to `_pretext-web.scss`.

This is still going to fail on the old epubcheck. Any use of pseudoclasses (`:is()`, `:not()`, etc...) fails that check, not just `:has()`.

parayou1978 (approved):

rbeezer:
Some notes:

*  Before/after for format "epub-mml-nozip", I don't see any real changes.  Should I be seeing a whole new CSS file, or something?  Just checking my understanding of some of your comments above.
*  I can do the update of the themes, as a part of merging.  Here it caused a merge conflict since I merged some other CSS stuff "out of order".  No problem, since I instead just dropped the last commit and regenerated later.
*  Claude complains about final newlines going missing - I don't think that really is an issue.  Though we should be consistent, I guess.
*  Claude says there is a `:has()` now, due to the CSS for #sidebyside.  Should not matter here, since a #sidebyside is not rendering well in EPUB anyway.

rbeezer:
> Should I be seeing a whole new CSS file

Ooops, lost my way.  I see #2841 now.

PR #2837, opened 2026-04-29
CSS: avoid blank page after printouts

oscarlevin:
Chrome for windows appears to have started respecting the `page-break-after: always` css for printing even if there is no content after.  This has resulted in all printouts getting an extra blank page (but only for chrome on windows, it appears).  Whether or not this is a bug in chrome, the current PR addresses this by adding a selector for the last `.onepage` element in a printout and specifying to avoid the page-break.


rbeezer:
Merged as-is.  Thanks for the fix.

PR #2839, opened 2026-05-04
Make auto-permalinks not selectable for copy-paste

skiadas:
When a user tries to copy-paste text from a document, they should not also be copying the permalink symbols.


rbeezer:
Thanks, @skiadas!  I expect there will be others, but maybe not?  Merged as-is.  Except: I generally update the CSS themes as a new commit, for two reasons: not necessary, but perhaps instructive for those who don't speak SCSS, and less chance of merge conflicts when I tackle several CSS PRs all at once.

A bit more relevant discussion coming over on `-dev`.

Issue #2838, opened 2026-05-01
list-of bug


siwelwerd:
I stumbled across what I believe is a subtle bug with `<list-of>` today.  We have a list of definitions in our book, but I tried adding a Prefigure image today that uses a `<definition>` and got the following error:
```
error: * PTX:ERROR:   An object (definition) lacks a serial number, search output for "[NUM]"
*              located within: "MX1-image-adjacency" (xml:id)
error: * PTX:ERROR:   An object (definition) lacks a structure number, search output for "[STRUCT]"
error: * PTX:ERROR:   An object (definition) lacks a serial number, search output for "[NUM]"
error: * PTX:ERROR:   An object (definition) lacks a structure number, search output for "[STRUCT]"
```

I believe the `<list-of>` is trying to pull in the `<definition>` out of the Prefigure image.  Here is kind of a minimal example that fails with this error, but works if you either remove the `<definition>s` from the Prefigure on line 8 and 9 or the `<list-of>` at the end.

```
<section>
  <definition><statement>Test</statement></definition>
  <figure xml:id="fig">
    <caption>Figure</caption>
    <image>
      <prefigure xmlns="https://prefigure.org"  label="prefig">
        <diagram dimensions="(300,300)" margins="5" xmlns="https://prefigure.org">
          <definition>f(x)=2.5-x^2/2</definition>
          <definition>a = 1</definition>
          <coordinates bbox="(-4,-4,4,4)">
            <grid-axes xlabel="x" ylabel="y"/>
            <graph at="graph" function="f"/>
            <tangent-line at="tangent" function="f" point="a"/>
            <point at="point" p="(a,f(a))" alignment="ne">
              <m>(a,f(a))</m>
            </point>
          </coordinates>
        </diagram>
      </prefigure>
    </image>
  </figure>
  <list-of elements="definition"  empty="no" />
</section>
```

rbeezer:
Very brief suggestion, while I can.

Try putting `pf:` as a prefix namespace on the problem definition element and see what changes?

siwelwerd:
It throws an error:
```
error: Error loading /workspaces/library/source/linear-algebra/source/04-MX/01.ptx: Namespace prefix pf on definition is not defined, line 387, column 25 (01.ptx, line 387)
error: Error assembling source file: Namespace prefix pf on definition is not defined, line 387, column 25 (01.ptx, line 387)
critical: XML syntax for source file is invalid
critical: Failed to build without errors.  Exiting...
[...]

PR #2840, opened 2026-05-07
Common: for "list-of" only match elements in the default namespace

rbeezer:
## Summary

Fixes the bug reported in #2838: a `<list-of>` would match elements by local name only, so foreign-namespace elements with the same local name as the requested PreTeXt element (e.g. a `<definition>` inside a `<prefigure>` diagram) were swept into the list and then triggered numbering errors (`[NUM]`/`[STRUCT]`).

## Changes

- `xsl/pretext-common.xsl` — three matching predicates inside the `list-of-content` template now also require `namespace-uri(.) = ''`, so only elements in the default (PreTeXt) namespace are considered.
- `examples/sample-article/sample-article.xml` — adds a new subsection with a `<list-of elements="definition" empty="no"/>`. The article already contains a `<prefigure>` diagram with a `<definition>` element of its own, so this list serves as a regression test in addition to demonstrating the feature.

## Test plan

- [x] `pretext/pretext -vv -c doc -f html …` on the sample article: clean build, no `[NUM]`/`[STRUCT]` errors.
- [x] The new `<list-of elements="definition">` lists the real PreTeXt definitions and excludes the prefigure-namespace `<definition>`.
- [x] The pre-existing `<list-of>` uses still produce the same output as before.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -15044,6 +15044,14 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 <list-of elements="theorem" scope="section"/>
             </subsection>
 
+            <subsection xml:id="subsection-list-of-definitions" label="subsection-list-of-definitions">
+                <title>Definitions in This Article</title>
+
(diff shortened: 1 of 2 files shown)


rbeezer:
Resolves #2838.  Root cause uncovered by Claude, but as suspected.  Tests nicely independently of Claude, with, and without, sample article test, just for HTML (but changes are in `pretext-common` so this should be enough).

PR #2841, opened 2026-05-07
Script: all non-kindle epubs get epub.css

ascholerChemeketa:
Fix for bug found while looking at https://github.com/PreTeXtBook/pretext/pull/2836

`epub-mml` or `epub-speech` don't currently get CSS

--------------------

This pull request updates the logic for selecting the CSS file used during EPUB generation in the `pretext/lib/pretext.py` script. The main change ensures that the `epub.css` stylesheet is used as the default for all math formats except "kindle", which continues to use `kindle.css`.

**EPUB CSS selection logic:**

* Updated the conditional so that `epub.css` is used for all math formats except "kindle", instead of only for "svg". This makes the EPUB styling more consistent across different math formats.


rbeezer:
Thanks!  Merged as-is.

PR #2846, opened 2026-05-12
JavaScript: make image zoom always full width

jjrsylvestre:
When an image is in a knowl or in a multi-column or side-by-side (or combinations of those...) the "zoom" function sometimes only makes the image as wide as the container. This edit always makes it the full-width of the content pane.

### DIFF
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -122,15 +122,33 @@ window.addEventListener("load",function(event) {
     /* click an image to magnify */
     $('body').on('click','.image-box > img:not(.draw_on_me):not(.mag_popup), .sbspanel > img:not(.draw_on_me):not(.mag_popup), figure > img:not(.draw_on_me):not(.mag_popup), figure > div > img:not(.draw_on_me):not(.mag_popup)', function(){
         var img_big = document.createElement('div');
-        img_big.setAttribute('style', 'background:#fff;');
+        const content_element = document.getElementById('ptx-content');
         img_big.setAttribute('class', 'mag_popup_container');
-        img_big.innerHTML = '<img src="' + $(this).attr("src") + '" style="width:100%" class="mag_popup"/>';
+        img_big.innerHTML = `<img src="${$(this).attr("src")}" style="width:100%;" class="mag_popup"/>`;
  // place_to_put_big_img = $(this).parents(".sbsrow, figure, li").last();
         place_to_put_big_img = $(this).parents(".image-box, .sbsrow, figure, li, .cols2 article:nth-of-type(2n)").last();
   // for .cols2, the even ones have to go inside the previous odd one
         if (place_to_put_big_img.prop("tagName") == "ARTICLE") {
            place_to_put_big_img = place_to_put_big_img.prev().children().first();
         }
+
+        // find ancestor so that place_to_put_big_img's position is relative to that ancestor
+        var img_big_parent = place_to_put_big_img[0].parentElement;
+        while (img_big_parent.id !== "ptx-content") {
+           const computed_position = getComputedStyle(img_big_parent).position;
+           if (computed_position !== "static") {
+              break;
+           }
+           img_big_parent = img_big_parent.parentElement;
+        }
+
+        const content_element_computed_style = getComputedStyle(content_element);
+        const content_padding_left  = parseFloat(content_element_computed_style.paddingLeft );
+        const content_padding_right = parseFloat(content_element_computed_style.paddingRight);
+        const img_big_offset = content_element.getBoundingClientRect().left - img_big_parent.getBoundingClientRect().left + content_padding_left;
+        const doc_width = content_element.offsetWidth - content_padding_left - content_padding_right;
+        img_big.setAttribute('style', `width:${doc_width.toString()}px; left:${img_big_offset.toString()}px;`);
+
         $(img_big).insertBefore(place_to_put_big_img);
     });
 
--- a/css/components/elements/_media.scss
+++ b/css/components/elements/_media.scss
@@ -66,6 +66,7 @@ img.mag_popup {
 }
 
 .mag_popup_container {
+  background: var(--content-background);
   width:100%;
   position:absolute;
   z-index:1001;


rbeezer:
@ascholerChemeketa - would you be able to take a look at this one?

js/pretext_add_on.js
```diff
@@ -122,15 +122,33 @@ window.addEventListener("load",function(event) {
...
+           }
+           img_big_parent = img_big_parent.parentElement;
+        }
+
+        const content_element_computed_style = getComputedStyle(content_element);
+        const content_padding_left  = parseFloat(content_element_computed_style.paddingLeft );
+        const content_padding_right = parseFloat(content_element_computed_style.paddingRight);
+        const img_big_offset = content_element.getBoundingClientRect().left - img_big_parent.getBoundingClientRect().left + content_padding_left;
+        const doc_width = content_element.offsetWidth - content_padding_left - content_padding_right;
+        img_big.setAttribute('style', `background:#fff; width:${doc_width.toString()}px; left:${img_big_offset.toString()}px;`);
```
ascholerChemeketa: Why white background?

I think this should apply `var(--content-background)` as the background so that if it is visible, it matches the background of the content area of the page (i.e. dark in dark mode).

ascholerChemeketa:
Other than the question about background color, this works as advertised across different themes. 

Looks good.

ascholerChemeketa (changes requested):

js/pretext_add_on.js
jjrsylvestre: That was existing code, so I left it as-is.

If I had to guess, that style is probably because SVGs produced by PreTeXt from TikZ code have transparent backgrounds but (default) black text/line colour, so need to have a white background in case of dark mode.

But it doesn't seem like this style is doing anything in live examples, even in dark mode. Removing that `background:#fff` from the `img_big.setAttribute` call makes no difference in how zoomed images are displayed in HTML output.

Should I remove it?

ascholerChemeketa: Sorry, didn't realize that was carryover.

Inline styles, especially hard coded colors, cause problems for the CSS. There is already CSS to apply a background to SVGs, so this is indeed redundant. Would be nice to drop it while the code is being touched.

jjrsylvestre: OK, done

rbeezer:
I think maybe something is still not right for backgrounds.

Try Figure 26.13 in the sample article.  Before PR: opaque when zoomed out.  After PR: transparent when zoomed out.

jjrsylvestre:
> I think maybe something is still not right for backgrounds.
> 
> Try Figure 26.13 in the sample article. Before PR: opaque when zoomed out. After PR: transparent when zoomed out.

Images in Figure 26.13 are opaque in dark mode, only transparent in light mode.

Probably the CSS only anticipated that a white background would be needed in dark mode...

jjrsylvestre:
To be honest, I can't find the rule that is causing the zoomed image to have a white background in dark mode. If I use the browser's developer tools to add `background:blue;` to the `element.style` of the `div.mag_popup_container` that is created when you click-to-zoom, then the background is only blue in light mode, and it remains white in dark mode...

`grep -r mag_popup pretext/css` was not enlightening...

rbeezer:
> I can't find the rule

Yes, SCSS is a good idea, but it makes drive-by contributions harder.  I wonder if a brief tutorial/explanation in the Guide would be helpful for those who know "plain" CSS well?  I have a new assistant who could make an initial draft.  ;-)

jjrsylvestre:
> Yes, SCSS is a good idea, but it makes drive-by contributions harder. I wonder if a brief tutorial/explanation in the Guide would be helpful for those who know "plain" CSS well? I have a new assistant who could make an initial draft. ;-)

SCSS isn't that difficult to understand. It's more that "drive-by" contributors like me might go months without thinking about CSS at all and get rusty.

I've found the entries responsible for the discrepancy between light and dark mode behaviour:

````
_color-vars.scss
79:  "ptx-image-bg": transparent,

_palette-dark.scss
67:  "ptx-image-bg": white,
````

@ascholerChemeketa Would you prefer those both to be white, or for a new `background` rule for `div.mag_popup_container` to be added somewhere? I imagine that the choice of `transparent` is to respect themes that have a light-mode `background` for `.ptx-content` that is not white?

ascholerChemeketa:
@jjrsylvestre IMO the popup should have a background that matches the page.  `.mag_popup_container` (`_media.scss` line 69) getting `background: var(--content-background);`  would do that.

The image background is a separate thing. Applying a white background to images in dark mode (via `ptx-image-bg`) is a bit of a hack to handle existing images with transparency. It eventually should go away, or at least allow for more control, so that if an author's images do support color schemes (https://cassidyjames.com/blog/prefers-color-scheme-svg-light-dark/), they aren't forced to have a white background behind their SVGs.

jjrsylvestre:
> @jjrsylvestre IMO the popup should have a background that matches the page. `.mag_popup_container` (`_media.scss` line 69) getting `background: var(--content-background);` would do that.

Sounds good, thanks for the help!

ascholerChemeketa:
@rbeezer I really don't think scss is the root of the problem you are worried about. 

Figuring out "why is this element styled like that" is a core CSS/HTML developer skill. Open the browser's inspector and figure out what styles are being applied and where they come from:
image

Doing that is necessary to make a reasonably informed PR whether the source language is CSS or SCSS.  And doing that points to exactly where in the scss the rules are. The scss files are for the most part just css.


Supporting light/dark and different themes is absolutely a source of complexity. But that usually just requires avoiding hardcoding in values like `white` or `black` and using the css variables. that are described in `_color-vars.scss`. That complexity would be the same (actually worse) if we were just using plain css. 



There is already a pretty significant Readme in the css directory. Maybe some of it could get hoisted into the guide.

jjrsylvestre:
I just want click-to-zoom to actually, you know, zoom my images when I click on them. I couldn't care less about the optimal structure of the SCSS/CSS, or the "core CSS/HTML developer skills". I'm a math instructor, not a web developer.

rbeezer:
> There is already a pretty significant Readme in the css directory.

Right.  Thanks for the reminder.  My assistant will like that.

I want to keep knowlegeable and helpful contributors, like @jjrsylvestre, contributing and being productive.  So I probably will develop some materials to help, and perhaps @ascholerChemeketa  can help with a review.

ascholerChemeketa:
That was no shade at you @jjrsylvestre 
Having the image zoom work in nested contexts will be a much needed improvement!

rbeezer:
Alright, wasn't certain this was ready to go, but @ascholerChemeketa let me know in DropIn.

Thanks, @jjrsylvestre - good to have this working much better, and we'll get some (hopefully helpful) documentation going (see #2863).

Merged as-is.

PR #2848, opened 2026-05-12
CSS: add decorations for objectives/outcomes/assemblages in printouts

oscarlevin:
As suggested by @davidaustinm, printouts will now get some basic styling for objectives and outcomes.  I also went ahead and added a simple box for assemblages.

One change to scss outside of the printout targets, to the  `box-mixin` simply adds a default border width so the printout can override this.  Should not impact any other styles.

### DIFF
--- a/css/components/chunks/helpers/_box-mixin.scss
+++ b/css/components/chunks/helpers/_box-mixin.scss
@@ -1,9 +1,10 @@
 // These values can be set on @use to avoid repeating values in each @import
 $pad: 10px !default;
 $border-radius: 0px !default;
+$border-width: 2px !default;
 
 // Generate styles for a surrounding box
-@mixin box($border-width: 2px,
+@mixin box($border-width: $border-width,
   $style: solid,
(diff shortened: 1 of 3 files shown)


davidaustinm:
Thanks, @oscarlevin!

rbeezer:
Quick question on `_chunks-worksheet.scss`, @oscarlevin — the new `@use 'components/chunks/helpers/heading-box-mixin' with ($border-radius: $border-radius);` doesn't seem to be invoked anywhere in this PR.  Is it forward-looking scaffolding for a follow-on commit (e.g., dressing the printout's `theorem-like`/`definition-like` headings), or can it be dropped?

oscarlevin:
Good point Claude.  That was left over from experimenting with different box styles.  Cleaned up in latest force push.

rbeezer:
I like that look!  Thanks @oscarlevin and @davidaustinm.

I split this into two commits - one for light mode and one for decorations.  Otherwise, as-is.

PR #2849, opened 2026-05-12
CSS: ensure print preview icon is clickable

oscarlevin:
In the edge case when a worksheet/handout title is long enough to push the right-floated print preview button to the next line, the following divs could cover it up and make it difficult to click.  This raises the z-index of the print-links class so that it is easier to access.


rbeezer:
Merged as-is, thanks!

PR #2850, opened 2026-05-13
HTML: revise and reformat block-production overview comment

rbeezer:
## Motivation

The long overview comment that introduces the "block production" framework
(currently around line 2444 of `xsl/pretext-html.xsl`) is the front door for
anyone trying to understand how a block — `theorem`, `remark`, `example`,
etc. — ends up rendered as one of three different things in HTML output:

- visible on a main page,
- a born-hidden `<details>`/`<summary>` clickable on a main page,
- the duplicate body of an external cross-reference knowl.

This is a genuinely complicated process: the same generic dispatch template
is entered in three different situations, and the right behavior depends on
two parameters threaded down through nearly every container template.  The
intent of this PR is to make that comment carry its weight for a new
maintainer, so the first time someone reads it they come away with an
accurate mental model and the right names to grep for.

No code changes; comment text and formatting only.

## Commit 1 — content revision

- Replaced the reference to a `make-efficient-knowl` template (which does
  not exist in the codebase) with the actual template names:
  `make-xref-knowls` (collects targets) and `manufacture-knowl` (writes one
  external file per target).
- Replaced the stale `TODO` on item (6) with a real description of the
  `wrapped-content` template, which has been in place for a long time.
- Dropped the 2023-09-23 nomenclature disclaimer.  The revised body now
  uses the current "born hidden" terminology throughout; the historical
  name "embedded knowl" is mentioned once, as context.
- Sharpened the (a)/(b)/(c) modes to state, for each, where the content
  lives, what HTML carries it, and what happens to identifying information.
- Spelled out the asymmetry that confused me when I read this: inside an
  xref-knowl, `b-original=false` flows down the whole subtree (so no `@id`
  is ever emitted in the file), while `block-type` resets to `visible` or
  `hidden` at each block boundary according to the descendant's own
  `is-hidden`.
- Noted the `xref`-to-`mrow` special case (knowl body is the containing
  `md`, not just the row).
- Rewrote item (5) "body" to state precisely when an `@id` is emitted
  (visible original on the wrapper; on the `<summary>` for a born-hidden
  original; never in a duplicate).
- Fixed typo "cross-refernce".

A couple of lower-value asides were dropped (the "(As boolean: …)"
implementation hint on item (1) and the "(eg display math)" exception aside
on item (2)) on the grounds that they confused more than they helped.
Happy to put either back if anyone disagrees.

## Commit 2 — reformatting

The revised block was originally written as a series of very long
single-line `<!-- … -->` paragraphs.  This commit reflows it into the
project's normal one-line-comment style, with every line in the block
padded to the same width so the closing `-->` aligns in a single column
throughout.  Pure whitespace change relative to commit 1.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -2444,40 +2444,112 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 <!-- Block Production, Knowls -->
 <!-- ######################## -->
 
-<!-- Generically, a "block" is a child of a "division"; see the schema for the precise rules.  Blocks also have "significant components": for example, a "solution" is a significant component of an "example".  A "p" can be a block in its own right, can be a significant component of a larger block, or can be neither. -->
+<!-- Generically, a "block" is a child of a "division"; see the schema for  -->
+<!-- the precise rules.  Blocks also have "significant components": for     -->
+<!-- example, a "solution" is a significant component of an "example".  A   -->
+<!-- "p" can be a block in its own right, can be a significant component of -->
+<!-- a larger block, or can be neither.                                     -->
 
-<!-- A block (or significant component) is rendered into HTML in one of three modes: -->
-<!-- (a) Visible and original.  Drawn directly on a main page, with full identifying information (HTML @id, etc.). -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Claude messed up and forgot to add a `Co-Authored by:` message to these commits, but the record here will establish that.

PR #2855, opened 2026-05-16
WeBWorK: write a representation file when server rendering fails

rbeezer:
Companion regression-fix to #2809 (per-exercise WeBWorK representation files).

Before #2809, when the WeBWorK server failed to render a problem (compile error, empty PG, malformed XML response, or missing `<statement>` — collectively "badness"), the script logged the error and built a minimal `<webwork-reps>` placeholder. Because that placeholder was a `SubElement` of a single shared `webwork-representations.xml` tree written once at the end of the loop, it always landed in the file, and assembly could still find a record for the failed problem.

After #2809, each `<webwork-reps>` is its own tree written from inside the loop. The leftover `continue` at the end of the badness block now skips the file write entirely. A failed problem produces no file, and assembly aborts with `PTX:ERROR: The WeBWorK problem with @assembly-id "..." could not be located` — with no useful diagnostic at the originating failure unless the author was watching the original generation log.

Two commits:

1. **Script** (`pretext/lib/pretext.py`) — replace the `continue` after badness detection with `else:` that scopes only the response-parsing block. The downstream rendering-data, `pg`, and per-exercise file-write blocks already contain `if badness:` branches that produce a base64-encoded shell PG with the failure message; those were unreachable before and now produce the intended "faux problem" file.

2. **Assembly** (`xsl/pretext-assembly.xsl`) — when a per-exercise representation file is loaded and its `static/@failure` is set, emit a `PTX:WARNING` naming the assembly-id and failure type, advising the author to re-generate the WeBWorK representations. Without this, a stored failure would be silently rendered as the placeholder in every later build.

### Test

Introduced a syntax error in `examples/webwork/sample-chapter/pg/Adding_SingleDigit_Integers.pg` (`xi:include`d by the `local-pg-file` exercise), re-generated representations against `https://webwork-ptx.aimath.org`. Confirmed extraction logs `PTX:ERROR: ...failed to compile`; `local-pg-file.xml` is now written with `<static seed="20" failure="compile">` plus a faux PG shell; assembly logs `PTX:WARNING: ...has a stored "compile" failure ... A placeholder is being rendered in its place. Re-generate the WeBWorK representations (and watch for errors) to retry.`

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -1839,100 +1839,103 @@ def webwork_to_xml(
             static.set("source", path[problem])
 
         # If there is "badness"...
-        # Build 'shell' problems to indicate failures
+        # Build 'shell' problems to indicate failures.  Fall through (no
+        # "continue") so that rendering-data, pg, and the per-exercise file
+        # write below all execute — those sections have explicit "if badness"
+        # branches that produce a minimal faux problem, and assembly relies
+        # on the file existing for every WeBWorK exercise.
         if badness:
             log.error(badness_msg.format(path[problem], seed[problem], badness_tip))
             static.set("failure", badness_type)
             statement = ET.SubElement(static, "statement")
             p = ET.SubElement(statement, "p")
             p.text = badness_msg.format(path[problem], seed[problem], badness_tip)
-            continue
+        else:
+            # Start appending XML children
+            response_root = ET.fromstring(bytes(response_text, encoding='utf-8'))
 
-        # Start appending XML children
-        response_root = ET.fromstring(bytes(response_text, encoding='utf-8'))
+            # This recursive function is needed in the case of nested tasks.
+            # It is written in such a way to handle task with no nesting, and even an exercise without any task.
 
-        # This recursive function is needed in the case of nested tasks.
-        # It is written in such a way to handle task with no nesting, and even an exercise without any task.
+            def static_webwork_level(write, read):
+                # (tree we are building, tree we take from)
+                # since write is a tree and read is a tree, we use deepcopy to make sure
+                # that when we append nodes we are appending new ones, not intertwining the trees
 
-        def static_webwork_level(write, read):
-            # (tree we are building, tree we take from)
-            # since write is a tree and read is a tree, we use deepcopy to make sure
-            # that when we append nodes we are appending new ones, not intertwining the trees
+                tasks = read.findall("./task")
+                if tasks:
+                    titles = read.xpath("./title")
+                    if titles:
+                        for ttl in list(titles):
+                            title = copy.deepcopy(ttl)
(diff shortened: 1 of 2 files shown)


rbeezer:
Chose to rebase on the merged #2854, but there were no changes in the commits.

rbeezer:
@oscarlevin - we should now be always writing out some sort of file for each WW problem.  It has an indication of any sort of failure at the server trip in `static/@failure` and this will get read when used.  Of course, we cannot protect from somebody removing generated files, but this should cover most cases and there should be reasonably good error messages (which will be upgraded in a minute).

@Alex-Jordan - wouldn't be a bad idea for you to review recent changes and verify that this PR is as good as it could be.  We can consider any tune-ups you wish to propose.

oscarlevin:
@rbeezer, where can I get a list of the file names we expect?  I looked at the version assembled source (which is what I was checking for the existence of webwork previously, but this only has the labels that were provided by the author.  Is there a "cheap" assembled source that would have all the labels?

rbeezer:
There's not another source iteration that short-circuits the whole process like the version version does.

We can't put the generated `assembly-id` onto elements as we make the version tree, since the tree is undergoing (big) changes.  We'd need to add a pass.

I really, really thought I was through jumping through hoops for WW in assembly and could fine-tune other parts.  :-( :-( :-(

Are you identifying other (similar) generated files in some way?  MOM or STACK, or?  There's the chance I don't understand exactly what you need or when?

Can we chat on Tuesday?  Maybe something will come to me.

Is it really so bad to run the assembly process?  This is not the only outstanding request to short-circuit it.

One alternative is to move the version version later in the process, especially since the info is now coming off the containing #exercise.

Thinking as I write, sorry.  Maybe use the "expensive" tree for now until we work out a replacment?

oscarlevin:
That gives me enough to go on for now.  Let's talk Tuesday to see if there are opportunities to speed things up.

oscarlevin:
Okay,  a few more thoughts while I have them.  As far as I can tell, my options for getting assembled source are to use the `method="version"` or `method="static|dynamic|pg-problems"`.  The former succeeds always (and is what we were using to test for the existence of particular elements in general).  The latter three options all fail if webwork has not already been extracted since they assume that we can include the generated xml.  

I see two options: 

1. Add a new method that would result in getting the same assembled source that the extract-*.xsl uses to gather the list of assembly-ids.  Then check these against the list of xml files that we expect in the generated assets.
2. Be less clever and just do a quick dry run of building the static or dynamic assembly-internal, and sniff the errors for "could not load WeBWorK representation", which would tell the cli to try to generate webwork.  

I think option 2 is somewhat brittle, but could do the job.  Option 1 is more work, but then could also be applied to all asset types, including images that don't cause failures, but do result in bad output.

PR #2856, opened 2026-05-16
WeBWorK: clean up diagnostics after switch to per-exercise files

rbeezer:
A small audit-and-cleanup pass over WeBWorK-related diagnostics and comments that had not caught up with #2809 (per-exercise representation files) and #2855 (placeholder file written on server failure). Four focused commits:

- **WeBWorK: extract-pg.xsl comments describe per-exercise representation files** — three documentation-comment blocks in `xsl/extract-pg.xsl` still described the pipeline as building "a single XML file (called `webwork-representations.xml`)". Rewritten to describe the per-exercise format (one file per exercise, named by `@assembly-id`).

- **Assembly: condense WeBWorK file-loading error messages** — in `xsl/pretext-assembly.xsl`, the two `xsl:when` branches that handle "no representations directory configured" and "representation file could not be loaded" emitted 4-line and 8-line `xsl:message` blocks. Each collapses to a single line. Structure (the two branches, the hollow-copy body) unchanged. The hint that a `webwork-representations.xml` in the directory indicates old-format files is kept, in shortened form.

- **HTML: remove redundant error template for unprocessed WeBWorK exercises** — `xsl/pretext-html.xsl` had a `<xsl:template match="webwork[*]">` emitting 4 message lines (with a "nees"→"needs" typo) when an unprocessed `<webwork>` survived into HTML conversion. That condition is already diagnosed at assembly time by the messages above, so this template is redundant and is dropped.

- **Publisher: error on deprecated source/@webwork-problems entry** — promote the `PTX:WARNING` for `source/@webwork-problems` in `xsl/publisher-variables.xsl` to `PTX:ERROR`. Managed directories are the expected configuration now.

A small adjacent fallback in `xsl/pretext-html.xsl` (three-way choice on `rendering-data/@passwd` / `@password` / `@course-password` for backward compatibility with old representation files) is intentionally left alone for this pass.

### DIFF
--- a/xsl/extract-pg.xsl
+++ b/xsl/extract-pg.xsl
@@ -39,8 +39,8 @@
 <!-- project. This ephemeral tree also has some global publisher variable  -->
 <!-- data for how PG should be processed. This tree will be used by        -->
 <!-- pretext.py in conjuction with some form of renderer (webwork2, local  -->
-<!-- PG, or the standalone renderer) to create an XML file called          -->
-<!-- webwork-representations.xml with various representations of each      -->
+<!-- PG, or the standalone renderer) to write, for each webwork exercise,  -->
+<!-- a file of representations of that problem (named by @assembly-id).    -->
 
 <!-- The top of the ephemeral XML tree has some global attributes.         -->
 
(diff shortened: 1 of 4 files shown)


rbeezer:
@Alex-Jordan: I'm done with the WW work.  It'd be a good time for you to catch up.  ;-)  Especially if we need to do some support work.  Thanks for your continued work to improve all this and keep the WW-server side up-to-date.

I'm now free to do some tune-up work on `pretext-assembly.xsl`, since I wanted to streamline and unify this some first.

PR #2845, opened 2026-05-11
Knowls - avoid flash of content when revealing

ascholerChemeketa:
Knowls currently can briefly flash the content before trying to shrink and then reveal it. This is most noticeable when the knowl contains complex content (e.g. MathJax) and has not been rendered yet.

This removes that flash of content and smooths the reveal process by better calculating the final height and padding of the element being revealed. There is still a brief hiccup as complex content is revealed as the content is loaded and rendered.

For issue identified by @siefkenj here: https://groups.google.com/g/pretext-dev/c/XgiX8KhH2Dw/m/eW9srLM8AQAJ?utm_medium=email&utm_source=footer

--------------------------------------

This pull request enhances the animation logic for revealing and hiding "knowl" content, focusing on smoother transitions and improved style management. The main improvements are the preservation and restoration of inline styles during animations, more accurate and visually consistent expand/collapse animations, and better handling of animation state to prevent conflicts.

Key changes include:

**Animation and Style Management Improvements:**

* Added methods to `SlideRevealer` to store and restore the inline styles (`overflow`, `height`, `paddingTop`, `paddingBottom`) of the animated element before and after animations, ensuring that original styles are not lost or corrupted during transitions. (`js/knowl.js` [[1]](diffhunk://#diff-6eecafd760dfec47f29caa1ee83a3ff5ac652dfeed3f1a3b84726cb9b1cda8d7R42-R79) [[2]](diffhunk://#diff-6eecafd760dfec47f29caa1ee83a3ff5ac652dfeed3f1a3b84726cb9b1cda8d7L118-R187)
* Refined the expand/collapse animation to use the element's natural padding and height values for more accurate measurements, and to animate both `paddingTop` and `paddingBottom` independently. The animation now uses the `ease-out` easing for a smoother effect. (`js/knowl.js` [[1]](diffhunk://#diff-6eecafd760dfec47f29caa1ee83a3ff5ac652dfeed3f1a3b84726cb9b1cda8d7R88-R144) [[2]](diffhunk://#diff-6eecafd760dfec47f29caa1ee83a3ff5ac652dfeed3f1a3b84726cb9b1cda8d7L95-R169)

**Animation State and Event Handling:**

[...]

PR #2859, opened 2026-05-17
JS: knowls no longer wait on mathjax rendering

ascholerChemeketa:
Follow up to [first PR](https://github.com/PreTeXtBook/pretext/pull/2845) for https://groups.google.com/g/pretext-dev/c/XgiX8KhH2Dw/m/vSAWz2dVCQAJ?utm_medium=email&utm_source=footer

Knowls will no longer wait on MathJax rendering. Makes them more responsive during MathJax processing on pages with significant amounts of math. Possible animation jank if mathjax is processing content while it is expanding. That is mostly a concern for cross-page knowls, but this also moves typesetting of them up a little, which may help avoid the issue.

Heads up to @jjrsylvestre

### DIFF
--- a/js/knowl.js
+++ b/js/knowl.js
@@ -99,24 +99,19 @@ class SlideRevealer {
       this.animatedElement.style.paddingTop = '0px';
       this.animatedElement.style.paddingBottom = '0px';
 
-      // Trigger the animation to expand or collapse the knowl.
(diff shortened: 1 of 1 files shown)


rbeezer:
I like that per-element typesetting!  Looks like just what we need.

Removed a blank line, but otherwise as-is.  Thanks for iterating through this one.

PR #2864, opened 2026-05-18
Reveal: thread heading levels so slideshow headings are hierarchical

rbeezer:
Reveal.js slideshows emitted heading tags that skipped levels: the title slide jumped `h2` → `h4` for event/date, and a `<section>` title was `h1` with its slides at `h3`, skipping `h2`. With no `<section>` in the source, every slide was an `h3` nominally parented to the subtitle (the situation the existing `slide` `TODO` flagged).

The slideshow is a single monolithic page, so heading levels are now threaded rather than chunked. The only level that genuinely varies — the slide title, and the slide content beneath it — is computed from whether the slideshow uses sections, reusing the existing `mode="hN"` helper. Fixed levels stay hardcoded: slideshow `h1`, subtitle `h2`, section `h2`, abstract `h2`, event/date `h3`.

Resulting outline (both with and without `<section>`): every step into a deeper level is exactly one level, with no skips.

Tested by building reveal.js from the sample slideshow, and from a copy with all `<section>` infrastructure removed, then checking the emitted `h1`–`h6` sequence in each.

### DIFF
--- a/xsl/pretext-revealjs.xsl
+++ b/xsl/pretext-revealjs.xsl
@@ -52,6 +52,24 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <!-- HTML conversion (preferably avoid using this).                -->
 <xsl:variable name="b-reveal-build" select="true()" />
 
+<!-- Reveal.js output is one monolithic page, so heading levels are -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Got a weak topic from Claude, so I edited the commit message.  Otherwise, as-is.

PR #2865, opened 2026-05-20
SCORM: first pass at implementing assignment tracking for scorm

oscarlevin:
For an instructor that wants to give their students a PreTeXt authored quiz online in their LMS, they could upload a worksheet or one-page article as a SCORM archive.  They could include WeBWorK questions, or multiple choice, or parsons, or any of the runestone components.  Students taking the quiz will have the percentage of responses answered correctly added to the gradebook.  They can come back and pick up where they left off, since this should restore their answers.

The only thing it doesn't do (and as far as I can tell, will never do) is let the instructor see the student work.  I have an idea for a non-technical workaround for this: include a "print work" button that generates a simple PDF that the student can upload to the LMS or email the instructor.  

Still a bit of a work in progress.  Some things that could be improved:

- [ ] Restore the correct re-randomized version of webwork questions and their answers.
- [ ] Parsons problems might not be restoring correctly
- [ ] FITB are rendering strangely (possibly this is unrelated to this work)
- [ ] Not sure what to do about "interact" assessments.

All that said, the new javascript will only get loaded if we are building scorm, so this could be merged as is with very little danger.


rbeezer:
Claude has some tiny things I will take care of before a merge.  And other suggestions that could be valuable.  Or not.  Coming next.

rbeezer:
Nice work, Oscar — and I agree with your read that the gating on `$b-build-scorm-manifest` makes this safe to land while the open issues get worked through. The 14-section structure and design-decision comments in the JS made the review pleasant to follow.

A few small asks before merge:

- **Filename**: `js/ptx-scorm-events.js` → `js/ptx_scorm_events.js`. Every other file in `js/` is snake_case (`pretext_add_on.js`, `ptx_search.js`, `lti_iframe_resizer.js`, …); easier to rename now than later.
- **Commit message**: trailing `scorm` → `SCORM` (initialism). Maybe also tighten the doubled "SCORM … scorm" — something like `SCORM: initial assignment tracking implementation`.
- **XSL comment alignment** at `xsl/pretext-html.xsl:13750-13755`: the `-->` markers in the body lines don't line up (three at column 79, two at column 80). Worth re-padding to a single column.

A couple of follow-up notes that don't need to block this PR:

- `suspend_data` 4 KB limit (SCORM 1.2): the comment in section 13 acknowledges it, but there's no runtime check. A `console.warn` when `JSON.stringify(_state)` exceeds 4096 chars would surface silent truncation in the field.
- `initSession()` is effectively dead in the normal path now that `loadRestoreData()` initializes eagerly at DOMContentLoaded — could be inlined/removed in a cleanup pass.
- Console log volume is high (two lines per submission plus the full `suspend_data` key list). Fine for first pass; before this stabilizes, gating behind a debug flag would keep students from seeing their own answer history in devtools.
- WeBWorK badge uses inline styles via `style.cssText`. The element already has a `.ptx-scorm-ww-badge` class — easy to migrate to a CSS rule whenever a SCORM-specific stylesheet comes along.

rbeezer:
I'll catch the "few small asks"

rbeezer:
Merged, with vey minor changes as noted above.

Wanted to say in Drop-In how impressive this is.  But the conversation took a more philosophical turn...

BTW, when the time comes, the publication file has a `host/@platform` entry for HTML (or is it the other way around?) that we might use as a way to specify a SCORM build.

PR #2868, opened 2026-05-20
XSL: provide enhansed source that includes assembly-id

oscarlevin:
To check whether all webwork representation files have been build, the CLI needs access to the `@assembly-id` of webwork elements.  These are in the assembled source, but not the "version" variant (that's too soon) and we cannot use the other representations since they embed the exercise representations that might not exist.  

This PR exposes a new version of "enhanced source" that includes the assembly-ids, but quits before loading the representations.

Adds this as an option to the assembly routines in python.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -5011,6 +5011,8 @@ def assembly(xml, pub_file, stringparams, out_file, dest_dir, method):
         stringparams["debug.assembly.exercise"] = method
     elif method == "version":
         stringparams["assembly.version-only"] = "yes"
+    elif method == "assembly-id":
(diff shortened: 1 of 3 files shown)


rbeezer:
Looks good visually.  I will sic Claude on it, once he comes back from his lunch break.  $20/mo Claude is a bit of a slacker.  I like his $100/mo older brother better.

oscarlevin:
The CLI has a PR ready when this is merged (https://github.com/PreTeXtBook/pretext-cli/pull/1154)

rbeezer:
Merged, as-is.  Typo in commit message, and I ended up rewriting it.

Thanks very much for picking up this one.

PR #2879, opened 2026-05-28
MathJax script load timing

ascholerChemeketa:
Addresses issue described here:
https://groups.google.com/g/pretext-support/c/6WPUi-lfsLM/m/weEOxubjAQAJ?utm_medium=email&utm_source=footer

`<script type="module"` is implicitly deferred and run in order with other `deferred` scripts. 

So we want to first include the `<script defer ...` that loads MathJax into the HTML, then include the `<script type="module"` that calls the startup.

Solution makes sense, and seems to reliably fix the reported issue. I just am not sure why the ordering issue does not break more things.


ascholerChemeketa:
Withdrawing for more testing...

PR #2880, opened 2026-05-28
MathJax in JSXBoard bugfix

ascholerChemeketa:
Fixes issue identified here:
https://groups.google.com/g/pretext-support/c/6WPUi-lfsLM/m/weEOxubjAQAJ?utm_medium=email&utm_source=footer

Question: Should we slap `process-math` on interactives more broadly?


rbeezer:
Thanks very much for getting this fixed!  And especially for adding an example to stress the situation.  A little bit of work on the example:
*  Indented the PreTeXt code.
*  Removed the `\arc` macro, which has no definition available.
*  Added a preview image, QR code, and sidecar file, so the PDF would build.
Updating website now.

PR #2882, opened 2026-05-31
Equation numbering: part-level counter fix and document-root @level

rbeezer:
Two small, independent fixes to equation numbering, surfaced while preparing a larger rework of the numbering passes. Both stand alone and are safe to land ahead of that work.

**1. LaTeX: part-level equation numbering no longer duplicates** *(commits "scope … to part", "reset … at the backmatter boundary")*

For a book with structural parts whose equations are numbered at the part level, the LaTeX/PDF output numbered equations `(I.1), (I.2), …` and **restarted them in every chapter of the part** — genuine duplicate numbers. `\numberwithin{equation}{part}` adds the part-level reset and the Roman part prefix, but the book class's per-chapter reset of the `equation` counter stays in force, so the counter still restarts each chapter.

The fix removes the chapter reset in this configuration (`\counterwithout{equation}{chapter}`), keeps `\numberwithin{equation}{part}`, and strips the prefix to a bare serial, so equations run continuously within each part. Since removing the chapter reset also means `\appendix` no longer resets the counter, an explicit `\setcounter{equation}{0}` is added at the backmatter boundary so the appendices restart at 1. Affects only documents numbering equations at the part level; HTML was already correct.

**2. Assembly: record `@level` on the document root** *(commit "Assembly: record @level on the document root")*

The assembly pass records `@level` on every division except the top-level `book`/`article`/…, though the documented level scale puts the root at level 0. This records it. The sole consumer of the division `@level` attribute (`new-level` in `-common`) matches `part|chapter|…` and never the root, so the change is **output-neutral** for all current conversions; it completes the recorded information and is a prerequisite for whole-document (level 0) numbering in the upcoming rework.

Validated during the equation-numbering work via HTML and PDF output diffs on a parts stress test (plus a parts-removed variant and the sample article).

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -2717,6 +2717,22 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     </xsl:copy>
 </xsl:template>
 
+<!-- The top-level division (book, article, ...) is the root of the   -->
+<!-- division tree, at level 0, which the catch-all does not record.  -->
+<!-- A level-0 numbering scheme counts continuously from the root.    -->
+<xsl:template match="book|article|slideshow|letter|memo" mode="augment">
+    <xsl:copy>
+        <xsl:attribute name="level">
+            <xsl:text>0</xsl:text>
+        </xsl:attribute>
+        <xsl:apply-templates select="node()|@*" mode="augment">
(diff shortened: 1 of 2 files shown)


rbeezer:
First commits from a major reworking of how XSL computes numbers - these are existing (independent) changes identified during extreme testing.

PR #2887, opened 2026-06-02
SCORM: add support for doenet exercises

oscarlevin:
With the changes here, doenet interactive exercises will be graded in SCORM, and anwers will be restored when a student returns to the assignment in their LMS.

This also fixes a bug from the previous Scorm PR, which changed the js file to use underscores but left hyphens in the file name for the generated HTML.

I considered also implementing a publication variable setting platform host="scorm".  I think I'll save that for the next PR when I can think through the consequences further (not sure how the cli and the pretext script should handle this).

### DIFF
--- a/js/ptx_scorm_events.js
+++ b/js/ptx_scorm_events.js
@@ -200,6 +200,27 @@
         answers: {}   // {divId: {answer, correct, percent}} — persisted across sessions
     };
 
+    // In-memory cache of Doenet activity state blobs, keyed by divId.
+    // Populated when the student answers a DoenetActivity (SPLICE.reportScoreAndState).
+    // Used to respond to SPLICE.getState requests when the Doenet iframe reloads.
+    // Also persisted to localStorage (not cmi.suspend_data — state blobs can be large).
+    var _doenetStates = {};
+
+    // Queue for SPLICE.getState requests that arrive before loadRestoreData() has run
+    // (i.e., before _learnerId is set and localStorage can be read correctly).
+    // In practice this never fires — external iframes have network latency — but the
+    // queue makes the code correct under any load order.
+    var _doenetRestoreReady = false;    // true once loadRestoreData() has initialized _learnerId
+    var _pendingGetStateRequests = [];  // queued { source } entries waiting for restore-ready
+
+    // When we respond to SPLICE.getState we record the timestamp so that any
+    // SPLICE.reportScoreAndState that arrives within the initialization window can
+    // be identified as an auto-report from Doenet's own startup sequence rather
+    // than a genuine new student submission.  Doenet reports its current (restored)
+    // score when it finishes initializing — we must not treat that as a re-answer.
+    var _doenetSentStateAt = {};        // divId → Date.now() when SPLICE.getState.response was sent
+    var DOENET_INIT_WINDOW_MS = 8000;   // 8 seconds: generous window for iframe init
+
     // Total auto-gradeable questions found on this page (computed at DOM ready).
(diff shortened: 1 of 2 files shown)


rbeezer:
Thanks, @oscarlevin!  I meant to merge this much sooner than this, since I know you are working on Doenet right now.  Sorry I wasn't faster.  I'll merge with a couple of cosmetic edits.  Claude has some things to say, which I can't really evaluate, so those are coming next.  Follow-up if urgent, fold in to later work if not urgent, ignore if they are noise.

rbeezer:
Reviewed this on Rob's behalf. The XSL side is clean, and the `ptx_scorm_events.js` rename is a real fix rather than cosmetic. The JS is intricate but clearly well-reasoned, and the queueing/guarding shows you've thought hard about load-order and restore races. Nothing here looks like a merge blocker — the points below are things worth confirming or picking up as follow-up, with the first being the one I'd most want a second look at.

**1. The init-report guard may not cover the non-Runestone path.** The guard that suppresses Doenet's post-restore auto-report keys on the event name:

```js
if (ev.event === 'SPLICE.reportScoreAndState' && score !== null) { … init-window guard … }
```

That label is only produced on the Runestone-hook path (via the new `RUNESTONE_TO_SCORM_TYPE['SPLICE.reportScoreAndState']` mapping). When the `window` `message` listener handles `SPLICE.reportScoreAndState` itself — i.e. when `window.RunestoneBase && __ptxScormHooked` is falsy — it relabels the event to `'splice'` before calling `recordInteraction`, so the guard never runs there. Since the auto-report is triggered by our own `SPLICE.getState.response` and `_doenetSentStateAt` is set regardless of Runestone, a SCORM build without the `logBookEvent` hook would route restore→auto-report through the listener as `'splice'`, bypass the init-window check, and could rewrite or zero the saved score on reload — in exactly the non-Runestone build this targets. Can you confirm `RunestoneBase.__ptxScormHooked` is guaranteed truthy in a SCORM package by the time Doenet fires that report? If it always is, this is a non-issue; if not, preserving the real event name through the listener (or applying the same init-window check on the `'splice'` path) would close it. Minor, same area: `_doenetSentStateAt[divId]` is only `delete`d on the accepted-report branch, so a report after the 8 s window leaves a stale timestamp — harmless, just noting.

**2. Repeated reports for the same `div_id`.** State now gets saved in two places (the hook-path early return and `if (ev.state) saveDoenetState(...)` inside `recordInteraction`), with several score-threshold guards. The score math looks net-neutral when the restored value equals the saved one, but I couldn't see the denominator/`_state.graded` accounting in the diff. Worth confirming that a second report for an already-recorded `div_id` — the init auto-report or the belt-and-suspenders save — can't inflate `_state.graded` or the score denominator, i.e. that a repeated `recordInteraction` doesn't write a fresh interaction index on every reload.

**3. `saveDoenetState` vs. the restore-ready queue.** You correctly queue `SPLICE.getState` until `_doenetRestoreReady` so `lsKey()` sees the right `_learnerId`, but `saveDoenetState()` calls `lsKey()` with no equivalent guard — a report arriving before `loadRestoreData()` runs would persist under the wrong learner key. Low-probability given iframe latency (and the comments acknowledge it), but it's an asymmetry: reads are gated on readiness, writes aren't. Draining writes through the same gate would close it; otherwise a one-line note on the assumption is probably enough.

None of these need to hold up the merge — happy to see them handled as follow-up if they pan out.

oscarlevin:
Thanks.  I'll release a new version of the CLI momentarily, and look into your assistant's comments soon.

PR #2888, opened 2026-06-03
Split WeBWorK and STACK out of pretext.py into sibling modules

rbeezer:
`pretext/lib/pretext.py` has grown to ~6,350 lines. This splits the two largest self-contained subsystems — WeBWorK and STACK processing — into their own modules, backed by a new `common.py` of shared low-level helpers. `pretext.py` drops from **6,352 → 4,442 lines**; the public interface is unchanged.

```
 pretext/lib/common.py  |  878 +++++++++  (new)
 pretext/lib/webwork.py |  955 +++++++++  (new)
 pretext/lib/stack.py   |  321 +++++++++  (new)
 pretext/lib/pretext.py | 2374 ++--------  (shrinks)
```

### The import graph is strictly one-way (no cycles)

```
     common.py        - imports nothing from its siblings
       ^  ^  ^
       |  |  +---- pretext.py
       |  +------- stack.py
       +---------- webwork.py
```

`webwork.py` and `stack.py` import **only** `common`; `pretext.py` imports all three. This is documented in a comment block at the head of `pretext.py`.

### What went into `common.py` — and what deliberately did not

`common.py` holds the **genuinely shared, foundational** layer: `xsltproc`, the path/executable/temp-directory accessors, `get_managed_directories`, `get_publisher_variable(_report)`, `get_platform_host`, the `sanitize_*`/`copy_*`/`download_*` helpers, `working_directory`, `targz`, etc. — 30 functions in all, used across multiple conversions and/or by the split-out modules.

Conversion-*specific* helpers were intentionally **kept in `pretext.py`** even though they're low-level: the theme builders, the four `individual_*_conversion` helpers, the runestone-services cluster, and the LaTeX/braille/biblio helpers. The test was "is this genuinely shared, or does it belong to one conversion?" — single-conversion code stays with its conversion.

Two structural constraints shaped the boundary:
- **Module state moves as a whole.** The `__ptx_path` / `__executables` / `__temps` globals and *all* functions that touch them had to land in the same module, or a writer and a reader would bind to different module globals and silently desync. Their accessors are therefore in `common.py`.
- **`common` cannot import upward.** Anything `common` calls must also be in `common` (verified: the set is dependency-closed), which is what guarantees the acyclic graph.

`__module_warning` is defined once, canonically, in `common.py`; `pretext`/`webwork`/`stack` alias it (`__module_warning = common.__module_warning`).

### Re-exports keep the public interface unchanged

Functions that moved to `common.py`/`webwork.py`/`stack.py` but are part of the public surface are **re-exported** from `pretext.py` (`get_ptx_path = common.get_ptx_path`, `webwork_to_xml = webwork.webwork_to_xml`, …), so every `ptx.NAME` call from the driver script resolves exactly as before. No call site in the driver changed.

### The `pretext-cli` repo was consulted

The set of names that must stay re-exported is driven by what calls them externally. Beyond the driver script, **`pretext-cli` was checked** — it calls these functions as `core.NAME` — to confirm the split hides nothing it depends on. Its calls are part of what makes the re-export set as broad as it is (e.g. `individual_*_conversion`, `query_runestone_services`, `assembly_internal`, `xsltproc`, `get_executable_cmd`).

### A failure worth memorializing: the `html` name collision

An automated step qualified every moved reference to a relocated helper (`xsltproc(...)` → `common.xsltproc(...)`). It also rewrote `html.fragment_fromstring(...)` in the STACK code to `pretext.html.fragment_fromstring(...)` — **wrong**: that `html` is `lxml.html`, not the top-level `html()` conversion. It collided only because the code used `from lxml import html`, which violates this codebase's own import rule (*always `import foo`; never `from foo import bar`*) and leaves a bare `html` that shadows the conversion function. A build of the STACK path caught it before it shipped.

The fix is the first commit: module-qualify those imports (`import lxml.html` → `lxml.html.fragment_fromstring`), which removes the shadowing entirely and means STACK needs nothing from `pretext` at all. The lesson is exactly why the house rule exists — `from foo import bar` makes it impossible to tell, locally, where a name comes from.

### Long-term benefits

- **`pretext.py` is ~30% smaller** and easier to navigate.
- **WeBWorK and STACK are now cohesive, independently readable/testable modules** rather than threads woven through a 6,000-line file.
- **`common.py` establishes an explicit foundational layer** with a one-way dependency rule, so future extractions have somewhere clean to stand and the import graph stays acyclic by construction.
- The split surfaced and fixed several pre-existing issues along the way (the import-rule violations, redundant `global` declarations, a latent `NameError`).

### Commits

1. **Module-qualify `lxml`/`coloraide` imports per house style** — fixes three `from foo import bar` violations; eliminates the `html` shadowing described above. Pure precursor.
2. **Drop redundant read-only `global` declarations** — 16 `global` statements that never assign (pyflakes-flagged); behavior-neutral cleanup, early so the moved code is clean.
3. **New `common.py`** — the 30 shared helpers, with public ones re-exported from `pretext.py`.
4. **Move WeBWorK into `webwork.py`** — `webwork_to_xml`, `webwork_sets`, `pg_macros`; imports only `common`.
5. **Move STACK into `stack.py`** — the `_stack_*` cluster and `stack_extraction`; imports only `common`.
6. **Document the one-way import graph** in `pretext.py`.
7. **Fix undefined names in the WeBWorK version-check error** — a pre-existing latent `NameError` (`ww_version`/`ww_domain` → `webwork2_version`/`webwork2_domain`), surfaced by linting the isolated module.

### Testing

Before/after (master vs branch) on the documented targets, all behavior-preserving (identical except build timestamps and nondeterministic generated images): **latex**, **html**, **assembly-static**, **stack** (`-c stack`), **webwork** (`-c webwork`, `examples/webwork/sample-chapter` — 75 representation XMLs identical), and **epub-svg-nozip**. New modules are pyflakes-clean (no unused imports).


rbeezer:
Well, now Claude thinks there is some sort of problem using this with the CLI.  Stay tuned, it may be a bit.

rbeezer:
Heads-up: I've force-pushed an update to this branch. Only one thing changed since the initial push, folded into the **"new `common.py`"** commit alongside the existing re-exports:

- **Re-export `set_ptx_path` and `build_info_message`** from `pretext.py`. Both relocated to `common.py` in the split. They aren't called by the driver script, but the **pretext-cli core wrapper** reaches them through the `pretext` namespace — it calls `set_ptx_path(...)` at import (via `from .pretext import *`) and monkeypatches `build_info_message` — so they need to remain re-exported for the CLI's wrapper to keep working with only a trivial file-copy change on its side.

No other content changed; downstream commit SHAs shifted only because of the fold.

Before re-pushing I re-ran the before/after build suite — latex, html, assembly-static, epub, stack, and webwork — and all outputs are behavior-preserving (identical except build timestamps and nondeterministic generated images).

PR #2890, opened 2026-06-04
Enable publisher page breaks for exercisegroup

sean-fitzpatrick:
Adds one line to the template for `exercisegroup` so that if the `xml:id` of an exercise group is listed in the publisher file in `latex/insersions/@pagebreak`, a page break is added before the title and introduction of an exercise group.

This is to avoid the situation where the title and introduction of an exercise group are at the bottom of one page, with all the exercises on the next page.

### DIFF
--- a/xsl/pretext-latex-common.xsl
+++ b/xsl/pretext-latex-common.xsl
@@ -6163,6 +6163,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         </xsl:choose>
     </xsl:variable>
     <!-- build it -->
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, @sean-fitzpatrick!  Merged, after I removed a trailing space character.

Needs to be in the Guide to keep documentation current, I had my assistant do that.  You will see the extra commit.

Thanks for working though this one on the `pretext-dev` list.

sean-fitzpatrick:
Thanks Rob. I didn't think of adding to the guide, mostly because I was already expecting this to work ;-)

PR #2892, opened 2026-06-05
GDScript code highlighting

DavidCooperWCU:
Made necessary changes for GDScript 2.0 highlighting. 

Documentation was updated to include runestone support, but runestone activecode isn't working yet. (Right now it's set to work with JOBE, but I'm leaning towards a client side solution more like sql.)

Prism has GDScript 1.0 highlighting, so I had to add some custom javascript to handle code with `@`.

Latex's `listings` library supports python, which is pretty close, but not a perfect match.


rbeezer:
Visually, I think this looks good.  A few comments, some about undocumented conventions (sorry!):

*  Generally, we would prefer you rebase your contribution on `master` privately before making a PR.  I think the merge commits are fine, they'll go away when I rebase, but usually we do not want them in the way.  Rebase conflicts are pretty rare, the code is quite modular.
*  With one (of few) commits, new work can be added on, I can combine as part of merging.  A force-push is usually fine.  We generally do not have others adding on to a PR.
*  Let's not inject the JS this way.  For one thing, it does not read so well, with `CDATA`, etc.  But more important, we have never used this device.  It should be in its own file that gets read in via a `script` tag.  Location is good, and the `before-tokenize` should be kept.
*  We could have a boolean variable like `$b-has-gdscript` that is a stricter test than `$b-has-program` and only load the JS as needed.
*  Let's add a code comment before we use Python for LaTeX `listings` - just so there is no confusion in the future.
*  Timing?  Should I wait for an all-clear on Runestone work, or should this go in as soon as it is ready?

Holler if any of this needs more explanation.  With changes, I can do a more thorough review.

DavidCooperWCU:
I added a separate gdscript-prism.js file to inject the syntax highlighting.

I'm not sure how to get syntax highlighting working in the activecode. It looks like it uses code mirror, and I think there is a gdscript module for code mirror. So, I'll need to look into that for a future version.

There are now 2 examples in the sample book as well as the godot-shell folder that holds the godot wasm files.

Sorry about the merge commits. In the future, I won't update the branch after making changes to it.

I submitted a pull request for the rs changes, and I'll let you know when it goes through.

DavidCooperWCU:
I also had some trouble with indentation with a code in a program in a statement for an activity. (Maybe it's all the nesting.)

DavidCooperWCU:
The JavaScriptBridge filtering is to address the largest security concern with using a Godot WASM file. 
This concern is that GDScript can call arbitrary javascript, which could potentially open up the site to known attacks.

The filtering restricts JavaScript calls that aren't specifically postMessages that have a JSON dictionary as a payload, and a known origin. As long as eval isn't used on the JSON dictionary, this should restrict calls to actual GDScript instead of allowing arbitrary javascript code to be run in the WASM.

rbeezer:
Is the Runestone side ready on this?  @bnmnetp was just asking.

Is this ready?  You could mark it as a Draft if not.  But post here when it comes off being a draft, as I don't always see that.  Thanks.

DavidCooperWCU:
@rbeezer, yes. This is ready, and the Runestone side is ready also, @bnmnetp

rbeezer:
Hi @DavidCooperWCU!  I've given this a first pass.  Some details to work out, like the correct XSL, publisher options, etc.  Nothing we can't get straight.

Binary blobs (`*.pck`)  make my ears prick up, for multiple reasons.  I am posting a few questions on our `pretext-dev` list.  I hope you can participate in that.

DavidCooperWCU:
@rbeezer I'm happy to make any changes you feel are necessary. Just let me know. I did respond to your message on `pretext-dev`.  I have added a modest amount of security guardrails for the binary code and the arbitrary gdscript that is allowed to run. The raw javascript that is allowed to run based on a javascript bridge is limited to those sending gdscript programs to be run, so no random javascript can be evaluated. Code needs to be either on the server hosting the iframe or on runestone.academy for it to be loaded, but we may want to do some adversarial tests on that.

DavidCooperWCU:
I have the .zip working in the wasm. The next step is to have a way of converting the directory for a gdscript activecode interaction to a zip during the pretext build.

DavidCooperWCU:
@rbeezer I made the necessary changes, but there are changes needed in the pretext-cli for the changes to work. I'll fork and make a PR for that momentarily.

bnmnetp:
Let me know when I should re-copy everything to the cdn

DavidCooperWCU:
@bnmnetp extra_dependencies/godot-shell-4.6.3 is ready to be re-copied.
@rbeezer please let me know if you need any other changes

rbeezer:
@DavidCooperWCU  - thanks for the progress on this one.  

> I did have to make a PR to the prefix-cli for everything to work.

"work" is always rather vague.  So I am uncertain in how far you've gone.  But for development, you should have the `pretext/pretext` script making HTML, the zip files placed, etc.  That is what I will test with.  The CLI is premature, I think.

There are some initial aspects that need to be worked on before I can do much reviewing.

*  Is everything in  examples/sample-book/ext/gdpractice/  going into the zip file?  Or should we have godot regenerate them from source under the control of the Python script?  Ideally, "ext\" will just have human-authored source.  The main raison d'etre of the script is to create files XSL cannot make itself.  Any filename with a generated hash looks suspect here.  `.DSStore` should not ever be there, it should not be precluded by .gitignore, it is up to an author (who may be using a Mac) to clean up that sort of stuff before it becomes part of source.

*  Python:
    - We like to use `.format()`, not "f-strings".
    - `log.debug/log.info` are used for progress, not `log.error` (this whole scheme may need some work on our end).
    -  Standard-library modules (zipfile, pathlib, itertools) are wrapped in try/except ImportError with `__module_warning` — that pattern is for optional third-party deps; import stdlib normally.
    -  Remove the dead commented line `#archive_root = source_path.parent` (and its trailing whitespace).

* XSL attribute construction in pretext-runestone.xsl. 
  The data-wasm / data-pck / data-scene attributes are built across multiple <xsl:value-of> / <xsl:text> lines with stray blank lines and inconsistent indentation inside <xsl:attribute> (e.g. data-wasm has a whitespace-only line before </xsl:attribute>). Build each value cleanly (a variable or concat()) and drop the blank lines. Also remove leftover commented-out  lines and the duplicated .zip-path block.

* Copyright header in xsl/extract-gdscript.xsl.
  We don't want copyright spread across many contributors.  It says "part of PreTexT" (should be PreTeXt — 4 occurrences) and "Copyright 2026 David G. Cooper" — PreTeXt's boilerplate uses holder Robert A. Beezer. Match the standard header.

* `pretext-html.xsl`, `<script src="{$html.js.dir}/gdscript-prism.js"></script>` has about 5 extra blank lines.

DavidCooperWCU:
> * Is everything in  examples/sample-book/ext/gdpractice/  going into the zip file?  Or should we have godot regenerate them from source under the control of the Python script?  Ideally, "ext" will just have human-authored source.  The main raison d'etre of the script is to create files XSL cannot make itself.  Any filename with a generated hash looks suspect here.  

Each folder in gdpractice is a separate exercise for activecode. If godot regenerates them, then that would require pretext to run a headless version of godot each time it builds the book. Is that preferred to having the files generated once for the book and saved in ext?

> `.DSStore` should not ever be there, it should not be precluded by .gitignore, it is up to an author (who may be using a Mac) to clean up that sort of stuff before it becomes part of source.

Adding .DSStore to the gitignore, just makes is easier for those who develop on a Mac. There is never a need to include .DSStore in a git repository, so, I see no downside in including it in the .gitignore

> * Python:
>   
>   * We like to use `.format()`, not "f-strings".

there are a bunch of f-strings in other parts of the code. Should I clean those up also?

>   * `log.debug/log.info` are used for progress, not `log.error` (this whole scheme may need some work on our end).

I removed the log.error that had do do with progress for code I added.
>   * Standard-library modules (zipfile, pathlib, itertools) are wrapped in try/except ImportError with `__module_warning` — that pattern is for optional third-party deps; import stdlib normally.

I moved the imports to before the optional ones.
>   * Remove the dead commented line `#archive_root = source_path.parent` (and its trailing whitespace).

removed.


> * XSL attribute construction in pretext-runestone.xsl.
>   The data-wasm / data-pck / data-scene attributes are built across multiple xsl:value-of / xsl:text lines with stray blank lines and inconsistent indentation inside xsl:attribute (e.g. data-wasm has a whitespace-only line before </xsl:attribute>). Build each value cleanly (a variable or concat()) and drop the blank lines. Also remove leftover commented-out  lines and the duplicated .zip-path block.

I cleaned up the code.

> * Copyright header in xsl/extract-gdscript.xsl.
>   We don't want copyright spread across many contributors.  It says "part of PreTexT" (should be PreTeXt — 4 occurrences) and "Copyright 2026 David G. Cooper" — PreTeXt's boilerplate uses holder Robert A. Beezer. Match the standard header.

fixed the copyright

> * `pretext-html.xsl`, `<script src="{$html.js.dir}/gdscript-prism.js"></script>` has about 5 extra blank lines.

removed extra lines

DavidCooperWCU:
@rbeezer I made most of the suggested changes you asked for (see previous comment and associated commits). 

I am still unable to use the pretext scripts without the pretext-cli, and xslt tools don't appear to work either, so I'm not sure how to test without the changes to the pretext-cli that I made, but if you are able to, then that's great.

If you believe that adding godot as a requirement for building a gdscript book is the way we want to go, then I'll look into modifying python to run godot headlessly. I am of the opinion that the current implementation where the files are in unzipped form so that they are inspectable is a good compromise between transparency and excess computation.

DavidCooperWCU:
@rbeezer I've added godot to pretext. The `common.get_executable_cmd` method was giving me errors, but since the script needs to check for the correct version anyhow, it just adds noise to the build. Once, there is a correct version that works with the method, my fallback will just check to make sure it's the right version and that the export pack is installed. If it's installed correctly, then it won't do any more work, but if either are missing, then they'll be downloaded once each.

For testing the activecode in the sample book, you'll need to have https running with localhost, and add godot.shell="https://wcu-cs-cooperlab.github.io/Fall2026GameDev/godot-shell-" so that ports 8129 and 8130 can be used for your web server. You also, will need to set up cors on the web server. Let me know if you'd like me to upload a python script for that.

I've done some pretty thorough testing with the prext-cli.

Please, let me know if there are any more changes you'd like.

DavidCooperWCU:
I had one other thought. Since godot is now generating the problems from source during the build of the book, I think it might make sense to specify a different location for the source than the source for the external directory. 

There are two reasons:

1. Right now, the content will be doubled, since it's put in 2 places but only used in the generated location.
2. There are solutions in the exercises, so hiding them as compiled binary code makes it a little harder to just look up the answer.

I suggest one additional attribute in the program that holds the directory name relative to the project.

I suppose another option to handle the solution exposure would be to delete the solution files that are copied to the export directory after the fact, or having some exclusion list for solution files.

rbeezer:
Earlier I said

> The CLI is premature, I think.

Maybe I wasn't clear enough.  I **never** use the CLI.  And we do not advise using it for development.  So I can't test if the current status requires the CLI.


Our goal is to make complicated setup easy for authors who are not technically proficient.  So opening up ports and setting up CORS are not in that spirit.

DavidCooperWCU:
> Maybe I wasn't clear enough. I never use the CLI. And we do not advise using it for development. So I can't test if the current status requires the CLI.

The current status doesn't require the CLI. I just didn't know how to use the python script without it. @ascholerChemeketa helped me figure out the command:

`python ../../pretext/pretext -vvv  -d out-test -f html -c all -p publication-structural.xml sample-book.xml`

Then run a server and point to the out-test directory.



Testing the activecode part will require waiting for the runestone part of this.

DavidCooperWCU:
I cleaned this up and started [a new pull request](https://github.com/PreTeXtBook/pretext/pull/3024).

DavidCooperWCU:
Closing this pr in favor of the one mentioned in the previous comment.

PR #2897, opened 2026-06-06
Search box accessibility update

ascholerChemeketa:
A start on:
https://github.com/PreTeXtBook/pretext/issues/2794

Moves the search dialog to a standard `<dialog>` element and thus provides standard keyboard navigation:
https://webaim.org/techniques/keyboard/
 
Also improves screen reader behavior for the search button and to announce search results. While searching, an invisible to the eye div gets updated to text like`"32 results found"`. This div is marked `aria-live` so that screen readers will announce it, giving the user some idea of what is happening down in the results panel without trying to announce all of the results.

@rbeezer I'm planning on making a number of small PRs as I get to various parts of the UI. This one is split into the usual CSS/JS/HTML commits. For a small commit that touches 2 or 3 files but all in a minor way, (add an `aria-XXX` to HTML and touch up a style or two) do you still want separate commits? Or just one `HTML: accessibility improvements for X`?

### DIFF
--- a/css/components/_pretext-search.scss
+++ b/css/components/_pretext-search.scss
@@ -1,35 +1,31 @@
 
 @use 'components/helpers/buttons-default' as buttons;
+@use 'components/helpers/accessibility' as accessibility;
 
-.searchbox {
+.ptx-search-box {
 
-  .searchwidget {
+  .ptx-search-widget {
     height: 100%;
   }
   
-  .searchresultsplaceholder {
-    position: fixed;
-    top: 5vh;
-    bottom: 5vh;
-    padding: 1em;
-    left: max(10vw, calc(100vw - 800px) / 2);
+  .ptx-search-dialog[open] {
     width: 80vw;
+    height: 100%;
     max-width: 800px;
     border: 2px solid var(--body-text-color);
     background: var(--knowl-background, #eaf0f6);
-    z-index: 5000;
     display: flex;
     flex-direction: column;
   }
 
-  .searchresultsplaceholder article {
+  .ptx-search-dialog article {
     width: 60%;
     margin-left: auto;
     margin-right: auto;
     font-family: sans-serif;
   }
 
-  .search-results-controls {
+  .ptx-search-dialog-controls {
     display: flex;
     justify-content: space-between;
     align-items: stretch;
@@ -38,26 +34,30 @@
     height: 35px;
   }
 
-  .ptxsearch {
+  .ptx-search-terms {
     flex: 1 1;
   }
   
 
-  .closesearchresults {
+  .ptx-search-close {
     @include buttons.ptx-button;
   }
 
-  .detailed_result {
+  .ptx-search-detailed-result {
     margin-bottom: 10px;
   }
 
-  .searchresults a:hover {
+  .ptx-search-results a:hover {
     text-decoration: underline;
     background: var(--link-active-background);
   }
 
+  .ptx-search-status {
+    @include accessibility.sr-only;
+  }
 
-  .searchresults {
+
+  .ptx-search-results {
     padding-left: 10px;
     margin-top: 0;
     overflow-y: auto;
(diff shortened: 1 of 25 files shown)


rbeezer:
Thanks, @ascholerChemeketa.  Let's keep HTML changes and CSS/JS changes on different commits.  So typically two commits on small related changes?  OK compromise?  ;-)

rbeezer:
Review coming from Claude, which might contain worthwhile suggestions, and might contain noise.  Sorry to do this - perhaps I should be asking @oscarlevin to expand his remit from just CSS?  Anyway (3) on the no-list for `@label`, etc is brand-new (and necessary).

rbeezer:
Really nice direction — moving to a native `<dialog>` gets you a proper focus trap, Esc-to-close, and keyboard navigation for free, and the `aria-live` result count plus screen-reader-available button names are clear wins. A few things worth addressing before merge:

**1. The dialog open/close is now fully declarative, with no JavaScript fallback.**
Opening depends on `command="show-modal"`/`commandfor`, closing on `command="close"`, and dismissal on `closedby="closerequest"`. The Invoker Commands API and `closedby` are very recent (2025 baseline), and the old JS that toggled the panel has been removed — so on any browser without Invoker Commands support, the search dialog won't open at all. Since the `<dialog>` element itself is broadly supported, the robust approach is to open/close from JS (`resultsDiv.showModal()` / `.close()` in the existing handlers), which works everywhere `<dialog>` does. The declarative `command` attributes are great as progressive enhancement, but probably shouldn't be the only path.

**2. The heading class rename looks like it breaks the legacy theme.**
The HTML changes `class="search-results-heading"` to `class="searchresultsheading"`. The modern SCSS doesn't style either name, but the legacy stylesheet (`css/legacy/pretext_search.css`) targets `.search-results-heading`, and this template is shared across themes — so the legacy heading loses its styling. Reverting the rename (keeping `search-results-heading`) is the simplest fix, unless the legacy CSS is updated to match.

**3. The new `search-status` id should be added to the reserved-id list.**
`<div id="search-status" …>` is a new fixed HTML id, but the `identifier-warning` reserved list in `xsl/pretext-common.xsl` (which already protects `searchbutton`, `searchresultsplaceholder`, `searchempty`, etc.) doesn't include it. Adding it there keeps an author from silently colliding with `xml:id="search-status"`.

**Minor:**
- The `i18next` stub is declared as an unconditional `class i18next`, which would collide with the real library if it's ever loaded for actual localization. Guarding it (e.g. `const i18next = window.i18next || { … }`) or using a distinct name avoids that.
- `type="button"` was dropped from the search button — harmless now (no `<form>` ancestor), but worth keeping so it can never implicitly submit.
- `searchInput.select()` was removed; if recalling previous terms, that previously let the user type over the selection — intentional?

ascholerChemeketa:
Good to know about #3. Will touchup based on that and other suggestions - some of those do look good.

Roger re: other similar PRs

ascholerChemeketa:
Question: should I update the ids with `ptx-` while I am in this code?

I can't imagine there are too many people overriding parts of the search interface, so it doesn't seem too likely to break anyone's book.

rbeezer:
> should I update the ids with ptx- while I am in this code?

Yes I think that would be great.  Should be safe, as you surmise.  And with more coverage, perhaps others will see the pattern.  Thanks.

ascholerChemeketa:
Updated with suggested changes.

rbeezer:
Looking good.  I'll post Claude's latest review here now.  I will fix the one missing space here as I merge this.

rbeezer:
Thanks for the rework — this is a big improvement, and it addresses everything from the earlier pass: the dialog now has a real JS fallback (`hasNativeInvokers` → `showModal()`/`.close()`) so it works everywhere `<dialog>` is supported, with the `command`/`commandfor` attributes left as progressive enhancement; the class rename is carried through all the legacy stylesheets in lockstep so the legacy themes keep their styling; the `i18next` stub is now guarded with `window.i18next || …`; and `type="button"` and `searchInput.select()` are both back. Nice.

One thing to fix before merge:

**The reserved-id list lost a token separator.** In `xsl/pretext-common.xsl`, the new line ends without a trailing space:

```
<xsl:text>... ptx-search-status ptx-search-empty ptx-search-close</xsl:text>
<xsl:text>light-dark-button papersize-select ...</xsl:text>
```

Since `reserved-html-ids` is consumed with `str:tokenize` (whitespace), the two pieces concatenate into a single token `ptx-search-closelight-dark-button`. The net effect is that **both** `ptx-search-close` *and* `light-dark-button` fall out of the reserved set, so an author using either as an `xml:id` would no longer get the collision warning. Adding the trailing space back (matching the other lines in the block) restores both.

A couple of smaller, non-blocking notes:

- **Screen-reader status text is hardcoded English.** The two `searchStatus.innerHTML = i18next.t('…')` strings (`'No results found for "{{terms}}".'`, `'{{count}} results found.'`) bypass PreTeXt's localization, unlike the results heading, which still goes through `type-name`/`string-id`. The `i18next` stub is clearly a forward-looking placeholder for exactly this, so it's understood — but as written, the `aria-live` announcements stay English regardless of document language. Worth a follow-up to route them through localization string-ids.

- **Vestigial display toggles.** `addResultToPage` still does `document.getElementById("ptx-search-dialog").style.display = null;` (twice). With a real `<dialog>` shown via `showModal()` and hidden via `.close()`, those inline-display writes are no-ops now — leftovers from the old div approach. Harmless, just cleanup.

rbeezer:
Thanks very much!  Great to see this improving (and that there will be more coming).  Thanks, too, for the ptx-* fixups.

Merged as-is.  Added the missing space mentioned above, and removed two trailing spaces elsewhere.

PR #2898, opened 2026-06-06
WeBWorK: group WeBWorK representations into a folder

Alex-Jordan:
Previously, the `generated/webwork/` folder had a few subfolders (`pg`, `images`) and one giant representations file. Now there are representations files for each exercise, and that folder is full of them. The presence of the `pg` and `images` folders is drowned out by so many individual representation files.

So this change makes a `representations` folder and puts all of the representations in there. The folder tree structure is now:

```
generated
    webwork
        images
            [image files]
        pg
            [pg files]
        representations
            [rep files]
```


Alex-Jordan:
I see that all of the representation files for the sample chapter and showcase article are committed to the repo. How would you like that handled with this PR? As a separate commit? I'm not sure what makes it easiest for you to do before/after testing that nothing really changed with those things.

Alex-Jordan:
For the three examples that have webwork exercises, I moved their repo representations files into the new folders this PR would create. That's on a separate commit, and I'm unsure if you'd prefer that they be merged.

rbeezer:
At first blush, I see some other changes that need to be made.  For example a message `A "webwork-representations.xml" in that directory indicates old-format files that need replacing.` is now incorrect since "that directory" is now the new one and not the old one.

I've not checked, but will image files be found correctly?  I can't tell if some paths would need to change.

Alex-Jordan:
OK, that message is reworded.

Yes, files are found correctly, since all of the looking for them passes through a publisher variable that is also adjusted here. I've successfully built HTML and PDF with this change, and with all old files in `generated` cleared out first.

rbeezer:
OK, this looks good.  But it will create a headache for the CLI, methinks, and @oscarlevin.
*  Assets will need a regeneration.  I was trying not to put authors onto a treadmill with repeated instances of this, as we bundled up some changes all at once already.
* The CLI looks for the one grand file as a hint for re-generation, I think.  Maybe that logic gets more complicated, looking now for some files at the top-level.

@oscarlevin is already hopping with other changes!  What is the outlook, Oscar - should this be scheduled, or sooner is better?  @Alex-Jordan - no other changes of this magnitude contemplated?

Alex-Jordan:
I haven't surveyed the landscape well following changes from this year. But
as far as file locations, I would not expect more changes.

Would it help to look for *any* .xml file at the generated/webwork/ level,
and report based on finding such a file? There shouldn't be such files
following this change.

Alex Jordan
Mathematics Instructor
Portland Community College

oscarlevin:
I'm pretty sure that the CLI looks for each file name that is needed and
regenerates everything if even one is missing.  So either folks would need
to move all the files to the new location, or regenerate everything and
would have two copies of each.

So I don't think there is much I would need to change (other than a line or
two changing where to look for the files), but it would be a hassle for
authors.

Not sure how useful it is to have the added file structure.  I don't think
there are too many people viewing the contents of generated assets anyway.
I think of that folder as mostly being organized to help the python tools
keep track of things.

oscarlevin:
In terms of timing, if this gets merged in the next few hours, I should be
able to get it out in the CLI today.  Otherwise it will have to sit for a
week.

Alex-Jordan:
If my suggestion to look for .xml files in the generated/webwork/ folder is
worth doing, I can do that. But not until this evening.

I'm not in a rush to get this in. Maybe it's not worth doing at all. It
seems "right" to me, but would have been "more right" if I'd been able to
follow along back when the change was made.

Alex Jordan
Mathematics Instructor
Portland Community College

rbeezer:
Well, I've been away, so missed Oscar's window it seems.

And this is really only if @Alex-Jordan wants it (which is fine).  I like structure.  ;-)

But we need to do it soon, before @oscarlevin embarks on any kind of caching to differentiate the all-or-none dichotomy.  If it sits for new week, then it needs to sit for two, and so should be marked off as a draft.

oscarlevin:
I think it would be fine to merge, it just might make the nightlies fail
until I get back from vacation.

Alex-Jordan:
OK. Now suppose an individual reps file is not inside `generated/webwork/representatons/`.

1. It checks if the file is in the parent `generated/webwork/` folder. Instructs user to move the file or rebuild reps.
2. Failing that, it checks if there is a `webwork-representations.xml` file in the parent `generated/webwork/` folder. Informs the user this is an old format, an instructs user to rebuild reps.
3. Failing that, a very generic error message is given to rebuild representations.

Alex-Jordan:
This had gotten lost and fell behind some conflicts. I've caught it back up.

rbeezer:
I mant to comment on this a couple weeks ago.  But my comment got lost as I try to multitask with two Claude instances and everything else going on.  Here it is verbatim.  I'm sorry I didn't get this to you before you did some catch-up.

> We had a chance to do this when the changes were first made.  I wasn't certain that I wanted to make/announce another disruption then, and now that I have hesitated, I am even more reluctant.

> I saved up two big set of changes for one disruption, I think we should do the same here.  Next time we want to change how representations are made, in a way that forces regeneration, we could consider this.  It seems to be working now just fine, though I am sympathetic to better organization (that does not look like "mixed content"!).

PR #2899, opened 2026-06-07
WeBWorK: clean up some remnants from when there was one big WeBWorK r…

Alex-Jordan:
…epresentations file

While looking over #2898, I saw these places where some comments have gone stale. Not directly related to #2898, so I put here in a separate PR.


rbeezer:
Thanks.  Merged as-is.  Should not conflict with #2898.  One comment is out of sync with "missing" #2898, but that won't be for long.

PR #2900, opened 2026-06-07
JS: remove unused ptx_search.js file

ascholerChemeketa:
Quite confident this file is cruft - it looks like an initial version of `pretext_search.js`.


rbeezer:
Thanks for the maintenance!  Claude agrees this is cruft. Merged as-is.

PR #2902, opened 2026-06-08
Improve keyboard navigation for embed button

ascholerChemeketa:
Change the embed nav button to native dialog and improves keyboard navigation handling.

Updates the id's for the HTML components to be `ptx-` prefixed. This potentially breaks after market styling, but that seems low likely hood and impact.

Related issue:
https://github.com/PreTeXtBook/pretext/issues/2794

### DIFF
--- a/css/components/_pretext-search.scss
+++ b/css/components/_pretext-search.scss
@@ -1,28 +1,19 @@
 
 @use 'components/helpers/buttons-default' as buttons;
+@use 'components/helpers/dialogs' as dialogs;
 @use 'components/helpers/accessibility' as accessibility;
 
 .ptx-search-box {
-
-  .ptx-search-widget {
-    height: 100%;
+  .ptx-search-dialog {
+    @include dialogs.ptx-dialog-modal;
   }
-  
+
   .ptx-search-dialog[open] {
-    width: 80vw;
-    height: 100%;
-    max-width: 800px;
-    border: 2px solid var(--body-text-color);
-    background: var(--knowl-background, #eaf0f6);
     display: flex;
     flex-direction: column;
-  }
-
-  .ptx-search-dialog article {
-    width: 60%;
-    margin-left: auto;
-    margin-right: auto;
-    font-family: sans-serif;
+    max-width: 800px;
+    width: 80vw;
+    height: 100%;
   }
 
   .ptx-search-dialog-controls {
@@ -37,10 +28,11 @@
   .ptx-search-terms {
     flex: 1 1;
   }
-  
 
   .ptx-search-close {
-    @include buttons.ptx-button;
+    @include buttons.ptx-button(
+      $border-width: 0
+    );
   }
 
   .ptx-search-detailed-result {
@@ -56,7 +48,6 @@
     @include accessibility.sr-only;
   }
 
-
   .ptx-search-results {
     padding-left: 10px;
     margin-top: 0;
@@ -64,6 +55,11 @@
     flex: 1 1;
     background: var(--content-background, white);
     border: 1px solid var(--page-border-color, #ccc);
+    color: var(--body-text-color, black);
+  }
+
+  .ptx-search-results-heading {
+    margin-top: 0;
   }
 
   .ptx-search-results:empty {
@@ -103,10 +99,6 @@
 
   .ptx-search-result-clip-highlight {
     background: var(--ptx-search-resultshighlight);
-  }
-
-  .ptx-search-dialog::backdrop {
-    background: var(--ptx-search-resultsbackground, white);
   }
 
   @media screen and (max-width: 800px) {
--- a/css/components/helpers/_buttons-default.scss
+++ b/css/components/helpers/_buttons-default.scss
@@ -1,7 +1,9 @@
 $border-radius: 0 !default;
+$border-width: 1px !default;
 
 @mixin ptx-button(
-  $border-radius: $border-radius
+  $border-radius: $border-radius,
+  $border-width: $border-width
 ) {
   font: inherit;
   display: flex;
@@ -12,7 +14,7 @@ $border-radius: 0 !default;
   min-height: 34px;
   color: var(--button-text-color);
   background-color: var(--button-background);
-  border-width: 1px;
+  border-width: $border-width;
   border-color: var(--button-border-color);
   border-style: solid;
   border-radius: $border-radius;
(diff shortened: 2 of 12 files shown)


ascholerChemeketa:
Marking as draft temporarily ... factoring out some common stuff. Going to try to make a simple dialog js/css component to reuse code explicitly

ascholerChemeketa:
@oscarlevin - question about Greeley...

A consolidated framework means that all dialogs, absent extra effort, will share the same design features (IMO a good thing and part of the reason to do it).

The old embed dialog used the navbar color as the background. Which is pretty vivid in Greeley and Salem:


It would be even more vivid if applied to other dialogs. Imagine a "Reader options" with a bunch of labels/inputs in a sea of bright blue.

For Salem, I am setting using a clean low color approach:


Another alternative would be to match the styling for the search box in Greeley:


Let me know if for Greeley I should:
1) Use the existing vivid colors
2) Match Salem
3) Match what search looks like

Obviously you can tinker to your heart's content. So this is mostly about should I add some logic to start you off at 2 or 3 instead of 1.

oscarlevin:
I like option 3 to start with.  Thanks.

rbeezer:
Works well for me.  Like the way these are going.  Will merge, after a rebase onto `master`.

Might be some nuggets in Claude's review.  If anything seems urgent, a follow-up PR would be fine.  That's next.

rbeezer:
I'll add the updated themes, too!

rbeezer:
Ooops, I see this is a draft.  I was all ready to go...  Review coming all the same.

rbeezer:
Nice refactor — consolidating the dialog handling into a shared `PTXDialog` is the right direction, and the keyboard/focus basics (autofocus into the dialog, focus return to the opener on close, `aria-label`led close buttons, Esc via `closedby`) look good. A few things I noticed reading through the diff:

**1. The open/close wiring is doubled up, and the two dialogs don't do it the same way.** Both invoker buttons keep a native `command`/`commandfor`, and `PTXDialog` *also* attaches a click listener unconditionally:

```js
if (this.openButton) {
    this.openButton.addEventListener("click", () => this.open());
}
```

Two wrinkles fall out of that:
- The search button uses `command="show-modal"` (the built-in invoker command), so on browsers with native invokers it's wired twice — native *and* JS. The previous code gated the JS on `!hasNativeInvokers` to avoid exactly this.
- The embed button uses `command="showModal"`, which isn't the built-in command name (`show-modal`), so the native invoker is inert there and only the JS opens it.

So the two dialogs take different open paths, and the `hasNativeCommandInvokers()` branch is mostly moot since both branches end up calling `showModal`/`close` from the JS listeners anyway. It'd be cleaner to pick one mechanism and apply it consistently — either drop the native `command`/`commandfor` and let `PTXDialog` own the wiring, or keep them and only attach the JS listeners when `!hasNativeCommandInvokers()` (and fix `showModal` → `show-modal` on the embed button). One sharp edge if the JS path stays as-is: the native-branch `this.open = () => this.dialog.showModal()` has no already-open guard, so it would throw `InvalidStateError` if ever called on an open dialog.

**2. The embed `<dialog>` carries `popover="auto"` alongside `showModal`.** Popover light-dismiss only applies when an element is shown *as a popover*, so on a dialog opened with `showModal()` it doesn't really do anything (Esc is handled by `closedby`, outside-click by the `kind: "light-close"` handler). The search dialog omits it — dropping it from embed keeps them consistent.

A few smaller things:
- The auto-generated close button gets `close()` bound twice — once in the `if (!this.closeButton && this.isModal)` block and again in the following `if (this.closeButton)` block. Harmless, but likely unintended. (Search and embed pass their own close button, so they only hit the second.)
- The embed popup keeps its old absolute-position offsets (`margin-top` / `bottom: $nav-height`) now that it's a centered modal `<dialog>` — worth checking it lands where you expect in top-nav, bottom-nav, and dark mode.
- `_dialogs.scss`: `ptx-dialog-modal` re-declares the same `::backdrop` rule it already inherits from `ptx-dialog`, and the file is missing a trailing newline.
- The i18next stub is duplicated in `pretext_add_on.js` and `pretext_search.js`, and it brings in i18next as a runtime-string localization path alongside the existing XSL/`$locale` mechanism — might be worth a deliberate call on whether that's the direction for JS-side strings before it spreads further.

For what it's worth, the id/class rename looks complete across the source (no dangling `embed-*` references), the reserved-id list stays in sync with the emitted HTML, and the `embed-button` publication option is correctly left untouched. Given the goal, the highest-value check from here is probably a manual keyboard + screen-reader pass over both dialogs — open, tab-trap, Esc, and focus-return.

ascholerChemeketa:
Updated with option #3 for @oscarlevin and Claude's latest suggestions.

Gemini didn't find any new issues... let's see what Claude says. ;) If he is happy, should be good to human review.

rbeezer:
Thanks.  Timing is good.  ;-)

rbeezer:
Thanks — this resolves everything from the last round. The open/close wiring now gates the JS listener on `!hasNativeCommandInvokers()`, so it no longer doubles up with the native invoker; the embed button uses the built-in `command="show-modal"`; `popover="auto"` is gone; the auto-generated close button no longer binds `close()` twice; and the duplicate `::backdrop` is removed. Search is clean too — `PTXDialog` owns open/close and the leftover button listener is just the restore-last-search bit. The non-modal `show()` path and the `getBoundingClientRect()` light-dismiss hit-test are nice touches, and both JS files still pass a syntax check.

A few small leftovers:
- `_dialogs.scss` and `greeley/_customization.scss` are each missing a trailing newline.
- Typo in the light-dismiss comment: "bounding rext" → "rect".
- The i18next shim is still defined in both `pretext_add_on.js` and `pretext_search.js` — harmless with the `||` guard, but worth a deliberate call on whether i18next is the path for JS-string localization before it spreads.
- The embed popup still applies `top: auto` / `margin-bottom` / `bottom: $nav-height` to what is now a centered modal `<dialog>` — probably fine given how much you've eyeballed the theming, but worth one confirm in top-nav, bottom-nav, and dark mode.

Nothing blocking from my side. The remaining real test is a manual keyboard + screen-reader pass over both dialogs (open, tab-trap, Esc, focus return).

rbeezer:
Merging as-is.

rbeezer:
Hah - merge conflict with reserved ids edits taken in out-of-order.  We need the `ptx-*` generalization yesterday!  ;-)

rbeezer:
Two trailing blank characters, and one line with four blank characters.  It is a small fiddle to find and fix these.  Do you have a tool you can use to find them before I do?

rbeezer:
Merged with whitespace fixes silently slipped in.  As before, thatnks for all the tune-ups and the accessibility push.

PR #2904, opened 2026-06-09
Compute block serial numbers once, during assembly

rbeezer:
Following the equation and footnote serial-number work in #2885, this moves the **block** family of numbers off the conversion-time `xsl:number` machinery and onto a single stamped `@serial`, computed once during assembly as each item's position among its numbering peers. The HTML conversion then reads that attribute instead of recomputing a count at every numbered item and at every cross-reference to it.

### What it covers

The interacting block counters: theorem-like blocks (the shared counter), `figure`s, `project`s (and `activity`/`exploration`/`investigation`), and inline `exercise`s. Each can run on its own counter through the publisher-file `@distinct` switches added in #2903; otherwise they share the one block counter. The structure-number side (the chapter/section prefix) is unchanged.

### Effect on output

Rendered block numbers are reproduced unchanged — checked against the sample article and the numbering stress-test document in both HTML and PDF — with two exceptions:

- **Fix:** a `figure` inside an *unnumbered* `exercises` division was given an incorrect serial in HTML; it now matches the PDF.
- An inline `exercise` placed directly in an `introduction` or `conclusion` is now left unnumbered (the same as the PDF), whereas master currently gives it an incorrect, colliding number. Whether such exercises should be numbered at all is a separate question — the Author's Guide advises against numbered items in an introduction or conclusion — and is recorded as a known issue, not addressed here.

LaTeX/PDF continues to use its native counters; this change is on the HTML side.

### Side benefit

Because numbers are read rather than recomputed per reference, the HTML conversion is faster: in local timing, roughly 15–25% less XSL transform time on the sample book and the cross-reference-heavy sample article.

### Scope and follow-ups

- The old `overall-blocks` / `atomic-*` templates are left in place (now unused by HTML) rather than removed; pruning can follow once the direction is confirmed.
- Open problems (`openproblem`/`openquestion`/`openconjecture`) are **not** numbered here — they keep their existing placeholder, deferred to a follow-on that adds them as a fourth distinct-capable counter.
- A second pre-existing issue is left unchanged and will be filed separately: a numbered item in a `conclusion` can receive different numbers in HTML and PDF (the conclusion's items continue the count in HTML but restart under LaTeX). Equations and footnotes already exhibit this.

### DIFF
--- a/xsl/entities.ent
+++ b/xsl/entities.ent
@@ -132,6 +132,13 @@
 <!ENTITY FIGURE-LIKE "figure|table|listing|list">
 <!ENTITY FIGURE-FILTER "self::figure or self::table or self::listing or self::list">
 
+<!-- Composite filters for the numbering "serial-stamp" pass.  The      -->
+<!-- "fundamental" blocks always share the overall blocks counter; a    -->
+<!-- "top" figure-like is a captioned display that earns a number (a    -->
+<!-- panel inside a captioned side-by-side earns a letter instead).     -->
+<!ENTITY FUNDAMENTAL-BLOCK-FILTER "(&DEFINITION-FILTER;) or (&THEOREM-FILTER;) or (&AXIOM-FILTER;) or (&REMARK-FILTER;) or (&COMPUTATION-FILTER;) or (&EXAMPLE-FILTER;)">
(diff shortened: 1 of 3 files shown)


rbeezer:
See #2905 and #2906.

PR #2909, opened 2026-06-10
Assembly: bug fixes and fewer passes in the pre-processor

rbeezer:
A round of cleanup in `pretext-assembly.xsl`, the assembly (pre-processor) stylesheet: three latent bug fixes, two reductions in the number of full-tree passes, and one match simplification.

**Bug fixes**
- **Duplicated template** — `match="*" mode="standalone-filename"` was defined twice, identically (an XSLT 1.0 conflict).
- **Attributes processed twice** — the `interactive` repair applied `@*` explicitly and again via `node()|@*`.
- **Self-referencing attribute paths** — the `titlepage` and front-`colophon` repairs selected attributes through a child path that does not exist with the element already as context, silently dropping authored attributes (including an `@xml:id` that may be a cross-reference target).

**Fewer passes**
- **Skip the private-solutions pass** when no private-solutions file is configured (the common case), rather than copying the whole source tree to no effect.
- **Fold the `language` pass into the `labels` pass** — both only stamp attributes, so this removes one traversal of the entire source. It also fixes a latent equal-priority template conflict by which a root element carrying an unsupported `@xml:lang` could skip the `en-US` fail-safe.

**Cleanup**
- **Collapse a 71-line match** for interactive-exercise representations (six element names × eleven interactivity values) into six lines with a fenced-string `contains()` test.

Rendered HTML and LaTeX output is unchanged. The `language`/`labels` fold does reorder two internal attributes (`locale-lang`, `unique-id`) on elements carrying `@xml:lang`; this is confirmed canonically identical (`xmllint --c14n`) and affects only the developer-facing `assembly-*` output, not any rendered format.

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -337,12 +337,6 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     </xsl:copy>
 </xsl:template>
 
-<xsl:template match="node()|@*" mode="language">
-    <xsl:copy>
-        <xsl:apply-templates select="node()|@*" mode="language"/>
-    </xsl:copy>
-</xsl:template>
-
 <xsl:template match="node()|@*" mode="augment">
     <xsl:param name="parent-struct" select="''"/>
     <xsl:param name="level" select="0"/>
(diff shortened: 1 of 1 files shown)


rbeezer:
Some similar work was split off into #2910.

PR #2910, opened 2026-06-10
Assembly: fixes and a skipped pass for dynamic exercises

rbeezer:
Three changes in `pretext-assembly.xsl` concerning dynamically-generated (fill-in-the-blank) exercises.

**Bug fixes** (affect documents that use dynamic exercises)
- **Over-broad `standalone-filename` match** — the template choosing the standalone-page filename for a fill-in exercise tested `//task` (an absolute path matching any `task` in the whole document) where `.//task` was intended, so it matched any such exercise whenever a fill-in `task` existed anywhere in the source.
- **Substitution lookup missed `@xml:id`-only exercises** — the static substitutions for a dynamic exercise are keyed by the same identifier used downstream (`@label`, falling back to `@xml:id`), but this early lookup consulted only an authored `@label`. Since the promotion of `@xml:id` to `@label` happens in a later pass, an exercise identified solely by `@xml:id` never matched. The lookup now consults `@label` or `@xml:id`.

**Fewer passes**
- **Skip the dynamic-substitution pass** when the document contains no dynamic markup. The pass is otherwise the identity, so a document with no dynamic exercises no longer copies the entire source tree to no effect. The presence test is paired with a source note so that a new template added to the pass is reflected in the gate.

Verified output-neutral on the sample article, whose fill-in exercises carry labels: the post-pre-processor `assembly-dynamic` source is byte-identical to master. The fixes change behavior only for the edge cases they address.

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -1718,8 +1718,12 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <xsl:template match="fillin[@ansobj]" mode="dynamic-substitution">
     <xsl:choose>
         <xsl:when test="($exercise-style = 'static') and not($b-extracting)">
+            <!-- The substitutions file is keyed by the late "visible-id",  -->
+            <!-- which prefers @label, then @xml:id.  This early pass runs  -->
+            <!-- before the promotion of @xml:id to @label, so consult both -->
+            <xsl:variable name="owner" select="ancestor::statement/.."/>
             <xsl:variable name="parent-id">
-                <xsl:apply-templates select="ancestor::statement/../@label" />
+                <xsl:apply-templates select="$owner/@label | $owner[not(@label)]/@xml:id" />
             </xsl:variable>
             <xsl:variable name="eval-subs" select="document($dynamic-substitutions-file,$original)"/>
             <xsl:variable name="object" select="@ansobj"/>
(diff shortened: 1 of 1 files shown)


rbeezer:
@dbrianwalton  - some items found by Claude Code.

1.  First commit is almost certainly a bug.  `//` is the root of the whole overall tree.
2.  I'm really uncertain about the second commit.
3.  The skip should be good, I just merged a very similar approach for "private solutions", which may only rarely be used.  Felt like you should know about this and that it belonged here, rather than on #2909.

I would love your comments before merging this.  Thanks.

rbeezer:
Forgot to say - I don't see these fixes influencing anything, but I am testing with the sample article, but I may have also tested the sample book.  Perhaps they are harmless bugs.  But that is part of why I split these off.

dbrianwalton:
1. Agree that was a bug.
2. This appears to be correct — adds xml:id if label was not present. However, I think the @label was expected to be required for Runestone. 
https://pretextbook.org/doc/guide/html/topic-interactive-exercises.html#topic-interactive-exercises-4-5
I don't recall if a warning was raised if label wasn't there.
3. I don't see a problem with skipping the pass if not needed. This is a good improvement.

rbeezer:
Thanks very much for the confirmations.  Claude has a plan for a better improvement to the situation in commit 2, but I will go ahead and merge that now - it seems a better pathway for testing the more involved approach.

rbeezer:
Force-pushed, just to get back on top of `master`

rbeezer:
I think a missing `@label ` on an interactive `exercise` only triggers a warning when building an HTML `id` for hosting on Runestone.  That is where it really matters, but perhaps a more universal warning would be good?

PR #2913, opened 2026-06-11
FITB: key the dynamic-substitution round trip on "@assembly-id"

rbeezer:
Computed fill-in-the-blank exercises (those with `fillin[@ansobj]` / `eval[@obj]` driven by a `setup`) use a two-pass round trip: an extraction pass writes each exercise's identity and the values to compute, and a later assembly pass reads them back to splice the answers into static (PDF/LaTeX) output. The two ends keyed on **different** identifiers — extraction wrote the late `visible-id` (`@unique-id`), while the substitution lookup keyed on the authored `@label`. They aligned only for labeled exercises; an exercise with no `@label`, or a fill-in inside a `task` (the owner is the unlabeled task), silently lost its computed answer.

This keys both ends on `@assembly-id`, the early identifier minted precisely so assembly passes can coordinate before `@unique-id` exists.

- **`extract-dynamic.xsl`** writes `@assembly-id` as the `exercise_id`.
- The substitution lookup reads `$owner`'s `@assembly-id` (via the `assembly-id` getter mode) instead of `@label`.
- The dynamic-substitution pass now runs **after** the `@assembly-id` stamp (it must still precede the representations pass that consumes the spliced answer). A side benefit: stamping before the static-mode removal of `setup`/`eval` makes `@assembly-id` independent of build mode, which is what lets the extraction (dynamic) and substitution (static) passes agree.
- A bare fill-in `task` now reports its `@assembly-id` from the getter (it is the lookup owner).

Compatibility: for any labeled exercise `@assembly-id == @label == visible-id`, so existing generated substitution files keep working unchanged; previously-broken unlabeled exercises and task fill-ins now resolve. Rendered HTML and LaTeX are unchanged.

**Testing.** Because every computed fill-in in the sample documents is labeled, the bug is invisible to them — so verification used two deliberately bug-inducing duplicates of an existing computed exercise: one with its `@label`/`@xml:id` removed, and one with the fill-in moved inside a `task`. Both render a blank answer on `master` and the correct answer with this change. To guard the identification contract (the substitution pass now runs between the `@assembly-id` and `@unique-id` stamps), the `assembly.debug` coherence check was run on the sample article before and after — the set of `@assembly-id`/`@unique-id` mismatches is byte-identical, so the change adds none. Regenerating the dynamic substitutions for the sample article yields a file byte-identical to the committed one, and the sample book's rendered HTML is unchanged.

### DIFF
--- a/xsl/extract-dynamic.xsl
+++ b/xsl/extract-dynamic.xsl
@@ -96,7 +96,10 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <xsl:text>,&#xa;</xsl:text>
     <xsl:text>{</xsl:text>
     <xsl:text>  "exercise_id": "</xsl:text>
-    <xsl:apply-templates select="." mode="visible-id" />
+    <!-- Key the round trip on @assembly-id, the early identifier the   -->
+    <!-- substitution pass reads back.  @label only exists for labeled  -->
+    <!-- exercises, so it would miss unlabeled exercises and tasks.      -->
+    <xsl:apply-templates select="." mode="assembly-id" />
     <xsl:text>",&#xa;</xsl:text>
     <xsl:text>  "exercise_setup": </xsl:text>
     <xsl:call-template name="dynamic-setup" />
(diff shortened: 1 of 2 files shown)


rbeezer:
@dbrianwalton - I think this really solidifies tracking FITB pieces from the generation pass to the incororation pass.  Testing clearly showed gthe potential for bugs, and these changes resolved them.  Apologies for how this has all shifteed over the years, but I think `@assembly-id` has worked out as a way to isolate a tricky problem.

dbrianwalton:
Wow, and thanks! This is a great resolution.

PR #2916, opened 2026-06-12
Runestone: prepare for RS user menu responsibility handoff

ascholerChemeketa:
This preps the way for Runestone to take control over its entries in the "bust" menu.

This marks the old menu items so RS can discard them and provides templates for RS to use to insert items. Should be 100% safe to deploy before RS code. Once RS code is in place, we can fully remove the old HTML.

PreTeXt will control the HTML in the templates and the location where RS places items. RS will instantiate template copies for whatever menu items it wants and customize the <a> href and text.

@bnmnetp - I don't know what menu options you want to give and in what order. So not doing a RS PR. but here is some sample JS for removing the old items and inserting ones you want:

```js

window.addEventListener("DOMContentLoaded", function(event) {

  const oldDropDown = this.document.querySelector(".dropdown-content[data-deprecated]");
  if( oldDropDown )
    oldDropDown.remove();

  const itemTemplate = this.document.getElementById("ptx-user-dropdown-content_item-template");
  const sepTemplate = this.document.getElementById("ptx-user-dropdown-content_separator-template");
  const menuContentArea = this.document.getElementById("ptx-user-dropdown_rs-content");

  if( !itemTemplate || !sepTemplate || !menuContentArea ) {
    console.error("Missing required template or content area for user dropdown");
    return;
  }

  function makeLink(url, title, ariaLabel = null) {
    // All we should assume about the item template is that it has an anchor element
    const link =  itemTemplate.content.cloneNode(true);
    const linkAnchor = link.querySelector("a");
    linkAnchor.href = url;
    linkAnchor.innerText = title;
    if(ariaLabel) {
      linkAnchor.ariaLabel = ariaLabel;
    }
    // return entire item template, not just link
    return link;
  }

  menuContentArea.appendChild(makeLink("http://google.com", "Go to google"));
  menuContentArea.appendChild(makeLink("http://github.com", "Go to github"));
  menuContentArea.appendChild(sepTemplate.content.cloneNode(true));
  menuContentArea.appendChild(makeLink("http://apple.com", "Go to apple"));
});
```

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -10048,6 +10048,7 @@ http://andrewmccarthy.ie/2014/11/06/swung-dash-in-latex/
         <xsl:text>print-first-page-footer-checkbox print-running-footer-checkbox </xsl:text>
         <xsl:text>calculator-toggle calculator-container geogebra-calculator </xsl:text>
         <xsl:text>embed-button embed-popup embed-code-textbox copy-embed-button </xsl:text>
+        <xsl:text>ptx-user-dropdown ptx-user-dropdown-content_item-template ptx-user-dropdown-content_separator-template </xsl:text>
         <!-- Runestone Services ids, banned always: these components can appear -->
         <!-- even on a non-hosted build.  (When actually hosted on Runestone the -->
         <!-- HTML ids are mangled, so the collision risk there is small.)        -->
--- a/xsl/pretext-runestone.xsl
+++ b/xsl/pretext-runestone.xsl
@@ -255,7 +255,17 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                 <xsl:with-param name="name" select="'person'"/>
             </xsl:call-template>
             <span class="name"><xsl:text>Profile</xsl:text></span>
-            <div class="dropdown-content">
+            <div id="ptx-user-dropdown_rs-content" class="dropdown-content">
+            </div>
+            <!-- clone template, modify href and innertext, and append to ptx-rs-dropdown-content -->
+            <template id="ptx-user-dropdown-content_item-template">
+              <a href="url here" >title here</a>
+            </template>
+            <!-- clone and append to dropdown contenet area add a separator within the menu -->
+            <template id="ptx-user-dropdown-content_separator-template">
+              <hr/>
+            </template>
+            <div class="dropdown-content" data-deprecated="true">
                 <xsl:text>&#xa;</xsl:text>
                 <xsl:text>{% if settings.academy_mode: %}&#xa;</xsl:text>
                 <a href="/ns/course/index">Course Home</a>


rbeezer:
I was unaware of the `template` element!  More tomorrow.

bnmnetp:
Thanks @ascholerChemeketa I don't see anything with those ids in the books or in the latest PreTeXt xsl.  Maybe I missed a PR for PreTeXt ?

ascholerChemeketa:
@bnmnetp You are commenting on an unmerged PR. :)

bnmnetp:
🤦‍♂️

ascholerChemeketa:
Forgot to add new id's to the reserved list. Can do it as a follow-up if this gets merged before I get a chance (on the road at events most of today/tomorrow).

ascholerChemeketa:
Had a window... added the new ids to the reserved list.

@rbeezer Starting to think that `ptx-*` should be off limit for authors. Would make it easier to list/check and for authors to have a convention instead of a list. Not saying we immediately change all existing ones immediately. Maybe we add a non-fatal warning for any authored `ptx-*` labels. After a while we can make it a stronger error and stop explicitly checking the existing `ptx-*` ids. New ones can be authored using that convention.

rbeezer:
> Starting to think that ptx-* should be off limit for authors.

In favor.  How about we start one step further back.  Ask on `-dev` if anybody does this, or if they have helped folks who do?  Unlikely, but I'd be mildly annoyed if some systematic naming scheme of mine got wrecked.  ;-)

rbeezer:
Posting from Claude in a second.  I presume the underscores are a conscious choice for CSS styling.  I will make trivial edits on a follow-on commit.

rbeezer:
Nice, clean handoff prep — and `<template>` is a good fit here (inert, valid HTML5).

**Reserved list doesn't match the emitted ids.** The follow-up commit reserves `ptx-user-dropdown`, but nothing emits that id (the button is just `class="runestone-profile dropdown"`). Meanwhile the id that *is* emitted, `ptx-user-dropdown_rs-content`, isn't reserved. The two template ids line up correctly — just swap `ptx-user-dropdown` → `ptx-user-dropdown_rs-content`.

**Transition safety.** Before the RS JS ships, `ptx-user-dropdown_rs-content` carries `class="dropdown-content"` and sits as a sibling of the still-live deprecated `.dropdown-content`. If the hover CSS targets `.dropdown-content`, that empty div may show as a second (empty) panel next to the real menu during the interim. Worth a hover-test on a hosted build, or hiding it until populated (e.g. `:empty`).

**Smaller things:**
- The item-template comment points at `ptx-rs-dropdown-content`, but the id is `ptx-user-dropdown_rs-content`; that comment also has "innertext", and the separator comment has "contenet" plus a garbled "dropdown contenet area add a separator".
- Placeholder `<a href="url here" >title here</a>` has a stray space before `>`.
- Commit message: "ids ro reserved" → "to".
- The new ids mix separators (`...content_item-template`, `..._rs-content`) — hyphens elsewhere, underscores here.

rbeezer:
@bnmnetp - in now!  Patience!  ;-)

@ascholerChemeketa - looks good.  Small follow-on commit with minor edits.  Let's pursue the `ptx-*` consolidation, that'll be good to have.

bnmnetp:
Patience is practically my middle name

PR #2919, opened 2026-06-14
Javascript: set up builder for upcoming modularization effort

oscarlevin:
This will subsume the draft (and soon to be closed) #2815, but doesn't go as far.  

This sets up an npm build script that will bundle the "core" javascript in `js/dist` to be included in each page.  No additional modularization is done, although once this is in, we can start splitting up the `add_on.js` monolith and possibly transition some of it to typescript.

The commit includes adding the built files in the `dist` subfolder, similar to how the css is handled.

One side effect: the webwork javascript had a hard coded reference to  `js/lib/knowl.js`, which hasn't existed for at least a while.  I am guessing that this was left over from before solution knowls became detail elements?  Or maybe it is supposed to be useful for born hidden knowls?  In any event, it will not have been working, and now it will, since this directs that link to an existing version of the knowl.js file in the dist directory.

One more thing: I tried to remove the jQuery.js file, but this caused a bunch of problems, including with activating webwork.  I think this is something we will want to look at soon.  For now, we just leave it where it is and include it via the xsl (it did result in adding a new variable for xsl, since it will not live in the same directory as the rest of the core javascript).


oscarlevin:
Should have mentioned, @ascholerChemeketa and @Alex-Jordan as being interested in this.

rbeezer:
> This sets up an npm build script 

Am I right that this is a developer-per-js-change task and not a author-per-test-build task?  Similar to the `cssbuiilder` script?  Maybe a review from my assistant in a moment

oscarlevin:
Right.  We would ship the minimized js just like we do for the css.  Maybe
eventually some additional customizations can be added that would *allow*
the author to run the js builder while building a project, but that is not
part of this PR.

rbeezer:
Thanks very much, @oscarlevin.  Review coming.  I was a bit startled to learn that `file://` previewing might not work, but perhaps that was overblown - everything seemed to checkout visually just fine.

Review seemed very careful.  Let me know if you want thhis merged now and catch some things on the next pass, or if you want to push (force-push?) some changes first.  I'll let you decide oon CLI release timing (of course!).  I'm a bit constrained on reacting to emergencies, but it should not be too bad.

rbeezer (commented):
Thanks for getting this off the ground, @oscarlevin — setting up the bundler now (with the `dist/` artifacts committed, mirroring the CSS builder) is a clean foundation for the modularization work to come. I went through the build mechanics and the XSL rewiring fairly carefully; overall it looks solid. Notes below.

## What I checked and looks good

- **The committed artifacts reproduce.** With the pinned esbuild (`^0.25.0`), running `npm run build` in `script/jsbuilder/` regenerates every file in `js/dist/` byte-for-byte against what's committed. The unminified full-build / minified single-target (`-t`) split matches the CSS builder, so the readable diffs are intentional.
- **The IIFE conversion looks safe.** Bundling `pretext.js` + `pretext_add_on.js` + `knowl.js` into one IIFE removes their top-level declarations from the global scope, so anything referenced across that boundary has to be re-exposed. Checking the rest of the XSL and JS, only two symbols actually cross it — `PTXDialog` (used by `pretext_search.js`) and `isDarkMode` (used by an inline script) — and both are assigned to `window` in `pretext_add_on.js`. I didn't find any other inline `<script>` or separately-loaded file reaching into those three.
- **Execution order is preserved.** In the bundle the three run `pretext.js → pretext_add_on.js → knowl.js`, which matches their original document order (the `pretext-js` block precedes the per-page `knowl` template), so `load`-listener registration order is unchanged.
- **The `$html.js.dir` redefinition is consistent.** It now resolves to `…/js/dist`, shifting every existing reference into `dist/`; each one (WeBWorK 2.16–2.20, LTI, STACK, search, SCORM, diagcess, `mathjax_startup`, and the new core bundle) maps to a real build target, and jQuery is correctly moved to the new `$html.js.root` since it's a third-party file rather than a built artifact.
- The `js/lib/knowl.js` → `js/dist/knowl.js` fix in the WeBWorK shims points at a file that now actually exists — good catch.

## Suggestions / things to confirm

1. **`script/jsbuilder/README.md` — `node_modules` claim.** The README says the `node_modules/` directory "is already present in this folder … checked in for offline use," but `**/node_modules/` is gitignored (same as `cssbuilder`), so a fresh clone won't have it. The cssbuilder README just tells you to run `npm install`; matching that here would avoid confusion.
2. **Packaging dependency worth confirming before this ships.** Since output now references `_static/pretext/js/dist/…`, whatever populates `_static` has to preserve the `dist/` subdirectory. The in-repo HTML build copies the whole `js/` tree recursively, so it's fine there — just worth verifying the CLI does the same so the new paths don't 404.
3. **A forward-looking note for the modularization.** The IIFE boundary means this becomes a standing rule: any symbol consumed by an inline `<script>` (or another separately-loaded file) must be on `window`. Today that's `PTXDialog`/`isDarkMode`; as more moves into the bundle it'd be easy to scope something away by accident. The comment in `js/src/pretext-core.js` covers adding new scripts — might be worth a line there about the reverse case too.

## Minor

- The `knowl` template still carries the comment "Variables are defined to defaults in knowl.js and we can override them," but `sagecellEvalName` isn't defined there anymore (it works as a plain global for the Sage Cell library). Since the template is touched here anyway, could tidy.
- The old `pretext_add_on.js?x=1` had a cache-busting query string; `pretext-core.js` has none. The rename busts caches for this transition, but as the bundle changes during modularization it may be worth a deliberate cache-busting story (this matches the other JS files today, so not new).

## Testing

I built the sample article to HTML and viewed it both served over a local web server and opened directly from disk as `file://`. Rendering and interactive behavior — navigation/TOC, knowls, dark-mode toggle, and math under the server — looked correct on visual inspection in Firefox. Note that I did **not** open the browser console, so this is a behavioral check rather than an error-log audit; a console pass (watching specifically for any `is not defined` originating in `pretext-core.js`, which would be the signature of a missing `window` re-export) would be a good final confirmation. As expected, the module-loaded MathJax and fetch-based features don't operate under `file://`, but that's a pre-existing browser limitation unchanged by this PR.

Nice work — modulo the README fix and confirming the packaging copy, this reads as ready.

oscarlevin:
Force pushed with fixes for two of the suggestions in the previous comment (fixing node_modules mention in README and cache-busting).  The rest of the comments are either out of scope or are indeed confirmed.

Merge whenever you are ready.

rbeezer:
Merged as-is after a rebase.  Thanks for the changes, and for getting this started!

Could you post a heads-up on `-dev` and maybe just a quick hint as to where this is going?  Not sure just who could be affected...

PR #2920, opened 2026-06-15
update to newest diagcess library

davidaustinm:
This brings in the latest `diagcess` library and some of its improvements for annotated PreFigure diagrams.


rbeezer (commented):
Reviewed and tested locally. Confirmed the change is self-contained: the only diff is `js/diagcess/diagcess.js` bumping 1.3.3 → 1.4.1. The library is loaded by path (no version pin or SRI hash elsewhere), so nothing else needs to move in lockstep.

Did a before/after HTML build of the sample article on both versions — identical output across all pages and every PreFigure asset, with the sole substantive change being the swapped library file. The diagcess DOM hooks the XSL emits are unchanged. Verified the annotated PreFigure diagrams render and the keyboard navigation works on 1.4.1 over a local server.

rbeezer:
Merged - edited the commit message to have the new version number and a topic.

Thanks for keeping this up-to-date!

rbeezer:
I got something out-of-order here, and now javascript goes through a build process (`script/jsbuilder`).  I guess you need to know this so you can test.

I left it out before, my fault.  Corrected at commit referenced just above.  Updating the bundle will be a **committer** job, but a reminder won't hurt.

davidaustinm:
Sorry I'm a little late to this, but I just want to check that I understand this issue.  For future updates of `diagcess`, I should run the bundler script to get the new version of the library in all the places it needs to go.  Is that correct?

PR #2929, opened 2026-06-16
CSS: Fix accessibility contrast issue with ice-fire quad color scheme

ascholerChemeketa:
One of the colors used in the "ice-fire" color scheme doesn't have quite enough contrast vs white. This gets it above a 4.5 ratio for WCAG AA.

### DIFF
--- a/css/colors/_palette-quad-chunks.scss
+++ b/css/colors/_palette-quad-chunks.scss
@@ -25,7 +25,7 @@ $schemes: (
   ),
   "ice-fire": (
     "color-main": hsl(210, 45%, 27%),
-    "color-do": hsl(208, 7%, 49%),
+    "color-do": hsl(208, 7%, 45%),
     "color-fact": #0f80b0,
     "color-meta": #a13838,
   ),


rbeezer:
Great!   I will merge just now.  One further suggestion coming from Claude, which you can handle as you see fit.

rbeezer:
Confirmed: this moves the white-on-`color-do` contrast from 4.16 to 4.80, clearing the 4.5 AA bar with a bit of margin. Nice catch and a cleanly scoped fix.

One question while you're in here: by the same vs-white measurement, the sibling `color-fact` (`#0f80b0`) comes out at ~4.44 — also just under 4.5. (`color-main` and `color-meta` are both comfortably above.) Since these colors are used as block-heading backgrounds with white heading text, and headings often qualify as WCAG "large text" where the 3:1 threshold applies, `color-fact` may well be fine in practice. But if the intent is to hold the scheme to 4.5 across the board, `color-fact` looks like it needs the same small nudge. Was leaving it as-is deliberate?

rbeezer:
Merged as-is, thanks.  Had to catch-up on CSS with a pre-PR update, I was behind on regular updates!  So the post-PR update is accurate.  Just for the record.

PR #2931, opened 2026-06-16
Spacing fixes

ascholerChemeketa:
Three fixes related to line heights from this discussion:
https://groups.google.com/g/pretext-dev/c/DiP5zmNxnO8/m/DBOaEIchBAAJ?utm_medium=email&utm_source=footer


ascholerChemeketa:
@Alex-Jordan heads up if you want to test the math CSS. But it is the same code you already gave a quick try to.

rbeezer:
Thanks for moving on this one!  A bit of discussion from Claude coming in a second.  Let me know your reaction and we can merge when you say so.

rbeezer:
Thanks for these — the line-height mechanism in all three commits looks correct, and switching prism's inline background from a hardcoded `#fdfdfd` to `var(--code-background)` is a nice dark-mode win. A few things worth a look before merge:

**1. Inline code may lose its scroll/clamp for long content.** In `.code-inline` the `display: inline-block` was dropped, but `overflow-x: auto` and `max-width: 100%` were kept. On a plain inline element those two properties have no effect (overflow and width don't apply to non-replaced inline boxes). Combined with the new `white-space: nowrap`, a long inline `<c>` token — a long path or URL, say — can no longer scroll inside a clamped box; it will overflow and can push horizontal scroll onto the whole page. Previously `inline-block` + `overflow-x: auto` let it scroll internally. Is dropping that intentional? If so, the two now-inert declarations could come out; if not, it's a regression for long inline code.

**2. `white-space: pre` → `nowrap`.** `nowrap` collapses runs of internal spaces that `pre` preserved (e.g. `<c>a    b</c>`). Probably rare and arguably fine for inline code, but it's a rendering change worth a conscious sign-off.

Minor:

- `z-index: -1` on the `::before` background works fine over ordinary prose, but it's worth eyeballing inline code inside a container that may establish its own stacking context (a knowl, callout, or sidebar with its own background) — the chip could end up behind that background.
- Prism's `.code-inline` sets `line-height: 1.35`, while the regular `.code-inline` now sets none and inherits. Small inconsistency — any reason prism needs the explicit value when the goal is to leave line height undisturbed?
- The new `mjx-container` rule in `_math.scss` is indented 4 spaces; the surrounding rules use 2.

ascholerChemeketa:
Reworked the inline code bits.

There is a new commit that adds a CSS variable with the line-height. It is future proofing/groundwork against possible future user selectable line-height.

rbeezer:
I'm going to merge this, and will post a couple follow-up comments from Claude Code.

rbeezer:
Two spots worth tightening before merge:

**1. Inline-code border, plain vs. Prism.** Dropping the border from `.code-inline` in `_codelike.scss` and re-adding the same border only under Prism (`&.code-inline` in `_prism.scss`) leaves two inline-code styles in one document: a borderless pill for plain `<c>`, a bordered one wherever syntax highlighting is active. If that split is intended, a note would help; otherwise it reads more consistently as border everywhere or nowhere. Related: the re-added Prism border is mixed from `--code-inline-background`, while the Prism fill now comes from `--code-background`, so border and background key off different vars.

**2. The 1.35 line-height isn't single-sourced.** `$line-height: 1.35 !default` is currently unused, and `--content-line-height: 1.35` duplicates the paragraph line-height in `_misc-content.scss` (`.para { line-height: 1.35 }`) rather than deriving from it — so 1.35 is now hard-coded in three independent places. Defining `--content-line-height: #{$line-height}` and pointing `.para` at the variable would let the inline-code `calc()` track the body line-height automatically and retire the unused SCSS var.

Minor: `calc(var(--content-line-height) - 0.15)` has no fallback, so a Prism inline span rendered outside `.ptx-content` would invalidate the line-height — `var(--content-line-height, 1.35)` covers it; and the reformatted `.latex-logo` block returns at 4-space indent in an otherwise 2-space file.

ascholerChemeketa:
OK, readability dialog will change up line-height stuff anyway.

Border difference is intentional. Formatted code needs more of a visual offset than unformatted code.

rbeezer:
OK, thanks for the reply.  At one of the few work areas at DTW airport and the lighting is really BRIGHT!  Tough to review this one.  ;-) See what more progress I can make...

Metrged as-is, just a rebase on `master` and a theme update.  Thanks again for working on these - I know it is fiddly.

PR #2932, opened 2026-06-16
HTML: route dialog close buttons through insert-symbol template

ascholerChemeketa:
Just noticed the close buttons for embed and search were never set up to use the `insert-symbol` template.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -12515,7 +12515,9 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
         <div class="ptx-embed-popup-controls">
             <p>Copy the code below to embed this page in your own website or LMS page.</p>
             <button class="ptx-embed-close-button button" id="ptx-embed-close-button" title="Close embed popup">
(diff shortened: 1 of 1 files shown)


rbeezer:
The change is a good consistency cleanup — these two static close buttons were the last ones emitting their glyph as a hardcoded `<span class="material-symbols-outlined">close</span>`, so routing them through `insert-symbol` brings them in line with how every other nav/dialog button produces its icon.

Worth noting it's a small accessibility gain too, not just tidiness: `insert-symbol` emits the glyph with `aria-hidden="true"` (and the shared `icon` class), so the decorative icon is hidden from the accessibility tree — appropriate here since both buttons already carry their own `aria-label`/`title`. The `'close'` name resolves cleanly (`entity="e5cd"` in `html-symbols.xsl`), so it renders the same glyph.

Reads correct to me.

rbeezer:
Merged with rebase and addition of PR number to commit message.

Out-of-order and that was my fault.  Had to rebase.  I think I got that sorted and "search close" seems to be visually OK and functional.

PR #2934, opened 2026-06-17
Validate PreFigure diagrams against PreFigure's own schema

rbeezer:
## Motivation

PreFigure content inside an `<image>` (a `pf:prefigure` element in the `https://prefigure.org` namespace) was accepted by a permissive wildcard (`AnyPFContent = text | element * { … }`) — any element, any attribute, any nesting. Authoring mistakes in a diagram passed validation silently. PreFigure now publishes its own RELAX-NG schema, so we can validate diagram structure for real.

## Approach

Integration is built into the literate program (`schema/pretext.xml`):

- A new **adapter** (`pf-adapter.rnc`, emitted by the literate program) declares `default namespace = "https://prefigure.org"` and `include`s the upstream `pf_schema.rnc`. The default namespace remaps PreFigure's no-namespace elements into the PreFigure namespace without restructuring the upstream file.
- `pf:prefigure`'s content model changes from the wildcard to `external "pf-adapter.rnc"`. Using `externalRef` keeps the two grammars isolated, so the many pattern names they share (`Label`, `Image`, `Caption`, …) don't collide — a plain `include` is impossible here.
- Re-syncing a future PreFigure release is a one-file copy.

## Upstream schema bugs (fixed in our copy)

PreFigure's shipped schema (commit `c529af4`) does not load under `jing` — it appears never to have been checked with a strict RELAX-NG validator. The second commit fixes the minimum needed to load, in both `.rnc` and `.rng`:

- Undefined reference `Stroke-Attributes` (should be `StrokeAttributes`)
- Duplicate attributes from elements that declare an attribute explicitly *and* pull in a pattern that already provides it: `grid`, `tick-mark`, `histogram`, `scatter`, `riemann-sum`, `vector`
- Overlapping pattern references in `riemann-sum` (`CommonAttributes` already contains `AnnotatedElementAttributes`)
- `network/@alignment` written as a sequence `{'vertical', 'horizontal'}` instead of a choice `{'vertical' | 'horizontal'}`
- Missing `tactile-dash` attribute (used in PreFigure's own `de-system.xml` example and in our sample article) — added to `StrokeAttributes`

These should be reported upstream to `davidaustinm/prefigure`.

## Testing

Validated the sample article with `jing` against `pretext-dev.rng`. The 201 pre-existing, unrelated errors are unchanged; **0** PreFigure errors. Positive control confirms real validation through the `pretext.rng → pf-adapter.rng → pf_schema.rng` chain: a valid diagram passes, while a bogus element and a bogus attribute are both rejected.

## Known limitation: XSD

`trang -O xsd` cannot follow `externalRef` ("sorry, externalRef is not yet supported"), so the XSD schema (`pretext.xsd`/`pf.xsd`/`xml.xsd`) is left on the old wildcard model and does not yet carry structural PreFigure validation. `jing`/RNG is PreTeXt's validation path, so this is acceptable for now; a comment in `build.sh` records it.

## Commits

1. `Schema: add upstream PreFigure RELAX-NG schema files` — upstream, verbatim
2. `Schema: fix bugs in the PreFigure schema so it loads` — the fixes above; the diff isolates exactly what was wrong upstream
3. `Schema: validate "pf:prefigure" content with a PreFigure adapter` — the adapter, literate-program and regenerated-schema changes, `build.sh`

### DIFF
--- a/schema/build.sh
+++ b/schema/build.sh
@@ -68,14 +68,22 @@ declare XSL=${PTX}/xsl
 # ******************
 
 # PreTeXt extraction of RELAX-NG compact schema
+# Also emits the PreFigure adapter, "pf-adapter.rnc"
 xsltproc ${XSL}/pretext-litprog.xsl pretext.xml
 
 # System trang conversion to RELAX-NG XML schema
 trang -I rnc -O rng pretext.rnc pretext.rng
 trang -I rnc -O rng pretext-dev.rnc pretext-dev.rng
 
+# PreFigure adapter, which "externalRef" reaches from "pretext.rnc";
+# it in turn "include"s the upstream "pf_schema.rnc"/"pf_schema.rng"
+trang -I rnc -O rng pf-adapter.rnc pf-adapter.rng
+
 # System trang conversion to W3C XSD schema
 # "abstract groups" make schema browser too obtuse
+# NOTE: trang cannot follow the PreFigure "externalRef" when emitting XSD
+# ("sorry, externalRef is not yet supported"), so the XSD path does not yet
+# carry structural PreFigure validation.
 trang -o disable-abstract-elements -I rnc -O xsd pretext.rnc pretext.xsd
 
 # And the same steps for the publication-schema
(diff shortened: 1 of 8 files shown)


rbeezer:
The upstream PreFigure schema bugs fixed in this PR are now reported at davidaustinm/prefigure#52.

Issue #2460, opened 2025-04-10
FITB: implicit and explicitly correct answers behave differently


ascholerChemeketa:
If a test is correct because it is the first in a block it behaves different than if it is correct because is has `correct=Yes`

```            <exercise label="test1">
              <statement>
                <p>The second, implicitly correct fillin breaks the widget.</p>
                <p><fillin mode="string" answer="double\s+x;"/></p>
                <p><fillin mode="string" answer="double\s+x;"/></p>
              </statement>
              <evaluation>
                <evaluate>
                  <test correct="yes">
                    <strcmp use-answer="yes"/>
                  </test>
                </evaluate>
                <evaluate>
                  <test>
                    <strcmp use-answer="yes"/>
                  </test>
                </evaluate>
              </evaluation>
            </exercise>

            <exercise label="test2">
              <statement>
                <p>Neither of these ends up correctly matching "foo();".</p>
                <p><fillin mode="string" answer="foo();"/></p>
                <p><fillin mode="string" answer="foo();"/></p>
              </statement>
              <evaluation>
                <evaluate>
                  <test correct="yes">
                    <strcmp use-answer="yes"/>
                  </test>
                </evaluate>
                <evaluate>
                  <test>
                    <strcmp use-answer="yes"/>
                  </test>
                </evaluate>
              </evaluation>
            </exercise>

            <exercise label="test3">
              <statement>
                <p>The correct regex recipe depends on if correct="yes" is specified.</p>
                <p><fillin mode="string" answer="foo\(\);"/></p>
                <p><fillin mode="string" answer="foo\\(\\);"/></p>
              </statement>
              <evaluation>
                <evaluate>
                  <test correct="yes">
                    <strcmp use-answer="yes"/>
                  </test>
                </evaluate>
                <evaluate>
                  <test>
                    <strcmp use-answer="yes"/>
                  </test>
                </evaluate>
              </evaluation>
            </exercise>
```

ascholerChemeketa:
Tag @dbrianwalton

Which behavior is the expected/correct one?

dbrianwalton:
Thank you for finding these. I think that the issue here is about the role of the `@answer` and the role of the test. `@answer` is intended to be a literal string that represents the answer. It will be used directly for the correct answer in a static build.

I did not anticipate that the text of `@answer` would itself contain regex expressions.

If a regex is required to test the entry, then the explicit regex needs to be provided.

[...]

PR #2936, opened 2026-06-17
Fitb literal strings

dbrianwalton:
Targeted to address #2460 

Description of changes:

- Added a template in pretext-text-utilities.xsl that takes a string and escapes regular expression token characters so that they would match literally in a regular expression comparison. \ . ^ $ * + ? { } [ ] ( ) |
- When a FITB answer checker for a string involves `@use-answer="yes"`, it compares with the regex-escaped answer string instead of the answer string itself
- The original code used `xsl:choose` for a case that no answer checker was provided for the correct response so that it used the provided answer. This introduced a second area that would have needed the correction, so instead we used the `mode="exercise"` during assembly-stage to insert a missing `<test correct="yes">` element so that the same branch of logic is followed whether the author provided the element or not.

Although I originally expected I needed exslt for regular expressions, the standard str:replace method using search/replace pairs did the job fine.


dbrianwalton:
[sec-fitb-string-tests.xml](https://github.com/user-attachments/files/29054790/sec-fitb-string-tests.xml)

This file contains example exercises that would not have worked previously that now work. (Claude helped construct the examples.)

rbeezer:
Thanks for this — the approach is sound, and I've confirmed it works: I rebased onto `master` and built your `sec-fitb-string-tests.xml`, and every synthesized/explicit correct test escapes properly across the metacharacters (`3.14`→`3\.14`, `C:\Users`→`C:\\Users`, `f(x) = x^2 + 1`→`f\(x\) = x\^2 \+ 1`), with the backslash-first ordering correct. Routing the "no explicit correct test" case through an assembly-synthesized `<test correct="yes">` is a nice consolidation. I've prepared follow-up commits for the items below, which land alongside your work:

**1. Positional `fillin` match.** In the synthesis template, `$exercise-parent/statement//fillin[$eval-position]` is the classic `//x[n]` pitfall — the predicate binds per parent, so it selects "the Nth `fillin` within each parent," not the Nth in document order. It's correct when all blanks share one parent (as in the examples) but silently fails for a multi-blank question whose `fillin`s sit in separate paragraphs. Wrapped in parentheses in `fbafffe4`: `($exercise-parent/statement//fillin)[$eval-position]`.

**2. Schema addition for `@literal`.** The new `strcmp/@literal` is used in the XSL and documented in the Guide, but `<strcmp>` wasn't given the attribute in the schema, so `<strcmp literal="yes">` wouldn't validate. Added to the (dev) schema source, with the derived files regenerated, in `1d2ccd8e`.

**3. The test file → sample book.** It's an excellent regression suite, and I wanted it living in the sample book so the standard build guards this permanently — added as a cleaned new section in `b5f18118`. It couldn't go in as-is, though: eight exercises pair the answer with an intentionally-unescaped `<strcmp>` whose feedback describes a bug ("the period is acting as a wildcard"). That's a great harness assertion, but the sample book shouldn't model broken behavior — extreme *inputs* are welcome, deliberately-wrong regexes and bug-narrating feedback are not. The cleaned form keeps every special-character answer (the literal match is the feature on display), rewrites the statements as ordinary prompts, and demonstrates the explicit `use-answer` and `literal="yes"` paths plus one synthesized-answer example. Two schema points the test file itself trips on, worth knowing: an empty `<evaluate/>` is invalid (a `<test>` is required), and `case` goes on `<strcmp>`, not `<fillin>`.

**4. Minor.** The new assembly comment block had misaligned `-->` closers and a trailing space — tidied in place.

rbeezer:
As promised, additional commits:

*  Small XSL fix
*  An addition to the schema
*  Incorporatd a bunch of new problems into the sample book.

Latter has @dbrianwalton as author.  I backed out the incorrect-feedback stuff.  That was a nice device, but teh samples should be free of errors.  We like edge cases and extreme situations, but not errors.

PR #2941, opened 2026-06-18
HTML: revert to MathJax version 4.1.1

Alex-Jordan:
For consideration. See my forum post about how MathJax 4.1.2 has a rather serious bug.

I'm not sure where else MathJax versions are specified (for braille extraction, etc). So maybe ignore this if there is a larger reversion to make.

Or maybe you don't want to make this reversion. But I thought it wouldn't hurt to get the ball rolling in case you agree it should be done.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -13629,7 +13629,7 @@ TODO:
     <!-- MathJax 4 CDN -->
     <script defer="true">
         <xsl:attribute name="src">
-            <xsl:text>https://cdn.jsdelivr.net/npm/mathjax@4/</xsl:text>
+            <xsl:text>https://cdn.jsdelivr.net/npm/mathjax@4.1.1/</xsl:text>
             <xsl:choose>
                 <xsl:when test="$debug.mathjax.svg = 'yes'">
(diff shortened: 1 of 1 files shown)


rbeezer:
I think a fix is here, just now

https://github.com/Speech-Rule-Engine/speech-rule-engine/pull/872

rbeezer:
I was prepared to do this - BUT, the CLI (@oscarlevin) would need to react, and publishers will need to rebuild,...

If the fix rolls out, all projects will be OK, I think.

Long-term, routinely fixing a version might be something to discuss.

rbeezer:
> See my forum post 

https://github.com/mathjax/MathJax/issues/3578

rbeezer:
https://github.com/mathjax/MathJax/issues/3578

https://github.com/Speech-Rule-Engine/speech-rule-engine/pull/872

rbeezer:
Posted a query about status on [#3578.](https://github.com/mathjax/MathJax/issues/3578).

rbeezer:
In light of https://github.com/mathjax/MathJax/issues/3578#issuecomment-4767623749 I am going to close this, but teh fix is still available for anybody who wants it.  Thanks, @Alex-Jordan!

PR #2945, opened 2026-06-18
PreFigure schema update

davidaustinm:
This fixes a few errors in the PreFigure schema and allows for specific tactile attributes that override visual attributes.

### DIFF
--- a/schema/pf_schema.rnc
+++ b/schema/pf_schema.rnc
@@ -18,7 +18,8 @@ Caption = element caption {text}
 # Annotations are constructed with a single #annotations element,
 # which contains a single #annotation
 Annotations = element annotations {
-    Annotation
+    Annotation*,
+    attribute text {text}
 }
 
 AnnotationAttributes = 
@@ -164,6 +165,9 @@ TransformElements = (
     }* &
     element scale {
         attribute by {text}
+    }* &
+    element change-basis {
+        attribute basis {text}
     }*
 )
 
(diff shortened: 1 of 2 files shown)


rbeezer:
Review coming just now.  Just one of the items is keeoing this rom a merge, but consider others while you are at it.

rbeezer:
Check that - everything can wait for a possible follow-up, or on the next update.  I'll merge just now.

rbeezer:
Took a pass through the schema changes — nice cleanup grouping the `tactile-*` attributes into their own defines. The trang round-trip is clean (regenerating the `.rng` from your `.rnc` reproduces the committed `.rng` byte-for-byte), so the two forms won't drift. A few things, with rough severity:

**1. `TactileFillAttributes` is never referenced — _substantive_.** The define is there:

```
TactileFillAttributes =
    TactileStrokeAttributes,
    attribute tactile-fill {text}?,
    attribute tactile-fill-rule {text}?
```

…but nothing pulls it in. `StrokeAttributes` ends with `TactileStrokeAttributes?` and `LabelAttributes` ends with `TactileLabelAttributes?`, whereas `FillAttributes` doesn't end with `TactileFillAttributes?` — so as written, `tactile-fill` and `tactile-fill-rule` are accepted by no element. RELAX NG allows unreferenced defines, so trang/jing won't warn about it. Is this meant to be wired into `FillAttributes` (parallel to the stroke/label pattern), or staged for a later PR?

**2. `annotations/@text` is now required, and `annotation` became optional/repeatable — _minor, confirm intent_.** `element annotations { Annotation*, attribute text {text} }` makes `@text` mandatory (an `<annotations>` without it now fails validation) and lets the element carry zero or many `<annotation>` children instead of exactly one. Both look deliberate for the new annotation model — just flagging the required-ness in case the generator doesn't always emit `@text`.

**3. Stale comment on `Annotations` — _cosmetic_.** The comment above it still says it "contains a single #annotation," which no longer matches `Annotation*` plus the new `@text`. Worth a one-line update.

**4. Double blank line before `Math =` in the `.rnc` — _cosmetic_.** Every other top-level define is single-spaced; trang normalizes this away so the `.rng` is unaffected, but tidy to match.

Only #1 seems worth resolving before merge — likely a one-line `FillAttributes` change plus a trang regen. The rest are small.

rbeezer:
Merged as-is with an edited commit message.  No rush on any changes - the sample article checks out with PreTeXt schema just fine (201 errors, like before).

PR #2946, opened 2026-06-18
Script: update trim_changed_trailing_whitespace to find files from repo root

ascholerChemeketa:
The trim whitespace fails to find files (but does not indicate there is a problem) if run outside root dir of repo. It tries to use the root relative path from the PWD.

This modifies it to always look relative to the repo root.


rbeezer:
Merged as-is.

Claude thought it would be fun to run the script on this PR.  ;-)

Thanks for the updates.

PR #2947, opened 2026-06-18
Accessibility updates for calculator

ascholerChemeketa:
Updates PTXDialog to do more of the lifting for accessibility attributes on affected elements.

Fixes accessibility issues for the calculator and moves it to the PTXDialog framework. Non-modal dialogs are really tricky vis-a-vis accessibility. And we don't control anything going on inside the calculator. But this definitely is an improvement over the old behavior. 

Also allows the calculator to be resized and repositioned. The code for repositioning is part of PTXDialog and can be reused for other non-modals.

Added a build of the JS incase anyone else wants to test the PR. @rbeezer - you may want to discard that last commit and rebuild yourself.

### DIFF
--- a/css/components/helpers/_dialogs.scss
+++ b/css/components/helpers/_dialogs.scss
@@ -11,10 +11,15 @@ $border-radius: 0 !default;
   box-shadow: 0 0 5px var(--dialog-shadow-color);
   color: var(--dialog-text-color);
 
-  & .ptx-dialog-close-button {
+  .ptx-dialog-close-button {
     @include buttons.ptx-button(
       $border-width: 0
     );
+  }
+
+  .ptx-dialog-topbar {
+    min-height: 15px;
+    user-select: none;
   }
 }
 
--- a/css/components/interactives/_calculators.scss
+++ b/css/components/interactives/_calculators.scss
@@ -1,19 +1,30 @@
 // GeoGebra calculator
+@use 'components/helpers/dialogs' as dialogs;
 
 $navbar-breakpoint: 856px !default;
 
 .ptx-calculator-container {
   position: fixed;
   z-index: 100;
-  bottom: 5px;
-  right: 5px;
-  width: 253px;
-  height: 460px;
+  top: max(10px, calc(100vh - 650px));
+  left: max(10px, calc(100vw - 350px));
+  padding: 0;
+  width: 330px;
+  height: 600px;
+  resize: both;
+  overflow: hidden;
+  flex-direction: column;
+  @include dialogs.ptx-dialog-modal;
 }
 
-@media screen and (max-width: $navbar-breakpoint) {
-  .ptx-calculator-container {
-    //assumes navbar moves to bottom of screen
-    bottom: 50px !important;
+.ptx-calculator-container[open] {
+  display: flex;
+}
+
+.ptx-geogebra-calculator {
+  flex: 1;
+
+  .button {
+    all: unset;  //get rid of ptx ".button" styles that interfere with geogebra's buttons
   }
 }
--- a/css/components/interactives/_calculators.scss
+++ b/css/components/interactives/_calculators.scss
@@ -2,7 +2,7 @@
 
 $navbar-breakpoint: 856px !default;
 
-.calculator-container {
+.ptx-calculator-container {
   position: fixed;
   z-index: 100;
   bottom: 5px;
@@ -12,7 +12,7 @@ $navbar-breakpoint: 856px !default;
 }
 
 @media screen and (max-width: $navbar-breakpoint) {
-  .calculator-container {
+  .ptx-calculator-container {
     //assumes navbar moves to bottom of screen
     bottom: 50px !important;
   }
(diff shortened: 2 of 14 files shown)


rbeezer:
I  used this as an excuase to re-build JS and see that I git Andrew's result.  Yes, I did, but for a recent upgrade to the diagcess library.  Good.

rbeezer:
Comments from Claude.  Don't rush, I think I'm out of time for a while.  As usual, I can't vouch for their utility.

rbeezer (commented):
Nice consolidation onto `PTXDialog`. A few things I'd flag before merge.

**Should address**

1. **Non-modal `aria-expanded` / `.open` only update on the Invoker-Commands path.** The live `aria-expanded` toggle and the control button's `.open` class are written only inside the `if (PTXDialog.hasNativeCommandInvokers())` open/close closures. `openDialogFallback` / `closeDialogFallback` don't touch them, so on browsers without the Invoker Commands API (currently Firefox and Safari) the calculator opens and closes but its toggle button reports `aria-expanded="false"` permanently and never gets the open-state styling — the opposite of this PR's goal. Lifting that state update into shared logic that both paths call (e.g. a small `setExpanded(bool)` helper invoked from the native closures *and* the fallback open/close) would keep them in sync.

2. **Mobile positioning regresses.** `.ptx-calculator-container` now always sets `top`/`left` alongside an explicit `height`, so the mobile media query's `bottom: 50px !important` is over-constrained (top + height win) and no longer pins the dialog to the bottom as it did before. Separately, `top: calc(100vh - 650px)` is negative on viewports under 650px tall, putting the top off-screen.

3. **Remove the now-dead `getScrollbarWidth()`** in `js/pretext_add_on.js` — the old calculator was its only caller, so it's unused after this change (the bundled dist already drops it).

**Could improve**

4. **Calculator init rides a second click listener.** `PTXDialog` wires a toggle click handler on the control button, and the calculator code adds its own click handler that inits GeoGebra `if (calcDialog.dialog.open)`. First-open init works only because PTXDialog's listener is registered first and `.open` updates synchronously — it's order-of-registration coupling rather than a contract. An `onOpen`/toggle callback on `PTXDialog` would give the consumer a clean hook for the lazy init.

5. **First-open applet sizing leans on a later event.** `initGeogebra` omits width/height, and the `ResizeObserver` is attached before `applet` is assigned, so its initial callback no-ops. An explicit `setSize` right after `inject` would size the applet on first open without waiting for a resize.

6. **`aria-expanded` is written as a toggle** (`setAttribute('aria-expanded', !isExpanded)`) on both open and close; asserting `"true"` on open and `"false"` on close is sturdier if those are ever called redundantly.

7. **Mixed invoker paradigm.** The open buttons dropped `command`/`commandfor` in favor of JS, but the search/embed close buttons keep `command="close"` while `PTXDialog` also wires a JS `close()` on them — harmless double-close on Chromium, but two paradigms side by side. Also each non-modal instance adds a `window` `resize` listener that's never removed (fine for the singleton calculator, looser for a reusable framework).

ascholerChemeketa:
Updated with 1-3, 6 and parts of 7.

May address 4 in future.
5 is not a functional issue. I believe the GGB code  handles the initial sizing.

rbeezer:
Took another pass after the force-push. The move to a real `<dialog>` plus the centralized `setExpanded()` cleans this up nicely — the cross-browser `aria-expanded` / open-state problem from the last round is resolved, since both the native and fallback open/close paths now go through the same helper. The mobile top/left clamping and the removal of the old `command`/`commandfor`/`closedby` attributes also address earlier comments. Nothing here is blocking.

A few smaller things I noticed this time, none urgent:

- For **modal** dialogs the native `open` closure doesn't call `setExpanded`, so the search/embed trigger buttons keep `aria-expanded="false"` on Chromium while the fallback path toggles it correctly. Same native-vs-fallback split as before, just lower stakes for modals — worth making consistent.

- `controlElement.commandFor = this.dialog.id` looks inert: the reflected IDL property is `commandForElement`, and there's no `command` value set, so it isn't actually wiring a native invoker. (No harm — it's also why there's no double-toggle against the JS listener — but it reads as intended-but-unfunctional.)

- Two `geoGebraDiv` locals are declared but never used — one in `initGeogebra`, one in the `ResizeObserver` callback.

- The non-modal `window` `resize` listener is never removed. Fine for the singleton calculator, but a leak if PTXDialog grows more non-modal instances.

- The calculator `<dialog>` has no accessible name (`aria-label`); cheap to add given the rest of this work.

- `.ptx-calculator-container` includes the `ptx-dialog-modal` mixin, whose only addition over `ptx-dialog` is the `::backdrop` rule that never renders for a non-modal `show()`. And `_calculators.scss` is missing a trailing newline.

One thing I still want to confirm by hand rather than by reading: first-open GeoGebra sizing. With `width`/`height` dropped in favor of the `ResizeObserver`→`setSize` path, and the observer bailing until the applet object exists, does the applet fill 330×600 on the very first open, before any drag-resize? If you've checked that on a live page I'm happy.

rbeezer:
Thanks very much for the adjustments.  Re-review coming, along your comments above, and will merge.  Re-review is just FYI.

rbeezer:
Merged as-is.  Thaks for all your work on this big push.  Regenerated JS and CSS.  Kept your JS update, for a clean merge, then added Diagcess library that I missed on #2920

Issue #1795, opened 2022-07-19
Screenreader interaction with permalinks


mitchkeller:
Screenreaders are not getting to see the thing that comes up saying a link has been copied to the clipboard when a user clicks on the permalink. (Also think about what exactly screenreaders are reading for these.)

rbeezer:
Closing as fixed. Permalinks are now handled for screen readers via the HTML readability menu (#2951): the auto-permalink carries `aria-hidden`, and the readability menu offers an "accessible permalinks" option so a screen-reader user can opt into permalink interaction.

*Claude Opus 4.8 (1M context), acting as a review assistant for Rob Beezer*

PR #2951, opened 2026-06-19
Add HTML reading options

ascholerChemeketa:
Adds the much awaited reading options for HTML.

Builds off of the calculator branch - everything before 1184ff59c45efcc600548cc74f6c04e7c168bc9f is really on that PR.

Finish and merge that first. Then you or I can rebase this to that point.


ascholerChemeketa:
Just added a touchup via force push.

Also dropped the commit with pretext-core.js. I think the merger needs to add that commit to avoid trying to reconcile merge conflicts in the build product.

rbeezer:
Thanks for getting this rolling!

*  A rebase and force-push will be welcome, not sure I trust myself to handle the old and new commits.  ;-)
*  My practice is to give authores and publishers lots of options turned on, so they know about them.  If they do not like them, they can turn them off.  So I would like permalinks on by default.  I think this was discussed publicly - let me know if I should go and add to that discussion.
*  Claude likes it, but has comments.  Maybe the only must-fix is a leftover debugging command.  ;-)  Coming up next.

rbeezer:
Thanks for getting this rolling!

*  A rebase and force-push will be welcome, not sure I trust myself to handle the old and new commits.  ;-)
*  My practice is to give authores and publishers lots of options turned on, so they know about them.  If they do not like them, they can turn them off.  So I would like permalinks on by default.  I think this was discussed publicly - let me know if I should go and add to that discussion.
*  Claude likes it, but has comments.  Maybe the only must-fix is a leftover debugging command.  ;-)  Coming up next.

rbeezer:
And I forgot - localizations need to go into every language file, as the exact same code as for `en-US` and in the exact same location, but with each line made into a comment.  Tedious, yes.  I've done it many times.  This makes it easy for language contributors to see what needs to be done and they don't make a mess.  Some will react almost immediately.

Also, the `en-US` file is the documentation.  Insert a comment just prior that these strings go on this new menu.  It should not be repeated in the other files.

I am assuming that an assistant will be able to mine our files for the HTML-only strings and those will go into a JS localization scheme that is similarly easy for contributors.  If I am wrong about that, then this might be a good time to initiate whatever scheme will be used and start with these 10 new ones.

rbeezer:
Nice feature, and the refactor is clean — I checked, and `pretext-dialog.js` is a verbatim move of the `PTXDialog` class (just added doc comments and the `window.PTXDialog` export), with the symbols fully removed from `pretext_add_on.js` and the bundle import order correct. The dialog markup is good and accessible: `fieldset`/`legend` for the theme radios, `label`/`for` on the sliders, `<output>` for the live values, and the theme group gated on `$b-theme-has-darkmode`.

One must-fix:

- A leftover `console.log("asdfasdf")` at the top of `setAutopermalinksAccessible` in `readability-options.js`.

Smaller things:

- `applyThemeChoice` calls `setDarkMode(window.isDarkMode())`, but there's a local `isDarkMode()` right there and `window.isDarkMode` isn't assigned until the bottom of the file. It works because the handlers fire post-load, but calling the local function is clearer and less order-dependent.
- `setAutopermalinksAccessible` does `permalink.querySelector('a')` then `link.setAttribute(...)` with no null guard — it'll throw if an `.autopermalink` ever lacks an `<a>`.
- "Reset All" removes the `lineHeight` key but then calls `applyLineHeight(defaultValue)`, so `--ptx-content-line-height` stays explicitly set to the slider default rather than reverting to the stylesheet's value. That's only equivalent if the slider default (1.35) matches the theme's CSS default; clearing the inline property would be more robust. (Font size's 100% is fine.)
- The comment in `js/src/pretext-core.js` still says "pretext_add_on.js defines PTXDialog" — that's `pretext-dialog.js` now.
- Minor: the `reading-settings` localization id is the odd one out against the `readability-*` family.

And agreed on rebasing onto current master as the next step. Otherwise this reads well.

ascholerChemeketa:
OK, will rebase and force push.

This only turns permalinks off for screen readers. They are a distracting mess while using one of those. And you can still use a permalink to get to the right location with a screen reader, you just won't be told about the autopermalink link at the end of each paragraph.

IMO, if they are going to stay on by default when using a screen reader, someone needs to spend some time figuring out a way to make them unobtrusive in that environment. And make sure that they behave correctly (see https://github.com/PreTeXtBook/pretext/issues/1795).

ascholerChemeketa:
re: languages...

I was also wondering about the JS vs PTX internationalization. Anything created from whole cloth by JS should be internationalized there. Anything cross platform (LaTex/HTML) needs to be handled in PTX. But what about HTML only strings? i.e. things like the nav bar buttons.

I think handling those in JS would simplify things in terms of being one place to find/modify both static and dynamic HTML specific stuff.

I agree that AI should make easy work of moving translations around. So I will update all the current files for now. JS/HTML internationalization cleanup can be a separate pass later.

rbeezer:
See related #2816.

ascholerChemeketa:
@rbeezer - Force pushed with fixes for Claude identified issues.

Has localizations strings as comments for all languages.
Suggestion: maybe have your assistant run a pass to normalize the order and insert missing ones. The existing localizations have some discrepancies. I started to do that and decided the only hard work was in judgement calls about things like outdated `md` and `me` localizations and you should make those.

rbeezer:
All sounds good re: localizations:
*  JS stuff is JS
* HTML interface strings, with no comparable equivalent, can go with JS
* Universal text (e.g. "Theorem") can remain in current approach.
* "normalization" sounds beneficial.  They are largely in good shape, but there has been some drift.  (Please have your assistant talk to my assistant?)

rbeezer:
Messing around en route to a merge.  Very nice.  I like the design and the results.  Great to see the changes in "real time", and to have a reset-button.

One idea.  Would it make sense to be able to drag the dialog around?  I felt like I wanted an unobstructed view to watch the rendering take place and settle on final values.  (Separate PR, already well into this one!)

rbeezer:
`js/src/readability-options.js:295` re-reads `localStorage` directly:

```js
setAutopermalinksAccessible(localStorage.getItem("accessiblePermalinks") === "true");
```

It could reuse `getSavedAccessiblePermalinks()`.

rbeezer:
One nit you could catch whenever, if desired.

rbeezer:
Real good.  Take the glory on `-dev` please with an annoucement?  Maybe close out that line-height post that ressurected this one.  Folks are going to be happy to see this.

Merged as-is, but with a couple follow-on commits with small additions.

ascholerChemeketa:
Yes, I'll do a cleanup pass with assorted late Claude suggestions.

BTW - I think we need a JS/CSS dist commit on master.

rbeezer:
> Yes, I'll do a cleanup pass with assorted late Claude suggestions.

Sounds good.  Thanks.

> BTW - I think we need a JS/CSS dist commit on master.

Four copies of the repo, and two Claudes.  I lose my place sometimes.  Like on #2966.

Re-generated, attributed to #2952.  Thanks.

PR #2952, opened 2026-06-20
Nav button cleanup

ascholerChemeketa:
Addresses accessibility and localization issues in nav buttons.

This builds on https://github.com/PreTeXtBook/pretext/pull/2951

Starting at b8567b577468ff0614a3f072fa75b053656ae960

After that PR is in I can rebase to that point.


ascholerChemeketa:
Just added some commits to address issues that your assistant identified at the end of: https://github.com/PreTeXtBook/pretext/pull/2947

ascholerChemeketa:
Rebased to work on https://github.com/PreTeXtBook/pretext/pull/2951 as of 6/23/2026

rbeezer:
Claude's review coming.  I will rebase to blow away old commits, and pick up the nl-NL edit.  With that, I think this is ready to merge, so I am headed that way now

rbeezer:
This builds cleanly on #2951 — nice pass at getting the nav buttons localized and the dialog/ARIA state right. A few notes.

One that I'd call a must-fix:

- **`nl-NL.xml`: `scratch-activecode` was added live (uncommented)** with the English value, while every other language — and every other new string in nl-NL itself — has it commented out:

  ```
  <localization string-id='scratch-activecode'>Scratch ActiveCode</localization>
  ```

  As-is, nl-NL would render English "Scratch ActiveCode" as a "translation" and the string skips the contributor workflow. It should be the commented form like the rest. (It's the only such anomaly — the other languages and the other new strings are all correct.)

The things I checked that look right:

- The `copy` → `cc` rename is safe: the only reference to `string-id='copy'` is the new embed Copy button (the "Copy" meaning), and nothing references the repurposed `cc` id, so the "(unused)" note holds. Languages with a real translation (e.g. nl-NL `kopie`) correctly carried it onto `cc`; no stragglers.
- The PTXDialog fix is correct and symmetric — `open()` now sets `aria-expanded` true for both modal and non-modal, and `close()` already sets it false with focus return.
- `insert-symbol` replacing the raw `<span>close</span>` in the embed/search close buttons, plus their new accessible `title`s, is a real a11y win.

Smaller, non-blocking:

- Localization is partial — button titles/labels are done, but a few adjacent strings stay hardcoded English: the embed popup's `<p>Copy the code below…</p>` (most visible), the textarea `aria-label="textbox"` (also a weak label), and the search dialog's `aria-label="Close search"` / `"Search term"` / placeholder `"Search terms"`. Up to you whether to widen scope here.
- "Calc" → "Calculator" widens the desktop navbar label (and the title goes from "Show calculator" to "Calculator"); fine, just noting — it's hidden on mobile anyway.
- `.ptx-embed-button` is now `display:none` on mobile but still listed in the `:is(…) .name { display:none }` selector — redundant now.

rbeezer:
Rebase had a merge conflict with the embed button string.  I think I got it right by looking at similar items close-by, but worth a look if anything related mis-behaves.

rbeezer:
The `nl-NL` localization string was a false positive, the string was in a **block** of localizations commented out.  Somebody should normalize those files!

rbeezer:
Thought I was going to edit, so a couple of minor edits on commit messages.  Otherwie, merged as-is after the rebase edit.

Looking good - nice to see this all progressing.

I've set my assistant on the localization files while I have dinner.

PR #2953, opened 2026-06-20
Update profile dropdown menu

ascholerChemeketa:
Adds PTXDropdown framework and uses that to make Profile menu accessible. 
Gets rid of hard coded RS menu items.

This builds on https://github.com/PreTeXtBook/pretext/pull/2952

Starting at f46495a1f32ac646d20c876375424489deff3f77

After that PR is in I can rebase to that point.

### DIFF
--- a/css/components/helpers/_buttons-default.scss
+++ b/css/components/helpers/_buttons-default.scss
@@ -51,37 +51,48 @@ $border-width: 1px !default;
   }
 }
 
-@mixin ptx-dropdown-button {
-  position: relative;
+@mixin ptx-dropdown-menu {
+  display: none;
+  position: absolute;
+  background-color: var(--dropdown-background);
+  min-width: 160px;
+  z-index: 100;
+  border: 1px solid var(--dropdown-border-color);
+  right: 0;
+  top: 100%;
+  text-align: start;
+  padding: 0;
 
-  .dropdown-content {
-    display: none;
-    position: absolute;
-    background-color: var(--dropdown-background);
-    min-width: 160px;
-    z-index: 100;
-    border: 1px solid var(--dropdown-border-color);
-    right: 0;
-    top: 35px;
-    text-align: start;
+  a {
+    display: block;
+    text-decoration: none;
+    color: var(--dropdown-text-color);
+    padding: 2px 8px;
+
+    &:is(:hover, :focus-visible) {
+      background-color: var(--dropdown-hover-background);
+      color: var(--dropdown-hover-text-color);
+    }
+  }
+
+  ul {
+    list-style: none;
     padding: 0;
+    margin: 0;
+  }
 
-    a {
-      display: block;
-      text-decoration: none;
-      color: var(--dropdown-text-color);
-      padding: 2px 8px;
+  hr {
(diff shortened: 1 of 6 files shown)


rbeezer:
I think this must need a rebase, and I suspect it will be of a bigger magnitude than I should be touching?  Along the way, Claude suspects one "dormant" bug and has some nits.    Coming next.

rbeezer:
The PTXDropdown approach is the right move — proper menu semantics and keyboard handling, and it fixes the invalid `<div>`-inside-`<button>` nesting the old profile menu had. A few things.

One actual bug:

- In the `PTXDropdown` constructor, the "close on escape" listener dereferences `this.controlElement` outside the `if (this.controlElement)` guard that wraps the other control-element listeners:

  ```js
  // Close on escape even if opened via click and focus is on button
  this.controlElement.addEventListener("keydown", (event) => {
  if (event.key === "Escape" && this.isOpen()) {
  ```

  Since `openButton` defaults to `null` and you use `this.controlElement?.contains(...)` elsewhere, the class is meant to tolerate a null control — so `new PTXDropdown(menuEl)` would throw here. The current call site passes a button, so it doesn't bite today. Folding this listener into the existing `if (this.controlElement)` block fixes the crash and also drops a duplicate Escape path (it's already handled in `handleKeydown`). That inner `if` is also mis-indented.

Smaller:

- `console.log("PTXDropdown: No dropdown element provided.")` reads more like a `console.warn`.
- `pretext-dropdown.js` has no trailing newline.
- XSL: `aria-labelledby="ptx-user-dropdown-button"  role="menu"` has a double space; and `aria-label="Runestone"` is the one hardcoded string in the new menu (probably fine as a product name).
- Worth a visual check: the old `ptx-dropdown-button` mixin set `position: relative` on the button, and the renamed `ptx-dropdown-menu` mixin (now on the menu element) uses `position: absolute; top: 100%; right: 0` without that anchor — confirm the menu still positions under the button rather than against a higher navbar ancestor.

ascholerChemeketa:
OK, I'll do a rebase and then look at Claude's suggestions.

ascholerChemeketa:
Rebased off of recent master and addressed Claude's suggestions.

Also added a throw away commit at the end with a chunk of testing code that forces the Runestone bust menu to be rendered even in a `platform host="web"` build.

rbeezer:
Merged as-is, sans testing commit.

Any speculation on more to do here, beyond #2975?

BTW, the `or true()` on the profile menu did not seem to work for me.  I added the new `dist/` and that did not seem to matter.  So this merge was not really fully tested on my end.  (Thanks for trying, all teh same!)

ascholerChemeketa:
Since your testing didn't work, I did one last test using the merged copy on my local RS server. All still looks good.

I think that completes the navbar tour other than any cleanup. I think you can close https://github.com/PreTeXtBook/pretext/issues/2794

Movable reading settings can come later - not a big lift, but some decisions to make in it.

There are parts of the content that need accessibility review: legacy knowl keyboard handling, image magnification. But those are separate strands of work (though image magnification can leverage the PTXDialog to do the work).

PR #2954, opened 2026-06-21
XSL-FO: bundle a symbol font, fix missing glyphs, add block end-marks

rbeezer:
## Summary

The XSL-FO / Apache-FOP route to a PDF (the `pdf-fo` format) sets text in Latin Modern, which carries few non-letter glyphs, so FOP was printing `#` for many symbols an author can legitimately use. This bundles a dedicated symbol fallback face, fixes a batch of missing glyphs, gives definitions and examples their end-of-block marks, and implements visible spaces.

## A bundled symbol font

A new top-level `fonts/` holds **`PreTeXtSymbols.otf`** (~230 KB), a subset of GNU FreeFont's **FreeSerif** covering the symbol Unicode blocks, alongside `make-symbol-font.py` (regeneration) and a `README.md` documenting the bundling criteria, provenance, and license.

- FreeSerif is the one open *serif* font (so it matches the Latin Modern body) that carries the whole set in a single face — the geometric end-marks `■ ◆ ▲`, currency including the Guaraní `₲`, primes, the maltese cross and six-pointed star, white brackets.
- License is **GPL-3-or-later with the GNU font exception**, so embedding it in an author's PDF does *not* place the document under the GPL. The subset is a modified work that extends the exception; the regeneration script and README record this.
- `fop.xconf` embeds it by a relative path; `pdf_fo()` copies `fonts/` into the build directory, where FOP resolves the path. The other embedded faces are unchanged.

## Glyph fixes

`fo:root` now names a font-family list (`body, symbol`), so any glyph the body font lacks falls back to the symbol face. FOP's list fallback is **per word**, so glyphs that occur *inside* a word are named explicitly instead. Fixed: the Guaraní sign, the square unordered-list marker, prime/double-prime (minutes & seconds), thin space, swung dash, angle brackets, and the copyleft (now a parenthesized reversed-c, previously a missing-glyph box). `kbd[@name]` keys use a `monospace, symbol` list so the ASCII-valued named keys still render.

## End-of-block marks

Definitions and examples now close with a mark, matching the LaTeX conversion — `◆` for a definition, `▲` for an example — and the proof tombstone moves from `∎` to `■` so the three are a consistent filled-shape set. The mark rides the block's final line when it ends in a paragraph (reaching through a closing `statement`), otherwise it takes a line of its own.

## Visible spaces

`<cd showspaces="all">` now renders each space as the OPEN BOX `␣` (U+2423), as in the HTML route; the FO route had ignored the attribute.

## Testing

The sample article and the `pdf-fo-development` dev article both build end-to-end as PDFs and pass veraPDF **PDF/UA-1** validation (106/106 rules, 0 failures), with the symbol font fully embedded.

## Known gaps (not addressed here)

- Greek and other non-Latin *letters in running text* remain unrendered: the body font lacks them and a symbol font is the wrong home for letters. Pre-existing; wants a script-capable body fallback.
- The DejaVu / Latin Modern / FontAwesome faces still reference absolute host paths in `fop.xconf`; bundling or per-host resolution is deferred.

---


rbeezer:
@oscarlevin - you will want to have a look at this.  New `fonts` directory that I think you will need to claim as part of the CLI distribution.  I am going to distribute a *limited* number of mature, free, non-bloated fonts, for out-of-the-box Tier 1 support.  Not there yet.  Publisher switch may change.

Tier 2 will be tested configurations.  BYOF.

More at Drop-In?

PR #2956, opened 2026-06-22
Fix typo in "Custom Versions" > "Versions"

sarangjo:
In the section "27.2.2 Forming a Version" in the PreTeXt guide, there is a small typo:

> This is accomplished with the `@include` attribute of the `source/include` element inside a publication file.

This should be attribute of the `source/version` element. This PR fixes this typo.


rbeezer:
Right!  Thanks very much @sarangjo!  Should go live later today when I rebuild the website.

PR #2964, opened 2026-06-23
Exercise group intro/conclusion spacing

ascholerChemeketa:
Addresses spacing issues identified by @jjrsylvestre in exercise group intro/conclusions.

I'm not sure the horizontal indent on conclusion makes sense, feels like it should match the intro, not the exercises, but that is what was there historically.

### DIFF
--- a/css/components/_spacing.scss
+++ b/css/components/_spacing.scss
@@ -25,13 +25,16 @@ $line-height: 1.35 !default;
 // Most spacing should be done with margin-top.
 
 @layer spacing {
+  // selectors for items that are expected to contain sequences of items
+  $containers: "section, article, .knowl__content, .introduction, .conclusion";
+
   // inter-item spacing high level containers
-  :is(section, article, .knowl__content) > *:not(:first-child) {
+  :is(#{$containers}) > *:not(:first-child) {
     margin-top: 1.5em;
   }
 
   // but tighten up spacing slightly for adjacent paragraphs in a section
-  :is(section, article, .knowl__content) > .para + .para {
+  :is(#{$containers}) > .para + .para {
     margin-top: 1em;
   }
 
--- a/css/components/chunks/_exercises.scss
+++ b/css/components/chunks/_exercises.scss
@@ -8,6 +8,10 @@
   font-size: inherit;
 }
 
+.exercise-like > *:is(.solutions) {
+    margin-top: 0.5em;  // tighten up spacing between answer, hint, solution, etc...
+}
+
 .exercisegroup {
 
   .exercise-like {
@@ -38,7 +42,7 @@
   }
 
   .conclusion {
-    margin-left: 40px;  // match the padding of the exercisegroup-exercises
+    margin-top: 1em;
   
     .heading {
       // exercise heading/numbers regular size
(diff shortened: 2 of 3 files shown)


rbeezer:
For PDF (since I was just doing it) we only indent the `exercise` themselves.

To keep the two PDF conversions in-sync, I have just started an area in `pretext-common.xsl` with "Distances and Colors".  Don't know if you want to mine that or not?  No colors yet, "Sage blue" might be the first.  (I think this has been pushed!)

ascholerChemeketa:
Removed the conclusion indent to match PDF and force pushed.

Leaving exercise indentation as is for now.

rbeezer:
All CSS, so I will ask @oscarlevin to handle this one.  Also, I don't know enough to follow Claude's reservations.  Posting that next.

rbeezer:
The exercisegroup spacing problem is real and the fix works for it, but two of the three CSS changes reach well beyond exercise groups — worth a look before merge.

**1. `.introduction`/`.conclusion` in `_spacing.scss` is document-wide, and it wins the cascade.** Both intro/conclusion templates set the class from `local-name()`, so `class="introduction"` lands on every division intro too — `<section class="introduction">`, not just the exercisegroup `<div class="introduction">`. The edited rule (specificity `(0,3,0)`) beats the high-level `:is(section, article, …) > *` rule (`(0,1,1)`) within the same `@layer`, so every chapter/section introduction and conclusion drops from 1.5em to **0.5em** inter-item spacing. The sample article alone has ~10 multi-paragraph section introductions that this re-spaces. The exercisegroup case genuinely needs the fix (those `<div>` intros matched nothing in the spacing layer), but scoping it — e.g. `.exercisegroup .introduction, .exercisegroup .conclusion` — would keep it off the division `<section>` intros.

**2. `.exercise-like > *:not(:first-child) { margin-top: 1em }` is also global.** It is at the top level of `_exercises.scss` (not under `.exercisegroup`) and outside `@layer`, so it overrides the layered 1.5em for *every* exercise — intra-exercise spacing (statement/hint/answer/solution) tightens to 1em document-wide. Reads as intended, just confirming that is the goal and not meant to be exercisegroup-only.

**3. Dropping the `conclusion` `margin-left: 40px` — agreed.** Flush-left intro and conclusion bracketing the indented exercises is a cleaner hierarchy than the historical indent; your instinct in the description is right.

Minor: `_spacing.scss` still has no trailing newline; sample-article test cases read clean.

Net: #3 is good; #1 and #2 work for exercisegroups but currently apply to the whole document, and #1 in particular silently re-spaces all division introductions via specificity — worth either scoping or making an explicit, tested decision.

oscarlevin:
I'm going to agree with Claude here and say that the fix for introductions/conclusions is too broad.  The introduction to a section gets half the top-margin on paragraphs that regular text gets.  I'm not sure why paragraphs in *any* introduction should be squished together more than other paragraphs.

ascholerChemeketa:
Force pushed with narrower focus for .introduction and .conclusion changes. Adjusted spacing for them to match "full spacing" containers.

oscarlevin:
I think this works for exercisegroup, but introductions to exercises and projects (before tasks) still do not have paragraph breaking.  We could merge this as is, or tweak to include that case too.  If you see how to do this right away, that would be great, or I can try to figure it out later today or tomorrow.

ascholerChemeketa:
Force pushed with update to include all introductions/conclusions.

Some are already `<section>` or `<article>` and this ends up being duplicative. But this now picks up anything that is `.introduction` or `.conclusion` instead of just those in exercise groups.

oscarlevin:
Excellent.  Thanks.  Merged as is.

PR #2966, opened 2026-06-23
XSL-FO: render captions as run-in paragraphs, not centered

rbeezer:
In the XSL-FO → PDF route, a figure-like block's caption — a `figure`'s `caption`, and the title of a `table`, `listing`, or `list` — was set in a `text-align="center"` block. A short caption survives this, but a long one renders as awkward centered, ragged, multi-line text. A caption is really a paragraph and should follow the same rules: a bold **run-in heading** (the type name and number) leading body-aligned prose.

This switches that block to the document's body alignment (`$text-alignment` — `justify`, or `start` when ragged-right), so the bold "Figure 5.3." / "Table 5.1." becomes a true run-in heading and the caption sets as ordinary prose. The paragraph following the figure already indents correctly via the `p` template's `preceding-sibling::p` rule, and a `caption` holds inline content, so there are no paragraphs *within* it to indent.

### Testing

Sample article: the `figure-long-caption` test (Figure 10.31) now reads as a justified paragraph led by a bold "Figure 10.31." run-in heading, rather than a centered ragged block; short captions and the table/listing/list titles get the same flush run-in treatment. veraPDF `--flavour ua1`: **106 passed / 0 failed**.

### Commits

- `XSL-FO: render captions as run-in paragraphs, not centered`

### DIFF
--- a/xsl/pretext-fo.xsl
+++ b/xsl/pretext-fo.xsl
@@ -3287,15 +3287,18 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 
 <!-- The FIGURE-LIKE blocks (a captioned "figure"; and "table",   -->
 <!-- "listing", "list", each titled) wrap their contents and      -->
-<!-- finish with a centered caption line; all share one counter.  -->
-<!-- The "caption" element is consumed here, so it is killed in   -->
-<!-- the default mode, with the other metadata.                   -->
+<!-- finish with a caption: a body-aligned paragraph led by a     -->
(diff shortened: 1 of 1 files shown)


rbeezer:
That Table of Contents commit was a mistake, it was meant to be part of #2937, and it should be gone now after a force-push.

rbeezer:
While fixing #2937, I messed up #2966 here via a force-push.  So the initial reported merge is gone, and I have re-done it as https://github.com/PreTeXtBook/pretext/commit/fb8d9b47db2d3a4543d7dfc91f52d29a5b30d0cb, which is the most recent report.

PR #2975, opened 2026-06-25
Accessibility cleanups for recent PRs

ascholerChemeketa:
Collection of small fixes/cleanups based on Claude parting shots from recent PRs.

Also bumps cssbuilder installed esbuild version to match jsbuilder. Both are now most recent 0.25 release (Nov 2025). It looks like to move to 0.27 we'll need to do some testing, but no urgent need to do that move.

@oscarlevin just in case he has notes/thoughts on versions of esbuild.

### DIFF
--- a/css/components/page-parts/extras/_navbar-btn-borders.scss
+++ b/css/components/page-parts/extras/_navbar-btn-borders.scss
@@ -3,16 +3,18 @@ $border-width: 1px !default;
 
 @mixin navbar-btn-borders {
   .ptx-navbar-contents {
+    gap: $border-width;
+    background: var(--page-border-color);
+
     & > .button,
     & > :not(dialog) > .button {
-      border-left-width: $border-width;
-      border-right-width: $border-width;
-      border-color: var(--page-border-color);
+      border-left-width: 0;
+      border-right-width: 0;
     }
-  
-    & > *:not(:first-child) {
-      //hide double borders
-      margin-left: -1 * $border-width;
+
+    :is(.treebuttons, .nav-runestone-controls, .nav-other-controls) {
+      gap: $border-width;
+      background: var(--page-border-color);
     }
   }
 }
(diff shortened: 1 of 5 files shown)


rbeezer:
Thanks - looks like Claude has a "must-fix", I'll post the full review.

rbeezer:
Nice sweep — it picks up the recent suggestions cleanly: ToC Escape-to-close now fires for the **mobile overlay** (`position: fixed`) and not just auto-collapse themes; `toggleTOCItem` drops the dead `targetElement` and hoists `groupName` out of the `if/else`; and the embed textarea and search-close `aria-label`s are now localized.

One must-fix, in `js/src/readability-options.js`:

```js
getSavedAccessiblePermalinks();
```

This replaced `setAutopermalinksAccessible(localStorage.getItem("accessiblePermalinks") === "true");`, but `getSavedAccessiblePermalinks()` is a pure getter — it returns the boolean and throws it away, so `setAutopermalinksAccessible(...)` is never called on load. The checkbox state is restored (the line just above), but the permalinks themselves aren't updated to match the saved preference until the reader toggles the control. It should be:

```js
setAutopermalinksAccessible(getSavedAccessiblePermalinks());
```

Minor: the embed textarea's new `aria-label="{$embed-localization} URL"` renders "Embed URL", but the box holds embed *code* (an `<iframe>`), not a URL — "code" reads better, and the trailing " URL" is hardcoded English on the localized part.

ascholerChemeketa:
Fixed both of those and rebased/pushed.

The aria-label for the Embed textbox is now just `"Embed"` so we don't add another localization. `"Embed Code"` would be better, but at some point we will want to tackle all the interactive localization in JS anyway. (And the long string in the Embed dialog is not localized at all.)

rbeezer:
Merged after a rebase.

Phew!  That was a great push.  Thanks.

PR #2981, opened 2026-06-26
Support descriptions on additional interactive

mitchkeller:
Sending this as three commits just to make clearer where the changes were. I think this now covers the various interactive elements (Desmos, GeoGebra, CalcPlot3D, CircuitJS, and arbitrary iframe) for descriptions. 

- Two changes in assembly: 
  - Ensure that `interactive[@geogebra]` with children has them copied when restructuring to use `interactive/slate`. 
  - Move an unstructured `description` to be a `shortdescription` when the child of `interactive`
- Changes to `pretext-html.xsl` to replicate earlier work done for `interactive[@platform]` to have descriptions.
- A mixture of test cases (no `shortdescription`, both `shortdescription` and `description`, unstructured `description`) added to the sample article.

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -9208,6 +9208,9 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 <figure>
                     <caption>GeoGebra: modified material ID</caption>
                     <interactive xml:id="geogebra-train-distance" platform="geogebra" width="75%" aspect="2:1">
+                        <description>
+                            This is a test of text-only description, which gets moved to shortdescription.
+                        </description>
                         <slate xml:id="train-distance" surface="geogebra" material="D4s2v4ft" aspect="2:1">
                             setAxisSteps(1, 1, 5, 1);
                             setCoordSystem(-0.7, 8, -6, 56);
@@ -9389,7 +9392,13 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
                 <figure xml:id="figure-circuitjs-attribute">
                     <caption>CircuitJS Example (source via an encoded attribute)</caption>
 
-                    <interactive xml:id="interactive-circuitjs-attribute" circuitjs="$ 1 0.000005 10.20027730826997 58 5 50 5e-11%0AR 464 32 464 -16 0 0 40 3.3 0 0 0.5%0Ag 464 224 464 240 0 0%0As 464 32 464 128 0 0 false%0As 464 128 464 224 0 1 false%0Ax 274 181 431 184 4 12 bottom\sswitch\s(to\sground)%0Ax 312 85 426 88 4 12 top\sswitch\s(to\sVdd)%0AO 464 128 544 128 1 0%0Ax 532 110 593 113 4 12 output\spin" width="90%" aspect="4:3"/>
+                    <interactive xml:id="interactive-circuitjs-attribute" circuitjs="$ 1 0.000005 10.20027730826997 58 5 50 5e-11%0AR 464 32 464 -16 0 0 40 3.3 0 0 0.5%0Ag 464 224 464 240 0 0%0As 464 32 464 128 0 0 false%0As 464 128 464 224 0 1 false%0Ax 274 181 431 184 4 12 bottom\sswitch\s(to\sground)%0Ax 312 85 426 88 4 12 top\sswitch\s(to\sVdd)%0AO 464 128 544 128 1 0%0Ax 532 110 593 113 4 12 output\spin" width="90%" aspect="4:3">
+                        <description>
+                            <p>
+                                An interactive CircuitJS embed with a description.
+                            </p>
+                        </description>
+                    </interactive>
                 </figure>
 
                 <figure xml:id="figure-circuitjs">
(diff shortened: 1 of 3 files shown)


rbeezer:
Thanks, @mitchkeller!  I'll  be able to look at this closer in a couple of days.

rbeezer:
As discussed, review from Claude coming next.

* Some typos to fix up.
* Some duplicated code that should be isolated in a new template.  let me know if you want me to do that, I can add it.

A force-push is fine.  Thanks for this one!

rbeezer:
## Review of #2981 — works end-to-end; one consolidation suggestion

I built the sample article on the branch (its edits are the test cases) and inspected the rendered output for all five interactive types. Everything lands correctly:

| Test case | Result |
|---|---|
| `desmos-natural-log` (short + long) | `title="An interactive graph of a function."`, long description block follows ✓ |
| `geogebra-right-triangle` (shorthand `@geogebra`, structured desc) | `title="described in detail…"` + `aria-describedby` + description ✓ |
| `geogebra-train-distance` (`@platform`, **unstructured** desc) | repaired to `shortdescription` → `title="This is a test…"` ✓ |
| `interactive-circuitjs-attribute` (structured desc) | `title` + `aria-describedby` + description ✓ |
| `calcplot3d-intersection-planes` (short + long) | `title` from short + description block ✓ |
| `calcplot3d-probability-wavefunction` (long only) | `title` + `aria-describedby` + description ✓ |

**The geogebra shorthand path is the subtle one and it checks out.** `interactive[@geogebra]` is restructured in assembly to `<interactive platform="geogebra">` (it explicitly adds `@platform`), and the new `<xsl:apply-templates select="node()" mode="enrichment"/>` copies the authored `<description>` across before the manufactured `<slate>`. By the time HTML sees it, it matches the existing `interactive[@platform]` description templates — which is why `@geogebra` correctly doesn't appear in the new HTML match list, and the copied description renders correctly in the output.

**The repair extension is sound.** Adding `interactive/description[not(*[not(self::var)])]` alongside `image/description[…]` correctly moves an unstructured `interactive` description to `shortdescription` (verified by `geogebra-train-distance`).

**Consolidation opportunity (optional — the PR is correct as-is).** This work brings the count of one accessibility block to five. The title/`aria-describedby` `choose` is now byte-identical across all five `mode="iframe-interactive"` templates — `interactive[@desmos]`, `[@calcplot3d]`, `[@circuitjs]`, `[@iframe]`, and the pre-existing `[@platform]`:

```xml
<xsl:choose>
    <xsl:when test="not(string(shortdescription) = '')">
        <xsl:attribute name="title"><xsl:apply-templates select="shortdescription"/></xsl:attribute>
    </xsl:when>
    <xsl:when test="description">
        <xsl:attribute name="title"><xsl:text>described in detail following the image</xsl:text></xsl:attribute>
        <xsl:attribute name="aria-describedby"><xsl:apply-templates select="." mode="describedby-id"/></xsl:attribute>
    </xsl:when>
</xsl:choose>
```

Since all five are identical and all are `mode="iframe-interactive"`, lifting this into one shared template and calling `<xsl:apply-templates select="." mode="iframe-accessibility-attributes"/>` from each site would be output-neutral and collapse 5 copies to 1. Two smaller repetitions sit alongside it in the four new templates and could fold into the same helper: the `<xsl:apply-templates select="." mode="iframe-dark-mode-attribute"/>` after the `choose` is a redundant second call (it's already applied earlier in each template — harmless, last-wins, but dead), and the `<xsl:apply-templates select="." mode="description"/>` after `</iframe>` repeats five times. None of this affects output; it's purely DRY-ing up the five near-identical templates while they're all being touched anyway.

**A few typos in the new sample-article descriptions:**
- `sample-article.xml:9633` (`geogebra-right-triangle`): "decomposed into four **pices**" → **pieces**; and "…shapes of the pieces in the left triangle. **on** the right…" → capital **On** (new sentence).
- `sample-article.xml:9705` (`calcplot3d-probability-wavefunction`): the description labels the surface `<m>\phi_{n_x,n_y}(x,y)</m>`, but the interactive's jaxlabel is `\psi` and the caption is "Probability wavefunction" — should be `\psi`, not `\phi`.
- `sample-article.xml:9666` (`desmos-natural-log`): "…at `x=0`, where the `y`-value is approximately **1.5**." For `ln(x^2+5)-3` that value is `ln(5)-3 ≈ -1.39`; the shape is described correctly, but the number looks off (likely `ln(5)≈1.6` without the `-3`).

**Non-blocking observation:** for the short+long combination, the long description is in the DOM but the iframe is `aria-describedby`-linked only when there's no `shortdescription`. That's the existing `@platform` behavior faithfully replicated, so it's not this PR's concern — just flagging for whenever the broader pattern is revisited.

Net: functionally complete and correct across all five interactive types. No blocking change; the consolidation is a nice-to-have given how much of these templates is now duplicated, and the description typos are worth a quick fix since this is sample content.

mitchkeller:
I've pushed the typo fixes, and I'd like to do the consolidation bit myself, as I think doing that will improve my understanding of how some of these things work. Questions before I dig into that:

- Is there a preferred location in `pretext-html.xsl` where I should put the `iframe-accessibility-attributes` template? Not sure if there's an overarching structure to where templates get located.
- I need to review more carefully, but it looks like there are two other attributes (`@html-id-attribute` and `@interactive-sizing-style-attribute`) that are addressed in all of these templates, but it's only the `@iframe-dark-mode-attribute` that accidentally got a redundant addition. Would it be better to leave all three of those attributes behind in each individual template or take them all along into the new template with the code Claude flagged? Maybe then accessibility attributes isn't the best descriptor though. (Perhaps `iframe-common-attributes` or `iframe-shared-attributes`?)
- Mention of the call <xsl:apply-templates select="." mode="description"/> after </iframe> that's repeated across the templates is a bit unclear to me. I can't take the `</iframe>` off into the new template, so is there any action to be done with this, or is that just Claude making an observation?

rbeezer:
* Put it close by.  I usually put it *after* use, so if a code-reader wants more detail they will get there eventually.
* Let's keep topical things together.  You could put those *three* together in a template with a generic name (I like `common`).
* "a bit unclear to me" - maybe me too.  I think that maybe lives outside the `iframe`, and needs to?  So there is no real benefit in grouping it with other repeats?  I saw in the diff and thought about it.  Maye just leave it as-is.

mitchkeller:
Two topical templates as suggested are now after the five places they're used. Left the `mode="description"` where it was, because I think we need to not be putting that inside the `iframe`. Should be ready for you/your assistant to review.

rbeezer:
Thanks @mitchkeller!  Good to have some more accessibility features.  Lots of editing:

*  We don't accept new commits with errors (like typos) and then correct them in subsequent commits at the same time.  An interactive rebase is a great tool -- worth learning.  I "fixed in" your typo corrections, and separately your consolidation via shared templates.  Back to three commits, and I think you will agree that reads better.
*  We do not put inaccurate markup into the sample article ("test of a text-only description").  Yes, you should test your new repair feature.  No, we should not carry the **old** unstructured `description` in the sample article forever, for the schema to trip up on.  There is an "errors and warnings" test document if you want to have a permanent example.  Wrapped in a `p` now and really doesn't say much as a description.
*  A `shortdescription` had a newline in it ("...in three-space"), which maybe would be sanitized, but I removed it anyway.
* Edited commit messages - worth comparing.  Soon there will be a guide in the Guide.

Thanks again!

PR #2982, opened 2026-06-26
Update the webwork version check approach.

drgrice1:
The current method makes a request to the deprecated `html2xml` endpoint to get the version.  Then falls back to retrieving and parsing the HTML for the landing page.  The problem with that approach is that the `html2xml` endpoint will eventually be removed.  Most likely this will happen with the next release of WeBWorK. See https://github.com/openwebwork/webwork2/pull/3034.

So this switches to first making a request to the `render_rpc` endpoint.  If that returns a 200 okay response, then get the version from the response.  Otherwise fall back to the `html2xml` endpoint and attempt to get the version from that response.  If that still fails, then try the landing page approach.

The `render_rpc` request will work for version 2.18 or newer of webwork2.  The fallback `html2xml` request will only work for webwork 2.17.  Any older version will not even send the `ww_version` in the `html2xml` request, so the landing page fallback approach is the only thing that could possibly work.

### DIFF
--- a/pretext/lib/webwork.py
+++ b/pretext/lib/webwork.py
@@ -209,8 +209,13 @@ def webwork_to_xml(
             disableCookies='1',
             outputformat='raw'
         )
-        # Always use html2xml for this; we don't know the server version yet
-        version_determination_json = requests.get(url=webwork2_html2xml, params=params_for_version_determination).json()
+        # Try render_rpc first.  If that fails, then fall back to html2xml.
+        render_rpc_response = requests.get(url=webwork2_render_rpc, params=params_for_version_determination)
+        if render_rpc_response.status_code == 200:
+            version_determination_json = render_rpc_response.json()
+        else:
+            version_determination_json = requests.get(
+                url=webwork2_html2xml, params=params_for_version_determination).json()
         if "ww_version" in version_determination_json:
             webwork2_version = version_determination_json["ww_version"]
             webwork2_version_match = re.search(


Alex-Jordan:
Thanks! I'm sure this is good, but can't test until roughly July 7. There is some stuff in `doc/guide` about setting up a WW server for PTX purposes, and it instructs everyone to enable `html2xml`. That should also be updated to only do that for when your WW server is 2.17.

drgrice1:
Okay.  I will update that.

drgrice1:
Actually, the only reference to the `html2xml` endpoint in the `doc/guides` directory is in the `doc/guides/publisher/webwork.xml` file, and it is regarding the setup for the `Access-Control-Allow-Origin` header.  The 2.17 setup for that obviously won't change.  The 2.18 setup shouldn't change either.  That does not affect the request from the `webwork.py` script, and the 2.18 javascript is still setup to use the `html2xml` endpoint, and so it still needs the `html2xml` setup for that.  The 2.19 setup already says to use the `render_rpc` endpoint and leave the `html2xml` part commented out.  There is no 2.20 setup instructions (although that would be the same as for 2.19).  So I don't think a change is needed for that file.

rbeezer:
Thanks, @drgrice1!  I'll be able to study this closer in a couple of days.

rbeezer:
Thanks again, @drgrice1!  I like it, and so does Claude.  Thanks for getting this our in front of upcoming changes.

Tested against current server at AIM, and results look good based on the new endpoint.

Merged as-is with minor edits to commit message.

drgrice1:
Darn! I know you are particular about the commit messages, and so I tried to figure out what it should be based on the existing commit messages. It should have been a prefix of `WeBWorK` not `Python`.

rbeezer:
No problem!  It was a good guess.  `Script` or `WeBWorK`.  A toss-up.  Thanks for trying.

PR #2983, opened 2026-06-29
Common: reserve all ptx-* ids for internal use

ascholerChemeketa:
Modifies `identifier-warning` to reserve all ids starting with `ptx-` for internal use and drops the existing enumeration of known instances.

### DIFF
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -10330,19 +10330,10 @@ http://andrewmccarthy.ie/2014/11/06/swung-dash-in-latex/
     <!-- identifier equal to one of these would collide and break the UI. -->
     <!-- Audited against the HTML output, 2026-06-04.                     -->
     <xsl:variable name="reserved-html-ids">
-        <xsl:text>ptx-masthead ptx-content ptx-content-footer ptx-page-footer </xsl:text>
-        <xsl:text>ptx-navbar ptx-sidebar ptx-toc-toggle ptx-toc mainmatter logo-link latex-macros </xsl:text>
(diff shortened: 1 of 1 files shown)


rbeezer:
A few unprefixed ids got chopped out accidentally, I will restore them.  I guess we could whittle down that list if anybody had the enthusiasim, or a willing assistant.

A `ptx-*` prefix in the WeBWorK sample chapter, which I will fix (FYI: @Alex-Jordan).

Assists from Claude, so a review coming next.

rbeezer:
Nice cleanup — collapsing the enumeration into a single `ptx-` prefix ban is the right move and removes a list that had to be hand-maintained. Two things before it lands:

**1. Three reserved ids got dropped that aren't `ptx-` prefixed.** This removed line —

```
ptx-navbar ptx-sidebar ptx-toc-toggle ptx-toc mainmatter logo-link latex-macros
```

— also carried `mainmatter`, `logo-link`, and `latex-macros`, which the new `starts-with(., 'ptx-')` test doesn't cover. They're still live HTML ids (`id="mainmatter"`, `id="latex-macros"`, `id="logo-link"`), so an author id of one of those would now slip past the guard into a silent collision. Worth keeping them in the surviving reserved list:

```xml
<xsl:text>mainmatter logo-link latex-macros </xsl:text>
```

**2. The blanket ban catches one of our own examples.** `examples/webwork/sample-chapter/sample-chapter.ptx` has `<exercise label="ptx-source-with-server-images">`, and `@label` is checked too — so this would become a FATAL and break the sample-chapter build. The label also flows into the generated WeBWorK filenames/ids under `my-generated/`, so it'd need renaming (drop the `ptx-`) and a regenerate in the same change.

Everything else removed did start with `ptx-` and is correctly absorbed by the new check.

PR #2984, opened 2026-06-30
Add `resources/@host` publication variable to use cdn versions of css/js

oscarlevin:
Since the reveal.js publication variables already allows selecting of local vs cdn hosting its javascript, I think this choice of publication variable name makes sense for html too.

This is mostly straight forward, but I did clean up the xsl and python a bit to use a consolidated `$b-cdn-resources` variable that is true when either `portable="yes"` or `host="cdn"` is selected.

### DIFF
--- a/doc/guide/publisher/publication-file.xml
+++ b/doc/guide/publisher/publication-file.xml
@@ -806,8 +806,26 @@
                 <cline>/publication/html/platform/@portable</cline>
             </cd>
             attribute to the value <c>"yes"</c>. 
-            This will result in the HTML using hosted CSS and Javascript files, rather than including them adjacent to the rest of your output.
-            This is especially useful in conjunction with setting the <xref ref="common-chunking-options">chunking variable</xref> to 0 to get a single html file.</p>
+            This will result in the HTML using hosted CSS and Javascript files, rather than including them adjacent to the rest of your output  (identical to selecting <q>cdn</q> as the <c>resources/@host</c> value, see <xref ref="online-resource-host-options"/>).
+            Further, the <xref ref="common-chunking-options">chunking variable</xref> will be set to 0 to get a single html file, and svgs produced from generated assets will be embedded directly in the page.</p>
+        </subsection>
+
+        <subsection xml:id="online-resource-host-options">
+            <title>HTML Resources</title>
+            <idx><h>resource hosting</h><h>HTML</h></idx>
+            <idx><h>host</h><h>HTML resources</h></idx>
+
+            <p>The
+            <cd>
+                <cline>/publication/html/resources</cline>
+            </cd>
+            element can have an attribute <attr>host</attr> with values:
+            <ul>
+                <li><c>local</c>: the default, all css and js files are included in the output directory.</li>
+                <li><c>cdn</c>: all css and js files are hosted remotely on a content delivery network.</li>
+            </ul></p>
+
+            <p>Note that since the css is delivered remotely, only the basic themes are available (you cannot customize the colors as described in <xref ref="online-style-options"/>).</p>
         </subsection>
 
         <subsection xml:id="online-style-options">
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -3636,7 +3636,7 @@ def html(xml, pub_file, stringparams, xmlid_root, file_format, extra_xsl, out_fi
     tmp_dir = common.get_temporary_directory()
 
     pub_vars = common.get_publisher_variable_report(xml, pub_file, stringparams)
-    include_static_files = common.get_publisher_variable(pub_vars, 'portable-html') != "yes"
+    include_static_files = common.get_publisher_variable(pub_vars, 'b-cdn-resources') != "true"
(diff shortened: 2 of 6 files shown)


rbeezer:
Comments coming next from Claude.  I would just do (1) if that was all there was.  I think (2) needs your touch.

What to do about (3)?  Is it documented that this is a problem?  Should we fail more gracefully?  Maybe we should test against something else (not the sample article), since I don't think Claude's review ever did a proper build.  Maybe we should inline those markers?  Just for a portable/single-page build?

rbeezer:
Thanks, Oscar — the `resources/@host` consolidation into `$b-cdn-resources` is clean, and I confirmed the feature works: `host="cdn"` produces the CDN prefix, the `.min.css` theme, and skips the local static files, matching `portable="yes"`. The schema regenerates faithfully too. A few things before merge:

1. **The Guide won't build — malformed XML.** The new `<subsection xml:id="online-resource-host-options">` is missing its closing `</subsection>`, so it runs straight into `online-style-options` and `publication-file.xml` no longer parses. Just add the closing tag.

2. **Reworded portable paragraph.** It now says the chunking variable "will be set to 0" and svgs "will be embedded" — but `portable`/`b-cdn-resources` doesn't affect chunking or svg embedding (only the CDN prefix, the minified theme file, and static-file inclusion). The previous "useful in conjunction with setting the chunking variable to 0" was accurate; I'd keep that sense.

3. **CDN mode + custom `<ol>` markers.** A custom ordered-list marker generates a per-document `ol-markers.css`, written via `exsl:document` to `$html.css.dir` — which in cdn mode is the remote `https://…` URL, so the build fails trying to write there. The sample article (which has such lists) can't build with `host="cdn"`. This pre-dates your PR (`portable="yes"` hits it identically), but since this PR promotes cdn as a first-class option it's worth handling: a per-document file like `ol-markers.css` can't live on the shared CDN, so it needs to stay local (or be inlined) even in cdn mode — or the limitation should be noted in the docs.

Only (1) is a hard blocker; the core change is in good shape.

oscarlevin:
Fixed issue 1 (my bad, I should be more careful).  Force pushed the fix (and rebased on master).

Issue 2 is a Claude mistake: see line 2956 of publisher-variables.xsl, where html-chunk-level is set: if $b-portable-html" we set the chunk level to 0.

Issue 3: as Claude says, this is an existing limitation that I thought was going to be resolved soon by another method (I've thought that for a year).  We should chat with @ascholerChemeketa  about it, so I suggest we not let this block now.

rbeezer:
Two follow-ups after a closer look.

**On point (2) — you're right, that was my mistake.** I claimed the reworded portable paragraph was inaccurate about chunking and SVG embedding, but `@portable` does exactly what the new text says. As you noted, `publisher-variables.xsl` sets `html-chunk-level` to `0` when `$b-portable-html` is true; and `pretext-html.xsl` (the `$b-portable-html and (latex-image|pf:prefigure…)` branch, ~line 6294) routes generated SVGs through `mode="svg-embedded"` so they're inlined. My error was reasoning about `$b-cdn-resources` — which only swaps in the CDN prefix and the minified theme — and carrying that over to `@portable`, which additionally forces chunk-0 and SVG embedding. The paragraph is accurate; withdraw the point.

**A schema blocker worth catching before merge.** The publication schema already defines a grammar pattern named `Resources` — it's the reveal.js `resources` element (local/cdn hosting of slideshow assets), referenced from `revealjs`. This PR adds a *second* `Resources =` definition for the new `html/resources`, so the generated `publication-schema.rng` ends up with two `<define name="Resources">`. RELAX-NG doesn't allow that without `combine`, and jing rejects the whole grammar:

```
publication-schema.rng: error: multiple definitions of "Resources" without "combine" attribute
```

Because it's a grammar-load failure, *every* publication file fails to validate against the schema — not just ones using `resources`.

Since your new pattern is identical to the existing one (`element resources { attribute host { "local" | "cdn" } }`), the simplest fix is to just recycle it: drop the new `Resources =` definition and keep the `Resources?` reference you added to the `Html` block — one definition, referenced from both `Html` and `revealjs`. I confirmed the schema then regenerates, jing loads it, and a `host="cdn"` publication validates. (If you'd rather keep the two contexts independent in case they diverge later, giving the new one a distinct name like `HTMLResources` works too — but recycling is cleaner while they're the same.)

rbeezer:
I see the "old" `Resources` pattern on current `master` and see a new one being added here.

*  Can you correct the duplication?
*  While you are at it, remove the generated *.rnc, *.rng, which will (a) (generally) increase the chance of merge conflicts, and (b) (here) get Claude off-track and not looking at `pretext.xml`?  Keep teh schema commit last so I can pile on.
*  And while you are rebasing, can you make the backticks in teh commit messages into double-quotes, as discussed at Drop-In?  Thanks.

oscarlevin:
Force pushed requested change.

rbeezer:
Merged as-is.  ThaNks for sticking with this one, and for the minor edits.  Can you record (3) as an issue perhaps?  Do we want to "inline" the ol-markers CSS for a single-page build?

PR #2990, opened 2026-07-02
HTML - use h1 for page specific titles

ascholerChemeketa:
As discussed (sort of) here:
https://groups.google.com/g/pretext-a11y/c/ylSiiOsbfyI

The normal HTML was an easy fix thanks to the well disciplined structured chunking/heading-level logic.

Some weird bits around single page documents, watermarks, and epub. Added a few switches to help with those.

Also, the root ToC has no title in the page body. Which is inconsistent from a user point of view. Sighted users see what looks like a table of contents and figure it out. I guess others would hear the list of links and guess what is going on. But this adds a heading with the translated "Contents" string to that page, then hides it from sighted users.

Tested:
- Across many pages of SA with header visualizing accessibility tool, including special cases like frontmatter
- On hello-world with/wo frontmatter and sections and various levels of chunking
- With revealjs - no changes
- epub - maintains h1 for initial title and h2 for later pages (as appears to be correct for an epub), fixes ToC title to be h2


rbeezer:
Nice — I built and inspected both the HTML and the EPUB. The core structure is right: in HTML every content page gets one `<h1>` (its own title) with clean `<h2>` nesting, and in EPUB the cover is `<h1>` with content pages at `<h2>`. Three things to fix:

1. **The ToC heading class is misspelled, so it isn't hidden.** The XSL emits `class="heading ptx-toc-headling"` (extra "l"), but the CSS `sr-only` rule targets `.ptx-toc-heading`. So the "Contents" heading shows to sighted users (HTML and EPUB), the opposite of the intent. Correcting the class to `ptx-toc-heading` fixes it.

2. **`b-is-single-page` mis-fires on a book with structural parts.** It checks `count($document-root/chapter) = 0`, but with `<part>`s the chapters live under them, so a full multi-page book gets flagged single-page. Confirmed against `examples/sample-book/sample-book-parts.xml`: its root page gets an `<h1>` banner and loses the "Contents" heading, whereas a non-parted book correctly gets a `<div>` banner and the hidden "Contents" `<h1>`. Counting parts' chapters too fixes it: `count($document-root/chapter | $document-root/part/chapter) = 0` (or `not($document-root//chapter)`).

3. **The "Contents" section-heading hardcodes `<h1>`, ignoring `chunk-heading-level`.** In EPUB (`chunk-heading-level=2`) it comes out as `<h1>` on a nested article's root page while that article's own sections are `<h2>` — e.g. in the sample-article EPUB, `derivatives.xhtml` has `<h1>Contents</h1>` while `derivatives-9.xhtml` is `<h2>`. So "h2 for later pages" isn't fully held. Use `$chunk-heading-level` for that heading (the same `html-heading` pattern you already use for the frontmatter title) instead of a literal `<h1>`.

Minor: comment typo "sited" → "sighted"; and `test="$b-is-single-page = false()"` reads a bit cleaner as `not($b-is-single-page)`.

Otherwise this is a solid accessibility win.

ascholerChemeketa:
Force pushed with Claude's fixes.

TIL nested articles are a thing.

rbeezer:
Thanks for the fixes — I rebuilt the HTML, the EPUB, and a parted book, and all three earlier points check out: the ToC heading class is now `ptx-toc-heading` (so it's actually hidden), `count($document-root//chapter)` handles the parted-book case (its root page gets a `<div>` banner + the "Contents" heading like any book), and the Contents heading now follows `chunk-heading-level`, so the EPUB's `derivatives.xhtml` is `<h2>` rather than `<h1>`. Nice.

One thing the new scheme surfaces, though: now that the banner is a `<div>`, pages without a page-specific title have no `<h1>` at all. Frontmatter gets one (your new commit), but **backmatter doesn't** — `backmatter.html` now starts at `<h2>` with no `<h1>`, i.e. a heading-level skip (it was `h1=1` on master, via the banner). The standalone media pages (e.g. `airborne-midnight-1.html`) similarly go from a banner `<h1>` to no heading at all.

Backmatter is the concrete one to fix — probably mirror the frontmatter treatment so it gets a page-level `<h1>`. The standalone media/knowl pages are more of a judgment call (a lone audio player arguably doesn't need a heading), but worth a deliberate decision.

Otherwise this is a solid accessibility improvement.

ascholerChemeketa:
Added a section-heading for the backmatter page. Visually hidden to match existing appearance and ToC page.

No heading for standalone pages is intentional.

Surfaced during testing - preexisting issue with backmatter in EPUB output: https://github.com/PreTeXtBook/pretext/issues/3005

rbeezer:
I've had Claude take another look, and it is good, but for a regression.  And this is **exactly** where I recently did a major, major refactor recently that I need to finalize.  So the review coming is really a heads-up, so you can prep.  I will prioritize getting the refactor in.

Also, did not get an automated check on this one.  A `book` must always have at least one `chapter` (otherwise it is not a `book`!).  Maybe safe to remove that from the single-page check.  Claude could check that on a fresh review.

rbeezer:
I rebuilt HTML and EPUB (before/after) to check the whole heading structure. The earlier points are all resolved: the ToC heading class is corrected and actually hidden, the parted-book `b-is-single-page` case is handled, the Contents heading follows the chunk heading level (the EPUB `h1` is now `h2`), and backmatter gets its hidden page heading. Content pages come out with a single semantic `h1` and clean `h2…` nesting. Nice work.

One real issue remains — a heading-level **skip** on the pages produced by the solutions machinery. On the sample article, `solutions-answers` builds as:

- HTML: `h1` (page title, "…All Solutions…") → `h3` ("4 An Interesting Corollary / 4.2 …") → `h4` …, with no `h2` between them.
- EPUB: `h2` (chunk title) → `h4` …, skipping `h3`.

Same story on the other generated pages — `solutions-backmatter`, `solutions-odd-answers`, `solutions-single-exercises` — and on pages with deep inline solutions (`section-advanced-numbering`, `section-interesting-corollary`, `section-worksheet-solutions-testing`). Regular content pages are clean.

The cause: the solutions page's division-grouping headings are seeded at an absolute document-depth level (a subsection → `h3`) rather than relative to the page's new baseline. On master it only lined up by luck — the masthead consumed the `h1`, so the page title sat at `h2`, one step above those `h3` groupings. Promoting the title to `h1` (the right move) opens the gap.

The wrinkle on fixing it: the seam this lands on is the shared solutions/exercise heading driver, and there's in-progress work reorganizing exactly that machinery across all the output formats. That refactor needs to go in first — the fix belongs in the converged driver, and patching it here now would just have to be redone afterward. So this is a good moment to survey the issue, but I'd hold rather than finalize: please wait for an all-clear once that other work lands, and plan on a rebase at that point.

Minor while here: the CSS comment added for the backmatter heading reads "visualy" (→ "visually"). And the standalone media/interactive pages dropping to no heading is noted as intentional — that's fine.

rbeezer:
OK, I have merged #3010.  I hope that does not complicate things for you too much.

rbeezer:
Now that the exercise/solutions driver consolidation is merged (#3010), this is a good time to pick this back up — the fix lives in one place instead of three.

The skip is at the seam between this PR's new page-title level and the solutions driver's grouping level:

- This PR makes the `solutions` page title render at `$chunk-heading-level` (so `h1` in HTML), and the page body is handed that same `$chunk-heading-level` as its base heading level.
- The solutions driver then seeds its first division-grouping heading at `$next-heading-level` — and for a specialized division in an unstructured parent that's **base + 2** (`pretext-common.xsl`, `mode="solutions"`, the `$next-heading-level` variable, ~L5902–5911). So the first grouping comes out at `h3` under the `h1` title, skipping `h2`.

The invariant to restore is: **the first grouping heading sits one level below the page title**, i.e. `$chunk-heading-level + 1`. The `+2` was calibrated for the old scheme, where the page title sat one level deeper than the chunk baseline; promoting the title to `$chunk-heading-level` removes that extra level, so the chunk-root solutions page needs one fewer. Reconciling it there (rather than at the title) keeps the offset correct for the nested/non-root cases, and because the driver is now single-sourced it fixes HTML, EPUB, LaTeX, and XSL-FO together.

ascholerChemeketa:
Claude seemed to want a fix in -common. But changing the logic there affected pdf-fo. So I implemented a fix in -html. Epub and HTML both appear to have correct heading levels for the generated solutions pages.

The multiple special cases around solutions pages don't feel great, but I don't want to much around with -common.

rbeezer:
OK, more back-and-forth with Claude.  The  comment https://github.com/PreTeXtBook/pretext/pull/2990#issuecomment-4907742515 turned out to be wrong.  I think I have a plan to knock this one out, so you will be clear.  So much work to just get the hN right....

rbeezer:
Done!  Merged with an extra (short) commit from Claude, and CSS updates.  Claude and I will pick up pre-existing skips in hN runs.  You are free, with a clear conscience.  ;-) Enjoy.

rbeezer:
Correction to [my earlier note](https://github.com/PreTeXtBook/pretext/pull/2990#issuecomment-4907742515): I pointed at `$next-heading-level` (the `+2` in the `mode="solutions"` driver) as the culprit. I implemented that and it changed nothing — that `is-specialized-division and not-child-of-structured` branch isn't even taken for these pages, so the suspicion was wrong. Apologies for the misdirection.

The actual cause: an inline `<solutions>` is rendered *as a division*, so it passes through the chunk-content dispatch that adds `$heading-level + 1`, and then the `solutions` modal templates add their own increment — a double count. The current fix bypasses that increment, but the `and $heading-level = $chunk-heading-level` clause limits the bypass to the standalone solutions page, so an inline `<solutions>` still double-counts and skips a level.

Relaxing the condition to just `self::solutions` fixes it. Measured on the sample article (pages with an HTML heading-level skip): **6 → 4**. Both inline-solutions pages go clean (`worksheet-solutions-testing`, `advanced-numbering`: the "Solutions X.Y" heading → grouping is `h2 → h3` now), the standalone pages stay fixed, and no new skips appear. It also lines up with the original intent — the comment there already says "the modal solutions templates increment … so we by-pass the increment here."

The 4 that remain are unrelated to this PR:
- `interactive-authored` ("Instructions." headings) and `side-by-side-gallery` (panel titles) skip on `master` too — pre-existing.
- The proof-in-solution skip (`further-reading`, `interesting-corollary`) is a separate, pre-existing quirk: a born-hidden solution knowl occupies a heading level but renders its label as a knowl `<summary>`, not an `<h>`, so its proof lands two levels below the exercise. It's masked on `master` by h6-clamping; the level shift here just exposes it.

I'll add a commit with the one-line `self::solutions` change and carry it to production. The unrelated skips we'll pick up separately — nothing needed from you on those.

PR #2992, opened 2026-07-03
Update PreFigure schema and sample diagrams

davidaustinm:
The first commit here provides the latest version of the PreFigure schema while the latter commits amend the PreTeXt sample documents so that all enclosed annotated PreFigure diagrams successfully validate against the schema.

There is no need to update the diagrams themselves since the amended PreFigure source still compiles to the same SVGs and annotations.

### DIFF
--- a/examples/epub/epub-sampler.xml
+++ b/examples/epub/epub-sampler.xml
@@ -247,14 +247,11 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                                     </point>
                                 </coordinates>
 
-                                <annotations>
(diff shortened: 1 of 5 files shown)


rbeezer:
Merged as-is - thanks for the updates.

Rebased, and edited some of the commit messages.

You have a trailing space at the definition of `TactileStrokeAttributes`.  I'll let you take care of your own hygiene.  ;-)

davidaustinm:
Thanks for letting me know about that now-deleted whitespace.  This evidence notwithstanding, I've become obsessive about this issue!

PR #2993, opened 2026-07-03
pdf-fo conversion: load SVG 1.1 compliant PreFigure images

davidaustinm:
Changes to PreFigure mean that when PreFigure assets are generated within the PreTeXt environment, an SVG 1.1 compliant version of each image is created in addition to the usual ones.  If the `prefigure` element has `label="foo"`, then the new SVG 1.1 compliant image is `foo-11.svg`.

This PR includes a number of commits, most of which simply add the SVG 1.1 compliant PreFigure diagrams into the appropriate generated assets directory of the sample documents.

The second commit `dcd81f2ec84245ea2056df47633f5a693bd0d9a1` is the significant one as it modifies the `pretext-fo.xsl` stylesheet to load 1.1 compliant PreFigure images.  

The function `_downgrade_svg2_for_batik` inside pretext.py does not properly handle arrowheads in PreFigure diagrams so this change sidesteps that issue.  That function is still needed, however, for two MOM problems that require SVG 1.1 remediation.

### DIFF
--- a/examples/epub/gen/prefigure/prefigure-tangent-11.svg
+++ b/examples/epub/gen/prefigure/prefigure-tangent-11.svg
@@ -1,0 +1,200 @@
+<svg xmlns="http://www.w3.org/2000/svg" id="prefigure-tangent-figure" width="310" height="310" viewBox="0 0 310 310">
+  <defs>
+    <clipPath id="prefigure-tangent-__clipPath-0">
+      <rect x="5.0" y="5.0" width="300.0" height="300.0"/>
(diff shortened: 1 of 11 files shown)


rbeezer:
Thanks very much for this.  Long discussion with my assistant, who still wants to try to repair these files in the `pretext/pretext` script.  We can do better.  See bb7e3517afe268ef5268eccd68a0f5270d5cd722.

Merged as-is, but I messed with commit messages, and rotated actual code to go first in the list.

PR #2994, opened 2026-07-03
CSS: Salem - decrease nested knowl background saturation

ascholerChemeketa:
Minor adjustment to nested knowl colors that only affects Salem. Some main colors end up pretty garish when lightened without dropping saturation.

@oscarlevin do you want to use the same logic for your theme family? If so I can adjust the default alt saturation to be -10%.


oscarlevin:
I wasn't able to find an example of where this shows up, but code looks good.  Merged as is.

PR #2995, opened 2026-07-03
Pretext based template for Runestone update

ascholerChemeketa:
A few minor adjustments to support the Runestone template page.

The harden commit will prevent malformed JS if a RS page fails to fill in some of the eBookConfig. It is a stopgap - I still think we want to simplify what PTX is responsible for there but I want to finish what is in process before making any significant changes on that front.

CSS changes affect all pages, but should have no visual impact on PTX pages.

### DIFF
--- a/css/components/_rs-template.scss
+++ b/css/components/_rs-template.scss
@@ -1,6 +1,13 @@
+@use 'components/helpers/buttons-default' as buttons;
+
 // styles for the runestone template page
 body.ptx-runestone-template {
     .ptx-main > .ptx-content {
         max-width: unset;
+    }
+
+    /* expose a basic button class for use in the Runestone template */
+    .ptx-button {
+        @include buttons.ptx-button;
     }
 }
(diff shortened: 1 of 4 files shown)


rbeezer:
Claude likes it!  Posting the review, just so you can bask in the reflected glow. :-)  Will merge next.

rbeezer:
Looks good — I built it to check. The `eBookConfig` defaults do the job, and nice call using the string `'false'` rather than `default(false)`: the boolean would render as `False` and break the JS, whereas `'false'` renders lowercase. `readings`/`activity_info` defaulting to `[]`/`{}` are valid JS too, and it's consistent with the existing `course_attrs|default({})` line.

On the CSS: builds cleanly. Moving the fixed-navbar space reservation from the footer `margin-bottom` to `body.pretext { padding-bottom }` is the right call — it clears all content, not just the footer. I confirmed removing `$nav-height` from `_footer.scss` is safe (navbar defines its own, and no target `@use`s footer *with* it), and the rebuilt dist shows only `padding-bottom: 36px` in / `margin-bottom: 34px` out, both inside the mobile `@media` — so desktop is unchanged. The `.ptx-button` export is scoped to `body.ptx-runestone-template`, so no effect on regular PTX pages, matching your note.

Two tiny things, take or leave:
- Double blank line before the new `@media` comment in `_navbar.scss`.
- `_rs-template.scss` still has no trailing newline (pre-existing).

LGTM.

rbeezer:
Merged with zero changes.  Thanks!

PR #2996, opened 2026-07-03
Schema: panel cleanup, pattern economies, validation-plus repairs

rbeezer:
A follow-on to #2989, reworking how the schema treats `sidebyside` panels and retiring duplicated or unreachable patterns, with the advisory checks strengthened in `pretext-validation-plus.xsl`. Eleven commits, meant to be read in order.

**Panels and the "NoNumber" series.** The five `*NoCaption` patterns are renamed to the `*NoNumber` series, matching `ParagraphsNoNumber` and naming their real purpose: keeping numbered items out of strictly unnumbered locations. The schema continues to reject the obvious constructions (a whole panel, or an immediate child, that is FIGURE-LIKE); a new validation-plus rule catches what the schema cannot — a numbered item nested at any depth, such as `p > ol > li > figure` inside an `assemblage`.

**Exercise panels.** An `exercise` may now be a panel of any `sidebyside` in the base schema, replacing the worksheet-only threading (`WorksheetSideBySide`/`WorksheetBlock`/`WorksheetPage`) from #2989 with a validation-plus rule restricting the arrangement to a `worksheet` or `handout`. A `task` is *not* a panel: task-as-panel dated to the dev-schema worksheet era, and no example document ever used it.

**Economies.** `PrintoutExercise`/`PrintoutTask` had become character-identical to `Exercise`/`Task` (both now carry `@workspace`) and are removed; `PrintoutBlock` collapses to `BlockDivision | Paragraphs`, as the surrounding prose always claimed (this admits `openproblem` and friends in printouts — the one deliberate widening); the unused `MetaDataSubtitle` and unreachable `Interlude` patterns are gone; the dormant `Event` pattern gains a comment explaining it awaits `slideshow` coverage.

**Validation-plus repairs.** The image-description accessibility check tested `description = ''`, which is false for an *absent* element — so images with no description at all were never flagged, while decorative images without one drew a bogus warning. Both directions fixed; on the sample article the corrected check surfaces ~190 genuinely description-less images (a cleanup for a separate PR).

**Verification.** At every commit, `pretext.rnc`/`pretext.rng`/`pretext-dev.rnc`/`pretext-dev.rng` regenerate exactly from `schema/pretext.xml`. The sample article and The Guide were assembled and validated with jing against both schemas before and after: error sets are identical throughout (sample: dev 27, production 334; Guide: dev 0, production 1), with jing's "expected element" suggestion text the only difference. The new advisory rules were exercised on purpose-built torture documents and are silent on the sample article except where intended.

### DIFF
--- a/schema/pretext-dev.rnc
+++ b/schema/pretext-dev.rnc
@@ -33,8 +33,8 @@
                       ImageDescription? &
                       (
                         Slate |
-                        SideBySideNoCaption |
-                        SideBySideGroupNoCaption
+                        SideBySideNoNumber |
+                        SideBySideGroupNoNumber
                       )* &
                       element script { attribute type { text }?, text }* &
                       element source { text }? &
@@ -78,7 +78,7 @@
                         (
                           Paragraph |
                           Tabular |
-                          SideBySideNoCaption |
+                          SideBySideNoNumber |
                           SlateInput |
                           AnyXhtml |
                           element xhtml:button {
@@ -127,7 +127,7 @@
                 (Interactive | Slate)+
             }
 
-            SideBySideNoCaption |= element sidebyside {
+            SideBySideNoNumber |= element sidebyside {
                 SidebySideAttributes,
                 (Interactive | Slate)+
             }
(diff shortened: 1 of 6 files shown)


rbeezer:
@oscarlevin - I think this touches a lot of schema items you have worked on.  Nothing wrong with all that, just time for some clean-up, I think.  Might be worth reading the summary above as we keep firming up the schema - thanks for all your work in that area.

PR #2997, opened 2026-07-03
Validate the contents of a "prefigure-preamble"

### LINKED ISSUE #2979: PreFigure preamble in schema
Mostly for input from @davidaustinm.

PreFigure items will go in the `prefigure-preamble`, with a namespace.  We would like to be able to validate that.  So I expect some subset of PF to be best here?

Claude has a scheme in mind.  Feedback appreciated.  We can run with the proposal, and tighten later - as one approach.

rbeezer:
Resolves #2979.

A `<prefigure-preamble>` (in `docinfo`) holds PreFigure diagram *contents* —
graphical and group elements — but without an enclosing `<diagram>`. The published
`pf_schema.rnc` starts at the `diagram` element, so that content can't be reached
through an ordinary `externalRef`, which exposes only a grammar's start pattern. Until
now the preamble was simply not validated (and tripped a `docinfo` "not allowed here"
error).

This adds a second PreFigure adapter, `pf-preamble-adapter.rnc`, that `include`s the
upstream grammar and overrides its `start` to expose exactly the drawing-content
patterns, reusing the upstream definitions verbatim so it stays in sync with PreFigure:

    default namespace = "https://prefigure.org"
    grammar {
        include "pf_schema.rnc" {
            start =
                ( GraphicalElements* & GroupElements* )
        }
    }

`pf:prefigure-preamble` is then added to the `docinfo` configuration via
`external "pf-preamble-adapter.rnc"`. Per @davidaustinm's guidance in #2979,
**definition elements are excluded** as diagram-specific rather than preamble material;
graphical and group elements are allowed.

Notes:
- The adapter is defined in the literate `pretext.xml` (as `pf-adapter.rnc` already is),
  and the derived `pretext.rnc` / `pretext.rng` and the new `pf-preamble-adapter.rnc` /
  `.rng` are regenerated from it.
- `prefigure-preamble` lives in the PreFigure namespace, mirroring `pf:prefigure`, so its
  content is authored unprefixed under the container's `xmlns`.

Verified against the sample article: its existing `<prefigure-preamble>` now validates,
and the content is genuinely checked — a stray `<definition>` in a preamble is rejected,
while graphical and group elements are accepted.

### DIFF
--- a/schema/pf-preamble-adapter.rnc
+++ b/schema/pf-preamble-adapter.rnc
@@ -1,0 +1,9 @@
+
+                default namespace = "https://prefigure.org"
+                grammar {
+                    include "pf_schema.rnc" {
+                        start =
+                            ( GraphicalElements* & GroupElements* )
+                    }
+                }
+            
--- a/schema/pf-preamble-adapter.rng
+++ b/schema/pf-preamble-adapter.rng
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<grammar ns="https://prefigure.org" xmlns="http://relaxng.org/ns/structure/1.0">
+  <include href="pf_schema.rng">
+    <start>
+      <interleave>
+        <zeroOrMore>
+          <ref name="GraphicalElements"/>
+        </zeroOrMore>
+        <zeroOrMore>
+          <ref name="GroupElements"/>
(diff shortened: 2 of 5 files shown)


davidaustinm:
Looks great.  Thank you for doing this!

PR #2998, opened 2026-07-03
Javscript: fix bug with embed parameter

oscarlevin:
With the recent refactor of the readability options, the `setDarkMode` function got isolated in its own module.  Since the function that detects the `?embed` urlParam sets dark mode, this was preventing the embed behavior of hiding navigation from working.

Took the opportunity to also factor out these functions into their own js file in js/src, a small step on the path to removing `pretext_add_on.js`


rbeezer:
Practice is going to be like for CSS, the committer should rebuild `dist/` files, so there are not crazy merge conflicts.  Somehow (via `pretext-cli`???) a `dist\` file got minified.  I'll take care of it, Claude will go on and on about it.  ;-)

rbeezer:
Thanks, Oscar — nice diagnosis. The fix is right: after the readability refactor only `isDarkMode` was exposed on `window`, so the embed handler's bare `setDarkMode(...)` call threw and aborted before it could hide the navigation. Re-exposing `window.setDarkMode` restores it, and moving the embed code into `js/src/pretext-embed.js` is a clean step toward retiring `pretext_add_on.js`.

One note on the bundle, no action needed from you: `js/dist/pretext-core.js` came out minified — the 2092-line diff is a format change, not code (2101 unminified lines → 20 minified), from the single-target `-t` build path rather than the unminified `npm run build` used for committed dist files. Regenerating the committed bundle is the committer's job, and Rob will set it straight when this goes in.

The source change looks good to me.

rbeezer:
Merged, with the minified/unminified swap.  Thanks for the one small step to no more "add_on".  ;-)

PR #3000, opened 2026-07-05
CSS: prevent scrollbar in table with MJ4 content

ascholerChemeketa:
Fix for issue discussed here:
https://groups.google.com/g/pretext-dev/c/fHv9ajZDxM8/m/IxHJ4ogkAgAJ?utm_medium=email&utm_source=footer


rbeezer:
Merged as-is.

BTW, a reference like "Figure 3.2" is not going to always survive.  Maybe indicating the `@label` could be a better practice.

Thanks for the quick action on this one.

PR #3001, opened 2026-07-05
CSS: restore italics to theorem-like .para

ascholerChemeketa:
Fix for issue identified in:
https://groups.google.com/g/pretext-dev/c/84jEB5ZGWxY/m/XXNlIcr0AQAJ

See Theorem 37.9 in SA for an instance


rbeezer:
Merged as-is.  Thanks!

PR #3013, opened 2026-07-08
Script: set executable bit on find_xml_pattern.py

ascholerChemeketa:
Commit staged in windows ate the executable bit of `find_xml_pattern.py`.

New personal record - single bit commit!


rbeezer:
Merged as-is.  Well, what was there to change, anyway?  Congratulations on the new personal best.

PR #3018, opened 2026-07-08
Sample article: add qrcode for interactive-jsx-arc-graph-2

ascholerChemeketa:
It looks like qrcode and url for this interactive didn't get committed.


rbeezer:
Thank-you.  Claude has been stumbling about on that one, but not enough to get my attention.  ;-)  PR will be a quick and easy fix.

rbeezer:
The `interactive` lacked a `@label` so files created had unstable filenames (`-2` suffix), and a preview image was missing.  (PDF would not build.)  So I just fixed that all up, now at 6606cc0857d57f9356f4a7b252e8b2ef20834a64.

Thanks for the push.  ;-)

PR #3019, opened 2026-07-08
Script: add format_xsl description to pdf and latex docstrings

oscarlevin:
While trying to update the CLI to use the new signature for latex and pdf, I had a tough time figuring out what `format_xsl` was supposed to do.  Added it to the doc string for `pdf()`, and expanded the doc string for `latex()` to describe the parameters.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -3902,7 +3902,24 @@ def latex_package(xml, pub_file, stringparams, dest_dir):
 # Instead, this is a conveience for developers who want to compare
 # different versions of this file during development and testing.
 def latex(xml, pub_file, stringparams, extra_xsl, out_file, dest_dir, format_xsl):
-    """Convert XML source to LaTeX in destination directory"""
+    """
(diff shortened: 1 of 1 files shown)


rbeezer:
Right.  An attempt ar reuse.  Not sure `None` for the status quo is such a bright idea.  Almost stopped to change it, but was well along on a merge (and a bit lazy).

Do you think we should pass in someting explicit for "the way it was"?

rbeezer:
What if the parameter were a string, "regular", "classic", "beamer", and the routine held a dictionary of stylesheets, indexed by those strings?

oscarlevin:
I think that is a good idea, and would generalize nicely.  I really don't need the distraction, but I've been thinking about the new more accessible latex slideshow package as a potential target.

rbeezer:
Adjusted, retains your authorship, merged, etc, at 25cdb45a01d5c088f41ffc495d901ab7296ae4fd

Thanks for calling me out on this one.  ;-)

PR #3023, opened 2026-07-09
Script: choose the LaTeX-family stylesheet by format name, not a path

rbeezer:
The Beamer work (#3012) gave `latex()` and `pdf()` a `format_xsl` parameter — a stylesheet *path*, passed as `None` by every caller except Beamer — together with a `beamer()` wrapper whose only job was to compute that path. Working against the new signature made clear (this is what #3019 ran into) that a bare path-or-`None` is hard to interpret at the call site.

This reworks that selection to name the format instead of threading a path:

- `latex()`/`pdf()` take a `latex_format` string (`"latex"` or `"beamer"`) in place of `format_xsl`. Every caller now passes a self-describing name rather than `None` or a filesystem path.
- The format's base stylesheet comes from a small lookup table, local to `latex()` (the only place it's used):
  ```python
  latex_format_stylesheets = {"latex": "pretext-latex.xsl", "beamer": "pretext-beamer.xsl"}
  ```
- The stylesheet selection now reads as one branch per decider, in precedence order: an author's extra XSL (`-X`) overrides everything; then a publisher's `latex/@style` (a styled variant of the regular `latex` format); otherwise the requested format's base stylesheet.
- The `beamer()` wrapper is retired — `-f beamer` is simply `pdf(..., latex_format="beamer")`, and `-f pdf`/`-f latex`/`-f …-plus` pass `"latex"`.

Output is unchanged; this is a call-site/readability refactor (the signatures change, so any external consumer of `latex()`/`pdf()` must pass the new argument). Verified: `-f beamer` builds the 43-slide Beamer PDF, `-f latex` uses `pretext-latex.xsl` (`\documentclass{article}`), and a publisher `latex-style` still resolves to `pretext-latex-<style>.xsl`.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -3901,7 +3901,7 @@ def latex_package(xml, pub_file, stringparams, dest_dir):
 # This is not a build target, there is no such thing as a "latex build."
 # Instead, this is a conveience for developers who want to compare
 # different versions of this file during development and testing.
-def latex(xml, pub_file, stringparams, extra_xsl, out_file, dest_dir, format_xsl):
+def latex(xml, pub_file, stringparams, extra_xsl, out_file, dest_dir, latex_format):
     """Convert XML source to LaTeX in destination directory"""
 
     # to ensure provided stringparams aren't mutated unintentionally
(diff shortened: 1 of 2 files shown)


rbeezer:
@oscarlevin - no more `beamer()`, and a more sensible argument to `pdf()` and `latex()`, currently `latex` or `beamer` as format strings.  Sorry for the run-around, I should be more careful.  Adjusted #3019 coming just now.

PR #3026, opened 2026-07-09
SVG downgrade: extend Batik compatibility fixes for mermaid diagrams

davidaustinm:
## Summary

Apache FOP renders SVG images via Apache Batik, an SVG 1.1 implementation. When a PreTeXt document includes mermaid diagrams (sequence, git, class), the SVGs mermaid generates use several SVG 2 and CSS3 features that cause Batik to abort rendering entirely, producing blank or broken images in PDF output. This PR extends the existing `_downgrade_svg2_for_batik()` function — which already handled `href`→`xlink:href` and `orient="auto-start-reverse"` — with five additional fixes.

## Changes

The commits are organized so each adds one independent fix:

1. **Preamble refactor** — Introduces shared helpers (`parse_style`, `emit_style`, `get_prop`, `tag()`) and compact namespace constants to reduce boilerplate in subsequent fixes. No behavior change.

2. **`context-stroke`/`context-fill` in marker content** — Replaces the previous simple twin-marker approach with a full context-paint resolver. For each marker using `context-stroke` or `context-fill`, a concrete copy is created per referencing element's stroke/fill color. Copies with the same color and orientation are deduplicated. The `orient="auto-start-reverse"` handling (rotate-180 twin) is subsumed here.

3. **Color and CSS property normalization** — Batik rejects `fill="transparent"` (a CSS3 keyword), `hsl()`/`rgb()` color functions, and CSS filter functions such as `drop-shadow()`. These are rewritten to SVG 1.1-compatible values: `none`, hex colors, and `filter:none` respectively.

4. **`dominant-baseline`/`alignment-baseline` on `<text>`** — Batik does not reliably support `dominant-baseline` on `<text>` elements, and `alignment-baseline` is not valid in SVG 1.1. For centering values (`central`, `middle`), the intended visual center is computed from the element's `y` + `dy` (with em→px conversion), a 0.35em alphabetic-baseline offset is added, and the result is written as a plain numeric `y`. All `dy` attributes are then removed from the `<text>` and its `<tspan>` children to avoid Batik's inconsistent dy/tspan handling.

5. **Mermaid git branch label placement** — Mermaid git diagrams render branch labels as `<text><tspan dy="1em">…</tspan></text>` with no `y` on the `<text>` element. Batik ignores a tspan's `dy` when the parent `<text>` has no `y`, placing text at `y=0` in the enclosing group's coordinate space (typically off the top of the diagram). The first tspan's `dy` is promoted to an explicit `y` on the `<text>`.

6. **`<foreignObject>` → SVG `<text>` for mermaid class diagrams** — Mermaid class diagrams render all text (class titles, attributes, methods) as `<foreignObject>` containing XHTML `<div>`/`<span>` elements. Batik cannot render `<foreignObject>`, so all class diagram text was invisible. Each `foreignObject` is replaced by an SVG `<text>` element whose position is derived from the element's `translate()` transform and `width`/`height` attributes. `classTitle` elements are centre-anchored; all others are left-aligned.

## Testing

Tested against the `examples/sample-article` document in this repository, which includes mermaid sequence, git, class, and ELK-layout class diagrams. All four diagrams render correctly in PDF output after this fix.

## Caveats

- The 0.35em centering offset in fix 4 is a heuristic for Latin text at typical font sizes; it may not be pixel-perfect for all fonts.
- The function currently processes all SVGs in the FOP staging directory, including PreFigure SVGs (which are already SVG 1.1-compliant). A future improvement could skip known-clean files.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -4340,6 +4340,8 @@ def _downgrade_svg2_for_batik(directory):
     * dominant-baseline="central/middle" and alignment-baseline on <text>:
       converted to an explicit numeric y coordinate (alphabetic baseline),
       with all dy attributes removed to avoid Batik dy/tspan quirks.
+    * <foreignObject> containing XHTML spans (mermaid class diagrams):
+      replaced by SVG <text> elements with equivalent content and position.
     """
     # TODO (2026-07-03) This function may modify SVG 1.1 files coming
     # from PreFigure that are known to be good for Batik to ingest.
@@ -4350,6 +4352,7 @@ def _downgrade_svg2_for_batik(directory):
 
     SVG           = "http://www.w3.org/2000/svg"
     XLINK         = "http://www.w3.org/1999/xlink"
+    XHTML         = "http://www.w3.org/1999/xhtml"
     ET.register_namespace("xlink", XLINK)
     tag           = lambda t: f"{{{SVG}}}{t}"
     xhref         = f"{{{XLINK}}}href"
@@ -4358,6 +4361,7 @@ def _downgrade_svg2_for_batik(directory):
     _HSL_RE       = re.compile(r"hsl\(\s*([\d.]+)\s*,\s*([\d.]+)%\s*,\s*([\d.]+)%\s*\)")
     _RGB_RE       = re.compile(r"rgb\(\s*([\d.]+)\s*,\s*([\d.]+)\s*,\s*([\d.]+)\s*\)")
     _FILTER_FN_RE = re.compile(r"(filter\s*:)\s*([^;{}]+)")
+    _TRANS_RE     = re.compile(r"translate\(\s*([-\d.]+)\s*,\s*([-\d.]+)\s*\)")
 
     def parse_style(el):
         d = {}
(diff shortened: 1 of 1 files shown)


davidaustinm:
Writing as a human here:  the first two commits handle the issues with MOM SVGs while the others fix a variety of problems with mermaid SVGs.  This result leads to clean output on the sample article.

rbeezer:
Review coming, perhaps your assistant can incorporate the suggestion (or refute it) and do a force-push with any changes.

rbeezer:
Nice work extending this to mermaid — I built the sample article to `pdf-fo` (`-c doc -f pdf-fo`) and it comes through cleanly: exit 0, no Batik abort, only the benign "Alternate text missing" warnings. The Mermaid Class Diagram (Figure 10.20) renders correctly — all class boxes, attributes, methods, fills, and the hollow inheritance arrowheads are present. The git and sequence SVGs are fully cleaned, and the prefigure arrow diagrams still render, so the reworked context-paint marker cloner is a clean generalization of the original auto-start-reverse handling rather than a regression. The code reads well and is nicely commented.

One thing I'd suggest tightening before merge — the `<defs>` lookup:

```python
defs = root.find(tag("defs")) or root.find("defs")
...
for m in new_mkrs.values(): defs.append(m)
```

Two concerns, both from the `or`:

- **Latent crash.** If an SVG carries svg2 markers but has no *top-level* `<defs>` (e.g. the markers are nested inside a `<g>`), both `find()` calls return `None` and `defs.append(...)` raises `AttributeError`. Every SVG in the current test set happens to keep its markers in a non-empty top-level `<defs>`, so this branch is never exercised by the build — but it's a real gap for SVGs from other tools.
- **lxml truthiness.** `root.find(...) or ...` truth-tests an Element, and an lxml element with no children is falsy (with a `FutureWarning` that this will change). So a present-but-empty top-level `<defs>` would wrongly fall through to the second `find()`.

The standard idiom sidesteps both:

```python
defs = root.find(tag("defs"))
if defs is None:
    defs = ET.SubElement(root, tag("defs"))
```

A few minor observations, none blocking:

1. The converter skips empty `<foreignObject>`s (`if not text: continue`), leaving them in the output. That's harmless — Batik ignores them and no label text is lost — but a one-line comment noting it's intentional would help the next reader.
2. `hsl()`/`rgb()` are rewritten inside `<style>` blocks but not in presentation attributes (`fill="hsl(...)"`). Not hit by current mermaid output, just an inconsistency if such attributes ever appear.
3. `_em_px` ignores a non-`em` (px) `dy`.

davidaustinm:
Thanks for the thorough review. All four points have been addressed in a follow-up commit:

- **`<defs>` lookup (blocking)**: replaced `root.find(...) or root.find(...)` with an explicit `None` check and `ET.SubElement` fallback, fixing both the lxml falsy-element pitfall and the potential `AttributeError` when no top-level `<defs>` exists.
- **`hsl()`/`rgb()` in presentation attributes**: the colour-function rewriting now also applies to `fill` and `stroke` presentation attributes, not only `<style>` blocks.
- **`_em_px` px handling**: the helper now returns the numeric pixel value directly for `px`-unit `dy` values, rather than falling through to 0.
- **Empty `<foreignObject>` comment**: added a short inline comment noting that empty `<foreignObject>` elements are intentionally left in place since Batik ignores them.

rbeezer:
Thanks for the fixes, David. I've folded the four review-feedback changes back into the commits that introduced each construct (`<defs>` handling, colour normalization, `_em_px`, the foreignObject conversion) rather than carrying them as a separate commit, and retitled all six under an `XSL-FO:` topic. The resulting tree is byte-identical to your last push — only the commit organization changed. I've force-pushed, so you'll want to reset your local branch to match before continuing.

rbeezer:
One more history tidy: PreTeXt commit messages are a single subject line plus the trailer, with no body paragraphs, so I've trimmed the six commit bodies down to their subjects. Content is unchanged. Force-pushed again — reset your local branch once more before continuing.

PR #3031, opened 2026-07-11
Asymptote: use offscreen rendering for pdf, png, and eps

sean-fitzpatrick:
The flag `-iconify` is now deprecated in recent versions of Asymptote.
Most users who are building Asymptote locally have an older version that ships with their TeX distribution, so the current command still works.
However, if we want to support Asymptote builds in GitHub codespaces (in particular, builds to PDF), this change is needed.

Notes:
- Testing suggests that this change has no effect on Windows and Linux for Asymptote circa 2.87
- To take advantage of offscreen rendering using Vulkan, Asymptote 3.13 or later is needed
- Asymptote 3.12 has a bug that causes OOM errors. There still seems to be a memory leak on Windows with 3.13, but I don't think there are many projects with more than 200 3D Asymptote images

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -475,9 +475,9 @@ def asymptote_conversion(
                 # perhaps an older version is being employed
                 asy_cli = asy_executable_cmd + ["-f", outform, "-noV"]
                 if outform in ["pdf", "eps"]:
-                    asy_cli += ["-noprc", "-iconify", "-tex", "xelatex", "-batchMask"]
+                    asy_cli += ["-noprc", "-offscreen", "-tex", "xelatex", "-batchMask"]
                 elif outform in ["svg", "png"]:
-                    asy_cli += ["-render=4", "-tex", "xelatex", "-iconify"]
+                    asy_cli += ["-render=4", "-tex", "xelatex", "-offscreen"]
             if method == "server":
                 alberta = "http://asymptote.ualberta.ca:10007?f={}".format(outform)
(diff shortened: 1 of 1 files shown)


rbeezer:
Reviewed — this is a clean, low-risk change. 👍

**Scope.** Only the `method == "local"` command construction changes; the Alberta server path is untouched. Both flag paths (`pdf`/`eps` and `svg`/`png`) are updated consistently. +2/−2, whitespace-clean.

**Compatibility on older Asymptote.** The flags aren't version-gated, so `-offscreen` reaches every local Asymptote regardless of the version detected. I confirmed the deprecated→new swap is graceful on an older TeX-distribution install: on Asymptote 2.85, `-offscreen` isn't even listed in `asy -h`, yet it's accepted silently — exit 0, and a PDF identical to the `-iconify` run. That matches your "no effect on ... 2.87" testing, so nothing breaks for users on the older bundled versions while enabling the headless/codespaces case.

**One optional follow-up (not for this PR).** The function already captures `asyversion` but never uses it, and the adjacent comment notes a real version check is wanted. Now that the two flags differ in intent, gating `-offscreen` (new) vs. `-iconify` (old) would be the robust end state. Not needed here, since `-offscreen` is a harmless no-op on versions that predate it.

*Claude Opus 4.8, acting as a review/coding assistant for Rob Beezer*

rbeezer:
Merged, after a rebase and a commit message edit.  Sorry this took a while, several big pushes have been underway here for a few days.

Thanks for working up this one!

PR #3036, opened 2026-07-14
Script: include PreFigure schema files for server validation

mitchkeller:
Realized after my meeting that it was only one more schema file that I needed to bundle. Not sure if this is the ultimate right destination, but with these changes, validation via server is working.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -5399,7 +5399,10 @@ def _validate_server(xml_source, out_file, dest_dir):
 
     # fresh schema from the PreTeXt distribution
     schema_filename = os.path.join(common.get_ptx_path(), "schema", "pretext.rng")
-    files = {'source': open(zip_filename,'rb'), 'rng': open(schema_filename,'rb')}
+    pf_adapter_filename = os.path.join(common.get_ptx_path(), "schema", "pf-adapter.rng")
+    pf_schema_filename = os.path.join(common.get_ptx_path(), "schema", "pf_schema.rng")
+    pf_preable_adapter_filename = os.path.join(common.get_ptx_path(), "schema", "pf-preamble-adapter.rng")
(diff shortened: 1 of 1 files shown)


rbeezer:
Thanks, Mitch!  Back to this now and will update on progress.

rbeezer:
Thanks, Mitch — nice to have server validation working for PreFigure. I've pushed two small follow-ons onto this branch:

- **Fix a typo and format the dict** — `pf_preable_adapter_filename` → `pf_preamble_adapter_filename` (harmless, since the path and the multipart key were both already correct), and broke the now-long `files = {…}` into a readable multi-line dict.
- **Close the schema files after the POST** — the entries in `files` open handles that are never closed (`requests` reads them but doesn't close them). This was already true of the original `source` and `rng` files, so it's a pre-existing issue, not something your change introduced — your PR just made it a natural moment to tidy it up. They're now all released in a `finally` after the request.

I left the bigger design point alone for now: both sides hardcode the PreFigure schema list — the client names each file here, and the server expects exactly these field names — so the two can drift apart if the schema's includes change. Worth a future look (bundle the whole `schema/` directory, or derive the list from the `<include>`s), but not needed to land this.

rbeezer:
Thanks, Mitch!  Server validation is working again for me.  My fault, I had not thought through that adding new files would impact this.  Merged with the two follow-on commits from my assistant, and a rebase onto current master.

Now I have some work to do to - incorporating this output into the grander report format I have worked up already for local processing.

Thanks for the PR and for spending more than 3 minutes on this!  Back at you on other channels in a bit.

mitchkeller:
I agree that there's likely a better approach of identifying the schema files to include or just sending the whole directory rather than the current hard-coding method, but hopefully this lets you do the dev work to incorporate into the validation flow and then that design decision can be made later.

Might be worth some testing to see if zipping up the source is necessary or helpful when sending the assembled version. Keeping the size of the POST down seems positive, but sysadmins will probably be wary of services that accept zip files. (Reading the source of `pretext.py` pretty much tells someone what they could try to send to the service, so the server side can't assume that what it receives is coming from the `pretext` script when it comes to security matters.) If all the files in the POST are just XML files, it feels like some checks could be done when processing the POST to make sure that they really are XML files and that malicious files aren't dropped into a temp directory.

rbeezer:
Right, just having my assistant draft a note about zipping the source, but I guess it applies tot he schema files, too.

PR #3040, opened 2026-07-15
SCORM: improve js so blackboard records scores

oscarlevin:
With the help of @davidaustinm and 16 rounds of testing, we now have SCORM javascript that correctly records student work in Blackbaord (and Canvas still works as expected).  There were multiple issues, all tracked in comments inside the file.  Squashed all the false steps down into a single commit.

The jsbuilder will need to be run of course.

### DIFF
--- a/js/ptx_scorm_events.js
+++ b/js/ptx_scorm_events.js
@@ -61,14 +61,35 @@
 (function () {
     'use strict';
 
+    // ── Debug logging (initialised first so it is available during API discovery)
+    //
+    // Enable by any of these methods before the page loads:
+    //   • Add  ?ptxscormdebug  to the page URL
+    //   • Run  localStorage.setItem('ptxscormdebug', '1')  in the console, then reload
+    //   • Set  window.ptxScormDebug = true  before this script runs
+    //
+    // When enabled, every Get/Set/Commit/Terminate call — and every step of the
+    // SCORM API discovery — is logged with its result and LMS error code.
+    var _debug = !!(
+        window.ptxScormDebug ||
+        (window.location.search.indexOf('ptxscormdebug') !== -1) ||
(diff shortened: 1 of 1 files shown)


rbeezer:
Really nice work — this is thorough, and the inline documentation makes the reasoning easy to follow. The root-cause diagnoses are convincing: the `suspend_data` SPM overflow silently sinking Blackboard's server-side persist; JSON responses tripping the `performance` type's 406 → `other`; the trailing-`Z` timestamp 406 on Canvas → local time; and `completed` mid-page finalizing the attempt. The deeper, cross-origin-safe API discovery and the opt-in debug logging are clean touches — and the double-unload is nicely guarded, since `Terminate()` clearing `_initialized` keeps `beforeunload` and `pagehide` from double-terminating.

A few small, non-blocking suggestions:

- **The `Terminate()` docstring.** It reads *"the signal most LMSes (Blackboard in particular) need before they will post the score to the gradebook,"* which is the opposite of what Sections 14–15 conclude — the score is committed while the session stays open, a *mid-page* `Terminate()` voids the attempt, and terminating is correct only at page exit. Since the comments here are effectively the spec, you might reword that one docstring to match, e.g. "Called only at page exit; terminating mid-page voids the Blackboard attempt."

- **The injected Submit button.** `addSubmitButton()`/`updateScoreDisplay()` hardcode English text ("Submit Assignment", "Your current score", "You may now close this window") and inline colors. Perfectly reasonable for LMS-only injected UI, but if SCORM ever ships in a non-English or themed/dark build, you might eventually route these through localization and PreTeXt styling — a someday follow-up, not for this PR.

Everything else looks good to me.

rbeezer:
Thanks, merged with code as-is.  Rebased on master, and edited the commit message.

16 rounds of testing!  Phew.

PR #3044, opened 2026-07-16
Add CSS to STACK JS questions rendering

geoo89:
This makes inputs and feedback in interactive STACK questions stand out a bit more and look prettier.

This PR builds on top of #3043 (only the last two commits are new), but I've made this into a separate PR as I'm not sure where exactly to ship the CSS. Maybe it should be integrated into the whole theming framework, but this seems like a bigger task, so maybe STACK-specific (theme agnostic) CSS is a decent temporary solution?


rbeezer:
Thanks — the two new commits are small and clean, and I've checked the path fix works. `$html.js.root` resolves to `_static/pretext/js`, `copy_html_js()` copies the whole `js/` tree there, and `js/dist/pretext-stack/` holds no `stack.css`, so the old link would have 404'd. A build of the `stack/minimal` example emits `href="_static/pretext/js/pretext-stack/stack.css"` with the file present at that path. The stylesheet is also scoped throughout under `.que.stack` and the link is guarded on `$b-has-stack`, both of which I appreciate.

I'm holding this one until #3043 lands. As it stands the branch sits on the older #3043, so its diff still carries `js/dist/pretext-stack/stackjsvle.js` — the file you took out of #3043 at my request — and it predates the radio-group, `vle_get_element` and `download()` fixes you pushed on the 28th.

**What would help:** a force-push carrying just the two new commits — `STACK: add CSS for matrix brackets and feedback/correctness boxes` and `STACK: fix stack.css link pointing at the JS dist path`. Rebase onto the current #3043 head if you'd like this looked at sooner, or wait until #3043 merges and rebase onto master. Either way this PR then becomes a 224-line, two-file change and is easy to take.

On where the CSS should live: the stylesheet hardcodes around thirty colour literals, and six of the shipped themes set `--body-text-color: #f2f2f2`, meaning dark backgrounds. On those themes the boxes render as black text on white inside a dark page.

On copyright and provenance: `js/pretext-stack/stack.css` carries no copyright line or license statement, and it is adapted from `moodle-qtype_stack` and Moodle core, both GPL v3. The opening comment does name the sources. This is the second file in that directory in that position — `stackjsvle.js` has a copyright line but no license and no README. I have a copyright and licensing pass touching notices across the repository that I'll be pushing later today.

One last thing, unrelated to your changes but it will bite anyone testing this: `examples/stack/minimal` will not build from a clean checkout, because its publication file names a `generated` directory that isn't shipped. It is the same problem you fixed in #3043 for `examples/stack/features/`, with the same remedy — a `.gitkeep` so git tracks the empty directory. It's on master, not in this PR.

rbeezer:
We have worked long and hard to put in a comprehensive system for CSS.  So I am reluctant to add some non-conforming CSS now on a "temporary" basis.  I think we should take the time **now** to do it right.

@ascholerChemeketa  - you could visually look at the last two commits here, that is the only new part, and offer your opinion on the best approach.  Perhaps when you are back to active development, you would be able to help the STACK folks chart a way through?

Somehow, I missed the new work on #3043 - sorry.  I'll go there shortly, so take the above AI review in light of that.

ascholerChemeketa:
Where/how:

There is a tradeoff between "extra" CSS files and building everything into the `theme.css` file. We currently do not have the theme builder do any inspection of pub variables or content to decide what to build into the theme for a book. So building stack CSS into the theme means it is cruft in any book that does not use stack (same is true of every other chunk of CSS). But, adding a link to a standalone file means that special handling might be needed in situations like Runestone grader, single file builds, etc...

For the limited amount of CSS here, I think bundling it makes sense.

You could add a `_stack.scss` to `css\components\interactives` and then `@use` that file from `_pretext-web.scss` to include it in the appropriate bundles.


Colors:

Either all of the colors play well with dark and light mode (preferrable), or the entire interactive should be rendered in light mode regardless of the overall page (by forcibly setting all background/foreground colors).

If you want to play well with the existing themes in dark/light mode,  `css\colors\_color-vars` defines a list of colors that are rendered into CSS variables. Something like:

```scss
"body-text-color": #000,
```

Will get rendered into a CSS variable `--body-text-color`

That will be set to the appropriate value in dark/light mode. There are other variables like `--content-background` and `--page-border-color`.

When setting colors for things like `.que.stack .feedback`, you likely need to set a light version and a dark version so you maintain proper contrast in both modes. Use can use:

```css
:root.dark-mode {
  ...
}
```

To set dark specific versions of CSS rules or variables that will be used when in dark mode.


I'm back as of yesterday, so feel free to ping me for follow ups.

PR #3049, opened 2026-07-17
Script: modify server validation to support changes to the service

mitchkeller:
This calls to `validate-dev` endpoint. See comments on the issue on the server repository for more info and plans.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -5436,36 +5436,36 @@ def _jing_server(schema_filename, assembled_source):
     # change this URL (and, if the reply format changes, the parsing below).
     server_url = "https://mathgenealogy.org:9000/validate"
 
-    # The service expects a zip of "source" with a named "mainfile", plus
-    # the schema and its PreFigure companions.  Here the "source" is the
+    # The service expects "assembled_source" and "schema" plus the PreFigure
+    # companions of the schema.  Here the "assembled_source" is the
     # single, already-assembled file, so the locations in the report line
     # up with the deposited assembled source, exactly as a local run does.
     tmp_dir = common.get_temporary_directory()
-    mainfile = os.path.basename(assembled_source)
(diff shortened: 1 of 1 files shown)


rbeezer:
Tested the reworked `server` method end-to-end against the live `/validate-dev` endpoint, with a clean file and a deliberately broken one.

- **Clean source** → both report sections empty (`(no messages)` for the `jing`/RELAX-NG schema and for validation-plus). The round-trip succeeds and the consolidated report is identical in shape to a local run.
- **Broken source** (a disallowed element inside a `<p>`, plus an `<image>` with no description) → the schema section correctly carried the server's `jing` message (`element "bogus" not allowed anywhere …`, `check: schema`), and the validation-plus section independently reported the missing image description. So real schema errors come back through the new string-field POST, and locations still line up with the deposited assembled source.

The functional change — multipart zip → file contents as string form fields, `/validate` → `/validate-dev`, and the renamed keys — behaves correctly for both cases. Two small non-blocking things:

1. Trailing whitespace on the new comment line ending `… as strings in form fields, `.
2. `open(path, "r")` reads in text mode with the platform's default encoding. Since the assembled source and schema can contain non-ASCII, an explicit `encoding="utf-8"` would avoid a locale-dependent decode. (Minor: a space after the comma for style.)

rbeezer:
Do you want me to merge this now?  Or is it just for testing?

mitchkeller:
Force pushed with the small things noted. I plan to come to drop-in today, so hopefully this can be merged and we can cutover then.

mitchkeller:
To make this ready to cutover, I force-pushed again with change of the endpoint from `validate-dev` to `validate`. Cutover will then involve merging this and me renaming `validate-dev` to `validate` in source and restarting the service.

rbeezer:
Merged as-is.  Thanks.

PR #3052, opened 2026-07-20
Guide: document landing page

bnmnetp:
Any errors or omissions are completely my own work.  No assistant was used in creating this PR. 😄

### DIFF
--- a/doc/guide/publisher/online-html.xml
+++ b/doc/guide/publisher/online-html.xml
@@ -26,7 +26,7 @@
                 <cline>platypus.mammal-institute.org/aota/html</cline>
             </cd>then most modern web servers will automatically return the page<cd>
                 <cline>platypus.mammal-institute.org/aota/html/index.html</cline>
-            </cd>So you can advertise the shorter version to potential readers.  What is in <c>index.html</c>?  Simple code to <term>redirect</term> to another one of your pages.  Which one?  Any one you like!</p>
+            </cd>So you can advertise the shorter version to potential readers.  What is in <c>index.html</c>?  Simple code to <term>redirect</term> to another one of your pages.  Which one?  Any one you like! <idx>landing page</idx></p>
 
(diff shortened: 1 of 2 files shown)


rbeezer:
@bnmnetp - thanks for this hand-crafted PR.  ;-)  That "Landing Page" section was meant to be about a website with a landing page for a book project!  We have documentation already for the "index page".  So I had my assistant fill out the Landing Page section and try to explain the difference, add index, entries, cross-reference, etc.  Sorry for the further confusion.  Hopefull it is all sorted over on #3060.

My assistant wrote a fuller explanation - coming next.

rbeezer:
@bnmnetp thanks for this, and for flagging that "Landing Page" was a bare stub worth filling in — you've put your finger on a real source of confusion.

The two things named here are genuinely distinct. The **HTML index page** (the `index-page` publication switch) is the `index.html` file PreTeXt builds *inside* your HTML output, so a directory URL redirects to a sensible first page of the book; it's part of one output, and it's already documented. A **landing page**, in the sense the Production chapter intends, is the home page of a separate little *website devoted to the whole project* — the front door you advertise, from which a reader can reach the online HTML, download the PDF, find errata, and so on. It sits above all of your outputs rather than inside one of them.

This PR wires "landing page" to the index-page docs, which merges the two ideas. I understand why: Runestone hosts a book directly and doesn't really have a spot for a separate project website, so from that vantage the index page *is* the landing page. But for authors who host their own site, keeping them separate matters.

So rather than merge this, I've filled out the Production-chapter stub to describe the project-website landing page proper — with a note on PreTeXt's own tooling (`site` folder + `pretext deploy` for multi-target sites), a suggestion to use GitHub Pages (handy for Runestone authors, since the book already lives in a GitHub repo), and cross-references in both directions so anyone who lands on the index-page docs looking for "landing page" (or vice versa) is pointed to the right place. That went in as #3063, so I'll close this one in favor of that.

Thanks again — a useful nudge.

PR #3055, opened 2026-07-21
Sample chapter: update output from 'make sample-chapter-representations'

Alex-Jordan:
This is simply the result of navigating to the `examples/webwork/` folder and running `make sample-chapter-representations`. I'm not sure what led to this, but somewhere some changes were made to XSL files or to the pretext python script, and the output from `make sample-chapter-representations` was not kept up. So now as I work on things like #3053, these changes show up in the diff even though they have nothing to do with that work.

I also get changes in the generated images in `sample-chapter/my-generated/webwork/images/`, but I assume those are due to different LaTeX versions making those images, different versions of the tools used to convert PDF to other image files, metadata in the files, etc. I'm OK with just ignoring all that.

### DIFF
--- a/examples/webwork/sample-chapter/my-generated/webwork/pg/Integrating_WeBWorK_into_Textbooks/2-Technical_Examples/2_3-Stress_Tests/2_3_1-PreTeXt_problem_source_with_servergenerated_images.pg
+++ b/examples/webwork/sample-chapter/my-generated/webwork/pg/Integrating_WeBWorK_into_Textbooks/2-Technical_Examples/2_3-Stress_Tests/2_3_1-PreTeXt_problem_source_with_servergenerated_images.pg
@@ -1,31 +1,3 @@
-<?xml version='1.0' encoding='utf-8'?>
-<webwork-reps version="2" webwork2_major_version="2" webwork2_minor_version="20" xml:id="extracted-pretext-source-with-server-images" assembly-id="pretext-source-with-server-images">
-  <static seed="82">
-    <task>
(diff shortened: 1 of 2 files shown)


rbeezer:
We are now using `ptx` as a prefix/identifier fot LaTeX macrtos, like a crude namespacing technique.  So I had to adjust an @xml:id for a problem if I recall right.  (And failed to update the generated stuff.)

rbeezer:
> for LaTeX macros

Moving too fast - also HTML `id` attributes for specific parts of the UI.

rbeezer:
c51bcb9c was the change that caused this

rbeezer:
Merged.  Apologies for the mess.  I added in the image files, those changes belong as well, as part of a complete update.

Thanks for sounding the alarm on this one.

Alex-Jordan:
I missed something here. The repo has the file `examples/webwork/sample-chapter/my-generated/webwork/pg/Integrating_WeBWorK_into_Textbooks/2-Technical_Examples/2_3-Stress_Tests/2_3_1-PTX_problem_source_with_servergenerated_images.pg`, which this update should have updated to `examples/webwork/sample-chapter/my-generated/webwork/pg/Integrating_WeBWorK_into_Textbooks/2-Technical_Examples/2_3-Stress_Tests/2_3_1-PreTeXt_problem_source_with_servergenerated_images.pg`.

Instead, I added the new file without deleting the old one. What's preferred to remove it? I could put it on a standalone commit in an upcoming PR about removing unused junk now that we are beyond PG 2.18?

PR #3056, opened 2026-07-22
Script: extend href→xlink:href downgrade to all SVG elements

davidaustinm:
## Summary

`_downgrade_svg2_for_batik()` currently replaces bare `href` with `xlink:href` only on `<use>` elements. SVG 1.1 requires `xlink:href` on any element that carries an `href` — and several SVG elements besides `<use>` can do so: `<image>`, `<a>`, `<textPath>`, `<pattern>`, `<linearGradient>`, `<radialGradient>`, `<feImage>`, and `<mpath>`, among others. Any of these could appear in SVGs from sources other than PreFigure that are included in a PreTeXt document.

This issue came to light while developing the PreFigure testing framework, where `<image href=…>` appeared in test SVGs and was silently dropped by Batik.

## Change

One line in `_downgrade_svg2_for_batik()`: change `root.iter(tag("use"))` to `root.iter()` so every element in the tree is checked. The guard `el.get(xhref) is None` prevents double-setting on elements that already carry `xlink:href`.

## Testing

Verified against SVGs containing `<image href=…>` elements; Batik renders the image correctly after the fix.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -4741,10 +4741,10 @@ def _downgrade_svg2_for_batik(directory):
         root    = tree.getroot()
         changed = False
 
-        for use in root.iter(tag("use")):
-            href = use.get("href")
-            if href and use.get(xhref) is None:
-                use.set(xhref, href); del use.attrib["href"]; changed = True
+        for el in root.iter():
+            href = el.get("href")
+            if href and el.get(xhref) is None:
+                el.set(xhref, href); del el.attrib["href"]; changed = True
 
         # "transparent" is a CSS3 colour keyword, not SVG 1.1;
         # hsl()/rgb() in presentation attributes are also rewritten here.
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -4626,7 +4626,7 @@ def _downgrade_svg2_for_batik(directory):
     * orient="auto-start-reverse": changed to "auto"; elements that use
       the marker as marker-start get a rotated (180°) copy (id suffix
       "-start").
-    * Plain href on <use>: migrated to xlink:href.
+    * Plain href on any SVG element: migrated to xlink:href.
     * fill="transparent" → fill="none"; hsl()/rgb() colour functions and
       CSS filter functions in <style> blocks rewritten to values Batik
       accepts.
@@ -4742,6 +4742,8 @@ def _downgrade_svg2_for_batik(directory):
         changed = False
 
         for el in root.iter():
+            if not (isinstance(el.tag, str) and el.tag.startswith("{" + SVG + "}")):
+                continue
             href = el.get("href")
             if href and el.get(xhref) is None:
                 el.set(xhref, href); del el.attrib["href"]; changed = True


rbeezer:
Reviewed at `9d1cf6a8`, against the merge-base `a6a89323`.  The underlying fix is right: `<image href=…>` and the other href-bearing SVG elements really are dropped by Batik today, since the old loop only visited `<use>`.  One scope point.

## `root.iter()` reaches into foreign namespaces

The title says "all SVG elements", but `root.iter()` visits every element in the tree regardless of namespace, and `el.get("href")` then matches the plain `href` on non-SVG elements too.  The docstring notes these files can carry `<foreignObject>` with XHTML, and an XHTML anchor uses a plain `href`, not the XLink one.  Running the patched function on such a file rewrites it:

```
<xhtml:a href="https://link.test">   ->   <xhtml:a xlink:href="https://link.test">
```

which breaks the link, since HTML does not read `xlink:href`.  The `foreignObject` handling further down only rescues the mermaid shape — it gathers text from `xhtml:span` descendants and rebuilds an SVG `<text>`, discarding attributes.  A `foreignObject` whose XHTML has an `href` but no span-wrapped text reaches `if not text: continue` and is written out still carrying the rewritten attribute.

In practice the effect is small: the function operates on the staged copy in `tmp_dir`, so the managed SVGs are untouched, and Batik ignores non-mermaid `foreignObject` content, so the mangled link does not reach the PDF.  It is a correctness/scope matter rather than a build break — but it becomes visible the moment this staged tree feeds another consumer, or Batik gains `foreignObject` support.

Scoping the loop to the SVG namespace matches the stated intent and leaves foreign content alone:

```python
for el in root.iter():
    if not (isinstance(el.tag, str) and el.tag.startswith("{" + SVG + "}")):
        continue
    href = el.get("href")
    if href and el.get(xhref) is None:
        el.set(xhref, href); del el.attrib["href"]; changed = True
```

Tested: this still converts `<image>`, `<use>`, `<a>` and the gradient elements, and leaves an XHTML `href` as a plain `href`.  The `isinstance(el.tag, str)` guard also skips comment and processing-instruction nodes (harmless in the current form, but explicit here).

## Docstring

Line 4629 still reads `* Plain href on <use>: migrated to xlink:href.`  With the wider scope it should name any SVG element, not just `<use>`, ideally in this same commit.

The diff is otherwise clean — one file, four lines, no stray whitespace, based on the current master tip.

davidaustinm:
Good catch on the namespace scope — the fix now guards with `isinstance(el.tag, str) and el.tag.startswith("{" + SVG + "}")` so only SVG-namespaced elements are touched. The docstring is updated in the same commit. Thanks for the review.

rbeezer:
The fix is correct.  On a rebuild, the SVG-namespaced elements (`image`, `use`, `a`, the gradients) still migrate to `xlink:href`, and an XHTML `href` inside a `foreignObject` is now left as a plain `href`.  The docstring update reads accurately too.  Thanks for the quick turnaround.

One workflow note for next time, not a request to redo anything here: when a change corrects something introduced earlier in the same PR, we prefer the contributor amend the commit that introduced it and force-push the branch, rather than add a follow-up "fix" commit.  That way each commit's diff is exactly its own topic and the branch reads as if the mistake never happened.

No need to do that on this PR — @rbeezer will meld the two commits himself when integrating, so it lands as a single clean commit rather than coming in as a two-commit merge.

PR #3059, opened 2026-07-22
WeBWorK: end support for 2.16, 2.17, and 2.18

Alex-Jordan:
As announced back in January on -announce, we are ending support for WeBWorK/PG at versions 2.18 and earlier. This does that.

I limited this to simply removing the things that are needed to support these old versions in the javascript and python. And of course, updating things in the guide. So what I did *not* do here is make certain changes in `extract-pg.xsl` that could now be done to clean up inert junk presently going into our PG files. The "inert junk" was meaningful with the old versions, but not with 2.19+. If I made such changes now, there would be significant diff to look at in the .pg and representations files. Once this PR is processed, I can remove the inert junk separately, perhaps with commits that really focus on one thing at a time.


rbeezer:
Review coming from my assistant - looks like one error-checking situation needs an adjustment.  A force-push is fine.

rbeezer:
Nicely scoped removal. The version branching collapses cleanly to the render_rpc-only path, and the dead clauses fall away correctly once 2.19 is the floor — the `webwork2_minor_version < 19` terms in the `source_key`/`source_value` logic were always going to be false, so dropping them changes nothing. I checked for loose ends and found none: no orphaned references to `html2xml`, the `base64` module, `pgbase64`, or `embed_problem_base64` (the surviving `base64` hits are the hardcoded `badness_base64` literals, not the removed import); the remaining version directories match the trimmed `WEBWORK_VERSIONS` list; and there are no stray whitespace changes. Deferring the `extract-pg.xsl` inert-junk cleanup to its own PRs reads well.

One thing to fix. The removal drops the `else:` html2xml fallback in the version-determination block, but leaves the `ww_version` check outside the status guard:

```python
render_rpc_response = requests.get(url=webwork2_render_rpc, params=params_for_version_determination)
if render_rpc_response.status_code == 200:
    version_determination_json = render_rpc_response.json()
if "ww_version" in version_determination_json:
```

If `render_rpc` returns anything other than 200, `version_determination_json` is never assigned, so `if "ww_version" in version_determination_json:` raises `UnboundLocalError`. That is caught by the enclosing `except Exception` and re-raised as `PTX:ERROR:   There was a problem contacting the WeBWorK server.` with the raw Python message appended.

The consequence is that the landing-page fallback just below — which parses `WW.VERSION` from the server's HTML and then reaches the `PreTeXt supports WeBWorK 2.19 and later ...` message — becomes unreachable whenever `render_rpc` is non-200. So a user pointed at a pre-2.19 server (or any server whose `render_rpc` endpoint 404s or 500s) would get the generic contact error instead of the version-specific one — which is the audience this change is aimed at.

Nesting the check inside the status guard preserves that fallback:

```python
if render_rpc_response.status_code == 200:
    version_determination_json = render_rpc_response.json()
    if "ww_version" in version_determination_json:
        webwork2_version = version_determination_json["ww_version"]
        webwork2_version_match = re.search(
            r"((\d+)\.(\d+)(\+develop)?)", webwork2_version, re.I
        )
```

Alex-Jordan:
OK, I made Claude's suggested change and force pushed.

rbeezer:
Thank, merged as-is.  Let's get the inert junk very soon, while our heads are in it.

Thanks for all the maintenance.

rbeezer:
There was an `-announce` thread warning about this, maybe six months or so ago?  Do you want to reply to it and say it is done now?

PR #3066, opened 2026-07-24
WeBWorK: remove beginproblem() calls, which have been inert since PG …

Alex-Jordan:
…2.19

One change to `extract-pg.xsl`, and then all the repo representation and .pg files updated accordingly.

The string `beginproblem` remains in only one place, in a file that I should have deleted with #3055. I could add that deletion here if that helps. Separate commit, or folded into this one.


Alex-Jordan:
My search for all instances of `beginproblem` must have been wrong (probably I used `beginProblem`. I found a few more from the Showcase example. Updated now. And in the process of updating those files, other things about them were updated: no dates in the header blurb and stable dates for the set def file.

rbeezer:
Bloody hell, I got these out of order, got turned around with the omissions and follow-ups.  Maybe best to just clearly tell me to regenerate with merges as this clean-up continues.  Easy for me to do.  Please do it yourself, to see if changes are as intended, but don't include in the PR.  Thanks.

rbeezer:
OK, I take that back.  So simple, a rebase went fine.  An alternative is to tell me what PRs a give one depends on.  Maybe you did that.  I just need all teh help I can get - too many balls in teh air at one time.  All the time.  ;-)

rbeezer:
Reviewed against the merge-base and checked how it composes with #3068 (which landed first). On the ordering: even though this branch predates #3067/#3068, it merges cleanly and correctly. Its only unique change to each representation is the single `TEXT(beginproblem());` line — orthogonal to #3068's reformatting — and its other regenerated churn (the dropped `### … stable commit` banner lines and the 2027 `.def` dates) already matches current master, so the 3-way merge keeps #3068's formatting and drops `beginproblem`, exactly as a fresh regeneration would. Verified on `tables.xml`; whitespace clean, no conflicts.

The `extract-pg.xsl` change is a clean one-line removal of the inert call.

Two notes on the leftover `beginproblem`:

1. The remaining extracted one is in `…/2_3_1-PTX_problem_source_with_servergenerated_images.pg` — an orphaned duplicate (superseded by the `…PreTeXt_problem_source…` file after the title rename), the #3055 leftover you mentioned. Worth deleting; a separate commit reads cleanly.
2. `beginproblem` also survives in `pretext/lib/webwork.py` — the hardcoded `pg_shell` for a badness/error placeholder still emits `TEXT(beginproblem());`. Inert so harmless, but a second spot if you're purging the call.

Otherwise looks good.

rbeezer:
Merged as-is after rebase.  A separate PR with just (1) and (2) from Claude (two commits) would be the best way to handle that.  Thanks!  Carry on.

Alex-Jordan:
Thanks, and that PR is coming up.

PR #3067, opened 2026-07-24
WeBWorK: remove inert embed_problem block

Alex-Jordan:
This should have been removed with #3059. The `embed_problem` variable it makes is no longer used anywhere.

### DIFF
--- a/pretext/lib/webwork.py
+++ b/pretext/lib/webwork.py
@@ -336,16 +336,6 @@ def webwork_to_xml(
             )
         log.info(msg)
 
-        # The code in pgdense[problem] may have `$refreshCachedImages=1;`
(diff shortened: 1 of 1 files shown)


rbeezer:
Confirmed — `embed_problem` has been dead since #3059: on current master it's assigned in this block and read nowhere, and its only former consumer (`embed_problem_base64`) went away with #3059. Removal is behavior-neutral: the live `pgdense` (with `$refreshCachedImages` live) is used only for static harvesting (the local socket and the server `rawProblemSource`), while embedded HTML problems reference `path[problem]`, not the live source — so the refresh-inert-ing this block did has no remaining consumer. `import re` stays in use; whitespace clean. Looks good.

rbeezer:
Merged as-is after a rebase.  Carry on.

PR #3068, opened 2026-07-25
WeBWorK: remove lots of b-human-readable dependency

Alex-Jordan:
This is another follow up to the retirement of PG 2.16, 2.17, 2.18 support.

In modern times, the "non human readable" PG does not serve much purpose. There was a time when we needed to optimize PG, keeping it short, so that its base64 encoding could be appended in a GET request without hitting the cap on the length of a GET request. But now we use POST requests. So why not simplify the code massively by removing `$b-human-readable`, and just make all PG code human readable?

So this does that as much as possible. Lots of net deleted lines in `extract-pg.xsl`.

There are two ways in which `$b-human-readable` is still important though, having nothing to do with human readable code. These remaining two topics make "human-readable" a misnomer that I will fix soon. 

- The places where we wanted human readable code coincide with the places where we want the `introduction` from an `exercisegroup` to be copied into the .pg. So `$b-human-readable` is still leveraged for that.
- The hack that we do to get LaTeX macros into a WeBWorK problem hinges on `$b-human-readable` too, and again there is a decision point that coincidentally aligned with where we wanted (or didn't want) human readable PG. So `$b-human-readable` is still leveraged for that too.

I'll come to a drop-in soon to discuss #2761. If we can resolve that (or do something different that achieves the same end goal) then the second bullet here melts away, and more `$b-human-readable` can go away. At that point the only thing left is the first bullet. I'll rename `$b-human-readable` to something appropriate, as well as related terminology (`pgdense`) in the extraction file and the python.

### DIFF
--- a/examples/showcase/generated/webwork/answer-type-variety.xml
+++ b/examples/showcase/generated/webwork/answer-type-variety.xml
@@ -92,9 +92,6 @@
   <pg><![CDATA[
 #############################################
 ###       Generated from PreTeXt source      
-###       on 2026-05-15T12:26:17-07:00       
-###   A recent stable commit (2022-07-01):   
-### 6c761d3dba23af92cba35001c852aac04ae99a5f 
 ###                                          
 ###         https://pretextbook.org          
 ###                                          
--- a/examples/showcase/generated/webwork/pg/PreTeXt_Showcase/6_1-Answer_Type_Variety.pg
+++ b/examples/showcase/generated/webwork/pg/PreTeXt_Showcase/6_1-Answer_Type_Variety.pg
@@ -1,8 +1,5 @@
 #############################################
 ###       Generated from PreTeXt source      
-###       on 2026-05-15T12:26:17-07:00       
-###   A recent stable commit (2022-07-01):   
-### 6c761d3dba23af92cba35001c852aac04ae99a5f 
 ###                                          
 ###         https://pretextbook.org          
 ###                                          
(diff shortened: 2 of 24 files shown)


rbeezer:
Reviewed `extract-pg.xsl` against the merge-base and spot-checked the regenerated outputs. The bulk is a clean, mechanical unification — dropping the `$b-human-readable` guards so PG is always human-readable (173 → 67 references), pure formatting. The two intended retentions are intact: the `pg-header` `not($b-human-readable)` branch and the `exercisegroup/introduction//latex-image` macro load, matching your note. Tables still render — the removed `table` template was dead for WeBWorK (problems use `<tabular>`, never `<table>`), and regenerated problems still emit `[@DataTable(...)]` with `niceTables.pl`.

Two things worth a look, neither blocking:

1. In `match="latex"`, the dense `[$TEX]*` is gone, unified to `[$LATEX]*` — the only non-whitespace change. Since dense was sent to the server pre-2.19, `<latex/>` may have rendered the TeX logo there, so this looks like a correctness fix; just confirming `$LATEX` is intended.
2. `git diff --check` flags trailing whitespace in four regenerated files (`tables.xml`, `2_2_5-Tables.pg`, …) — all `[PGML('…`]  ` with two trailing spaces, a PGML hard line break in a multi-line table cell. Faithful output, but committed trailing whitespace will trip a whitespace linter if CI runs one.

Otherwise looks good.

rbeezer:
Merged as-is, after a rebase.  Phew- much cleaner now without that switch/option!  Looking forward to more clean-up.  Thanks!

PR #3070, opened 2026-07-25
Cleanup

Alex-Jordan:
This is the follow-up from #3066.

Side note: the `beginproblem` here in this instance was not even right. It was `TEXT(beginproblem());` which just prints "beginproblem()". Looks like an 11-year old mental slip I made.


Alex-Jordan:
Errr... not an 11-year old mental slip. A modern day mental slip. That was the right way to use `beginproblem()`, it's just been too long since it mattered and I forgot :)

Alex-Jordan:
I added a third, related small cleanup. I realized more `beginproblem()` were hidden in the base64 encoded shell problems. Cutting those, I saw more inert code to trim in the same place.

That's all for today, if you are processing these. I have more things in mind for this spring cleaning, but will get to it over the coming week. Hoping to make a drop-in session as well.

rbeezer:
Reviewed all three commits — clean, and it closes out the two loose ends from #3066 plus a bonus.

- **Remove orphaned PG file**: deletes the stale `2_3_1-PTX_problem_source_…` (superseded by the `…PreTeXt_problem_source…` file after the title rename); the live counterpart is untouched and already `beginproblem`-free.
- **Remove `beginproblem()` from `webwork.py`**: drops the inert `TEXT(beginproblem());` from the badness `pg_shell`. With this, `beginproblem` is fully purged from the repo.
- **Remove inert `file_empty` handling and tighten base64 shells**: `file_empty` was always false on 2.19+ (a missing file is indistinguishable from a compile error, so `no_compile` catches it — its message already reads "either empty or failed to compile"), so that branch was dead and removing it loses no reporting. The three `badness_base64` shells are still live (used as `source_value` on badness); I decoded the new ones — valid PG with the right messages (`DOCUMENT(); loadMacros('PGstandard.pl'); TEXT('…'); ENDDOCUMENT();`), just trimmed of the inert wrapping.

Whitespace clean, commits nicely separated. Looks good.

rbeezer:
Thanks.  A true merge, since I didn't rebase.  ;-)

I don't care much about PR **titles** here, but let's not get too generic, I often end up searching for old stuff and we don't do much tagging.  "WeBWorK cleanup" would have been fine here.

Thanks for all the "spring" cleaning.

PR #3073, opened 2026-07-26
Fix duplicate-label error for image elements inside cardsort responses and premises

### LINKED ISSUE #2927: PreFigure diagrams in a cardsort exercise are flagged incorrectly as having duplicate labels
When PreFigure diagrams are included in a `response` element inside a [`cardsort` problem](https://github.com/davidaustinm/ula/blob/6feefcaf078beb8678eddd67fae58159118bd615/src/section2-6.xml#L2113), the build process will sometimes report that the diagram's labels are duplicated.  Upon building immediately afterwards, the error no longer appears, and the diagrams appear in the build output as expected.

davidaustinm:
Fixes #2927.

When a \`cardsort\` response or premise contains an image element with a \`@label\` attribute (\`pf:prefigure\`, \`latex-image\`, \`asymptote\`, \`sageplot\`, or \`mermaid\`), the static conversion emits the same element — and the same label — in two places: once in the statement tabular and again in the solution list. The assembly-phase duplicate-label check fires on a clean build with the message "should be unique, but is authored multiple times."

The error appears only on a clean build. On subsequent incremental builds the generated assets are already cached, so the extraction step is skipped and the error is not surfaced — making the problem seem intermittent.

The fix changes the cardsort and matching solution templates to use \`apply-templates\` with a new \`cardsort-sol-copy\` mode instead of \`xsl:copy-of\`. An additional template in that mode rewrites the \`@label\` attribute on image-source elements by appending \`-sol\`, giving each solution copy a unique label while keeping the images visible in the rendered solution.

### DIFF
--- a/xsl/pretext-runestone-static.xsl
+++ b/xsl/pretext-runestone-static.xsl
@@ -743,7 +743,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                         <ul>
                             <xsl:for-each select="premise">
                                 <li>
-                                    <xsl:copy-of select="."/>
+                                    <xsl:apply-templates select="." mode="cardsort-sol-copy"/>
                                 </li>
                             </xsl:for-each>
                         </ul>
@@ -777,7 +777,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                             <!-- the test for adjacency -->
                             <xsl:if test="contains($the-ref-fenced, $response-xmlid-fenced)">
                                 <li>
-                                    <xsl:copy-of select="."/>
+                                    <xsl:apply-templates select="." mode="cardsort-sol-copy"/>
                                 </li>
                             </xsl:if>
                         </xsl:for-each>


rbeezer:
Thanks for this — the diagnosis in the description is exactly right (statement + solution each get a copy, and the caching of generated assets is what makes it look intermittent), and it pushed the investigation on #2927 in the right direction.  Reviewing the branch as pushed, though, there are two concrete problems, and a larger question about the approach.

**1. The `cardsort-sol-copy` templates are not in the commit.**  The diff converts two `xsl:copy-of` calls to `apply-templates` in `mode="cardsort-sol-copy"`, but no template in that mode exists anywhere in the file — the `@label`-rewriting template the description relies on appears to have gone uncommitted.  The consequence is worse than a no-op: with no templates in the mode, XSLT's built-in rules take over, which recurse through elements and copy only text.  Exercised against the reproducible example on #2927, the assembled static solution comes out with the premise's `<image>`/`<latex-image>` gone entirely and the raw `\begin{tikzpicture}…` source sitting in the prose.  The premise-side duplicate error disappears, but only because the image has been destroyed — on a quick build this reads as success.

**2. The reported case is not covered.**  Only the two *premise* copies are rerouted.  The solution templates also copy every *response* — `<xsl:copy-of select="response"/>` in the cardsort solution, and the response's `<xsl:copy-of select="."/>` in the matching solution — and #2927 is specifically about PreFigure diagrams in a `response`.  At this branch's head, the example's `response-image` still errors.

**3. The approach may be unnecessary.**  Two findings from the #2927 investigation bear directly on the design here.  First, `@label` is the generated file's basename, so an appended `-sol` means every affected diagram is generated twice into byte-identical files.  Second, the current pipeline — false error aside — already does the desirable thing: both copies share the label, extraction produces exactly one file, and both renderings reference it.  The remedy under discussion on the issue is to move the authored-duplicates check *earlier* in assembly, before the `representations` pass manufactures the copies, which would make the error vanish without touching labels, filenames, or these templates at all.  It may be worth holding this PR until that direction is settled, since it could shrink to nothing — or, if a copy-rewriting mode does turn out to be needed somewhere, it would want to cover premises and responses symmetrically, across both the `cardsort` and `matching` solution templates.

davidaustinm:
Closing this in favor of the alternative fix suggested in issue #2927 and consolidating discussion there and on -dev.

rbeezer:
PR #3080 relocates the duplicate-identifier checks onto the version tree — commit 5be196d11d0206269e1c8040fa6338c27bc1786c, "Assembly: check duplicate identifiers on the version tree" — immediately after version resolution, before any machinery manufactures copies of authored content. The card-sort reproduction from the discussion above now passes cleanly, while a document with a genuine authored duplicate still draws the error. The guard proposed here therefore becomes unnecessary: manufactured interior duplicates are never examined at all. If that lands, this issue can close alongside #2927.

PR #3082, opened 2026-07-28
Resolve Runestone and bibliography lookups on the assembled tree; reinstate Runestone identifier warnings

rbeezer:
Five commits that complete the demotion of the authored source tree (`$original`) begun in #3080, reinstate a pair of Runestone author warnings that the newly-fixed bug had made unreliable, and label the sample book's own components that those warnings correctly flag.

**Runestone component pointers — the detailed story, for Runestone developers.**

A Runestone component (an activecode `program`, an interactive `exercise`, a `datafile`, …) is known to the database by its `@label`. Older projects wrote only `@xml:id`, so for backward compatibility the assembly phase copies `@xml:id` into `@label` when no label was authored. That copy exists only on the assembled tree built in memory — the authored source is never changed.

A name is needed in two situations. First, when a component renders it asks for its own name — this always worked, because rendering walks the assembled tree, where the copy has happened. Second, one component can name another: select questions (`@questions`) and a `program`'s `@include`, `@add-files`, `@compile-also` name their targets by `@xml:id`, and we look each target up to write *its* database name into the emitted HTML (`data-include`, `data-questionlist`, …). That second lookup searched the *authored source* instead of the assembled tree — it found the target, but read a `@label` from a tree where the backward-compatibility copy never happens. If the target's author wrote only `@xml:id`, the label was blank, and the HTML came out `data-include=""` — broken wiring, no message, ever. Projects that write `@label` on every component were unaffected, which is why it went unseen. This has been so since the first `@include` (direct commit `4b77cb2f`, February 2023 — there is no PR to cite); every later attribute of the same kind inherited it. The first commit here resolves those lookups on the assembled tree.

**Citations respect the version in play.** Three bibliography-citation lookups also consulted the authored tree: the CSL extraction's citation vetting, the assembly's is-this-a-citation classifier, and the external-bibliography (`references/@source`) cited-check. Under a document version that excludes a cited `biblio` the consequences were real — generating references *crashed* (`ValueError: 'bib-gone' is not in list`, the bibliography half of the extraction respecting the version while the citation half did not), and a full conversion *silently deleted* the citation from the paragraph (the unmatched replacement copies an empty node-set). Now the extraction is coherent, an excluded citation draws the standard missing-target placeholder and its error, and an external-bibliography entry cited only from excluded content is no longer realized. Unversioned documents are byte-identical.

**`references/@source` resolves again.** Surfaced by testing the citation fix: the external bibliography was fetched with `document(@source, .)`, anchored on a node in a constructed tree that has no base URI — so any *relative* `@source` failed outright. Anchoring on the authored source document restores the documented behavior.

**Reinstating the Runestone identifier warnings.** Two author-facing messages built into the database-name routine — an error when a component has no name at all, a gentle note when it has only an old-style `@xml:id` — were disabled in February 2024 as unreliable. There were two causes, both now removed. The first was the pointer bug above: a named target was handed to the routine as an authored-tree node, which carries no assembly stamp, so it looked unlabeled even when it was not — and the warnings cried wolf. The second is that the routine is also asked, purely in case a name exists, about plain static code listings that are not interactive and need no name; an empty answer there is correct, not a mistake. The fix is a single condition: a `program` that is not `activecode` or `codelens` is a plain listing and is exempt — the one element reaching the routine that legitimately has no database identity (the schema restricts `@interactive` to `codelens`/`activecode`/`no`). The message texts are unchanged from 2024.

**Labeling the sample book.** The reinstated messages correctly flag the sample book's own interactive components authored without a `@label` — a Runestone Academy build of the book would fail on the unnamed ones. Twelve components that already carried an `@xml:id` get a `@label` copied verbatim from it, which preserves the database id exactly (rendered output byte-identical; only the assembled tree records the labels as authored). Two experimental dual/Doenet exercises that carried neither identifier get a fresh matching `@xml:id`+`@label`; these previously emitted an empty database id (`id="rs-"`, the Academy failure mode), now `id="rs-doenet-average-velocity"` and the like.

Five commits, in reading order: (1) Runestone pointers on the assembled tree; (2) citations on the version tree; (3) `references/@source` relative resolution; (4) reinstate the identifier warnings; (5) label the sample book's components.

Verification. The sample article and sample book build byte-identical to `master` (`assembly-static`, `assembly-dynamic`, `latex`, `html`) through the first four commits — none of their behavior changes is reachable by the plain samples — so each is proven on a purpose-built document: a two-`program` pointer test (empty id versus resolved label), a version-excluding CSL citation test (crash versus clean generation; vanished citation versus placeholder), and a version-excluding external-bibliography test (ghost entry realized versus dropped, with a relative `@source`); the CSL tests run the full pipeline through citeproc. The fifth commit takes the Runestone Academy build of the sample book from twenty-eight messages to zero, changes no prose (the sample-book output differs only by the labels and the two new exercises' anchors), and leaves validation exactly as on `master`.

### DIFF
--- a/examples/sample-book/rune.xml
+++ b/examples/sample-book/rune.xml
@@ -104,7 +104,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
         <listing xml:id="program-activecode-c">
             <title>An C program, interactive on a <pubtitle>Runestone</pubtitle> server</title>
-            <program xml:id="c-hello-world" interactive='activecode' language="c">
+            <program xml:id="c-hello-world" label="c-hello-world" interactive='activecode' language="c">
                 <code>
                 #include &lt;stdio.h&gt;
 
@@ -120,7 +120,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
         <listing xml:id="program-activecode-java">
             <title>A Java program, interactive on a <pubtitle>Runestone</pubtitle> server</title>
-            <program xml:id="java-hello-world" interactive='activecode' language="java">
+            <program xml:id="java-hello-world" label="java-hello-world" interactive='activecode' language="java">
                 <code>
                 public class HelloWorld {
                     public static void main(String[] args) {
@@ -135,7 +135,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
         <listing xml:id="program-activecode-javascript">
             <title>An interactive JavaScript program, using <pubtitle>Runestone</pubtitle></title>
-            <program xml:id="javascript-hello-world" interactive='activecode' language="javascript">
+            <program xml:id="javascript-hello-world" label="javascript-hello-world" interactive='activecode' language="javascript">
                 <code>
                 document.write('Hello, world!');
                 </code>
@@ -188,7 +188,7 @@ along with MathBook XML.  If not, see <http://www.gnu.org/licenses/>.
 
         <listing xml:id="program-activecode-octave">
             <title>A simple Octave program</title>
-            <program xml:id="octave-simple" interactive='activecode' language="octave">
+            <program xml:id="octave-simple" label="octave-simple" interactive='activecode' language="octave">
                 <code>
                 x = 2 + 2
                 printf("%d\n", x)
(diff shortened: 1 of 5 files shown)


rbeezer:
@bnmnetp:

You really should try to read the above, to see how hard Runestone database identifiers are.  ;-)

Short version: folks not using `@label` have not been getting good warnings.  And if they do not have an `@xml:id` they are creating empty strings.

**IF** they only have an `@xml:id`, then they should add a `@label` with exactly the same string as the `@xml:id` to keep the database consistent.

PR #3084, opened 2026-07-28
Fix uncapped height of brandlogo when $docinfo/brandlogo/@url is not set

StevenClontz:
The cap on the height of a brandlogo set at https://github.com/PreTeXtBook/pretext/blob/2f098cffab023cac9feab7b1da865d11a8cfae07/css/components/page-parts/_banner.scss#L59-L70 is not being applied when $docinfo/brandlogo/@url is not set.

This small PR adds a dummy `<span class="logo-link"/>` to apply the CSS as intended even when the brandlogo dosn't have a link.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -14566,7 +14566,9 @@ TODO:
                 </a>
             </xsl:when>
             <xsl:otherwise>
+                <span id="logo-link" class="logo-link">
                     <img src="{$location}" alt="Logo image"/>
+                </span>
             </xsl:otherwise>
         </xsl:choose>
     </xsl:if>


StevenClontz:
Before:

image

After:

image

rbeezer:
@StevenClontz — confirmed, and the fix does what it says. Measured in Chrome against a build of the sample article with `@url` removed from its `brandlogo`, so the tall book-cover image exercises the case:

| | wrapper | logo height | banner height |
|---|---|---|---|
| master | none — a bare `` | **239px** (natural size, uncapped) | 260px |
| this PR | `<span class="logo-link">`, computed `height: 80px`, `display: flex` | **80px** (the 5em cap) | 101px |

So `.logo-link { height: 5em }` now applies in the no-URL case exactly as it already did with a URL, and the logo stops driving the banner height.

No regression on the ordinary path: a full HTML build of the unmodified sample article, which does carry `@url`, is byte-identical between master and this branch apart from build timestamps — the `<xsl:when>` branch is untouched.

Two notes for anyone reading the diff and wondering:

The `id="logo-link"` looks like it might be a duplicate-id hazard, and isn't. `logo-link` is on the `reserved-html-ids` list in `pretext-common.xsl` (audited 2026-06-04), so PreTeXt deliberately owns that identifier and refuses it to authors as an `@xml:id` or `@label`; the `<a>` branch has carried the same `id`/`class` pair all along, and only one branch can execute per page. Mirroring it here is the consistent choice.

The diff also quietly corrects a pre-existing indentation slip — the `` in the `<xsl:otherwise>` was over-indented by four spaces, and now reads correctly as the span's child.

One observation, not a request: nothing in `css/` or `js/` actually selects `#logo-link` — only the class is styled — so the `id` appears to be defensive or historical on both branches. Worth knowing if anyone ever tidies that, but consistency with the existing `<a>` is right for this PR.

rbeezer:
Very good!  Thanks for the clean-up.  Merged as-is with a rebase and commit message edit.

@ascholerChemeketa - I've got a todo to investigate the `logo-link` id - I think it can maybe just vanish.  Mostly tagging you as an FYI, no need to react unless absolutely necessary.

PR #3086, opened 2026-07-29
WeBWorK: change descendent reference to child reference

### LINKED ISSUE #1265: Finish deprecation of webwork/setup
Look for `.//pg-code` and replace with `pg-code`.  This will be correct once the "assembly" pre-processing step is part of the WW pipeline AND the source is "corrected" to no longer have a `setup` element.

Better details and exact changes can be found at #1261

Alex-Jordan:
Following up on #1265. I don't believe this will overlap with other upcoming things, so best to cleanly make the replacements now.

Don't forget to close #1265 when this s processed.


rbeezer:
Claude likes it, and so do I.  Merged as-is.  Closes #1265.

PR #3092, opened 2026-07-30
Braille Formatting Changes

MedicalHuffy:
Removed extraneous whitespace inside of <md> environments.

Removed punctuation doubling at end of <md> environments.

### DIFF
--- a/pretext/lib/braille_format.py
+++ b/pretext/lib/braille_format.py
@@ -665,7 +665,7 @@ class PageComposer:
             emit("7" * self.layout.text_width(position, 0))
         elif blk.box == "nemeth":
             emit(translate_print(NEMETH_OPEN))
-            emit('')
+            #emit('')
 
         for child in blk.xml.xpath("segment|block"):
             if child.tag == "segment":
@@ -702,7 +702,7 @@ class PageComposer:
         if blk.box == "standard":
             emit("g" * self.layout.text_width(position, 0))
         elif blk.box == "nemeth":
-            emit('')
+            #emit('')
             close_brf = NEMETH_CLOSE
             if blk.punctuation:
                 close_brf += blk.punctuation
--- a/xsl/pretext-common.xsl
+++ b/xsl/pretext-common.xsl
@@ -1913,7 +1913,7 @@ Book (with parts), "section" at level 3
     <xsl:variable name="math-punctuation">
         <xsl:choose>
             <!-- drop punctuation after display math, if moving to math -->
-            <xsl:when test="$b-include-display and contains($clause-ending-marks, $first-char) and preceding-sibling::node()[1][self::md[mrow]]">
+            <xsl:when test="$b-include-display and contains($clause-ending-marks, $first-char) and preceding-sibling::node()[1][self::md]">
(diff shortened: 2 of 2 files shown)


rbeezer:
Welcome, and thank you — this is a good first contribution. It fixes a real defect, and the diagnosis behind it is right.

The heart of it is the `md[mrow]` → `md` change. A single-line display equation, `<md>p^4 + q^4 = r^4</md>`, has no `mrow` at all, so the old predicate could not see it. And the asymmetry was already there in the same file: the templates that pull punctuation *into* the math test a bare `self::md`, while the branch that drops it from the following text required `md[mrow]`. Punctuation was therefore being added in one place and not removed in the other, which is exactly the doubling you saw. Making the two agree is the right fix.

I built the electronic braille before and after, on `examples/braille/braille-test-book.xml`:

```
BEFORE                                     AFTER
_%                    ← Nemeth open        _%
                      ← blank              p~4"+q~4" .k r~4
p~4"+q~4" .k r~4      ← the math           _:4
                      ← blank              ,adju/ to a s+le 4play$ equa;n1 b au?or$
_:4                   ← close + period     z a s+le `<mr{`> )9 a `<md`>1 )|t
4  ,adju/ to a s+le…  ← second period,
                        then the text
```

804 lines down to 787. All three effects show there: the blank line after the opening Nemeth indicator, the blank line before the closing one, and the duplicated period. With the stray `. ` gone, the continuation text also reflows to fill the line properly.

Worth noting the indentation comes out right as well. That passage is a single `<p>` holding three displayed equations, so the text resuming after each one is a runover and correctly sits at cell 1, while genuine new paragraphs keep their cell-3 indent. The fix does not disturb that.

Since `pretext-common.xsl` is shared by every conversion, I checked the change does not leak into other formats. It does not: with the sample article, which has 57 `md` blocks lacking `mrow` and 14 of them followed by clause punctuation, the LaTeX output is byte-identical and the HTML differs only in its build timestamp. Same for the braille test article in LaTeX, HTML and text. The change is visible only where it was meant to be.

I will remove the commented-out `emit('')` lines.

rbeezer:
Nice job, @MedicalHuffy (Michael D!).  Good catch on the `[mrow]`.

(Reviews from Claude Code are now a routine part of our workflow.)

It'd be nice if we could find a BANA or Nemeth rule about all this - I like to comment the braille code with those.  Do a follow-up PR if you have one.

We never leave dead code around, it is always in the GitHub history - so I will remove those two "commented-out" `emit()` when I merge.

Should get merged in the next 24 hours or so.

rbeezer:
Thanks again, @MedicalHuffy.

* Silently dropped the commented-out code.
* Added "Braille:" topic to commit messages (we alwyas do this)
* Edited commit messages, since I was there anyway.

New code all merged as-is.  Great contribution!

MedicalHuffy:
Hey Rob,

Thank you for the corrections, totally forgot to remove the testing comments before committing. I can't find a specific rule for the whitespace update, but it certainly is more inline with the convention of using indentation instead of whitespace in a formatted brf. At any rate, it would synergize well with future work in fully working on BANA indentation.

PR #3093, opened 2026-07-30
Print project-like with workspace like a worksheet

oscarlevin:
As we discussed in Drop-in a few weeks back, authors of textbooks focused on active learning might include projects and their like as things for readers to actually do.  These might live directly in the middle of a division, so it does not make sense to break into a worksheet for this purpose.  Instead, a PROJECT-LIKE element can be used. 

With this PR, if the author gives `@workspace` on tasks, such elements will now get the Print Preview button just like a worksheet or handout (in the HTML conversion).  In LaTeX, these are also treated like worksheets (with the exception that workspace does not grow/shrink to fill a page the same way it would for a workspace; that will require a future PR).


mitchkeller:
Very exciting to see! Just checking that setting `@workspace` on the PROJECT-LIKE will trigger the print preview in these cases:
- the PROJECT-LIKE is structured with `task` but the author wants the same `@workspace` for each `task` 
- the PROJECT-LIKE has `@workspace` and there are no `task` children

rbeezer:
This is a pleasure to review. Seven commits, one topic each, no commit bodies, no `js/dist` or `css/dist` in the diff, clean whitespace — and the `mode="is-*"` boolean follows the idiom already used three dozen times in `pretext-common.xsl`. Pulling `printout-margins-attribute` and `printout-header-footer-attributes` out as named templates so both paths share them is exactly right, and the comments explain *why* rather than what, including why the preview icon is a floated sibling rather than living inside a project's heading.

The change I looked hardest at is `sanitize-workspace`, since widening a condition in shared `pretext-common.xsl` is where a regression would hide. It's clean. The raw diffs look alarming at first — 379 changed lines in the sample article's LaTeX, 171 of them removals — but masking every digit reduces that to **zero removals and 37 additions**. All the apparent removals are cross-reference renumbering caused by inserting a new section. HTML tells the same story: 88 files differ, and once numbering is masked the only difference is the new section's entry in the table of contents. Schema errors are 52 before and 52 after.

The LaTeX geometry is handled properly. Computing `$b-own-pages` once so `\newgeometry` and `\restoregeometry` cannot fire under different conditions is the right instinct, and the sample article confirms it: 18 of each, final nesting depth zero.

Output checks out too — two projects carry `standalone-printout`, `data-margins="0.75in 0.75in 0.75in 0.75in"` is emitted, both preview links resolve, and the third project in the demo, the one without workspace, correctly gets nothing. Good demo design.

One thing worth tightening. The standalone test now has a single source of truth in `mode="is-standalone-printout"`, but `pretext-html.xsl` line 505 spells the same condition out inline for `b-has-printout`:

```
descendant::*[&PROJECT-FILTER;][not(ancestor::worksheet|ancestor::handout)][@workspace|.//@workspace]
```

Two expressions of one rule, and they will drift the first time the definition changes. That site could use the mode instead.

The branch is 133 commits behind master, so a rebase before merging is worth it.

rbeezer:
Should merge soon.  Not too concerened about teh rebase, my assistant has been busy.

But you may want to consolidate that doubled condition?

oscarlevin:
Unless you are in the middle of merging right now, I'll take care of the rebase and double condition now.

rbeezer:
At Mariners game.

oscarlevin:
Rebased and force pushed.  I looked at the duplicate logic the claude found and did not see an obvious way to consolidate: the instance in pretext-html is part of a param that looks for the conditions on descendents along with worksheets and handouts.

rbeezer:
Re-reviewed after the rebase.

**On the duplicated condition — you are right, and I withdraw the suggestion.** I had only looked at the two expressions side by side, not at the call site. It is a single-expression parameter:

```xslt
<xsl:with-param name="b-has-printout"
    select="descendant-or-self::worksheet or descendant-or-self::handout or
            descendant::*[&PROJECT-FILTER;][not(ancestor::worksheet|ancestor::handout)][@workspace|.//@workspace]"/>
```

`mode="is-standalone-printout"` answers for a single node, whereas this asks whether *any descendant* qualifies, as one of three disjuncts. XSLT 1.0 cannot invoke a mode from inside a predicate, so reusing it means a `for-each` into a variable and a `contains($flags,'true')` test — the `nav-companions` idiom — which turns a one-line `select` into roughly eight lines and forces the caller to be restructured. That is a worse trade than the duplication removes, so the code should stay as it is.

For the record the duplication is genuine — `&PROJECT-FILTER;` and `&PROJECT-LIKE;` are the same four elements — so if you ever want cheap insurance against drift, a one-line comment at each site naming the other costs nothing. Entirely optional.

The force-push was a clean rebase: diffing each side against its own merge-base gives an identical contribution, 335 insertions and 28 deletions either way. Since the base moved 111 commits, I re-ran everything rather than trusting the earlier numbers.

- **LaTeX**: 383 raw changed lines, but with every digit masked, **37 additions and zero removals** — the apparent removals are all renumbering.
- **Geometry**: `\newgeometry`/`\restoregeometry` go 16/16 to 18/18, balanced, depth never negative, ending at zero. Exactly two new pairs, one per new project.
- **HTML**: 172 to 173 pages. Of the shared pages, 59 change by addition only, each gaining the same table-of-contents entry. The only removals anywhere are four tokens on the two pages either side of the new section, whose previous/next buttons repoint to it.
- **Schema**: 315 errors before, 315 after, none new.
- Whitespace clean, seven commits with one topic each and no bodies, and the Guide covers both the HTML and LaTeX behaviour with the worksheet-options entry widened and cross-linked.

Nothing outstanding from me.

rbeezer:
> I looked at the duplicate logic the claude found 

Yes, I was not totally grokking that one either - thought maybe it would resonate better with you.

rbeezer:
Want to announce on `pretext-announce`?  Should be of wide interest.  I just rebuilt the website, but did not verify this was present.  If you see any problems, be sure to holler.  Thanks,

PR #3103, opened 2026-07-31
Read aloud

oscarlevin:
While a commercial screen reader is likely the preferred tool for a blind reader, there are many people who might benefit from an easy way to have the text in a pretext book read to them (including students who would like to hear what a particular math symbol is called). 

This adds an embedded text-to-speech player that will do a quite serviceable job reading the content of a page.  Some basic settings (reading speed, voice, whether to read knowled content, whether to jump to the next page) are available for readers as well.

Publisher variable to turn on or off (with on as the default). 

How it works: All javascript magic.  Modern browsers provide text-to-speech technology, often falling back to the speech engine provided by the operating system.  We provide some additional logic to get the reading order, and handoff between regular text and mathjax.  Basic keyboard navigation is also used to jump up and down the page.

### DIFF
--- a/css/components/page-parts/_navbar.scss
+++ b/css/components/page-parts/_navbar.scss
@@ -32,6 +32,11 @@ $center-content: true !default;
 
 @use 'components/page-parts/extras/navbar-btn-borders';
 
+@use 'components/page-parts/read-aloud' with (
+  $nav-height: $nav-height,
+  $navbar-breakpoint: $navbar-breakpoint
+);
+
 .ptx-navbar {
   position: sticky;
   top: 0;
@@ -188,7 +193,7 @@ $center-content: true !default;
     display: none;
   }
 
-  :is(.runestone-profile, .activecode-toggle, .ptx-search-button, .ptx-calculator-toggle, .ptx-embed-button, .ptx-readability-options-button) .name {
+  :is(.runestone-profile, .activecode-toggle, .ptx-search-button, .ptx-calculator-toggle, .ptx-embed-button, .ptx-readability-options-button, .ptx-read-aloud-button) .name {
     @include accessibility.sr-only;
   }
 
--- a/css/components/page-parts/_read-aloud.scss
+++ b/css/components/page-parts/_read-aloud.scss
@@ -1,0 +1,231 @@
+// Styles for the read-aloud (text-to-speech) feature: the in-navbar player,
+// its settings dialog, and the current-sentence highlight.  Included via
+// _navbar.scss so every theme with a navbar gets it; the markup only exists
+// when the read-aloud publication option is on.
+
+$nav-height: 36px !default;
+$navbar-breakpoint: 800px !default;
+
+@use 'components/helpers/dialogs' as dialogs;
+
+// ---------------------------------------------------------------------------
+// Player bar
+//
+// The player is emitted inside .nav-other-controls and simply unhides in
+// place, growing the navbar row to the right of the button that opened it.
+// It is deliberately *not* a floating bubble: a second styled surface hovering
+// over the page competed with the navbar it was anchored to, and on narrow
+// screens (where the navbar moves to the foot of the page) it had to be moved
+// again to avoid covering the content.  Living in the row, it inherits the
+// navbar's own button styling and needs no positioning of its own — only
+// below the breakpoint, where it lifts onto a second row, does it position
+// at all.
+
+.ptx-read-aloud-player {
+  display: flex;
+  align-items: center;
+  align-self: stretch;
+
+  &[hidden] {
+    display: none;
+  }
+
+  // The toggle button carries both glyphs; data-state picks one.
+  //
+  // `display: contents` rather than inline/block: prev and next hold their
+  // .icon as the button's direct flex child, where align-items centers it.
+  // A wrapper of its own would be the flex item instead, and as a blockified
+  // line box it stood 29px tall with the 24px glyph pinned to its top —
+  // leaving the play triangle riding ~2.5px above its neighbours.  Dropping
+  // the wrapper from the box tree makes this button structurally identical to
+  // the other two rather than merely tuned to match.
+  .ptx-read-aloud-icon-play {
+    display: contents;
+  }
+
+  .ptx-read-aloud-icon-pause {
+    display: none;
+  }
+}
+
+.ptx-read-aloud-player[data-state="speaking"] {
+  .ptx-read-aloud-icon-play {
+    display: none;
+  }
+  .ptx-read-aloud-icon-pause {
+    display: contents;
+  }
+}
+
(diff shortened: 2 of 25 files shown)


rbeezer:
This is a substantial piece of work and it shows. Four commits, one topic each, no bodies, correct prefixes. No `js/dist` or `css/dist` in the diff. The new `jsbuilder.mjs` entry is registered and commented properly, and the bundle builds cleanly — 61 KB from a fresh `node jsbuilder.mjs`. Best of all, **73 unit tests, and all 73 pass** under `node --test` in `js/src/read-aloud`. Contributions arriving with their own test suite are rare and very welcome.

One blocker, and it is a single step in an XPath.

In `xsl/publisher-variables.xsl`:

```xslt
<xsl:apply-templates select="$publisher-attribute-options/html/read-aloud/pi:pub-attribute[@name='read-aloud']" .../>
```

`read-aloud` is declared as a direct attribute of `html`, next to `short-answer-responses` and `design-width`, and those are selected as `html/pi:pub-attribute[@name='…']`. The extra step is only used for genuine sub-elements such as `calculator`, `knowl` and `annotation`. As written this selects an empty node-set, so `$read-aloud` is empty and `$b-read-aloud` is false — always, including under the documented default of `yes`.

Building the sample article with a default publication file gives **0 of 172 pages** carrying any read-aloud markup. Removing the `read-aloud/` step and rebuilding gives **140 pages** with the button, the player and the script tag. So the fix is to delete one path step, and the feature then does what the description says.

Two smaller things.

**The Guide has the reference entry but nothing conversational.** `publication-file.xml` gains `online-read-aloud-options`, stating the attribute and its two values, and that is the only mention of the feature anywhere in the Guide. `online-html.xml` is where the comparable options get their prose — `online-embedded-calculator`, `online-favicon-configuration`, `online-html-annotation` — and read-aloud has no counterpart there, so the new reference entry cross-references nothing. A publisher currently learns that a switch exists but not what it switches: nothing describes the player, the reader-facing settings, or the keyboard navigation, all of which the PR description explains well. That description is most of the missing text.

**The localization reasoning is worth revisiting.** The new template hard-codes ten English strings, and the comment above it argues that `xsl/localizations` is for content reaching more than one output format, whereas this is an HTML-only control surface with no LaTeX or EPUB counterpart. The question was clearly thought about, but the premise does not quite hold — `en-US.xml` already carries exactly that kind of chrome:

```
<localization string-id='toc'>Contents</localization>
<localization string-id='feedback'>Feedback</localization>
<localization string-id='calculator'>Calculator</localization>
```

with a note at line 313 that `toc` serves the navigation sidebar. None of those has a LaTeX or EPUB counterpart either, so as things stand the sidebar toggle, the Feedback button and the Calculator button are translatable while the read-aloud controls would not be. The twenty strings in `strings.js` are a separate matter and the eventual JavaScript-side layer you describe sounds right for them, but the ten in the XSL could use the existing mechanism today.

The branch is 51 commits behind master, so a rebase before merging is worth it.

rbeezer:
Really looking forward tlo this one - sorry it has sat for a few days.

Did you really mean `html/read-aloud/@read-aloud` (not according to the documentation)?

```
$publisher-attribute-options/html/read-aloud/pi:pub-attribute[@name='read-aloud']
```

Note to self: remember to regenerate publication-schema.

For publisher file switches I like to always have two discussions with two links:  the terse technical info for the switch, and some narrative description, so please do add at least a short paragraph about the feature where the HTML conversion is discussed.  And we have a general "accessibility" discussion - definitely should go there.

OK, to ignore the internationalization of strings, but let's not let this get too far removed.  A JS-HTML solution for the "chrome" should be a (boring) priority now.  @AXXXXXXXXXXX  should ahve ideas.

Did I see "direct child" somewhere?  No uuch thing as an indirect child.  ;-)  Try teaching your AI that!  Good luck.

oscarlevin:
Thanks!  Fixed the bad xpath and added more documentation.

It would be a lot of fun to show this off at MathFest, if you happen to have time to look at this again before Thursday.

Is "direct child" really the "most unique" error you noticed? ;)

oscarlevin:
Oh, I also switched the `$b-read-aloud` to be true when the attribute is not no, rather than it is yes, which I think is more in line with a "yes" default should work (I don't know if that matters given the later claim that default is yes).

rbeezer:
Re-reviewed at `935a1942`.

The XPath fix works, and the switch now works in both directions. On the sample article the default gives 61 pages with the controls and 140 loading the script — the 79 without controls are all standalone audio and video pages, where a toolbar would make no sense — and setting `read-aloud="no"` gives zero and zero.

Your change of `$b-read-aloud` to `not($read-aloud = 'no')` is the better form, and more idiomatic than what it replaced. The negated shape is the house majority here, 25 uses against 11, and the closest analogues are exactly this pattern:

```xslt
<xsl:variable name="b-has-calculator" select="not($html-calculator = 'none')"/>
<xsl:variable name="b-has-scratch-activecode" select="not($… = 'none')"/>
```

both being "on unless explicitly switched off", which is what this is. The value cannot surprise you either, since `set-pubfile-variable` falls back to `@default` when the attribute is absent and validates against `@options` otherwise.

The documentation now does what was asked, and moving it into its own `Guide:` commit rather than alongside the code is the right split. The narrative section covers the player, knowl handling, keyboard navigation and the settings menu, the accessibility discussion gains its Built-in Screen Reader item, and the two are cross-linked in both directions.

One small thing, which Rob will fix while merging rather than sending you round again: the new narrative says the feature works "via the `<tag>read-aloud</tag>` element of the publication file", but it is an attribute on `html` — `schema/publication-schema.xml` line 432 declares `attribute read-aloud { "yes" | "no" }?`. So `<attr>read-aloud</attr>` and "attribute". Nothing else in the paragraph needs touching.

On "direct child": that one is not yours. Both occurrences are pre-existing in `doc/guide/author/topics.xml`, in the Runestone `program` and `label` prose, unchanged at the merge-base, and this branch adds none. There is a third in `xsl/latex/pretext-latex-texstyle.xsl`. Three in the repository, worth a separate sweep.

Re-verified the rest: 73 unit tests still passing, the bundle still builds at the same size, whitespace clean, five commits with one topic each and no bodies.

rbeezer:
All set, ready for MathFest!  Have fun.  Definitely needs a "pretext-announce" post from you!  I just re-generated the website and checked that this is present on the sample article.  Maybe pick a good math-heavy page for reading?

BUT, not working for me in Firefox.  A play/pause button will toggle - that's as far as I get.  No sound, no moving highlight.  But it is working in Chrome.  Maybe a round of testing frpm "pretext-dev" now that this is live on the website?

oscarlevin:
I will definitely announce.  Firefox on Linux seems to be a particular
issue, and there doesn't seem to be any way to fix that.

PR #3104, opened 2026-08-01
WeBWorK: one js file to rule them all

Alex-Jordan:
Before merging, this should get approval from @bnmnetp (question posted in -dev "consolidating pretext-webwork.js") as well as approval from @oscarlevin (for some changes to `jsbuilder` where I flew blindfolded, guessing that my edits make sense).

This takes the `js/pretext-webwork/2.20/pretext-webwork.js` file, and moves it to `js/pretext-webwork.js`. And then deletes the `js/pretext-webwork/` folder. From here on out, I/we will only manage one `pretext-webwork.js` file.

This has always been a pain, and with 2.21 coming out next week, it would be nice not to have to deal with. I don't foresee structural changes with WeBWorK/PG happening anymore like happened from 2.15 through 2.18, and it should be easy to handle any need for subtle differences within one file. Although, I believe there are no such differences at all to worry about with 2.19, 2.20, 2.21.

- PTX HTML built long ago relies on these "versioned" pretext-webwork.js files that live on an AIM server. OK, those old files aren't going anywhere.
- PTX HTML built since we started building `_static` is unaffected: will use the javascript files in `_static`.
- PTX HTML built after this is merged will now rely on the lone `pretext-webwork.js` in its `_static` folder. Probably some `_static` folders out there won't be "cleaned" and will retain the folder with the "versioned" files. Not clean, but harmless.

I'm unsure if something on the Runestone side needs to know the actual file path to `pretext-webwork.js`, but I doubt it.


oscarlevin:
I would suggest moving the source javascript into the js/src folder, maybe inside js/src/webwork/.  Eventually, it might make sense to modularize the file, and this would make that easy.

bnmnetp:
Today, the `runestone-manifest.xml` gives us a `webwork_js_version` which we use to construct the path to the `pretext-webwork.js` file this is needed for instructor pages like the assignment builder and grader and for student pages like the doAssignment page.

A change like this needs to be coordinated with a change to Runestone to ensure that all of the above continue to work.

Its unclear to me how the larger jsbuilder project will impact the above pages.

oscarlevin:
@bnmnetp, I think it would be worth changing the construction of the path to be `runeston/stuff/__static/pretext/js/dist/pretext-webwork.{v}.js` rather than `runestone/stuff/__static/pretext/js/pretext-webwork.{v}.js`.  We still copy over more than we need to, so things have continued to work.  Assuming that the path only gets set when the book rebuilds, this shouldn't break anything.

As for what happens when you don't get a version, that's a @Alex-Jordan question.  Maybe it makes sense for the manifest to send the relevant relative path instead of the version.

bnmnetp:
I think I'll wait on doing anything until we know where this goes.  It sounds to me like @Alex-Jordan is saying I won't need a version at all anymore.

Alex-Jordan:
@oscarlevin If I understand right:

- I should not delete any files or folders from `js/dist/`. There is a future where the cleanup will happen automatically when `npm run build` is run.
- I moved `js/pretext-webwork.js` to `js/src/pretext-webwork.js`
- What changes should be made in this PR to `script/jsbuilder/jsbuilder.mjs` and `script/jsbuilder/README.md`?

@bnmnetp That's right. The plan here is that PTX HTML pages that currently have like:
```
<script src="_static/pretext/js/dist/pretext-webwork/2.20/pretext-webwork.js">
```
will now have something like:
```
<script src="_static/pretext/js/dist/pretext-webwork.js">
```
although I am not 100% sure about that path yet. The one and only `pretext-webwork.js` will have some references to eBookConfig, like the 2.20 one does now. This one file will be maintained to work over whichever small subset of WeBWorK/PG versions PTX supports. (Currently 2.19 and 2.20, soon to include 2.21).

Maybe we plan for:

- Some upcoming Friday* after decisions are made, Rob merges this.
- Later in the day, Oscar has it in the latest CLI.
- Later in the day or Saturday morning, Brad is rebuilding RS books.

*I'm just spitballing a plan. I'm not sure what day of the week is best for this sort of thing.

rbeezer:
This coming week is good, the next one is not.  But I can review (lightly) and let @oscarlevin take the lead if that is workable.

Alex-Jordan:
When @oscarlevin is able, I just need the confirmation (or correction) about the changes here within `script/jsbuilder`. If those are good, this is ready. And then I will post more to try and organize the coordination.

oscarlevin:
@Alex-Jordan, those changes to the jsbuilder script and readme look right to me.

rbeezer:
Sounds like @oscarlevin is clear.  @Alex-Jordan: do you have a plan/agreement in place with @bnmnetp?  I'd like to merge this as soon as it is ready.

rbeezer:
@Alex-Jordan - @bnmnetp should post here and release you, then ball will be in your court

Alex-Jordan:
@bnmnetp What this changes is that

1. The PTX repo no longer has the `js/pretext-webwork/` folder, with subfolders like `2.19/` and `2.20/`, each with their own `pretext-webwork.js`. Instead it has one file: `js/pretext-webwork/pretext-webwork.js`
2. HTML pages look for the javascript at `{$html.js.dir}/pretext-webwork.js` instead of `{$html.js.dir}/pretext-webwork/2.XX/pretext-webwork.js`

The file itself is like what it has been, with a few lines that rely on `eBookConfig`.

Does any of that affect you on the Runestone side?

> Today, the runestone-manifest.xml gives us a webwork_js_version which we use to construct the path to the pretext-webwork.js file this is needed for instructor pages like the assignment builder and grader and for student pages like the doAssignment page.

So it sounds like you can ignore that `webwork_js_version` from the manifest. And when this change happens, simplify those instructor pages.

Will there be a time when some books at Runestone are built after this merge, and others before? If that were the case, then for those instructor pages, could you look in two places and load whichever file is successfully found?

bnmnetp:
I'm ready to change my template to point to `<script src="_static/pretext/js/dist/pretext-webwork.js">` whenever @oscarlevin tells me its ready.  This is not a big change and won't effect Runestone until all the books have been rebuilt and have this new place for webwork.

If that is not where `pretext-webwork.js` is going to land, just let me know the right path.

Alex-Jordan:
On this branch, HTML is being built to look for the file at `_static/pretext/js/dist/pretext-webwork.js`. So yes, @bnmnetp , that is right.

However, a reminder for @rbeezer. When this is merged, it sounds like you need to run jsbuild before the final commit to the repo, so that the file will actually exist there.

I believe this is ready for a 1-2-3 punch.

1. @rbeezer merges (including run jsbuild).
2. @oscarlevin releases the CLI with the latest pretext.
3. @bnmnetp gets the latest CLI and rebuilds books. Does a quick check that the sample chapter and/or Active Calculus is working.

It's hard to coordinate with an asynchronous conversation, so I'll toss out a proposal. Rob starts this chain Wednesday afternoon. Does that work for Oscar and Brad to follow in turn? If this is not a good time for anyone, can you propose an alternate start time?

Alex-Jordan:
I haven't heard from @rbeezer about starting today, or from @oscarlevin, @bnmnetp. So maybe I was too ambitious with that timeline. Is everyone OK with it starting on Friday?

oscarlevin:
I don't think we need a timeline.  The only question is whether Brad wants
to update the CLI and make the change or not.

rbeezer:
No timeline - not necessary and we don't need to schedule the work.  Brad will react to whatever is in the CLI once Oscar gives him a heads-up about a release.

@Alex-Jordan - please remove the code that puts this in the manifest, or maybe that is gone already.  In any event, post here when this is ready, and I will work on merging.  Ball is in your court and you are free to complete this.

Alex-Jordan:
@bnmnetp Can you confirm there is no other reason to document the WW version in the manifest?

Alex-Jordan:
This now has the WW version omitted from the manifest. And then at this point, there is no need for the `webwork-m(in|aj)or-version` variable in -html, so that is removed too.

bnmnetp:
I'll have some code to remove that parses it, but now that we don't rely on it for the .js file there is not use for it.

Alex-Jordan:
OK, thanks @bnmnetp.  Then @rbeezer, passing it off to you to process.

rbeezer:
Got it - thanks.

rbeezer:
The consolidation itself is clean — no dangling references anywhere, and the Python side is correctly left alone: `webwork.py` still derives the versions, still stamps `@webwork2_major_version` and `@webwork2_minor_version` onto `webwork-reps`, and still refuses a server older than 2.19. Only the XSL consumers go, which is the point. One commit, no body, no `js/dist` in the diff, whitespace clean.

The `jsbuilder` half needs another pass, though, and the first item stops the build.

**The file and the bundler disagree about where it lives.** It moved to `js/src/pretext-webwork.js`, while the target says

```js
in: path.join(jsRoot, `pretext-webwork.js`),
```

so `node script/jsbuilder/jsbuilder.mjs` fails:

```
✘ [ERROR] Could not resolve ".../js/pretext-webwork.js"
Build failed with 1 error
```

esbuild aborts the whole run rather than skipping the one target, so no dist file can be rebuilt at all while this is applied.

The path looks like the right half to keep and the file placement the thing to change. Your description says "moves it to `js/pretext-webwork.js`", the README table you updated says the same, and every `group: 'script'` source already sits at the `js/` root — `pretext_search.js`, `ptx_scorm_events.js`, `lti_iframe_resizer.js`, `knowl.js`, `prism/gdscript-prism.js`, `pretext-stack/*` — while `js/src/` holds only the `bundle` entry points. Moving the file to `js/pretext-webwork.js` builds cleanly and produces a 27064-byte `js/dist/pretext-webwork.js`.

Three smaller leftovers from the removed `.map()`, none of which a build would catch:

- `WEBWORK_VERSIONS` at line 181 is now unreferenced.
- The header comment at line 28 still describes the old scheme: `pretext-webwork/2.X/pretext-webwork   (versions: 2.19–2.20)`.
- `name: \`pretext-webwork\`` and ``path.join(jsRoot, `pretext-webwork.js`)`` keep their backticks now that there is no `${v}` to interpolate; the other 23 `name:` entries use single quotes.

Alex-Jordan:
@oscarlevin Do you agree with Claude that line 301 of `jsbuilder.mjs` should be 
```
      in: path.join(jsRoot, `src/pretext-webwork.js`),
```
? I'm going ahead with pushing that change.

@rbeezer, I made the other cleanup changes that Claude suggested.

rbeezer:
Build works now, verified end to end: `jsbuilder` completes, and the WeBWorK sample chapter emits `src="_static/pretext/js/dist/pretext-webwork.js"` with that file present in the build. The three `jsbuilder` leftovers are all fixed, the manifest removal is complete, and the Python side is untouched.

Two small things remain, both from keeping the file in `src/`:

- `README.md:64` still gives the source as `js/pretext-webwork.js`, which no longer exists. The header comment in `jsbuilder.mjs` was updated to `js/src/pretext-webwork.js`, so the two now disagree.
- Removing `WEBWORK_VERSIONS` left a doubled blank line above `function getAllTargets()`. No others in the file.

On the placement question now with @oscarlevin: `pretext-webwork` is the one `group: 'script'` target sourced from `js/src/`. The other seven — `pretext_search`, `ptx_scorm_events`, `lti_iframe_resizer`, `knowl`, `prism/gdscript-prism`, and both `pretext-stack` entries — sit at the `js/` root, and `js/src/` otherwise holds only the bundle entry points, `pretext-core` and `pretext-read-aloud`. So the question is whether that split carries meaning or is incidental. Nothing breaks either way, and I have no objection to the current arrangement.

One thing to flag, since the question quotes me for the opposite position: what I suggested was keeping `js/pretext-webwork.js` as the bundler path and moving the file to it, not changing the path to `src/`.

Alex-Jordan:
Claude is steering for `js/pretext-webwork.js`, while I believe @oscarlevin recommended `js/src/pretext-webwork.js`. I assume Oscar is right, and the things Claude notes will also eventually migrate into `src/`. But I'd like to give Oscar the final word before I try this once more.

oscarlevin:
I don't think it really matters, but I suggest that our policy should be, for the time being, that javascript that we have "fixed up" got in `js/src`, and javascript that we haven't really touched stay where it is.  Eventually we might want to organize things inside `js/src` more, but for now it will be helpful to keep this distinction, I think.

One day maybe the webwork javascript would should be modularized to make it easier to to maintain, which we can easily do now.  I'm saying that without looking at the contents of the file, so I might be completely wrong here.  But it doesn't matter.

Alex-Jordan:
OK, this is ready to review again.

rbeezer:
Verified end to end: `jsbuilder` completes, the sample chapter emits `src="_static/pretext/js/dist/pretext-webwork.js"` with that file present in the build, and building the merge-base for comparison differs on exactly that one line and nothing else. The Runestone manifest no longer carries `webwork-version`, and a repo-wide sweep for `webwork-major-version`, `webwork-minor-version`, `WEBWORK_VERSIONS`, `pretext-webwork/2.`, `js/pretext-webwork` and `webwork_js_version` comes back empty. Both items from last round are fixed, and the placement is settled — Oscar's policy answers it and `js/src/` is right.

One thing worth recording, since it supports the consolidation: the 2.19 and 2.20 files differ in only two hunks, both Runestone base-course resolution, and 2.20 falls back to `eBookConfig.basecourse` — exactly what 2.19 did unconditionally. So a 2.19 server served the single file gets 2.19 behavior. The two never differed by WeBWorK version at all; they differed by Runestone feature date.

Two nits:

- `script/jsbuilder/README.md:166` says "The WeBWorK files in `js/src/pretext-webwork/`", but that directory does not exist — `js/src/` holds the file itself. `js/src/pretext-webwork.js` reads right, and singular now that there is one file.
- The new target is the only entry in `getAllTargets()` with no comment above it. Dropping the "one file per server minor version" note took the blank line with it, so it now runs on from the STACK pair and reads as a third STACK entry.

rbeezer:
Merged with a rebase and light editing of the commit message.  Thanks, @Alex-Jordan  for the maintenance, and for sticking with this one.

The 2.20 files (source and dist) got a git rename into the unversioned forms.

@bnmnetp, @oscarlevin: I stopped tracking the `dist/` forms for 2.16-2.19.  But there were also `*.map` files that are git-ignored (I believe) so I still had some directory structure laying around.  Not sure how important the hygiene is, but maybe we don't want to spew a lot of stuff into output?  Maybe you can just `rm -rf` on `js/dist/pretext-webwork` as a (necessary) precaution?

Does this argue for not having untracked files in `dist/`?  Or for blasting out the entire contents before a rebuild?  Or is it just me, since I do the (re-)building? @oscarlevin ?

oscarlevin:
Having the `*.map` files is very useful for development and testing, but they don't need to be distributed, thus they are gitignored.  You can do whatever you want with them to keep your local copy clean, or not.  Or maybe I don't understand the question.

Alex-Jordan:
> Merged with a rebase and light editing of the commit message.

Thanks!  I sometimes don't want to agonize over a commit message, and I'm happy if you change it. I hope it doesn't simply transfer the burden of wordsmithing a commit message to you though.

PR #3105, opened 2026-08-01
chore: fix unintentional(?) missing borders for blocks in Greeley theme

StevenClontz:
`border-left: 2px solid var(--block-border-color)` is now generated for `.remark-like` (and .example-like, .project-like, .openproblem-like, .computation-like, .commentary), matching default-modern's L-border treatment.

Root cause: css/targets/html/greeley/_chunks-greeley.scss imported the L-mixin helper and even had knowl-content reset rules assuming a left border existed (`border-left: none` override for knowls), but the actual `@include L-mixin.border;` call was missing from the .example-like etc. rule — it was left out since the theme's original commit (aa56f70e).

Fix: added `@include L-mixin.border;` to that rule block, mirroring how default-modern applies it.

### DIFF
--- a/css/targets/html/greeley/_chunks-greeley.scss
+++ b/css/targets/html/greeley/_chunks-greeley.scss
@@ -44,6 +44,8 @@ $border-radius: 8px !default;
 .openproblem-like,
 .computation-like,
 .commentary  {
+  @include L-mixin.border;
+
   > .heading {
     display: block;
     font-size: 1.1em;


rbeezer:
Thanks, @StevenClontz.  I'll let Oscar handle this change.

@oscarlevin - please edit the commit message and regenerate the CSS on a fresh commit.  Thanks.

oscarlevin:
The style already renders as intended.  The goal of the `greeley` still was to get the same outer page look as `denver` but with latex-like blocks: no boxes at all.  

The extra mixins that are added to the top of the file are remnants of copying over from denver and probably should be removed.

StevenClontz:
I'd encourage you to reopen this, or an issue related to it, because as is Greeley looks broken. There's no reason that 

```
<remark><p>foo</p></remark><p>bar</p>
```

and 

```
<remark><p>foo</p><p>bar</p></remark>
```

should render identically, because semantically they are very different.

StevenClontz:
>  The goal of the `greeley` still was to get the same outer page look as `denver` but with latex-like blocks: no boxes at all.

Let's not repeat the mistakes of LaTeX.

PR #3107, opened 2026-08-01
WeBWorK: a copied problem again shares the original's PG file

rbeezer:
Addresses a report on `pretext-dev` from Alex Jordan: a `webwork` element using `@copy` had started generating its own `.pg` file instead of sharing the file of the problem it copies, which defeats both purposes of the mechanism — less duplication in the source, and fewer files to deploy to a server. [thread](https://groups.google.com/d/msgid/pretext-dev/CA%2BR-jreZ4owd21FO_R7qyZYExtGbwYvB_V3gtbO1-8oyt2vfDg%40mail.gmail.com)

Confirmed and bisected. Building the WeBWorK sample chapter's problem sets gives 86 `.pg` files at the commit before `1a1259b1`, and 88 after it. The two extra files are exactly the two copied problems, and the set definition changed accordingly — the second entry of `set1_1-Arithmetic.def` used to point at `1_1_1-Adding_SingleDigit_Integers.pg` and had begun pointing at a `1_1_2-Declaring_a_Problem_Seed.pg` of its own. The report named that one; there was a second, `1_2_4-Copy_a_Problem_with_Tasks.pg`.

`1a1259b1` moved provenance stamps into the internal namespace, renaming the assembly stamp `@copied-from` to `@pi:copied-from`. It updated `extract-pg.xsl` to read the new name. It did not update `pretext-ww-problem-sets.xsl`, whose three templates still matched the un-prefixed `@copied-from` — one to redirect a copy's relative filename to the original's, one for the absolute filename, and an empty one to suppress writing a file at all. All three quietly stopped matching, which is why a second file appeared and the set definition followed it. XSLT reports nothing when a predicate simply never matches, and the stylesheet did not even declare the `pi` prefix, so the rename could not have taken effect there by accident.

## Two attributes, one name

The reason this was easy to get wrong is that `copied-from` names two different things holding two different values, and the conversion has to know which one it is looking at.

**In the assembly tree**, `@pi:copied-from` is stamped by `pretext-assembly.xsl` when it resolves `webwork[@copy]`. Its value is whatever the author aimed `@copy` at — the `@xml:id` of the target `webwork`, for instance `webwork-add-numbers`. This is why the lookup beside it, `webwork[@xml:id = $copied-from]`, resolves.

**In the representations file**, `@copied-from` is a plain attribute written by `extract-pg.xsl`. Its value is not the `@xml:id` at all; it is the assembly identifier of the *exercise* containing the copied problem, for instance `integer-addition`. That is an interchange format read by the Python routines, which know the attribute by its plain name.

`pretext-ww-problem-sets.xsl` imports `pretext-assembly.xsl` and assembles from source, so the attribute reaching it is the tree stamp, never the representations one. A comment added to that file in `1a1259b1` asserted the opposite — that its `@copied-from` arrives via the representations file — and believing that is presumably why the three templates were left alone. The comment is corrected here to say which of the two is in play and why.

The fix is confined to that file: declare the internal namespace, match `@pi:copied-from` in all three templates, replace the comment.

## Verification

Back to 86 `.pg` files, and the whole output tree — every `.pg` and every set definition — is byte-identical to a build from before the regression. Both entries in `set1_1-Arithmetic.def` again name the single shared file.

Also exercised against the public WeBWorK server named in the examples' publication files, which is what `make sample-chapter-representations` does: it completes without error, the regenerated set definition shares the one file, and no duplicate is created. The minimal example serves as a control, with no copies in it and identical output either way. Nothing else imports this stylesheet.

A second commit regenerates the sample chapter's committed representations from that server. The substance is four lines: a project and an exploration had drifted to different exercise numbers, so two `.pg` files are renamed with their content untouched, and the paths naming them are updated. Two files left unreferenced by the renumbering are removed. The image files carry new timestamps; their content is unchanged.


Alex-Jordan:
One thing about this: all of the committed changes to image files. Just now, after pulling `master`, I ran `make sample-chapter-representations`. And again there are diffs for all the image files. Probably due to metadata in those files, or some different version of texlive on my machine versus yours. Can you remind me if I should always commit such changes to these image files every time they come up? My instinct has been to ignore them (leave them out of a PR) when I know that the PR's topic should not affect these.

Another thing. Are we sure about the numbering? This renumbers
```
Project-1_1_1-Inside_a_project.pg
Exploration-1_1_2-Inside_an_exploration.pg
```
```
Project-1_1_9-Inside_a_project.pg
Exploration-1_1_10-Inside_an_exploration.pg
```
Basically it is putting the two project/explorations on the same counter as the inline exercises in the sample chapter. Is that how things work? Does project-like run on the same counter as inline exercise?

PR #3108, opened 2026-08-01
Update PreFigure schema (pf_schema.rnc)

davidaustinm:
This PR syncs `pretext/schema/pf_schema.rnc` with the updated version in the prefigure repository. There are no changes to the underlying PreTeXt schema.

A recently introduced snapshot testing framework in the prefigure repo, combined with validating two large corpora — the ULA linear algebra textbook and the prefigure documentation source — against the schema surfaced a number of places where the schema rejected real, engine-valid PreFigure diagrams. The fixes fall into three areas: a few typos that prevented the schema from compiling at all, interleaved content models that were rewritten in an equivalent but less expensive form, and missing attributes that the engine accepts but the schema didn't know about.

\`pf_schema.rnc\` remains namespace-free so PreTeXt's own adapter continues to supply the namespace unchanged.

### DIFF
--- a/schema/pf_schema.rnc
+++ b/schema/pf_schema.rnc
@@ -82,21 +82,27 @@ Read = element read {
     attribute delimiter {text}?,
     attribute quotechar {text}?,
     attribute string-columns {text}?,
-    attribute type {text}
+    attribute type {text}?
 }
 
+# template elements only carry default attributes, so no
+# attribute is required inside a template
 Templates = element templates {
-    GraphicalElements+
+    TemplateElement+
+}
+
+TemplateElement = element * {
+    (attribute * {text} | text | TemplateElement)*
 }
 
 # the next elements group other elements together
 GroupElements =
 (
-    Coordinates? &
-    Group? &
-    Repeat? &
-    Clip? &
-    Transform?
+    Coordinates |
+    Group |
+    Repeat |
+    Clip |
+    Transform
 )
 
 # coordinates are special in that they can contain everything a diagram can
@@ -105,12 +111,13 @@ Coordinates = element coordinates {
     attribute destination {text}?,
     attribute aspect-ratio {text}?,
     attribute preserve-y-range {"yes"|"no"}?,
+    attribute scales {text}?,
     (
-        DefinitionElements* &
-        GraphicalElements* &
-        GroupElements* &
-        TransformElements*
-    )
+        DefinitionElements |
+        GraphicalElements |
+        GroupElements |
+        TransformElements
+    )*
 }
 
 Group = element group {
@@ -119,11 +126,11 @@ Group = element group {
     attribute transform {text}?,
     CommonAttributes,
     (
-        DefinitionElements* &
-        GraphicalElements* &
-        GroupElements* &
-        TransformElements*
-    )
+        DefinitionElements |
+        GraphicalElements |
+        GroupElements |
+        TransformElements
+    )*
 }
 
 Repeat = element repeat {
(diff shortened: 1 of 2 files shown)


rbeezer:
Checked this against the merge-base, `bc9b7062`.

The thing I looked at hardest was whether the resync would quietly undo the fixes PreTeXt applied locally when the schema was first integrated — the ones that were needed to make it loadable under `jing` at all. It does not. All five are fixed upstream: the `Stroke-Attributes` typo in `grid`, the duplicate attributes on `grid`, `tick-mark`, `histogram`, `scatter`, `riemann-sum` and `vector`, the `riemann-sum` overlap between `CommonAttributes` and `AnnotatedElementAttributes`, `network/@alignment` written as a sequence rather than a choice, and the missing `tactile-dash`. So the sync brings those home rather than reverting them, which is the outcome we wanted.

The change is purely additive. Attributes go 189 to 209 and elements 71 to 73, the new ones being `center` and `set-eye`, and comparing the sorted name sets shows nothing dropped. No existing document can be invalidated by this.

`trang -I rnc -O rng pf_schema.rnc` reproduces the committed `pf_schema.rng` byte for byte, so the derived file is honest.

Validating the assembled sample article and sample book against both schemas gives **zero** new errors:

| | merge-base | this PR | new |
|---|---|---|---|
| sample article | 315 | 197 | 0 |
| sample book | 689 | 322 | 0 |

Worth being precise about that reduction, because it is not what it looks like: no error was fixed. Every construct rejected before is still rejected. The merge-base was reporting many of them *twice* — `justification` and `interactive`, for instance, each appearing as a pair. Removing `interleave` changed `jing`'s error recovery across the whole document, so the duplicates collapse. That is a genuine and welcome side effect of the de-interleaving: roughly 40% and 53% less noise in a validation report, in documents that have nothing to do with PreFigure.

PreFigure content itself is clean in both runs, on a sample article carrying 19 `prefigure` elements. As a positive control I planted a bogus element and a bogus attribute inside a real diagram, and both are still rejected through the full chain, so the schema is genuinely enforcing rather than merely loading.

Nothing else needs regenerating. `pf-adapter.rng` is a four-line live `include` of `pf_schema.rng` and `pretext.rng` reaches it by `externalRef`, so neither inlines the grammar and neither goes stale; `pretext-dev.rng` includes `pretext.rng`, so the development path picks this up too. `pf_schema.rnc` is still namespace-free, and `git diff --check` is clean with no trailing whitespace or tabs on any added line.

Two small things.

The commit subject is `update PreFigure schema with typo fixes, de-interleaved content models, and missing attributes`. Commits here take the form `Topic: short description`, capitalized — `PreFigure:` is already in use. The message is a single line with no body, which is right.

The copy of `pf_schema.rnc` in this repository does not record which upstream revision it came from — the file begins directly at `grammar {`.

That makes each resync harder than it needs to be. To see what actually changed upstream, someone has to diff the two copies of the schema against each other and work backwards, which is what I did here. If the file instead carried the upstream revision it was copied from, that same question becomes a single `git diff <recorded-revision>..HEAD` inside the PreFigure repository, which shows the upstream history rather than just the net effect.

It also answers a second question that this repository cannot otherwise answer: whether the PreTeXt copy has drifted from upstream. That is not hypothetical — when the schema was first integrated, several defects had to be patched here locally, so for a while the two copies genuinely differed, and nothing in the file said so. A recorded revision makes such drift visible instead of something to be rediscovered.

One comment line at the top would do it, naming the commit hash or a release tag and date. Possibly this belongs in the PreFigure repository, as a version marker in the file itself, rather than being added on the way in.

rbeezer:
Merged with a small rebase and an edited commit message.  Thanks for maintaining this properly.

Not sure I follow Claude's provenance question, and I certaily don't think you need a version number on your PF schema.  But maybe when you update here, a PF commit for the last time it was touched in the PF repo would give Claude more to go on?  We can chat at Drop-In.

PR #3110, opened 2026-08-03
Managed directories are the only scheme

rbeezer:
Directory management, introduced in mid-2021 as an opt-in feature, is now how every project is built.  This grew out of a pretext-dev report of assets going missing: the scheme required the external and generated directories to be declared together or not at all, so a project with nothing to declare on one side lost the benefits of the other.  The change is governed by an ownership principle: the **generated** directory is machine-owned — declared in the publication file or defaulting to `generated` next to the main source file, created on demand, never fatal when missing — while the **external** directory is author-owned — declared in `docinfo`, absent means a project simply has no external assets, and a declared-but-missing directory of an author's own files halts the build.

By commit:

* **Script: a generated directory is machine-owned, never fatal.**  `get_managed_directories` no longer verifies the generated directory (a debug note replaces a fatal error) and supplies the `generated` default when nothing is declared, so its returned pair never has an empty generated slot.  Its signature and return contract are unchanged.  Copying into the build directory skips a source directory that does not exist yet.
* **Publisher variables: managed directories are the only scheme.**  The internal `$b-managed-directories` gate is deleted along with the both-or-neither requirement and its fallback message; the generated location becomes declared-or-default; empty declared values remain fatal.  The internal directory names collapse to the constants `external/` and `generated/`, which is all they ever resolved to under the managed scheme.
* **Publication schema: the `@generated` directory declaration is optional.**  The publication-file schema now matches the behavior: the attribute, or the whole `directories` element, may be omitted.
* **Schema: "property of the source" replaces "fact of the source".**  A wording improvement in two literate-schema annotations; the other occurrences of the phrase are corrected within the commits that touch those files.
* **Deprecate: the `directory.images` string parameter is ignored.**  The parameter's default empties and a dated deprecation entry warns on any supplied value, pointing at the two real declarations.  A supplied value is the marker of legacy unmanaged use, so this commit is the complete retirement announcement for the old scheme.
* **Conversions: retire the unmanaged-directories branches.**  All nineteen consumer sites of the deleted gate collapse to their managed arms across six stylesheets; no reference to the variable remains.
* **Script: create the destination directory of `-d` when missing.**  An output destination is machine-owned, so the script creates it rather than aborting.  This also fixes a latent crash: the component-default path used a non-recursive directory creation that would fail under a fresh default generated directory.
* **Guide: managed directories are the only scheme.**  The directory-management section describes the two homes and the `generated` default, states that generated assets are placed automatically, and rewrites the backward-compatibility discussion around ownership; three other chapters lose now-vacuous "if you are using managed directories" conditionals.

Tested with same-day before/after builds: the sample article's LaTeX output is byte-identical to master and its HTML differs only in timestamps; the WeBWorK sample chapter builds with its external declaration removed, consuming the committed representations; the fonts example builds with no publication file at all; a declared-but-missing generated directory builds with a routine debug note; and the `directory.images` warning fires exactly when the parameter is supplied.  The Guide validates at its usual dev-schema baseline.

The pretext-cli is unaffected today — the `get_managed_directories` interface it uses is unchanged.  As a follow-on it may want to relax its own reading of the publication file, which currently requires both directory attributes (a file may now omit `@generated`, and `@external` lives in `docinfo`), and to adopt the same ownership rule in `ensure_asset_directories` — creating the generated directory freely, but not the external one, since silently creating an author-owned directory can mask a typo in its path.


oscarlevin:
This looks fantastic.  Tested with the CLI (using the new simpler functions, not the patches I put in temporarily) and all work as expected.  

Once you merge, I can release the new version of the CLI right away (no need for an issue there).

rbeezer:
Thanks, @oscarlevin for the quick look.  Holler if you need any help/insight for the CLI.

PR #3111, opened 2026-08-03
CSS: improve runestone colors for printouts

oscarlevin:
Currently if you have a parsons problem in a worksheet and view it when starting in dark mode, the blocks are unreadable.  This adds constants for colors that runestone components use that will be used regardless of the starting light/dark mode.

### DIFF
--- a/css/targets/print-worksheet/print-worksheet.scss
+++ b/css/targets/print-worksheet/print-worksheet.scss
@@ -62,12 +62,47 @@ $light-colors: map.merge(
   )
 );
 
+// Runestone ships its own dark-mode CSS variables that are independent of
+// PreTeXt theme variables; force those to light values in worksheet previews.
+$runestone-light-colors: (
+  "background": #fff,
+  "outerBackground": #eee,
+  "links": #0645ad,
+  "bodyFont": #000,
+  "tooltip": #fff,
+  "grayToWhite": #333,
+  "navbar": #f8f8f8,
+  "navbarFont": #707070,
+  "navbarFontHover": #000,
+  "alerts": #d9edf7,
+  "completionButton": #f19711,
+  "completionButtonHover": #ff9f19,
+  "completionButtonText": #000,
+  "admonition": #fcf8e3,
+  "admonitionBorder": #fbeed5,
+  "codeButtons": #474949,
+  "codeButtonsBorder": #474949,
+  "dangerAlerts": #f2dede,
+  "successAlerts": #dff0d8,
+  "componentBgColor": #f0f8ff,
+  "componentBorderColor": #000,
+  "questionBgColor": #e1efff,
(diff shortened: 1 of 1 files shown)


rbeezer:
Small and tightly scoped. `cssbuilder` runs clean and touches **exactly one** dist file, `print-worksheet.css`, adding 26 lines and removing none, so no other theme is disturbed. The generated CSS is precisely the intended block — the 24 custom properties under `:root.dark-mode`, each `!important`, and nothing else. One commit, `CSS:` topic, no body, no `css/dist` in the diff, whitespace clean.

The approach already has a precedent here. `css/targets/html/legacy/wide/shell_wide.scss` line 38 does

```scss
:root {
  --questionBgColor: var(--componentBgColor) !important;
}
```

under a comment reading "temp override runestone injection", so forcing Runestone-injected variables this way is established rather than new. That is also the only one of your 24 names that appears anywhere else in the CSS sources, and it is in a different stylesheet, so the two never meet.

Worth recording one thing that reads as an inconsistency and is not: the neighbouring PreTeXt block spans `:root, :root.dark-mode`, while this one is `:root.dark-mode` alone. Since these are Runestone's *dark-mode* variables, light mode already supplies light values and there is nothing to override, so the narrower selector is the minimal correct choice. I nearly raised it before re-reading your comment.

I built the sample book with `publication-noparts.xml` and confirmed the components are real and local on `worksheet-groupwork` — `data-component="parsons"` and `data-component="multiplechoice"`, plus `answer` and `feedback` — and that the rebuilt stylesheet carrying the override is what `pretext_add_on.js` pulls into the print preview. The remaining judgement, whether the blocks are now legible from dark mode, is yours; I did not put eyes on the rendered page.

Two minor things, neither blocking.

The new loop re-implements a mixin the file already imports. `colorHelpers.scss-to-css`, used four lines above, is exactly `@each $name, $value in $colors { --#{$name}: #{$value}; }` — the only difference here is `!important`, and SCSS would let that be a parameter (`@mixin scss-to-css($colors, $important: false)`). Against that, changing a shared mixin touches every caller to save five lines, so hand-rolling is defensible; a word on why would close the question either way.

Nothing records where the 24 values came from. Presumably they are Runestone's light-mode defaults, but if Runestone changes one, the copy here drifts and there is no way to notice. A comment naming the source file would make a future mismatch traceable.

rbeezer:
Thanks - merged with a rebase and a commit message edit.

What are the odds of working out CSS so all Runestone components, everywhere, look like they belong to the relevant theme?

oscarlevin:
> What are the odds of working out CSS so all Runestone components, everywhere, look like they belong to the relevant theme?

Yeah, wouldn't that be great?  Maybe we should chat on Tuesday.

PR #3112, opened 2026-08-03
Gaurd addKnowls against born-hidden-knowl elements missing a summary

kylemonette:
First PR attempt here; hopefully everything goes smoothly! 

While preparing print-preview handouts for my course, I ran into a JavaScript bug in PreTeXt's built-in print-preview feature that causes solution content to occasionally (in Safari, almost always) render as raw, un-typeset LaTeX source.

Print-preview mode converts each solution's details/summary into a plain div (in rewriteSolutions(), pretext_add_on.js) but leaves the born-hidden-knowl class on it. addKnowls() (knowl.js) then matches these divs, finds no summary child, and throws inside new SlideRevealer(), aborting the load handler.

Separately, rewriteSolutions() never calls MathJax on the newly-created solution content, and pagination (createPrintoutPages()/adjustPrintoutPages()) measures element heights immediately after. Since solution math is normally typeset lazily on knowl-open (never triggered by print preview), whether it renders becomes a pure race against MathJax's own unrelated auto-typeset pass — reproducible almost every time in Safari, intermittently in Chrome.

This PR: (1) guards addKnowls() to skip elements missing a summary/contents pair instead of throwing, and (2) makes rewriteSolutions() async, awaits MathJax.typesetPromise() before returning, and awaits it at the call site — so pagination always runs after solution math is fully typeset.

Tested via repeated fresh reloads (cache-busted) of a live print-preview page in both Chrome and Safari; no console errors and no unrendered LaTeX after the fix.

### DIFF
--- a/js/dist/knowl.js
+++ b/js/dist/knowl.js
@@ -10,6 +10,9 @@ function addKnowls(target) {
   for (const bhk of bornHiddens) {
     const summary = bhk.querySelector(":scope > summary");
     const contents = bhk.querySelector(":scope > summary + *");
+    if (!summary || !contents) {
+      continue;
+    }
     new SlideRevealer(summary, contents, bhk);
   }
 }
--- a/js/dist/pretext-core.js
+++ b/js/dist/pretext-core.js
@@ -40,6 +40,9 @@
         for (const bhk of bornHiddens) {
           const summary = bhk.querySelector(":scope > summary");
           const contents = bhk.querySelector(":scope > summary + *");
+          if (!summary || !contents) {
+            continue;
+          }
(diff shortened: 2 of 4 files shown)


oscarlevin:
This is fantastic!  Good sleuthing and the cause makes perfect sense.  

@rbeezer, let me know if you want me to manage this through.

oscarlevin:
Thank you very much for this contribution, it is very valuable.  I split your commit in two, the first with your work, the second with the building of the js dist files, and then slightly edited your commit message.  Both of those are now merged into the master branch, with you as the author on the contribution.

With this workflow we just close the PR, since the commits are now in master.  Thanks again.

PR #3114, opened 2026-08-04
Re-run pagination when toggling solutions in print preview

kylemonette:
Solution/hint/answer visibility toggle in handouts and worksheets, e.g., doesn't re-paginate print-preview pages, causing content to bleed past page boundaries.

Toggling the hide-hint/answer/solution checkboxes in print preview only adds/removes a "hidden" class and calls adjustWorkspaceToFitPage() — which balances blank workspace height but never recomputes page breaks. When revealing a solution adds real height, its page doesn't grow or reflow, so content spills past the fixed-height .onepage boundary. The code already had a commented-out attempt at this (//adjustPrintoutPages();), but that function alone only handles content outside the first/last page, not overflow within an existing page.

This PR adds resetPrintoutPagination(), which unwraps existing .onepage sections back to flat content (also removing their header/footer divs, which get freshly rebuilt), then re-runs the real createPrintoutPages() + addHeadersAndFootersToPrintout() pipeline — effectively replaying initial pagination against the current (post-toggle) content heights. The checkbox handler now calls this before adjustWorkspaceToFitPage(), matching the original page-load setup sequence.

### DIFF
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -1432,6 +1432,7 @@ window.addEventListener("DOMContentLoaded", async function(event) {
                         } else {
                             elem.classList.add("hidden");
                         }
+                        // Recompute layout once, after all elements of this type have been toggled
                         adjustWorkspaceToFitPage({paperSize: paperSize, margins: margins});
                     });
                 });
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -642,10 +642,14 @@ function pageOverflows() {
     return false;
 }
 
-function adjustWorkspaceOrRepaginate({paperSize, margins}) {
+function adjustWorkspaceOrRepaginate({paperSize, margins, fullRecompute = false}) {
     adjustWorkspaceToFitPage({paperSize, margins});
     if (pageOverflows()) {
-        addSpilloverPages(margins);
+        if (fullRecompute) {
+            resetPrintoutPagination(margins);
+        } else {
+            addSpilloverPages(margins);
+        }
         adjustWorkspaceToFitPage({paperSize, margins});
     }
 }
@@ -1002,8 +1006,10 @@ function findPageBreaks(rows, pageHeight) {
         }
     }
     // Backtrack to find the actual page breaks based on nextPageBreak
-    // Note: the nextPage = 1 is not an indexing mistake; we always assume that row 0 is a title and will go on the same page as row 1.
-    let nextPage = 1;
+    // Note: nextPage used to be set to 1.
+    // This meant the very first page's title height was never counted against that page's budget during optimization, even though it still occupied real space once pages were built.
+    // This caused the first page to sometimes exceed capacity, and the resulting correction to cascade into large wasted-space gaps on later pages.
+    let nextPage = 0;
     while (nextPage < rows.length) {
         pageBreaks.push(nextPageBreak[nextPage]);
         nextPage = nextPageBreak[nextPage];
@@ -1267,6 +1273,7 @@ function toPixels(value) {
 // Event listener for page load to handle print preview setup
 window.addEventListener("DOMContentLoaded", async function(event) {
     const urlParams = new URLSearchParams(window.location.search);
+    let pendingSettle = Promise.resolve();
     // We condition on the existence of the papersize radio buttons, which only appear in the printout print preview.
     if (urlParams.has("printpreview")) {
         const printableSectionID = urlParams.get("printpreview");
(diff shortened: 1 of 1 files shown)


davidfarmer:
It may also be necessary to re-run MathJax when revealing a hint or 
solution.

kylemonette:
David -- This might be the case; I can look into it further. When I was doing some testing earlier, however, it didn't seem necessary to rerun MathJax. Perhaps because of PR #3112 ?

oscarlevin:
This one is tricky.  There is a reasonable hope that many instructors have that solutions will "eat up" the workspace for their exercise when they are revealed.  I *think* we accomplish this by recomputing workspace but NOT recomputing pages.  Then anyone who wants to keep the original amount of workspace can reload after toggling solutions.

Perhaps a better solution is to add a "recompute page breaks" button or link that does just that?  I understand that when a multi-page scorm file is used, reloading the browser has bad side-effects.

kylemonette:
Hi Oscar - thank you for the input here! I like the idea of recomputing the page breaks with a button, but my concern is that students who want to print the worksheet/handout may not do this, and then have things cut off at the bottom. This was the original reason I started to investigate this.

My (albeit limited) investigation here is that sometimes when solutions are displayed, the workspace tends to remain below the solution (does not get eaten up) and this extra white space lengthens the document unnecessarily. But this is just my opinion here; perhaps this is desirable for some instructors.

kylemonette:
I looked more into this, and have an alternative solution, in the commit above.

Toggling to reveal a solution now tries to absorb the extra height into the existing workspace first, exactly like the current behavior, so in the normal case nothing changes and solutions just eat up the workspace, as expected. It only falls back to recomputing page breaks if a page would still overflow even with workspace shrunk to zero — so students printing without reloading should never see cut-off content, but instructors won't see extra pages appear for typical solutions.
Also, it seems necessary to always recompute page breaks when turning solutions from visible to invisible, since otherwise pages could get stuck until refresh.

oscarlevin:
Nice.  I'll take a look soon.  This makes me think that we should probably do something with pages that would overflow anyway (regardless of solutions).

oscarlevin:
I think we are getting closer, but there are still some common use cases this seems to break.  If you try it on the sample article and look at the 4th worksheet of section 37, you will see what I mean.  That's a worksheet that has authored pages.  Removing the onepage elements and recomputing destroys the authors intended page breaks.  This happens even if there are no solutions to try to fit.  

The issue is that when an author specifies pages, we simply shrink workspace to accommodate. But if there are not authored pages, then we use the originally authored workspace to guess the page breaks.  So a worksheet that fits on two pages with too-much authored workspace might end up jumping to 3 or 4 pages as a result of clicking the toggle.

What do you think about a function that just added a "spillover" page for each page that has too much content?  That would mean pages always start at the same spot, and some pages only have a couple of lines on them (the "extras").

kylemonette:
This was a great catch! A new commit was added.

Instead of recomputing every page, the fix now only touches a page if it actually overflows. Revealing a hint/answer/solution first tries to absorb the extra height into the existing workspace area (unchanged from before, since that's the default behavior we want to preserve), and only if a page still overflows even with workspace shrunk to nothing does it spill the excess content onto one new page inserted right after — every other page, authored or not, is left completely alone. Hiding content again reverses this by merging a spillover page back into its predecessor once the content fits again, rather than recomputing everything.

I tested the previous commit on Worksheet 37.4 in the sample article and noticed the error you were describing. This appears to be resolved with the new version.

kylemonette:
Fixed several remaining issues in the print-preview pagination logic. Withstood a lot of testing from worksheets designed specifically to have problematic authored/non-authored page breaks.

- Infinite page-creation loop — a single content row taller than one page caused addSpilloverPages() to loop forever creating new pages, freezing the tab. Added a guard so it stops trying to split a row that's inherently too tall, instead of retrying endlessly.
- That guard was too blunt — it gave up on an entire page if the first row didn't fit, even when later rows could still move to a fresh page. Refined so only the oversized row is left behind; everything after it still gets relocated.
- Overflow uncorrected on a fresh page load — the overflow-handling logic only ran on checkbox toggle, never at initial setup, so a page loading with solutions already visible (e.g. a returning visitor) could show uncorrected overflow. Routed initial setup through the same safety net.
- Timing race in that fix — some content (large matrices especially) could still be rendering when the check ran. Replaced a single deferred check with a short poll (every 100ms for up to 2s) on both the initial-load and toggle paths.
- Hiding solutions after a visible-from-load page didn't collapse correctly — for worksheets with no authored page breaks, the initial layout naturally uses more pages when solutions start visible, and hiding them only undid spillover pages, not that. Now tracks authored vs. computed pagination and does a full, safe recompute on hide for computed worksheets (preserving the existing structure-safe behavior for authored ones).

kylemonette:
Building on the earlier spillover/margin fixes, this commit addresses two deeper, related issues.

1. **Fixed a title-row bug in `findPageBreaks`.** The dynamic-programming page-break optimizer correctly computed a cost-minimized chain for every row, but its final backtracking step started from row 1 instead of row 0 (an intentional-looking shortcut to keep the title with the first page's content). This meant the very first page's title height was never counted against that page's budget during optimization, even though it still occupied real space once pages were built — causing the first page to sometimes exceed capacity, and the resulting correction to cascade into large wasted-space gaps on later pages. Fixed by starting backtracking from row 0, letting the DP's existing cost function handle the title like any other row.
2. **Fixed a race condition between overlapping pagination operations.** Several safety-net operations (initial-load settling, reveal-time overflow correction) ran as untracked, fire-and-forget async work. If a user toggled solutions while one of these was still in flight, two pagination passes could run concurrently and step on each other. Replaced ad-hoc retry delays with a single shared `pendingSettle` promise: every settling operation now registers itself there, and every new action (checkbox toggle, etc.) awaits it before proceeding.
3. **Extended structure-aware repagination to the reveal path.** Hiding solutions already chose between a structure-preserving local patch (authored pages) and a full recompute (computed pages). Revealing solutions always used the conservative local patch (`addSpilloverPages`), even on computed-pagination worksheets with no structure to protect — so overflow correction there just stranded the overflowing block on its own new page instead of optimizing the whole layout. `adjustWorkspaceOrRepaginate` now takes a `fullRecompute` flag, and both the reveal handler and initial-load setup pass `!hasAuthoredPages`, so computed worksheets get a genuine global re-optimization on reveal.

rbeezer:
@kylemonette - thanks for your work on this one.  When you have it ready for @oscarlevin to review, please add "topics" to each commit message, as is house style.  `JavaScript:` likely suffices for all of them.  ;-)

kylemonette:
I believe this will be the last commit here, so it should be about ready for review, @oscarlevin . I apologize if this is getting too long! Things kept popping up... Fingers crossed that this will do it. And, there were some cleanup fixes here too.

The most recent commit fixes authored-page-breaks failing to spill overflowing exercise/task content onto new pages, and failing to collapse spillover pages back down when that content shrinks again (e.g. toggling solutions off). Seems that tasks nested inside an exercise were never split into independent units for authored pages (only for computed pagination), so one oversized task solution could strand an entire exercise as an unsplittable block. The collapse logic decided whether a merge "fit" before workspace boxes got a chance to shrink back down to size, so valid merges were rejected based on stale layout.

Also fixed a footer-ordering bug during page merges and unified a few functions onto the shared `getPrintout()` lookup for consistency.

kylemonette:
I found some issues when an `<introduction>` for an exercise in a worksheet exceeded the page break, and dived into this again.

1. `flattenTasksIn`'s loop skipped the first `.task` in an exercise, so a single-task exercise stayed glued to its introduction as one unsplittable block.
2. a dead ``printout.style.width = toString(...) + 'px'`` assignment (`toString` isn't a real function in that scope) meant row heights were measured at the printout's full width instead of the narrow width the algorithm assumes, so text wrapped less during measurement than in the final padded page and `findPageBreaks` grouped rows that didn't actually fit.
3. neither `.introduction` nor a `task`'s `.solutions` block were ever split for pagination purposes, so a long introduction or a task with a large solution could overflow a page by itself even after the above fixes. The fix adds two new functions (`flattenIntroductionsIn`, `flattenSolutionsIn`) mirroring the existing `flattenTasksIn` pattern — pulling paragraph-level content out to independent, independently-placeable rows — and wires them into both the computed-pagination path (`createPrintoutPages`) and the authored-pagination path (`adjustPrintoutPages`).
4. on authored-page documents, content that's still settling into its final size right after load could get split onto a spillover page by the overflow check that runs after load. Nothing ever revisited that decision once the content settled down, since the only recovery path (`addSpilloverPages`) only ever adds pages, never merges one back. This showed up as a task sitting alone on a page with hundreds of pixels of empty space while its solution sat on the next page, and only went away on reload. Now, it tries `collapseSpilloverPages()` before every re-check, same pattern the "hide solutions" toggle already used.

oscarlevin:
Thanks very much for this work.  I've squashed your commits into one and followed up with a few needed tweaks before merging into the master branch.

I especially like that we now handle page overflows in general.

PR #3115, opened 2026-08-04
Tablenotes css

ascholerChemeketa:
Changes discussed on -dev

One additional change - I turned the tablenotes into a list (instead of a div). Semantically it is more accurate and improves keyboard/screen reader navigation. (Leaving the table into the list will announce "List - table notes" and allow easy skipping past it.)

----------------

This pull request updates the way table notes are rendered in HTML output, improving both accessibility and semantic structure. The changes include switching from `div` elements to semantic lists for table notes, adding ARIA labels for better accessibility, and updating CSS for the new structure.

**Accessibility and HTML structure improvements:**

* Table notes are now rendered as an ordered list (`<ol class="tablenotes">`) with each note as a list item (`<li class="tablenote">`), replacing the previous use of `div` elements. This improves semantic HTML and screen reader support.
* ARIA labels have been added to the table notes list and note references to improve accessibility for users with assistive technologies. [[1]](diffhunk://#diff-d6cc23bab210bdd0e195069306f6a2bc727005926cc003aa3cbe7b790713519fL8169-R8209) [[2]](diffhunk://#diff-d6cc23bab210bdd0e195069306f6a2bc727005926cc003aa3cbe7b790713519fR2756-R2772)
* Table note references now use more descriptive `href` and `aria-label` attributes and improved ID generation for better navigation and accessibility. [[1]](diffhunk://#diff-d6cc23bab210bdd0e195069306f6a2bc727005926cc003aa3cbe7b790713519fR2756-R2772) [[2]](diffhunk://#diff-d6cc23bab210bdd0e195069306f6a2bc727005926cc003aa3cbe7b790713519fL8169-R8209)

**Styling updates:**

* New CSS rules for `.tablenotes` and `.tablenote` are added to style the ordered list and its items, ensuring the new structure is displayed correctly.

### DIFF
--- a/css/components/elements/_tables.scss
+++ b/css/components/elements/_tables.scss
@@ -114,6 +114,17 @@
   margin-right: auto;
 }
 
+.tablenotes {
+  margin: 0.5em 0 0;
+  padding-left: 0;
+  list-style: none;
+  font-size: 80%;
+
+  .tablenote {
+    margin-top: 0;
+  }
+}
+
 // More specific types of tables
 
 table.notation-list {
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -2753,14 +2753,23 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <!-- specific notes at 3.80).  Adjacency is the presentation: a note   -->
 <!-- sits just below its table, so there is no knowl.                  -->
 <xsl:template match="tabular//tn">
+    <xsl:variable name="note-number">
+        <xsl:apply-templates select="." mode="number"/>
+    </xsl:variable>
     <sup class="tablenote-mark">
+        <xsl:apply-templates select="." mode="html-id-attribute"/>
         <a>
             <xsl:attribute name="href">
                 <xsl:text>#</xsl:text>
-                <xsl:apply-templates select="." mode="unique-id"/>
+                <xsl:apply-templates select="." mode="html-id"/>
+                <xsl:text>-note</xsl:text>
+            </xsl:attribute>
+            <xsl:attribute name="aria-label">
+                <xsl:text>Go to table note </xsl:text>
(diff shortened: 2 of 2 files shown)


rbeezer:
The semantic list is a clear improvement, and the wiring is sound. Across the sample article's table notes I checked **16 links: none broken, no duplicate ids**. The mark carries `id="<html-id>"` and points at `#<html-id>-note`; the item carries `id="<html-id>-note"` and points back. Two commits, one topic each, no bodies, no `css/dist` in the diff, whitespace clean.

Two things I looked at that turned out fine. The id effectively moved from the note block to the mark, which would matter if a table note were a cross-reference target — it isn't, and the deprecation message in `pretext-common.xsl` says so outright: "a table note ... is not a target for a cross-reference." And `mode="html-id"` just delegates to `mode="unique-id"`, so swapping which one the `href` uses changes nothing. The three new literal `aria-label` strings are consistent with the file, where 13 of the 14 aria-labels are literal English.

One follow-up, not a problem with this code. A `div` needs no CSS to render correctly, whereas an `<ol>` needs the `list-style: none` you added — and that component doesn't reach quite everywhere PreTeXt HTML is rendered. The modern themes, EPUB and Kindle are covered; Reveal.js, Jupyter and the legacy themes aren't, so a table note there would pick up the browser's own list marker alongside its letter. Jupyter is the awkward one, since its CSS is written inline by `pretext-jupyter.xsl` and can't take anything from `css/components/`.

Nothing for you to do here — I'll open an issue for it shortly.

rbeezer:
Looks real good.  Merging and then posting a follow-up issue.

rbeezer:
Merged as-is, just a rebase onto new master.  Love the improved sematics and the twin-linking.  And it looks much better with the extra CSS - thanks for taking this one up.  See https://github.com/PreTeXtBook/pretext/issues/3121

PR #3116, opened 2026-08-04
LaTeX: footnotes are native, even within boxed environments

rbeezer:
A footnote inside a block implemented as a `tcolorbox` (an `example`, `remark`, `figure`, and many more) has long been handled by a two-part dance: a bare `\footnotemark{}` at the location, then a second pass after the outermost enclosing environment that re-scanned its contents and emitted `\footnotetext[N]` with the number hard-coded from the assembly's footnote serial.  Correctness depended on two hand-synchronized twenty-entry lists of block types, and they had drifted again — the second list was missing the open-problem blocks, the same class of defect as the 2023 `biblio/note` bug remembered in the code comments.

Footnotes are now native LaTeX everywhere.  The `fn` template is simply `\footnote{}`.  Each block that can harbor a footnote wraps its environment in the `savenotes` environment of the `footnotehyper` package (TeXLive-standard, the hyperref-safe successor of the venerable `footnote` package), which delays each footnote's text to the true page bottom; the wrapping is transparent when no box intervenes and nests correctly, so a footnote works at any depth of boxes-within-boxes.  A small companion command restores the standard footnote counter inside each box (a box otherwise switches to the letter-marked minipage counter); it rides on the existing paragraph-style hook, which every paragraph-organized box already runs at exactly the right moment, plus one hook for lower box parts, where figure captions typeset.  The dance, the hard-coded serials, and both drifting lists are deleted, and LaTeX output no longer consumes the assembly footnote serial at all.

With the mark/text split gone, the reason hyperref was loaded with `hyperfootnotes=false` evaporates, so the second commit removes it: electronic PDFs gain a hyperlink from every footnote mark to its text.

Numbering is unchanged in practice: before-and-after builds of the sample article (292 pages) and its solutions-manual variant (291 pages) show every page-bottom note identical in mark, text, and page — native numbering lands on exactly the numbers the old scheme hard-coded.  One behavior is unchanged and now documented precisely in the Guide's print-compromises list: a footnote's text is held until its block finishes, so a footnote early in a multi-page example lands at the bottom of the page where the example ends.

The classic (journal-ready) conversion inherits all of this through `pretext-latex-common.xsl` and may be the biggest beneficiary.  The hard-coded serials were computed for the styled conversion's numbering, so under classic's own footnote counter the old scheme produced texts whose numbers disagree with their marks, one duplicated note, and a malformed run-in, all visible in a sample-article build; with native numbering one counter produces both mark and text, so they cannot disagree, and the same build renders every footnote correctly.  Where the wrapping surrounds one of classic's plain (non-box) environments it is verifiably inert: a footnote there still sets at the bottom of its own page, even when the environment spans a page break.

The sample article gains "Footnotes in Nested Boxes," extending its test-footnote series: one footnote in an `example`'s paragraph, one in an `aside` within it, and one in a description-list item within it — three box implementations, two levels deep, numbering continuously with the surrounding series.  Along the way it was confirmed that a `fn` inside a `caption` or `title` is not schema-valid (the long-text union carries no footnote), so the old caption `\protect\footnotemark` special case was guarding an unauthorable construct.


rbeezer:
@oscarlevin - see above relative to the "classic" conversion.  Sounds like footnotes were a bit messed-up, at least for the sample article (which I know is really out-of-scope).

I think you may see `\begin{savenote}` with an `\end` in classic output.  I think mostly it is doing no harm.  For more-classic output, maybe we want to skip it.  You have no use of `tcolorbox`, right?  We could throw a flag in the classic conversion, and condition on it in -common, to not write this environment ever.

There is one small pramble change, LaTeX package `hyperfootnote` (or similar, and I think you want that.

PR #3120, opened 2026-08-05
Embedded page reading controls

oscarlevin:
When an instructor embeds a pretext page in their LMS (or using any iframe) we offer the `?embed` query parameter that hides navigation, masthead, and footer.  This works well, but now that we have reading controls and read-aloud for pages, those new conveniences are lost inside the LMS.

This PR moves just the "other" navigation controls (the two above, and also the embedded calculator if it is used) to a floating and hopefully unobtrusive group of buttons in the top-right of the page.

### DIFF
--- a/css/components/page-parts/_embed.scss
+++ b/css/components/page-parts/_embed.scss
@@ -1,0 +1,133 @@
+// Styles for a page loaded with the "embed" URL parameter, as an instructor's
+// LMS iframe loads it.  pretext-embed.js hides the masthead, footers and
+// sidebar outright; the navbar is the one page part it keeps, because the
+// reading controls live inside it and those are worth keeping.  Everything
+// here restyles that surviving navbar into a small floating cluster.
+//
+// Included via _navbar.scss so every theme with a navbar gets it.  The rules
+// only bite when JavaScript has put .ptx-embedded on the body.
+
+$nav-height: 36px !default;
+
+// The cluster's inset from the top-right corner of the iframe, and the corner
+// rounding of the cluster itself.  Both are deliberately independent of the
+// theme's button radius: navbar buttons are square-cornered in most themes
+// because they sit in a full-width bar with nothing to round against, which
+// stops being true the moment the bar becomes a floating object.
+$embed-controls-inset: 0.75rem !default;
+$embed-controls-radius: 5px !default;
+
+body.pretext.ptx-embedded {
+  // The navbar reserves space for itself at the foot of the page below the
+  // breakpoint, and read-aloud reserves a second row on top of that.  Neither
+  // reservation means anything once the bar is floating at the top, and both
+  // would show as dead space under the last line of content.
+  //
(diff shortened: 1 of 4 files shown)


rbeezer:
This works nicely. I rebuilt both `css/dist` and `js/dist`, served the sample article, and looked at it: `?embed` gives the three-button cluster in the top-right — calculator, read-aloud, readability — with the masthead, sidebar, both footers and all navigation gone and the content filling the frame. Opening "Aa" shows line height, font size, assistive technologies and reset, with no theme group, so the URL's choice really is enforced and the reader really cannot override it. Without `?embed` the page is unchanged.

The rest holds up too. Every class the new CSS targets is present in the output — `nav-other-controls`, `ptx-navbar-contents`, `ptx-readability-options-group-theme`, `ptx-embed-button`, `ptx-read-aloud-player`, each on 62 pages. The new `import` is sound rather than duplicating a module, since `pretext-core.js` pulls in `readability-options.js` at line 18 and `pretext-embed.js` at line 21, so they share a bundle. Two commits, one topic each, no bodies, no `dist` in the diff, whitespace clean. The comments are better than most: the specificity arithmetic behind matching the `:has()` rule, and why the cluster is `fixed` rather than `sticky` inside a 1000px iframe, are both worth having written down.

Three small things.

The comment on the new export in `readability-options.js` describes behaviour the code does not have:

> pretext-embed.js needs to know whether the reader has chosen a theme of their own before it applies the one the embed URL asks for

Nothing checks that. `applyEmbedTheme` maps the parameter to `"dark"` or `"light"` and calls `applyThemeChoice`, which consults the saved preference only in its `"system"` branch — never reached from here — and the theme control is hidden outright. Your comment in `pretext-embed.js` has it right: "We always enforce this theme and remove the choices from the reader." It is the export comment that wants correcting.

`_embed.scss` line 120 runs to 136 characters where the file's other 56 comment lines wrap near 78, and it ends `LMS iframe.,` with a stray comma after the period.

And `import {applyThemeChoice }` is missing its opening space; every other named import under `js/src` uses `import { x }`.

One thing I checked and am not raising. Moving the navbar-hiding from JavaScript into CSS means embed view now depends on rules that reach only the modern themes — there are ten `ptx-embedded` rules in `theme-default-modern.css` and none in any `*-legacy`. Under a legacy theme `?embed` shows the full navigation bar, a stuck-open dropdown and a floating search panel. But building the merge-base the same way gives an identical page, so legacy themes did not support `?embed` before this either. Pre-existing, and not something this PR needs to carry.

rbeezer:
Very good!  A couple of nits in Claude's review you could pick up sometime.  Merged as-is after a rebase, CSS and JS updated.

Should I be able to drag the floating toolbar around?  There is nothing for my mouse to latch onto - it seems to be all button.  Maybe a few pixels of raw panel around the edges?  Or maybe that is not the intent?

oscarlevin:
Dragging is not intended. I don't know how useful that would be.

Issue #3121, opened 2026-08-05
Table notes need their list marker suppressed in Reveal.js and Jupyter


rbeezer:
Follow-up to #3115, as promised in review there.

Table notes are now marked up as an ordered list:

```html
<ol class="tablenotes" aria-label="Table notes">
  <li class="tablenote" id="...-note">...</li>
</ol>
```

The previous markup was a `div`, which needs no CSS to look right. An `<ol>` does: without `list-style: none` the browser adds its own `1.` `2.` beside the note's own letter. #3115 supplies that rule in `css/components/elements/_tables.scss`, and the maintainer dist rebuild carried it to nine stylesheets — `epub`, `kindle`, `print-worksheet` and the six modern themes. That is every target whose CSS is compiled from `css/components/`, and it leaves two that render PreTeXt HTML by other means.

## Reveal.js

`css/targets/revealjs/reveal.scss` deliberately takes almost nothing from the shared components — one line, `@use "components/elements/list-styles"` — so it has no `.tablenotes` rule. Adding a `<tn>` to the table already in `examples/sample-slideshow` and building `-f revealjs` emits the `<ol>`, and none of the stylesheets that slideshow loads (reveal's `reset.css` and `reveal.css`, prism's three, `external/custom.css`) mentions `tablenotes`.

## Jupyter

`xsl/pretext-jupyter.xsl` writes its own stylesheet into every notebook as string literals in `load-css`, scoped to the legacy `.mathbook-content` prefix. Nothing under `css/components/` can reach it. Building `-f jupyter` gives the same `<ol>`, and that 3690-character `<style>` block has no tablenote rule; its one `list-style: none` belongs to `.summary-links`.

## Suggested fix

Pulling `components/elements/tables` into Reveal.js would work but brings the rest of that file with it — `.ptx-content` table borders, `.tabular-box`, `table.notation-list` — imposing PreTeXt's whole table look on slides. Extracting the rule keeps one source and a small blast radius, and matches how `reveal.scss` already takes a single focused partial.

A new `css/components/elements/_tablenotes.scss` (GPL header omitted here):

```scss
// Table notes are marked up as an ordered list, so every target that
// renders PreTeXt HTML must suppress the list marker: the note carries
// its own letter.  Kept apart from  tables  so a target can take this
// alone, without the rest of the table styling (reveal.js does).

.tablenotes {
  margin: 0.5em 0 0;
  padding-left: 0;
  list-style: none;
  font-size: 80%;

[...]

PR #3127, opened 2026-08-06
Jupyter/RevealJS: add css for tablenotes

ascholerChemeketa:
For https://github.com/PreTeXtBook/pretext/issues/3121#issuecomment-5198029461

Untested for Jupyter. Had to add sample to the reveal slide show.

### DIFF
--- a/css/targets/revealjs/reveal.scss
+++ b/css/targets/revealjs/reveal.scss
@@ -170,3 +170,15 @@ pre[class*="language-"] {
 dfn {
   font-weight: bold;
 }
+
+.reveal .tablenotes {
+  margin: 0.5em auto;
+  width: fit-content;
+  display: block;
+  font-size: 80%;
+  list-style: none;
+}
+
+.reveal .tablenote sup i {
+  color: var(--r-link-color);
+}
--- a/examples/sample-slideshow/sample-slideshow.xml
+++ b/examples/sample-slideshow/sample-slideshow.xml
@@ -313,13 +313,13 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                         <col/>
                         <col/>
                         <row bottom="minor">
-                            <cell>Format</cell>
+                            <cell>Format<tn>A table note.</tn></cell>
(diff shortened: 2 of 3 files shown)


rbeezer:
Both halves check out, including the Jupyter one you flagged as untested.

**Reveal.js**: built the slideshow and looked at the slide — no list markers, notes at 80%, centered under the table. Deleting just the two new rules through the CSSOM, which is a faithful "before" since nothing else in any loaded stylesheet mentions `tablenotes`, brings back the `1.` `2.` markers with the block flush to the left edge while the table stays centered.

**Jupyter**: the selector does reach. The notebook cell holding the `<ol class="tablenotes">` opens with `<div class="mathbook-content">`. Measuring that cell against the notebook's own style block gives `list-style: none` and `padding-left: 0` with your line, `decimal` and `40px` without it. Worth noting that `padding-left: 0` is needed in Jupyter and unnecessary in Reveal.js, where reveal's own reset already computes it to `0` — so the difference between your two rules is right.

One question. Does `.reveal .tablenote sup i { color: var(--r-link-color); }` do something you were after? The `<i>` sits inside the note's back-link,

```html
<sup><a href="#…" aria-label="Back to table note reference a"><i>a</i></a></sup>
```

and has no color of its own, so it inherits from that `<a>`, which the theme already colors from `--r-link-color`. With and without the rule I measure `rgb(38, 139, 210)` both ways on the solarized theme the sample slideshow loads. Were you working around a different theme, or aiming at the mark in the table cell rather than the one in the note?

On the design, since I wrote the issue: what it proposed was a shared `_tablenotes.scss` partial for Reveal.js to `@use`, and your target-specific rule is the better call. Slides center their tables, so the shared rule's `margin: 0.5em 0 0` would leave the notes flush left under a centered table — which is exactly what the "before" looks like. A shared partial could not have served both without parameterizing it.

rbeezer:
Merged as-is, with a rebase and a commit message edit.  Thanks very much for taking on this one.  Closes #3121

PR #3132, opened 2026-08-06
WeBWorK: the sample chapter and the schema agree

rbeezer:
The same harmonization the Runestone chapter received in PR #3128, now
for WeBWorK: the sample chapter carried 57 development-schema errors,
and every one traced to the schema trailing forms the code has always
honored — or, in one case, to a deliberate tripwire living in the wrong
document.  After this branch the chapter validates at zero.

A correction along the way: an earlier survey claimed the `@copy`
mechanism was never exercised.  Wrong twice over — the raw source
exercises it twice, and better still, the schema *rejected* the form,
treating `@copy` as an optional rider on a full problem rather than
what it is: an empty, lightweight reference the assembly resolves.
(The validation stage runs before that resolution, so the schema must
accept the authored form.)

- a4cbaba1 (Schema) Four forms, each verified against its consumer in
  `extract-pg.xsl` before modeling: the empty `@copy` reference
  (optionally reseeded); `var/@data="pgml"` (the variable holds PGML
  markup to be processed); `array` joins the `@form` enumeration (an
  explicit extractor branch); and the PG-drawn image — `image` with
  `@pg-name` — admitted schema-wide on the established `var` pattern,
  since it appears deep inside list items where a grammar cannot see
  the `webwork` ancestor, with a new rule in the validation-plus
  stylesheet confining it to WeBWorK problems (mirroring the `var`
  rule).  A `shortdescription` may carry an inline `var`, same pattern.

- 3e6c2a59 (Chapter) The six bare-text image descriptions are authored
  as explicit `shortdescription` — their true identity: a 2023 repair
  has been silently converting text-only descriptions to
  `shortdescription`, which is the only reason they ever produced PG
  alt text.  The obvious fix — wrapping them in `p` to satisfy the
  long-description model — validates perfectly and silently destroys
  the alt text (the wrapper blocks the repair, and the extractor has no
  long-description path); a before/after build of the problem sets
  caught it.  The final form produces PG output byte-identical to
  master.

- f908784b (Errors) The deliberate empty-statement exercise — "It
  should throw a PTX warning during the representations build" — moves
  to the errors document, joining its Broken WeBWorK Problems section:
  the samples never deliberately raise errors.  The chapter's problem
  sets lose exactly that one problem, with a pure renumbering of its
  successor and nothing else.

- 92e6b70d, 2f38daa1 (Schema, Guide) The vague phrase "additional
  validation" is retired everywhere in favor of naming the
  validation-plus stylesheet plainly.

- 0a5cc5d5 (Chapter) The two remaining coverage gaps closed: a new
  section, "Answer Formats and Extra Macros", exercises every
  `@category` value — a nine-flavor gallery over the standard contexts
  (angle, decimal, exponent, logarithm, limit, syntax, interval,
  point, vector), plus the three needing macro files (fraction,
  inequality, quantity) loaded through the previously-unexercised
  `pg-macros`/`macro-file` elements.  Every generated blank carries
  its AnswerFormatHelp link, and both problems compiled cleanly on the
  public WeBWorK server: the committed representations are
  regenerated, including the position-derived seed shifts of the
  problems that follow.

With this, the chapter exercises every WeBWorK element and attribute
the schema defines, and every enumerated value of `@category` and
`@form`.

Testing: the chapter validates at zero development-schema errors; the
sample article and book are unaffected by the base-schema changes
(zero and the one sanctioned residue); `webwork-sets` output is
byte-identical to master before the tripwire's removal, and differs by
exactly that removal after; the two new problems were validated by the
representations build against the public WeBWorK server, and the
chapter's HTML build renders them, warning-free, with their help
links; schema regeneration is idempotent.


Alex-Jordan:
This added an exercise that uses the spectrum of `@category` values for answer types. I want to note that on the PG side, what this causes is not great. It has very outdated information from 20+ years ago. It should be updated of course on the PG side. It's just not the sort of thing I would "show off" to PTX users writing WeBWorK. Not yet anyway.

Not saying it should be reverted. But this could have been coordinated with some work on the PG side.

rbeezer:
> what this causes is not great

Sorry if this was a surprise.

> It's just not the sort of thing I would "show off"

Not showing off - this is for the comprehensive testing purposes of this document.  I'm on a mission to make the schema reliable, for the first time in a long time.  These have been in the schema since 2018, hence the assumption they were stable.  We could move #var into the "dev" version of the schema if you don't think it is complete.

> But this could have been coordinated

Moving fast on this one, the schema is close.  Getting it whipped into shape so we can keep it that way.  Improvements to the code always welcome.

PR #3024, opened 2026-07-09
gdscript highlighting and packaging 

DavidCooperWCU:
A new prism format was added for gdscript 2.0
A new component was added to generate gdscript packs as zip files The documentation was updated to describe the new gdscript highlighting and active code capabilities 1 program example and 3 activecode examples were added to the Runestone Testing section of the sample book.

PR #3136, opened 2026-08-08
Fix bug with gdscript work: ensure godot_helper functions use os functions correctly

oscarlevin:
Just prior to merging the gdscript work (PR #3024) we replaced pathlib calls with their equivalent os functions, to be more consistent with how the rest of the python libraries work.  In a few spots this left behind incorrect methods that resulted in errors when the script tried to download and install necessary helper files.  This PR fixes those bugs so the godot_helper script correctly uses the os modules.

We didn't see this when testing since the godot helper files were already installed prior to the pathlib -> os conversion.

### DIFF
--- a/pretext/lib/godot_helper.py
+++ b/pretext/lib/godot_helper.py
@@ -84,7 +84,7 @@ def _export_templates_asset_name() -> str:
     return "Godot_v{}_export_templates.tpz".format(GODOT_VERSION_TAG)
 
 
-def _template_install_dir() -> str | os.PathLike[str]:
+def _template_install_dir() -> str:
     system = platform.system()
     if system == "Linux":
         return os.path.join(os.path.expanduser("~"),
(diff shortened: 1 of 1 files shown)


rbeezer:
Confirmed on both paths, with `_download` and `_verify_checksum` stubbed and `CACHE_DIR` pointed at a temporary directory, so no network was involved:

| | master | this PR |
|---|---|---|
| `ensure_export_templates` | `AttributeError: 'str' object has no attribute 'mkdir'` at `:266` | installs `top.txt` and `sub/deep.txt`, `templates/` prefix stripped, `.tpz` removed |
| `_download_and_extract_godot` | `AttributeError: 'str' object has no attribute 'unlink'` at `:214` | returns a `str`, executable bit set for user, group and other, zip removed |

So both halves of the download were dead, not just one.

The conversion looks complete — no `.unlink(`, `.mkdir(`, `.read_text(`, `.parent`, `.joinpath(` or `Path(` left in the file. Since `pathlib` is not imported and never was, each of these was a certain `AttributeError` rather than a working alternative. Dropping the `str()` wrappers in `resolve_godot` is right as well: `pretext.py:1651` uses the value as a subprocess command.

One latent edge, not reachable today: `_download` does `os.makedirs(os.path.dirname(destination), exist_ok=True)`, and a destination with no directory component would give `os.makedirs("")` → `FileNotFoundError`. All three call sites pass `os.path.join(CACHE_DIR, …)`, so nothing hits it — `os.path.dirname(destination) or "."` would close it permanently.

Also worth having: `open(..., encoding="utf-8")` in `_get_release_sums` is better than the `read_text()` it replaces, which would have used the locale default.

rbeezer:
Merged as-is - thanks for the catch-up.

PR #3138, opened 2026-08-08
Update Italian localization strings

valelav:
Updated Italian localization strings for various document types and UI elements.

### DIFF
--- a/xsl/localizations/it-IT.xml
+++ b/xsl/localizations/it-IT.xml
@@ -59,10 +59,10 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <localization string-id='displaymath'>Matematica in display</localization>
     <!-- Types of documents, mostly for informational messages -->
     <localization string-id='volume'>Volume</localization>
-    <localization string-id='book'>libro</localization>
+    <localization string-id='book'>Libro</localization>
     <localization string-id='article'>Articolo</localization>
     <localization string-id='letter'>Lettera</localization>
-    <!-- <localization string-id='slideshow'>Slideshow</localization> -->
+    <localization string-id='slideshow'>Sequenza di diapositive</localization>
     <localization string-id='memo'>Memo</localization>
     <localization string-id='presentation'>Presentazione</localization>
     <!-- Parts of a document -->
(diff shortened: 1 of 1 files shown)


rbeezer:
@valelav - if you haven't noticed, I am now using Claude Code **all the time** and running every PR through.  It wants to help refine your translations!  Not sure how I would feel about that, I hope you do not mind.  Comment coming in a minute.  Welcome to delay changes until your next update.  Nothing looks too out of line to me.

rbeezer:
Measured on the sample article set to `xml:lang="it-IT"`: 1187 translation failures before, 78 after. Structurally the file is unchanged where it matters — 211 string-ids in the same order as en-US, nothing dropped or reordered, LF endings kept, `xmllint` clean. Removing the instructional comment above `permalink-tooltip` was the right call.

A few suggestions.

`top` is worth adding while you are here. It titles the back-to-top button in every page footer, so it accounts for 126 of the bracketed placeholders still in the build — more than everything else remaining put together. The others visible are `subexercises` (10) and `observation` (8).

Three wordings I would ask about, you being the Italian speaker and not me:

- `embed` and `embed-this-page` → "Immergi" / "Immergi questa pagina". The sense here is placing the page inside another site, normally "Incorpora"; "immergere" reads as submerging something.
- `permalink-tooltip` → "Copia descrizione e permalink per". There is a doubled space before "permalink", and the en-US string is "Copy heading and permalink for" — the tooltip sits on a heading, so "titolo" may be closer than "descrizione".
- `calculator` → "Calcolatore", where "calcolatrice" is the tool rather than the machine.

One to confirm rather than change: `or` and `either` both became "O". They are distinct in en-US, so if Italian does not need the distinction that is fine — just checking it was deliberate.

rbeezer:
Merged as-is, but for a rebase and a commit message edit.  As always, thanks very much for keepin this current!

valelav:
Dear Rob,
I use AI all the time for rewriting my emails, preparing my tests and so
on, and it makes a better work than I could do by myself, so no problem.

Unfortunately, I am now and for the next few days in a place without a
stable connection.  I can just read and write emails from my phone when I
move to a close town but I cannot use my computer to connect to Github.

Anyway, the corrections suggested by Claude are fine.

About 'either...or...', I can confirm that the corresponding translation Is
'o...o'

About the remaining untranslated strings:

Top Is a bit problematic as it is requested to be short to fit in the
button. The shortest word that It comes to my mind is 'Inizio' (which
strictly speaking means 'beginning') but I think It is still too long.

I translated 'Remark' with 'Osservazione': this Is standard in italian
Matematics textbooks. For this reason I didn't translate 'Observation'
which, by the way, I think I have never found in English mathematics books
(if it is meant as the result of an experiment, I could search for and
appropriate translation, maybe asking to my chemist colleaugues)

I could easily translate 'Subexercises' with 'Sottoesercizi' which sounds a
bit strange to me. It looks likes the plural of 'Subexercise' (which makes
me think to what Is called 'Task' in Pretext language), while it indicates
a subsection of 'Exercises' (does It?).

For now, that's all. I hope to have soon a better connection for writing on
the phone Is a pain.

Cheers,
Valerio








































Il dom 9 ago 2026, 18:51 Rob Beezer ***@***.***> ha scritto:

PR #3140, opened 2026-08-08
Assembly: ensure dynamic-subs is not required for assembly-id-only

oscarlevin:
The recent-ish update to webwork generation (in individual files) required a second later report from assembly to give the assembly-ids before trying to read in other xml files, such as dynamic substitutions.  I thought I had this sorted correctly, but must have missed a case, because trying to build a document that has both webwork and FITB can't generate webwork, since the assembly-id-only bailout still requires the dynamic substitutions.  

This PR fixes that.

### DIFF
--- a/xsl/pretext-assembly.xsl
+++ b/xsl/pretext-assembly.xsl
@@ -658,8 +658,13 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
 <!-- them it is the identity, so we skip the full-tree copy.  NB: a         -->
 <!-- new template in the mode must be reflected in this test.               -->
 <xsl:variable name="b-has-dynamic-markup" select="boolean($assembly-label//setup | $assembly-label//numcmp | $assembly-label//strcmp | $assembly-label//jscmp | $assembly-label//mathcmp | $assembly-label//logic | $assembly-label//fillin[@ansobj] | $assembly-label//eval[@obj])"/>
+<!-- libxslt evaluates global variables eagerly, so this pass would run     -->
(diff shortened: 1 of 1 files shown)


rbeezer:
I'm looking into this one.  A problem of timing with **just** WW and FITB seems ad-hoc to me.  Trying to get a grip on what the bigger design flaw is.

rbeezer:
Looking like we "bail out" and stop doing the passes in pass 8, but we have everything we need in pass 6.  dynamic-subs seems to happen in the gap, in pass 7.  Or something like that.

rbeezer:
Closing in favour of #3144, which fixes the same defect a level up.

Your diagnosis was right, and so was the mechanism you identified: libxslt computes `$dynamic-rtf` even in `assembly-id-only`, where nothing refers to `$dynamic` at all. The narrower question was where the test belongs. `assembly-id-only` emits `$assembly-label`, which is pass 6, so everything from pass 7 on is discarded — but the existing short-circuit was written at pass 8, in `$representations-rtf`. That covered passes 8 through 14 by leaving them an empty tree, and left pass 7 to test the condition on its own account. `dynamic-substitution` was the only pass in that gap, and the only one there that opens a file, so it was the only thing that could fail this way.

#3144 moves the test to immediately after pass 6 and feeds the rest of the chain from it, so the `<xsl:choose>` in `$representations-rtf` is unreachable and comes out too. Two tests become one, and a pass added between 6 and 8 later inherits the behaviour instead of having to repeat it.

Verified equivalent to this PR: with the substitutions file absent, `assembly-id-only` succeeds and a normal build still fails, which is correct — that build does need the file. With the file present, the assembled tree is byte-identical to master in both modes.

PR #3146, opened 2026-08-10
Validation: allow for multiple opts

ascholerChemeketa:
This makes the -M flag accept multiple values to specify the three concepts: where to run; which schema to use; how to format the output.

So you might specify:

... -V -M dev terse
or
... -V -M server dev verbose

local-dev is kept as an alias for "local dev"


rbeezer:
This is closed by #3165, where I have chosen to stick to single-valued arguments.  And managed to drastically improve how the validation reports behave.

PR #3148, opened 2026-08-11
Favicon scheme svg option

ascholerChemeketa:
SVGs are considered the best format for favicons in modern browsers and are a "baseline" feature with 90+% support : https://caniuse.com/link-icon-svg

This adds "svg" as an option instead of "simple" for favicon image.


rbeezer:
OK, Claude has a lot to say.  Some are nits, and about half of this I saw before asking Claude for a review.  These were mine, beyond small typos:

*  `ressvg` did not make any sense to me.  I see now, but it still needs to change.
*  The HTML commit has some publisher file work in it, I've put that on "Publisher variables:" commits.
*  The publication schema commit needs the derived files.
*  I've never been a fan of using PTX in public-facing situations.  I don't even like it as a file extension.  OK as a code prefix, but not as the name of the project.  Could the PreTeXt logo be used instead of a "PTX" wordmark?  It might be sorta illegible, but even with that I'd like it better.

From Claude's analysis:

*  Does it make sense to somehow make three images possible?  Prefer an SVG and have the PNG fallbacks for Safari?  If SVG is going to leave out older Safari (at 9% audience), I think that is a big enough omission to mention it in the Guide writeup.

ascholerChemeketa:
Force pushed with requested changes (and a typo fix in commit message).

Any idea where Claude got 9%?
My linked reference shows unsupported desktop safari as 0.9% and mobile as 3.5%.

For a "nice touch, but not essential" feature that authors need to actively switch to use, I don't think we need to engineer in backwards compatibility for that audience.

rbeezer:
Thanks for this — the feature is sound and the wiring is complete. Reviewed at `067958a9` against its merge-base, `1c2ae8eb`.

## Confirmed working

- Building the sample article to HTML emits `<link rel="icon" type="image/svg+xml" href="external/favicon/favicon.svg">` on every page, and the build copies `favicon.svg` into `external/favicon/`. Chrome requests the SVG and nothing else — no PNG, no `favicon.ico`.
- The `simple` scheme is unaffected by the restructure into `xsl:choose`: rebuilt with `favicon="simple"`, the same two `link` elements come out in the same order with the same paths.
- The legacy `docinfo/html/favicon` element still reaches the PNG branch when no publication attribute is set.
- The option registration in `publisher-variables.xsl` is right. With a bogus value the build reports *"Possible values are: `none`, `simple`, or `svg`. The default, `none`, will be used instead"* and emits no `link` element, and `favicon="svg"` draws no warning.
- The two Guide edits validate against the development schema.
- `name="favicon"` has a single caller, in the HTML head cache, so no other conversion is touched.

## Two things to fix

**The derived publication schema was not regenerated.** The change is in the literate source `schema/publication-schema.xml` only; `schema/publication-schema.rnc` and `schema/publication-schema.rng` still say `"none" | "simple"`, so the sample article this PR ships does not validate:

```
examples/sample-article/publication.xml:98:44: error: value of attribute
"favicon" is invalid; must be equal to "none" or "simple"
```

Regenerating from the branch's own source is clean — one line changes in each derived file and nothing else moves:

```
cd schema
xsltproc ../xsl/pretext-litprog.xsl publication-schema.xml
trang -I rnc -O rng publication-schema.rnc publication-schema.rng
```

Those two files belong in the same commit as the `.xml`, which is the shape the earlier publication-schema commits use.

**Wrong filename in the sample article's comment.** In `examples/sample-article/publication.xml`:

```
<!--     favicon/favicon-svg   -->
```

The build looks for `favicon/favicon.svg`. That publication file is the one publishers copy from, so the wrong name will travel.

## A question about the scheme's scope, for the maintainer

As written, `svg` and `simple` are mutually exclusive, so choosing `svg` emits one `link` element and no PNG fallback.

caniuse — the source linked in the description — puts SVG favicons at 91.14% globally, and the missing share is almost entirely Safari: SVG `rel="icon"` arrives in Safari 26 on both desktop and iOS, and Safari on iOS 3.2 through 18.7 does not support it at all. So a publisher who selects `svg` shows no favicon to a reader on an iPad or an iPhone below iOS 26, which is a meaningful part of a textbook audience.

The usual answer, and the one realfavicongenerator.net recommends — the same source this template's comment already cites — is to emit both and let the browser prefer the SVG. That could be `favicon="svg"` emitting the SVG link plus the two PNG links, with all three files required of the publisher; or a separate value for the combination; or leaving the behaviour as it is and saying plainly in the Guide that `svg` is for modern browsers and `simple` remains the broadly compatible choice.

Worth recording either way: `simple` keeps in-repo coverage through `examples/showcase/publisher/publication.xml`, so moving the sample article to `svg` costs nothing in testing.

Related, in the Guide: the replacement sentence "Try one of these schemes first, as they cover common modern browser behavior" drops the concrete claim it replaced without giving a publisher any basis for choosing between the two. Whatever is decided above is the sentence that belongs there.

## Smaller items

**Comment alignment.** The closing `-->` is aligned in a column throughout these files. Two of the new blocks are not. In `xsl/pretext-html.xsl`:

```
<!-- Expects publisher to provide -->
<!--     favicon/favicon.svg      -->
<!-- in the external images directory -->
```

The `simple` branch immediately below it is aligned, and the block replaced in `examples/sample-article/publication.xml` was too. (The typo repair on the template header does hold its column — that one is right.)

**`favicon.svg` has no trailing newline.** Every other committed `.svg` in the repository ends in one, and a past sweep made that a repository-wide convention for text files that are not build output.

**The favicon README is now inaccurate.** `examples/sample-article/media/favicon/README.md` says the favicon is the test image from RealFaviconGenerator. There are now two different images in that directory: the PNGs are that stock test logo, and the SVG is a **PTX** wordmark. Updating the README in the same commit would keep the provenance record straight. While you are there — do you know which font produced the wordmark outlines? They are baked-in paths, so the file itself does not record it.

**Generator debris in the SVG.** The file carries two rules that both say `filter: none`, one under `prefers-color-scheme: light` and one under `dark`, plus an RDF metadata block, all on a single 1499-character line. Since this is a hand-maintained asset rather than build output, it can be trimmed and wrapped. Removing those two inert rules also opens the door to making the icon adapt to a dark tab strip, which is worth considering: the background is a `#ffffff` to `#d4dade` gradient, so it is a bright square in a dark browser.

**How it reads at tab size.** At 16 pixels the three letters are about four pixels wide each — legible as a shape rather than as letters. The circular arrow it replaces survives the size better. That is a judgement call rather than a defect, but since this also changes the sample article's branding as well as its file format, it is worth a deliberate decision rather than a side effect.

## Commit hygiene

- `Guide; document SVG favicon option` — the separator should be a colon.
- `Schema: add publisher option for SVG favicon` — changes to the publication-file schema take the topic `Publication schema:`; plain `Schema:` is the document schema.
- `HTML: add SVG option for favicon-scheme` — `favicon-scheme` is the internal XSL variable name; the publisher-facing name is `"@favicon"`.
- `ressvg` introduces a new abbreviated identifier. Its neighbours `res32` and `res16` are older, and a new variable does not have to follow them.

No commit carries a body, and the whitespace is clean throughout.

ascholerChemeketa:
It looks like that is reviewing the old branch, not the new force push. `067958a9` is no longer a commit on the branch.

Already commented on png as backup... And Claude's math is not mathing. There is no way to get 9% from old versions of Safari.

rbeezer:
> It looks like that is reviewing the old branch

Hmmm.  ;-)  I'll start over.  Sorry for the noise.

rbeezer:
I missed a permission request, so the original review sat overnight, I think.  The newer commits came in while I was working through this anyway, I think.

rbeezer:
Re-reviewed at `5bbcf6fea`, head re-confirmed against the branch immediately before posting this time.

## The 9% was wrong — thanks for pushing back

@ascholerChemeketa asked where it came from, and the answer is that I took `100 − 90.48` from the caniuse summary and attributed the remainder to Safari. Both halves of that were wrong. The complement of the support figure is not the unsupported share: caniuse's table only accounts for 96.69% of global usage, so roughly 3.3% is browsers it does not attribute at all. And the part that genuinely lacks support is not mostly Safari.

From the caniuse dataset today:

| | share |
|---|---|
| Safari on iOS, unsupported | 3.60% |
| Safari desktop, unsupported | 0.90% |
| **Safari family** | **4.50%** |
| UC Browser, older Chrome, IE, QQ, Android Browser, Samsung | 1.70% |
| **total unsupported** | **6.20%** |

So 4.5% for Safari, which matches your 0.9% and 3.5% almost exactly, and half of what I said. That materially weakens the case I made for PNG fallbacks: for an opt-in decoration a publisher has to deliberately switch on, 4.5% is a reasonable thing to decline to engineer around, and I would not press the three-file design.

The separate request to note the limitation in the Guide is cheap either way, and the Guide text is byte-identical to the previous head, so that one is still open.

I also owe an apology for the timing of the last review: it went up seven minutes after the force push, describing a commit that was no longer on the branch. I had read the PR at the start of the session and never re-confirmed the head before posting. That is on me and it is fixed in how I did this one.

## Confirmed fixed

- `ressvg` is now `favicon-svg`.
- The publisher-file change is its own commit.
- The schema commit carries the derived files. Regenerating from the branch's own source produces `publication-schema.rnc` and `.rng` **byte-identical** to what is committed, and the sample article's publication file now validates against the derived schema with no favicon error.
- `Guide;` is now `Guide:`.
- The repeated "empty when not using managed directories" comments are hoisted into one aligned block above the `xsl:choose` — nice tidy-up.
- The artwork is the PreTeXt logo, dark-mode aware.

## One new thing, and it is the important one: the logo renders clipped

At every icon size I tried, Chrome draws the book's top curve, the spine bars and the `</>`, and then stops. There is no bottom curve, and the mark sits in the upper part of the square with empty space beneath it. Reproduction needs nothing from PreTeXt:

```
<img src="favicon.svg" width="32" height="32">
```

Same result at 16, 32, 64 and 128, and the same when the file is opened directly at its own viewBox aspect ratio.

What I can state, and what I cannot. Measuring the rendered ink, the complete artwork spans user y 3010..9820, and the committed `viewBox="338 3000 8772 6866"` reaches y 9866 — so the box does enclose the drawing, and simple cropping arithmetic does not explain this. Yet the render stops at about y 9000. A copy of the file differing **only** in the viewBox, changed to `"0 3000 9500 8000"`, draws the complete closed book. So the viewBox is implicated, but I could not identify the mechanism and would rather say so than invent one. You have the source artwork, so you will almost certainly see it faster than I did.

Separately, the colour mechanism is sound: forcing the dark rule active renders the strokes near-white as intended, so the custom property and media query do work. Whether a browser re-rasterizes a favicon when the system switches to dark is a different, browser-dependent question that I did not verify in a real tab strip.

## Still open from the previous round

- `favicon/favicon-svg` in the sample article's publication comment — the build looks for `favicon/favicon.svg`.
- That comment block is misaligned: lengths 41 / 38 / 44, where the block it replaced was 44 / 44 / 44 / 44.
- `favicon.svg` has no trailing newline, and now also opens with a stray leading space before `<svg` on line 2.
- The favicon README still describes the image as "the test image from RealFaviconGenerator", which is no longer what the file is.

## Smaller

- `<!-- All images references below ... -->` — "All image references".
- `<!-- Expects publisher to provide favicon/favicon.svg   -->` pads to align with nothing.
- The commit topic `Publisher:` appears 3 times in the history against 71 for `Publisher variables:`.

## Verified working

`favicon="svg"` emits `<link rel="icon" type="image/svg+xml" href="external/favicon/favicon.svg">` and the build copies the file into place. `favicon="simple"` is untouched by the restructure into `xsl:choose` — same two `link` elements, same order, same paths. Whitespace is otherwise clean.

ascholerChemeketa:
Fixed the whitespace items. 

The rest was a waste of time and tokens. 

Chrome/FF/Edge all render the favicon just fine.

rbeezer:
The favicon clipping report was wrong. @ascholerChemeketa is right that Chrome, Firefox and Edge all render it correctly, and I am sorry for the detour.

The fault was in how it was tested, not in the file. The check embedded the icon as ``. The SVG root carries `height="100%"` and no `width`, and in an `` context that combination makes Chrome clip the drawing — whether or not the box is square, and `object-fit: contain` does not rescue it. Opened directly in a browser, or used the way a favicon is actually used, the complete logo renders. A favicon is never loaded through ``, so the measurement never had any bearing on the thing it claimed to be about. That should have been checked against the real use before it was written up.

Three small things I will take care of when I merge, nothing needed from you:

- Change the topic on the publisher-file commit from `Publisher:` to `Publisher variables:`, which is the topic that change has carried historically.
- Fix the markdown link in the favicon README, which has the URL and the label in each other's slots, so it currently points at a relative path that does not exist. It predates this branch and just happens to sit on the line the branch edits.
- Add a sentence to the Guide noting that the `svg` scheme leaves readers on browsers without support — Safari before version 26, most of the remainder — with no icon rather than a fallback, and that `simple` is the more widely supported of the two. That is the note I asked for earlier, and it is a fair thing for a publisher to know when choosing between the schemes.

rbeezer:
Merged, with promised touch-ups.

Thanks for the improvement - I hadn't known we were so far behind.  ;-)

PR #3151, opened 2026-08-13
JavaScript: Ensure workspace is collapsed when solution is revealed

kylemonette:
This PR deals with two workspace/solution-pagination fixes, continuing off the changes made in PR #3114 .

1. **Workspace didn't collapse when a solution was revealed**
Workspace was always stretched to fill unused space on the page, even after the question's hint/answer/solution had been revealed. This left a large gap between a question and its "Solution" block.

2. **Embedded videos reloaded unnecessarily during repagination**
Toggling solution visibility triggers a settle loop that can move page rows (and therefore any `<iframe>` inside them) between `.onepage` containers several times before layout stabilizes. Each move caused the iframe to reload, so a single checkbox toggle could reload a video multiple times.

**New functions:**

- **`anySolutionTypeVisible()`** — Returns whether any hint/answer/solution is currently visible on the page. Checks *effective* visibility (the element itself and every ancestor free of the `hidden` class), not just `:not(.hidden)` on the element alone. This distinction matters because a live PreTeXt knowl nests its content in a `.knowl__content` div that never carries `hidden` directly — only its `.born-hidden-knowl` ancestor wrapper does. A naive `:not(.hidden)` check matches that inner div regardless of the collapsed ancestor, which caused every workspace on a page to incorrectly read as "a solution is visible" from the moment the page loaded, before the reader had revealed anything.

- **`withIframesDetached(fn)`** — Wraps a batch of repagination work. Detaches every `<iframe>` in the printout (swapping each for a same-sized placeholder div, so height measurements stay accurate) before `fn` runs, then reattaches them once `fn` resolves. This caps a solution-visibility toggle at reloading an embedded video once, instead of once per intermediate row move during the settle loop.

**Modified functions:**

- **`setInitialWorkspaceHeights()`** — Now collapses workspace to `0px` instead of its authored height whenever `anySolutionTypeVisible()` is true, freeing that space for pagination once it's no longer needed as blank writing room.

- **`adjustWorkspaceToFitPage()`** — Skips the pass that stretches workspace boxes to fill unused page space whenever `anySolutionTypeVisible()` is true, leaving workspace at its (now-collapsed) authored height instead.

- **`collapseSpilloverPages()`** — Once a solution is visible, workspace is pinned at its collapsed floor and will never shrink further, so the normal "merge first, let the caller re-split if it doesn't fit" order no longer converges — it just oscillates every settle tick. This function now checks fit *before* merging in that case (via a new `workspaceIsFixed` flag), and puts a spillover page's content back rather than flip-flopping if it doesn't actually fit.

- **`applySolutionVisibility()`** — The whole toggle body now runs inside `withIframesDetached()`, since the repagination/collapse work it triggers can move a row several times over the course of one toggle.

### DIFF
--- a/js/pretext_add_on.js
+++ b/js/pretext_add_on.js
@@ -462,6 +462,40 @@ function waitForImages(container, timeoutMs = 5000) {
     ]);
 }
 
+// Moving a row between .onepage containers makes any <iframe> inside it
+// (e.g. a YouTube video or GeoGebra applet) reload, even when it lands back
+// in the same parent. A single settle loop (repagination/spillover-collapse
+// running repeatedly until layout stabilizes) can move a row several times,
+// so detach every iframe -- swapped for a same-sized placeholder, to keep
+// height measurements accurate -- before a batch of repagination work and
+// reattach afterward, capping each toggle at reloading a video once instead
+// of once per move.
+async function withIframesDetached(fn) {
+    const printout = getPrintout();
+    const parked = [];
+    if (printout) {
+        printout.querySelectorAll('iframe').forEach(iframe => {
+            const rect = iframe.getBoundingClientRect();
+            const placeholder = document.createElement('div');
+            placeholder.className = 'iframe-placeholder';
+            placeholder.style.width = rect.width + 'px';
+            placeholder.style.height = rect.height + 'px';
+            placeholder.style.display = getComputedStyle(iframe).display;
+            iframe.parentNode.insertBefore(placeholder, iframe);
+            iframe.remove();
+            parked.push({iframe, placeholder});
+        });
+    }
+    try {
+        return await fn();
+    } finally {
+        parked.forEach(({iframe, placeholder}) => {
+            placeholder.parentNode.insertBefore(iframe, placeholder);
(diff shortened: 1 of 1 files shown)


kylemonette:
Follow-up: Revealing a solution can also free up space, not just use it

The previous fix made workspace collapse once a solution is shown. But `applySolutionVisibility()` still assumed that showing a solution could only ever need *more* room on the page, so it only tried to reclaim space when *hiding* solutions, never when *showing* them.

Showing a solution can now shrink a page too, because the workspace collapses at the same time. So a page that had already spilled its content onto a new page can end up with room to fit that content back.

Both showing and hiding now try to pull spilled-over content back onto the previous page first, and only push content to a new page if it still doesn't fit. This is safe in both directions because pulling content back only happens if it actually fits (from the previous fix).

oscarlevin:
Some of the work here might be subsumed by that of #3154, which should be merged first I think.

rbeezer:
#3154 is now closed

oscarlevin:
Refactored and merged just the iframe detachment fix from this PR.  The rest of the fixes landed in #3154.

PR #3152, opened 2026-08-13
HTML: inline ol-marker styles when using cdn resources

oscarlevin:
For both the portable html build and cdn resources publisher options, we have been ignoring the case of custom list markers.  For portable, this just failed silently.  For cdn resources, this was causing a build failure.

This PR unifies these two cases and in both of them writes the custom css directly into the html header.

We could do this always, but that would slightly increase the size of each html file, so I don't know what is better.

### DIFF
--- a/xsl/pretext-html.xsl
+++ b/xsl/pretext-html.xsl
@@ -351,7 +351,9 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
         <xsl:apply-templates select="." mode="make-xref-knowls"/>
     </xsl:if>
     <!-- custom ol marker css production -->
-    <xsl:if test="not($b-subsetting) and not($b-portable-html)">
+    <!-- A CDN cannot host these styles, since they are derived from the -->
+    <!-- source.  CDN builds get them inline instead, see "css-common".  -->
+    <xsl:if test="not($b-subsetting) and not($b-cdn-resources)">
         <xsl:call-template name="ol-marker-styles"/>
     </xsl:if>
 </xsl:template>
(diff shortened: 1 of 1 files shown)


rbeezer:
Reviewed at `78dfc094`, whose parent is the current `master` tip, so no rebase is needed.  Built `examples/sample-article` with `-c doc -f html` in four configurations, before and after, in the same working tree.

| publication setting | before | after |
|---|---|---|
| default, local resources | builds | byte-identical apart from build timestamps |
| `<resources host="cdn"/>` | **build aborts** — `xsltDocumentElem: unable to save to https://cdn.jsdelivr.net/.../ol-markers.css` | builds; 14 rules inlined, rule set identical to the file the local build writes |
| `<platform portable="yes"/>` | builds, but every page links to a CDN URL that 404s, so markers are silently lost | builds; rules inlined, including in the `-if.html` iframe pages |

Both reported failures reproduce on the merge base and are gone on the branch.  The fix looks right to me.

`$b-cdn-resources` is the correct condition here for a reason stronger than the comment gives.  `pretext/lib/pretext.py` reads that same publisher variable to set `include_static_files`, and when it is false it never copies the static tree, so `_static/pretext/css` does not exist at all.  The XSL condition and the Python condition are now the same statement: a local static CSS directory exists if and only if `$b-cdn-resources` is false.  That is worth having in one of the two comments, in place of "portable HTML has no place to link out to anyway" — a portable build links out plenty, all of it to the CDN; what it lacks is a local `_static`.

Three small things.

**The two conditions are no longer complementary.**  The writer is conditional on `not($b-subsetting) and not($b-cdn-resources)`, the reader on `$b-cdn-resources`, and the experimental subtree build falls between them:

```
pretext -c doc -f html -r lists -p examples/sample-article/publication.xml ...
  -> _static/pretext/css/ol-markers.css is not written
  -> <link href="_static/pretext/css/ol-markers.css" ...> is still emitted
```

Pre-existing, not introduced here, and `-r` is experimental — but it is the same defect this PR just removed from the portable path, surviving because two separate boolean expressions have to be kept in agreement by hand.

**Escaping, for the record.**  `mode="ol-marker-style"` does not escape the author's marker text for a CSS string context.  That is already true on `master` and format-independent: `<ol marker='"a)'>` writes `content: """counter(list-item,lower-alpha)") "; }` into the file, which is invalid CSS.  The inline form adds one new way to fail — `<style>` content is not escaped by the HTML output method, which is exactly what makes this PR work, so a marker containing `</style>` closes the element early and spills the rest of the rule into the page.  Both are contrived and neither is a reason to hold this up; noting it because inlining more widely would make the second one more reachable.

**Commit subject:** `cdn` reads better as `CDN`, matching the file's own comments.

**On doing it always:** yes, I think so — but as a separate follow-on commit, not folded in here.  The measured cost is much smaller than it sounds (the sample article, with an unusually high fourteen distinct markers, pays about 175 gzipped bytes per page and 1.2% of total HTML bytes; a document with one marker is roughly byte-neutral against the `<link>` it replaces), and it would delete the file, the `exsl:document`, and the pair of conditions above.  I will write that up properly rather than argue it here.  This PR fixes two live failures and stands on its own.

rbeezer:
Merged as-is, with edited commit message.  Opus 5 thinks universal inlining is a good idea and I have a design document.  I'm asking Fable 5 about whether there is evn a better way to handle the custom markers.

Issue #1194, opened 2019-10-21
Broken link to MathJax documentation in Guide


sean-fitzpatrick:
Section 4.8 of the guide points to the following URL:
http://docs.mathjax.org/en/latest/tex.html#supported-latex-commands

This link no longer appears to work.
The correct link (I think) is http://docs.mathjax.org/en/latest/input/tex/macros/
but it's also version-dependent. (This is for MathJax 3.0; there's also one for 2.7.)

Apologies for the issue rather than a pull request -- trying to meet other deadlines at the moment.

rbeezer:
Thanks very much for the heads-up!  Fix in place at 05e8001a72f1aabea04a719a8bcf5c1449a753f8 but no website update just yet.

PR #3153, opened 2026-08-16
Guide: correct broken external links

skiadas:
### Summary

Fixes 11 broken external links in the Guide (`doc/guide`), each verified to resolve over HTTP before committing:

- **MathJax** — Accessibility Features → `docs.mathjax.org/en/latest/basic/accessibility.html`; Custom Extension (webpack) → `docs.mathjax.org/en/latest/advanced/webpack.html`
- **GitHub** — schema directory → `github.com/PreTeXtBook/pretext/tree/master/schema`
- **Unicode input** — `cs.tut.fi/~jkorpela/fui.html8` → `jkorpela.fi/fui.html8` (author's site moved)
- **LaTeX Font Catalogue** (×2) — `tug.dk` → `tug.org/FontCatalogue` (legacy host gone)
- **A First Course in Linear Algebra** — `linear.ups.edu` → `linear.pugetsound.edu`
- **sagehill (DocBook XSL)** — drop `www.`; `sagehill.net` resolves, `www.sagehill.net` does not
- **Wyman Indexing** — `wymanindexing.com` → `wymanindexing.wordpress.com`
- **AMS-LaTeX page** — `ams.org/.../tex/amslatex` → `ams.org/arc/resources/amslatex-about.html`
- **textcomp** (a source comment reference) — → `ctan.org/pkg/textcomp`

### Still open (not fixed here)

- `idpf.org/a11y` — IDPF merged into W3C; likely successors are `w3.org/publishing/a11y/` or the EPUB Accessibility spec, but both need a browser check.
- `rellek.net` — returns 500; the linked author is still active, so we're checking with them for a current URL.

### Related

- #1194 — earlier MathJax documentation link breakage (same recurring theme).

The remaining audit entries (StackExchange, w3.org, AMS, sourceforge, debian, etc.) returned 401/403/429 to the automated check but were verified working in a browser; and `platypus.mammal-institute.org` URLs in `web-hosting.xml` are illustrative examples, not live links.


skiadas:
I had my assistant go through and suggest fixes to some broken links in the guide. Two it could not resolve: 

1. A mention to @mitchkeller  at https://pretextbook.org/doc/guide/html/schema-products.html pointing to https://rellek.net/home/  which doesn't seem to be working. Hoping Mitch can point to a better spot.
2. Accessibility docs at the very bottom of  https://pretextbook.org/doc/guide/html/topic-accessibility.html point to https://idpf.org/a11y for EPUB accessibility, but that is dead. Maybe this would be a replacement?  https://www.w3.org/TR/epub-a11y-11/

rbeezer:
Checked all 22 URLs — the 11 replaced and the 11 replacements — and built the Guide before and after.  This is good, careful work; nothing here needs changing.

**Every new URL returns 200 with no redirect, and each is the right page.**  Titles fetched: `CTAN: Package textcomp` · `AMS-LaTeX` · `Making a Custom Build of MathJax` · `Wyman Indexing` · `Full Unicode Input utility` · `Accessibility Features — MathJax 4.0` · `The LaTeX Font Catalogue` (front page and Iwona).  The two pages without titles check out on content: `sagehill.net/docbookxsl/DuplicateIDs.html` is "Using a module more than once in the same document", which matches the `xpointer()` remark it is cited for, and `linear.pugetsound.edu/html/section-ILT.html` contains both "Kernel and Pre-Image" and `KPI.svg` — the exact diagram the caption credits.

**Every old URL is genuinely gone:** 404 (GitHub schema, both MathJax pages, wymanindexing, cs.tut.fi after its redirect), DNS failure (`www.sagehill.net`, `linear.ups.edu`), connection timeout (the textcomp PDF host), and expired TLS certificate (`tug.dk`, both).

**The build confirms the change is exactly what it claims.**  Guide HTML before and after: 473 files each, 13 differ, and every difference is an `href` or the matching `@visual` text in the search index — except two files that differ only by build timestamp, because the Guide demonstrates `<today/>` and `<timeofday/>` live.  No collateral effect anywhere.

Worth calling out: the `@visual` attributes were updated in step with every `@href`, and the prose literal `<c>jkorpela.fi/fui.html8</c>` was caught too.  That last one is easy to miss.

## Three small things, all for a follow-on

Rob will pick these up in a follow-on commit — no need to spin this PR for them.

1. **One dead link the audit missed.**  `doc/guide/appendices/editors.xml`, in the Sublime Text appendix: `docs.sublimetext.info` no longer resolves in DNS.  The passage is about column/block selection, so the replacement is `https://www.sublimetext.com/docs/column_selection.html`, which returns 200 and is titled "Column Selection".
2. **A link text that drifted.**  The old MathJax link carried a `#custom-extension` fragment; the new URL has none, and MathJax has since renamed the concept from *extension* to *component*, so the Guide's link text "A Custom Extension" now matches no heading on the target.  The nearest anchor is `#custom-component`.
3. `http://docs.mathjax.org/en/latest/input/tex/macros/` sits on the same line as the AMS change and stayed on `http`.  Guide-wide there are 28 plain-`http` links against 223 `https`, so a general upgrade is separate work — that one is just a neighbour.

## On the "Still open" list

Both entries are right, and I scanned the Guide independently — 211 distinct URLs — so the list is complete apart from item 1 above.

- `idpf.org/a11y` — 404 confirmed.  Since IDPF itself no longer exists, that sentence needs rewording rather than a URL swap, so it is a judgement call rather than a lookup.
- `rellek.net` — 500 at every path, including the bare domain, which redirects to `/home/` and errors there.  The domain is live and the site is broken, so asking the author is exactly right.

The 403 and 402 responses elsewhere are bot-blocking, as the description says: `inkscape.org` and `metacpan.org/pod/App::cpanminus` both returned 200 once sent full browser headers, and `w3.org` serves a Cloudflare interstitial.  I could not clear the `stackoverflow.com`, `stackexchange.com`, `runestone.academy` or `sourceforge` ones from a script, so those rest on the browser checks already done.  The three `localhost:8128` entries are CLI examples, not links.

## Commit

The commit carries a ten-bullet body.  House convention here is a subject line and the trailer only — the material is already in the PR description, which is where it belongs.

rbeezer:
Well. that is a very useful task for your assistant.  I try to minimize links, since they rot.  Did not know I had about 200 of them!

Follow-on commits added, as promised.  Your commit merged as-is.

`rellek.net` and accessibility link still outstanding.  And Claude wants me to do a `http/https` sweep at #3158

PR #3155, opened 2026-08-17
CSS: typos

Alex-Jordan:
I'm trying out a typo catching tool. So that a diff won't be too hard to review, just going one commit per folder at a time, starting with css.

These are all comment typos, except the ones with "asside" will make an impact.


Alex-Jordan:
Now with `doc/` typos too. Here, it's almost entirely typos in the guide. At least one is a typo in an `idx`.

rbeezer:
So I should expect more?  Let me know when it is ready.

Alex-Jordan:
It's ready any time, unless you plan to squash the commits. I would just open a new PR when I came back to more. But if you plan to squash, I'll just do the rest as one more commit.

rbeezer:
I do not plan to combine the commits.  But also do not want to get a fresh PR just a couple days from now.  So this can sit until you think you are done for a good long while.  Thanks.

Alex-Jordan:
OK. Not sure when I will return, but you could apply the same tool I was using:
https://github.com/crate-ci/typos

Maybe AI can use it more effectively. I tired of manually reviewing to leave out false positives.

PR #3160, opened 2026-08-18
Update Italian localizations for various strings

valelav:
Accepting some changes suggested by Claude (see https://github.com/PreTeXtBook/pretext/pull/3138) plus some other minor updates


rbeezer:
Merged, with commit body removed (we just use single line messages).

A bare `Icorpora` was missing an `n`, so I silenty added it.

As usual, thanks for keeping this current!

PR #3162, opened 2026-08-18
External links: repair the dead ones and request the rest over "https"

### LINKED ISSUE #3158: Upgrade external links from "http" to "https"
Links in the repository are a mix of `http` and `https`, with no pattern behind which is which — #3153 touched a line where the two sat side by side.  Most `http` hosts now redirect to `https` anyway, so the plain form costs a redirect and looks dated in a printed Guide, where the `@visual` text is what the reader sees.

Counting distinct `@href` values that begin `http://`: **74 repo-wide, 25 of them in `doc/guide`**.  Setting aside `localhost` examples and the deliberate `example.com` URL-escaping fixtures in the sample article, I probed 65 of them over both schemes:

- **52 are reachable over `https`** — a mechanical swap, and the bulk of the work.  `@visual` needs essentially no attention: nearly all of them omit the scheme, and the eleven that carry one are already `https` apart from the `localhost` and `example.com` fixtures.
- **3 serve `http` only** and must be left alone: `www.chromevox.com`, and `www.literateprogramming.com` (two links).
- **10 are dead over both schemes.**  These are link rot, not a scheme problem, and should not be folded into this sweep.  Two are already known (`idpf.org/a11y`, `rellek.net`), but the rest are worth their own look — notably `abstract.pugetsound.edu` (no longer resolves; used as the example book URL across seven `examples/sample-book` files) and the four `clips.vorwaerts-gmbh.de/big_buck_bunny.*` media URLs that the sample article's video examples depend on.

Suggested shape: one commit for the 52, split by area if that reads better, and a separate issue or commit for the dead links.

Related: #3153, which corrected eleven broken external links in the Guide.

rbeezer:
Five commits, each reviewable on its own.  Four concern external links in `doc/guide/` and `examples/`; one is an XSL fix that the audit turned up.

## The sweep, and the link rot behind it

`8e6a48a8` does what #3158 asked: **53 URLs across 24 files** move from `http` to `https`.  Every one was fetched over `https` first and swapped only on success.  Only `href=` and `url=` attributes were rewritten, so nothing functional moved — no `server=` or `src=` appears in the diff, and the two publication-file changes are `brandlogo` link targets.

Nine URLs deliberately stay on `http`, and this is the complete list:

- the six `example.com` fixtures in the sample article, where the literal string is the point of the test
- three `localhost:8128` CLI examples
- `www.chromevox.com` and `www.literateprogramming.com` (twice) — these **serve `http` only**; `https` fails outright
- `www.fsf.org` and `www.gnu.org/copyleft` — **verbatim GFDL licence text**, which is not ours to edit
- `idpf.org/a11y` and `rellek.net` — dead either way, so a scheme change would only disguise link rot

Probing for the sweep also found targets that had simply vanished, repaired in `44ceff2d` and `34f1f07b`:

| was | now |
|---|---|
| `camendesign.com/…/test.htm` | `…/test.html` — a **typo**; the XML comment on the line above already had it right |
| `abstract.pugetsound.edu` (8 references) | `judsonbooks.org/abstract/` — no DNS record; the book has its own site now |
| `www.cs.tut.fi/~jkorpela/fui.html8` | `jkorpela.fi/fui.html8` — the relocation #3153 fixed in the Guide, missed here |
| `www.imagemagick.org/script/binary-releases.php` | `imagemagick.org/download/` |
| `www.cs.tut.fi/~jkorpela/dashes.html` (a source comment) | `jkorpela.fi/dashes.html` |

`idpf.org/a11y` is left alone on purpose: IDPF no longer exists, so that sentence needs rewriting rather than a new URL.  `rellek.net` errors at every path and is being chased with its author.

## The sample article's videos

`abfa397e` replaces seven video demonstrations whose host, `clips.vorwaerts-gmbh.de`, no longer resolves — the same host the upstream "Video for Everybody" test page still points at, so its own demo is broken too.

They now use the **Sintel trailer**, CC BY 3.0, Blender Foundation: WebM and MP4 from `media.w3.org`, and the `.ogv` from Wikimedia Commons.  Labels move `big-buck-bunny-*` → `sintel-trailer-*`, with QR assets regenerated to match and a real 960×540 16:9 frame from Commons as the preview image.

Two judgement calls worth flagging:

- **The `.ogv` figure is kept and captioned as expected not to play.**  Firefox retired Theora at v133 and Chrome did likewise, so no choice of host fixes it.  The figure now documents that, which is a fair thing for a torture-test document to do.
- **The two caption files carry explicit demonstration text, not a transcript**, with a `NOTE` saying so.  They also differ from one another, which makes the two-track figure meaningful for the first time — the prose there previously admitted the test was "a bit lame" because both tracks pointed at the same file.

## One XSL fix

`360a5097` corrects `file-extension`, which took everything after the last period **in the whole string** — so for any complete URL it caught the host's dot.  It now reduces to the final path segment first.

The visible symptom was the sample article's automatic-best-format figure, captioned "(temporarily broken)" for years: it emitted **zero `<source>` elements**, and now emits four.  The feature only ever worked for dotless local paths.

119 of that commit's 131 added lines are tests.  `xsl/tests/README.md` asks that a utility's behaviour be documented before it is changed, so nine `file-extension` cases were written and run against the unmodified code first.  Exactly one failed, which narrows the defect: a URL *with* an extension always worked; only the extension-less case broke.

The fragment identifier is untouched — a `@href` carrying `#t=10` would still fold that into the extension.  Same defect class, but there is no demonstration of it in the corpus, so it is left alone.

## Verification

Before and after in a single working tree, never a second worktree.

- **`file-extension`** reaches five stylesheets, so it was built both ways on **both sample documents** across HTML, LaTeX, Markdown, EPUB and XSL-FO.  Sample book: build timestamps only, every format.  Sample article: the four new `<source>` elements plus one caption, nothing else.  For EPUB the page was token-diffed rather than line-counted; every `media-type` in `package.opf` is unchanged.
- **The sweep** was built on Guide, sample article and sample book.  Every differing line had its scheme normalised and checked to pair with a partner, so the diff is provably scheme-only; the sole non-URL line was a build timestamp.
- **The video commit** was checked with `-V` before and after: 2408 validation messages each side, nothing added or removed.
- Each replacement URL returns 200, with page titles confirmed where a soft-404 was possible.

Closes #3158


rbeezer:
@ascholerChemeketa - some work here on text utilities, and a bunch of new related tests.  Thought you might want a heads-up about changes here.

https://github.com/PreTeXtBook/pretext/pull/3162/commits/360a50975d87c35a42f76f702f44e084b9237975

PR #3164, opened 2026-08-19
French translations

JeanSebastienTurcotte:
New readability settings and read-aloud.

### DIFF
--- a/xsl/localizations/fr-CA.xml
+++ b/xsl/localizations/fr-CA.xml
@@ -224,7 +224,7 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     <localization string-id='copyright'>Copyright</localization>
     <localization string-id='keywords'>Mots-clés</localization>
     <!-- HTML clickables (lowercase strings to click on) -->
-    <!-- <localization string-id='permalink'>permalink</localization> -->
+    <localization string-id='permalink'>permalien</localization>
     <localization string-id='incontext'>Contexte</localization>
     <!-- <localization string-id='standalone'>Standalone</localization> -->
     <!-- <localization string-id='embed'>Embed</localization> -->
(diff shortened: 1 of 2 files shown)


rbeezer:
Merged, as-is, except:
* One line lost its indent, silently fixed.
* We just use one-line commit messages, so I got rid of the commit message bodies.

As usual, thanks for keeping this current!

My assistant (Claude Code) looks at everything these days, some **suggestions** about shades of meaning coming next.  That can be a new PR, now, or you can just blend it in to something later.  Or ignore them all!  Your call, at your convenience.

rbeezer:
Thank you for these — the readability menu was entirely untranslated in both files, so this fills a real gap.

A few notes purely on shades of meaning.  You are the native speaker and I am not, so please read these as questions grounded in what each control actually does in the HTML rather than as corrections.

**`reset-all` → "Options par défaut".**  This one is a `<button>`, carrying a `restart_alt` icon, that undoes the reader's adjustments.  "Options par défaut" names a state, where the English "Reset All" names an action, so beside a restart arrow it may read as a label for the section rather than something to press.  Would "Tout réinitialiser" sit better with the icon?

**`readability-expose-permalinks` → "Exposer les permaliens".**  This is a checkbox inside the *Technologies d'assistance* fieldset, and its element id is `ptx-readability-accessible-permalinks` — the intent is making permalinks perceivable to a reader who cannot hover for them.  "Exposer" is the literal rendering, but in interface French "Afficher" is the usual verb for making something visible, and *exposer* can pull toward "put on display" or "expose to risk".  Given the accessibility framing, would "Afficher les permaliens" carry the intent more directly?

**`readability-content-font-size` → "Taille de police".**  There is only one font-size control here, a slider running 0.8 to 1.5, and the word *content* in the English is doing a job: it scales the body text, not the surrounding interface.  Dropping it may lead a reader to expect the whole page furniture to grow.  "Taille du texte" would keep the label short while still pointing at the content — or "Taille de police du contenu" if the distinction is worth the length.

**`readability-settings` → "Paramètres de lecture".**  Reading settings rather than readability settings.  It may well be the friendlier phrase for a menu, so this is the lightest of the four — I mention it only because *lisibilité* is available if you want the exactness.

The rest read very naturally to me.  **"Interligne"** is exactly the typographic term and better than a literal rendering of "line height" would have been; *Thème / Système / Clair / Sombre* and *Technologies d'assistance* are the conventional forms; and in `fr-CA` the lowercase **"permalien"** correctly matches the "lowercase strings to click on" grouping it sits in.

PR #3167, opened 2026-08-20
GDScript : allow multiple architectures for linux

DavidCooperWCU:
If pretext is running on a linux with arm 64 instead of x86 it currently won't download the right version of godot. This code fixes that.

### DIFF
--- a/pretext/lib/godot_helper.py
+++ b/pretext/lib/godot_helper.py
@@ -204,8 +204,10 @@ def _cached_binary_path() -> str | None:
         candidate = os.path.join(CACHE_DIR,
                                  "Godot_v{}_win64.exe".format(GODOT_VERSION_TAG))
     else:
+        machine = platform.machine().lower()
+        arch = "arm64" if machine in ("aarch64", "arm64") else "x86_64"
         candidate = os.path.join(CACHE_DIR,
-                                 "Godot_v{}_linux.x86_64".format(GODOT_VERSION_TAG))
+                                 "Godot_v{}_linux.{}".format(GODOT_VERSION_TAG, arch))
     return candidate if os.path.exists(candidate) else None
 
 


rbeezer (commented):
This is right, and it fixes more than the title suggests.

`_asset_name_for_platform()` already selects `linux.arm64` when `platform.machine()` reports `aarch64` or `arm64`, while `_cached_binary_path()` named `linux.x86_64` unconditionally. On an arm64 Linux machine the correct archive is therefore downloaded and extracted, and then `_download_and_extract_godot()` asks `_cached_binary_path()` for the executable, receives `None`, and raises `Extracted … but couldn't locate the Godot executable`, which points at the archive's internal layout rather than at the name being looked for. So this is a bug fix rather than an enhancement — Godot support does not work at all on arm64 Linux today — and it is a no-op on x86_64.

The two architecture strings match what Godot publishes: `Godot_v<tag>_linux.arm64.zip` and `Godot_v<tag>_linux.x86_64.zip`. Releases also carry `arm32` and `x86_32`, which neither the download side nor this one selects, so a 32-bit machine still resolves to `x86_64`. That is unchanged behaviour, worth knowing only because the title says "multiple architectures".

One thing for later rather than for this pull request: the architecture expression now appears twice, identically. That the two copies must agree is exactly what went wrong here, so a small `_linux_arch()` called from both would keep them in step.

On the commit message: the subject wants `Topic: description` with no space before the colon, and `Linux` capitalised. `GDScript:` is also not quite the topic — this concerns downloading the Godot binary rather than the scripting language — and it has no precedent in the history, whereas `Script:` is well established for changes under `pretext/lib/`.

rbeezer:
Merged with code as-is, commit message edited to fit expectations noted above.  A consolidating routine (also noted above) could be added later.

PR #3168, opened 2026-08-20
Permalinks for inline headings

ascholerChemeketa:
Cleanup for inline heading logic in CSS based on:
https://github.com/PreTeXtBook/pretext/pull/3150

Ties link "hide permalink" for affected elements to the fact that a particular element has an inlined heading, not on its type (`.solution-like`, `.proof`, etc...).

### DIFF
--- a/css/components/chunks/_discussion-inline.scss
+++ b/css/components/chunks/_discussion-inline.scss
@@ -26,10 +26,6 @@
       content: "\2009";
     }
 
-    & + .para {
-      display: inline;
-    }
-
     .space,
     .codenumber,
     .period {
--- a/css/components/chunks/_exercises.scss
+++ b/css/components/chunks/_exercises.scss
@@ -18,6 +18,7 @@
 //********************************************************************
 
 @use '../helpers/cols';
+@use './helpers/inline-heading-mixin';
 
 // generate multi column rules for exercises
 @include cols.allow-cols('.exercise-like');
@@ -32,6 +33,7 @@
 }
 
 .exercisegroup {
+  @include inline-heading-mixin.heading;
 
   .exercise-like {
     margin-top: 1em;
@@ -41,15 +43,6 @@
     font-size: 1.10em;
     line-height: 1.05em;
     margin-top: 0.75em;
-    display: inline;
-
-    & + .introduction {
-      display: inline;
-
-      & > .para:first-child {
-        display: inline;
-      }
-    }
   }
 
   // push the actual exercises down from any possible heading/intro
--- a/css/components/chunks/helpers/_inline-heading-mixin.scss
+++ b/css/components/chunks/helpers/_inline-heading-mixin.scss
@@ -32,19 +32,21 @@
     // For now limit to logical paragraphs and introductions as they seem to cover
     // the plain text cases. Not clear we would want same treatment for other
     // kinds of content
-    & + :is(.para, .para.logical, .introduction) {
+    & + :is(.para, .para.logical, .introduction):nth-child(2) {
       display: inline;
-      // possible first child that is the actual text
-      & > .para:first-child {
-        display: inline;
-        }
-      // an introduction could have a para.logical with a para
+      // Disable the autopermalink for this first child that is now inline
+      & > .autopermalink {
+          display: none;
+      }
+      
+      // possible first child that is the actual text also needs to be inlined
+      // and have permalink suppressed.
+      & > .para:first-child,
       & > .para.logical:first-child > .para:first-child {
         display: inline;
-      }
-      // Disable the autopermalink for this first child that is now inline
-      & .autopermalink {
-        display: none;
+        & > .autopermalink {
+          display: none;
+        }
       }
     }
   }
--- a/css/components/chunks/helpers/_inline-heading-mixin.scss
+++ b/css/components/chunks/helpers/_inline-heading-mixin.scss
@@ -42,6 +42,10 @@
       & > .para.logical:first-child > .para:first-child {
         display: inline;
       }
+      // Disable the autopermalink for this first child that is now inline
+      & .autopermalink {
+        display: none;
+      }
     }
   }
 }
--- a/css/components/elements/_permalinks.scss
+++ b/css/components/elements/_permalinks.scss
@@ -86,14 +86,19 @@ li > .autopermalink {
   left: -2.9em;
 }
 
-
-.axiom-like, .example-like, .exercise-like, .solution-like, .assemblage-like, .definition-like, .remark-like, .project-like, .openproblem-like, .computation-like, .theorem-like, .proof, .case, li, dd {
+// Hide the autopermalink for the first para of a list item or definition term
+li, dd {
   > .para:first-of-type > .autopermalink {
     display:none;
   }
+}
 
+// An unheaded solution-like or assemblage-like block and its first paragraph
+// occupy the same starting position.  Show only the block permalink there.
+:is(.solution-like, .assemblage-like):not(:has(> .heading)) {
+  > .para:first-of-type > .autopermalink,
   > .introduction > .para:first-of-type > .autopermalink {
-    display:none;
+    display: none;
   }
 }
 
--- a/css/components/elements/_permalinks.scss
+++ b/css/components/elements/_permalinks.scss
@@ -18,7 +18,7 @@
 //********************************************************************
 
 // TODO - refactor
-$opacity: 0.0 !default;
+$opacity: 1 !default;
 
 /*
 So that we can position things (like .autopermalink) absolutely wrt these items.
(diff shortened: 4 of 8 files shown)


rbeezer:
I'm working on the assumption this replaces https://github.com/PreTeXtBook/pretext/pull/3150 - that is not 100% clear to me from reading comments. Report from Claude, just now.  I'm not concerned about degrading legacy themes.

rbeezer:
The direction here is right — suppressing the permalink where a heading is genuinely inline, rather than by a list of block names, is the version that will keep working as themes change. Bringing `.exercisegroup` and the three themes' `.solution-like` onto the shared mixin is a real simplification, and the exercisegroup case it fixes is the one reported.

Building the branch and comparing it against master on identical HTML, though, the rule that comes out of `_permalinks.scss` was doing more work than the mixin currently replaces.

That rule hid the first-paragraph permalink across a dozen block types. The mixin only fires where a block actually has an inline heading, which is the point — but a few blocks were relying on the old rule and do not get the mixin. On the sample article with `theme-default-modern`:

| first-paragraph permalink | master | this branch |
|---|---|---|
| `exercisegroup` | 1 visible (the bug) | 0 visible ✓ |
| `solution-like` | 0 visible, 3 hidden | **3 visible** |
| `assemblage-like` | 0 visible, 1 hidden | **1 visible** |
| theorem, remark, example, proof, definition, computation, project, exercise | hidden | hidden ✓ |

`.solution-like` picks up the mixin in denver, greeley and tacoma, but not in default-modern, which is why solutions fall through in the default theme. Three more appear inside knowl content on the exercises page — not visible until a reader opens the knowl, but they are there.

Those blocks all carry their own permalink already, so a second one on their first paragraph is duplication, which is presumably what the old rule was for. Either extending the mixin to cover them, or keeping a narrowed version of the old rule for blocks that have a permalink of their own, would close it.

One other thing worth knowing, since it bears on whether this closes the issue by itself rather than on anything in the diff. The exercisegroup fix rides on the component layer, and the legacy themes do not pull that in, so it does not reach them:

| built theme | exercisegroup suppression |
|---|---|
| `theme-default-modern` | present |
| `theme-default-legacy` | absent |

#3150 addresses the same symptom in the stylesheet instead, where the permalink is never generated and every theme is covered. The two are not in conflict — if that one lands, it simply makes the exercisegroup half of this redundant, and the consolidation here is still worth having on its own terms.

ascholerChemeketa:
Yes, this is offered as a deeper fix for the problem #3150 addresses.

The reason permalinks for `exercisegroup`'s first `p` need to be suppressed is that the `exercisegroup`s have an inline heading. An alternate presentation of `exercisegroup` with a stand alone heading would not need to suppress the heading. So we don't want to say *"excercisegroup's first p never needs a permalink"* as that statement is overly broad.

Updated to address issue Claude found. That fix revealed another wrinkle - `default-modern` had some constructs getting styles for both inline and not inline headings.

rbeezer:
> Updated to address issue Claude found

Did you update the commits?  I'm not seeing them here.

ascholerChemeketa:
> Did you update the commits? I'm not seeing them here.

I did. But I guess that doesn't help much if I don't also push them. :)

Sorry about that!

rbeezer:
Claude has a lot to say - I hope it is not noise.

rbeezer:
Thanks for pushing these. The `exercisegroup` fix is solid now: on `section-further-reading.html` the five visible first-paragraph permalinks go to zero, and that holds in all four modern themes.

The regressions from the last round are still there, though, and the branch adds one. Building the branch and master into the same HTML and swapping only the compiled theme, on `section-interesting-corollary.html` with `theme-default-modern`, master shows **zero** visible first-paragraph permalinks and the branch shows nineteen:

| block | element | its own heading | master | this branch |
|---|---|---|---|---|
| `solution-like` | `div` | present, renders block | hidden ×13 | **visible ×13** |
| `assemblage-like` | `article` | present, renders block | hidden ×2 | **visible ×2** |
| `exercise-like` | `article` | none | hidden ×4 | **visible ×4** |
| `solution-like` | `div` | none | hidden ×47 | hidden ×47 ✓ |
| `exercise-like` | `article` | inline | hidden ×38 | hidden ×38 ✓ |

The thirteen solutions and two assemblages are on the page itself, not inside knowls; the four exercises are knowl content. The set differs by theme, because each theme attaches the mixin differently:

| theme | newly visible on that page |
|---|---|
| `theme-default-modern` | 19 |
| `theme-denver` | 10 |
| `theme-greeley` | 4 |
| `theme-tacoma` | 44 |
| `theme-default-legacy` | unchanged — the component layer is not pulled in |

Three separate causes, all the same shape:

1. **`solution-like` is a `div`, not an `article`.** In `_chunks-default.scss` the mixin goes to `.paragraphs, article:not(...)`, so a solution never receives it in default-modern — which is why denver, greeley and tacoma pick it up from their own `.solution-like` blocks and default-modern does not. With a heading present, the new `:not(:has(> .heading))` rule does not apply either, so it falls between the two.

2. **The `assemblage-like`/`goal-like` exclusion in commit 3 also removes their permalink suppression.** Putting `.autopermalink { display: none }` inside the heading mixin ties "suppress the permalink" to "receives the inline-heading mixin", so opting a block out of the mixin for a heading reason silently opts it out of the permalink rule too. Worth knowing: on the sample article those headings compute as `block` in default-modern on master *and* on the branch, so the exclusion changes their heading rendering not at all here — only the permalink.

3. **`exercise-like` with no heading** gets the mixin (it is an `article`) but the mixin keys on `& > .heading:first-child`, which is absent. The old blanket rule listed `.exercise-like`; the new rule lists only `solution-like` and `assemblage-like`.

Separately, the new suppression reaches further than the comment claims:

```scss
& + :is(.para, .para.logical, .introduction) {
  & .autopermalink { display: none; }
}
```

`& .autopermalink` is a descendant selector, so inside an `introduction` it hides *every* paragraph's permalink, not the first one's. The first `exercisegroup` on `section-further-reading.html` has a two-paragraph introduction; on master both permalinks show, on the branch both are hidden. Only the first is inline with the heading. The comment above it says "this first child that is now inline", which is the intent — a child combinator, or scoping to the first `.para`, would match it.

One small thing while I was in there: the `li, dd` comment says "definition term", but `dd` is the description and `dt` is the term.

No `css/dist` in the diff, which is right, and no whitespace damage.

ascholerChemeketa:
I'm not convinced the newly visible permalinks are incorrect.

Take this old rendering with Tacoma:
image

There isn't a principled way to explain what paragraphs have permalinks and which don't. Most first paragraphs after headings do not. But the one after 4.2.2 does.

For consistency, I would say every one of those paragraphs should have a link.

Permalinks are a mess. We have a mostly ad-hoc list of places where they get emitted in the XSL. We have a mostly ad-hoc list of places where they are suppressed or repositioned in the CSS and much of that logic is based on correlative criteria and not root causes. (e.g. disabling the permalink on first para of any exercise - that makes sense if the exercise has no heading, or an inline heading, but only makes sense because of those reasons, not because it is an exercise.) Trying to impose more principled logic about what is turned off and why is going to necessarily make some changes.

I think there are some legitimate issues claude found, I'll go review them. But the metric for success here should not be "no permalinks other than the initial target change".

ascholerChemeketa:
Pushed with intentionally unclean commits to show new work and make visual debugging easier.

`FIXUP for consolidate inline heading` fixes a genuine unintended change to Tacoma rendering

Visually audited default-modern/tacoma/denver. This does result in new permalinks, but I would argue that they now follow much clearer logic and the new ones are not ones that visually overlap.

If we are happy with the results I can do a rebase to move the fixups into existing commits and drop the temp.

rbeezer:
Your metric is the better one, and the branch holds up on it. I re-measured, asking of every permalink the branch newly reveals whether it lands on its block's left border or intrudes past the padding edge into the text column:

| page | theme | newly visible | on a border | intruding into text |
|---|---|---|---|---|
| `section-interesting-corollary` | default-modern | 19 | 0 | 0 |
| | denver | 49 | 0 | 0 |
| | greeley | 4 | 0 | 0 |
| | tacoma | 111 | 0 | 0 |
| `section-further-reading` | tacoma | 34 | 0 | 0 |
| `section-worksheet-solutions-testing` | all four | 2–20 | 0 | 0 |

Not one bad placement, so the additional permalinks are not colliding with anything. The `exercisegroup` suppression still holds — five on `section-further-reading` in every theme.

Both of the things I raised are fixed. The descendant selector is now a child selector with `:nth-child(2)`, and the two-paragraph `exercisegroup` introduction on `section-further-reading` now hides the first permalink and keeps the second, where before it hid both. And the Tacoma rendering change you caught yourself.

No need to rebase. I will fold the two fixups into their commits and drop the TEMP when I take this in, so leave the branch as it is.

rbeezer:
OK, thanks for sticking with this one.  Maybe Claude has the message now.

I cleaned-up the commits, so we could finish this off, and now merged.  See you on #3201 in a bit.

Issue #1504, opened 2021-06-28
Keyboard based image magnification


Alex-Jordan:
The current method to magnify an image relies on mouse interaction (I think). The user must click on the image to get the larger version.

If the image (or its containing div) had `tabindex="0"`, it could attain focus through tabbing, and the magnified image javascript could be modified such that hitting Enter would activate the magnified version of the image.

ascholerChemeketa:
Also need ESC to close.

Might be good to bundle into a proper `<dialog>` element:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog

rbeezer:
Closed via #3170.

PR #3170, opened 2026-08-21
Image popups for HTML builds improvements

ascholerChemeketa:
Fixes some accessibility issues, expands the size images are allowed to grow, incorporates PreFigure and Mermaid.

Addresses: https://github.com/PreTeXtBook/pretext/issues/1504

Relevant discussion here: https://groups.google.com/g/pretext-dev/c/dsizBweE5kA/m/bce9MYI7DAAJ


ascholerChemeketa:
Added features requested on -dev:
1) Asymptote handling
2) Enlargement of low resolution raster images

rbeezer:
Much nicer with this change!  Thanks for putting this together.

I had to work to rearrange some of these commits, one reason was Oscar's recent work to spit out some JS into a seperate file.  Full report in a minute.

I aslo added a commit so that low-res raster images got teh exact same treatment as every other image - that was the change I had been after.  Follow-on commit, in my name.

rbeezer:
Replaces the old click-to-magnify behavior on images with an accessible dialog, and extends it to the graphics formats the old one never handled.

An image in HTML output can now be opened in a dialog by click or from the keyboard, and the dialog understands raster images, SVG, Mermaid diagrams, PreFigure diagrams, and both two- and three-dimensional Asymptote. Three-dimensional Asymptote keeps its drag-to-rotate behavior inside the dialog rather than treating a drag as a click. The jQuery magnifier and its styles are removed.

Merged, with thanks — this is your work, transplanted onto current master and landed from a rebuilt branch. Recording here what changed on the way through, so the log and this thread agree.

### Everything but the last commit is your work, unaltered

The branch predated the printout work, so it was rebuilt on today's master rather than merged. **The code itself is untouched.** The stylesheets and the new script are byte-for-byte what you wrote; the only two files that read differently are the two master moved under us, and there they carry both changes rather than either replacing the other.

What changed is procedural, and none of it alters behavior.

The built bundle `js/dist/pretext-core.js` had been committed on one of the commits, and was left out. Generated files are rebuilt at merge rather than carried in a branch, and this is a tidy illustration of why: git merges two independently built bundles textually without complaint, and the result parsed fine but came out several kilobytes short of what a real build of the same sources produces. Only sources went in, and the dist files have been rebuilt since.

The new figure gained a `shortdescription`, which put validation back exactly where master sits — without one it was the single image in the sample article carrying an `image-description-missing` warning.

Three small text corrections went into the new prose, the notable one being a sentence that had lost its ending: "larger than the , then it will be scaled down" now reads "larger than the content width".

Commits that touched both stylesheets and scripts were split into a CSS commit and a JavaScript commit, matching how the printout work was organized, so each commit carries a single topic already established in the log.

### The last commit is a real change, and the one I have been asking for

Everything above is bookkeeping. This one is not.

The original limits a raster image in the dialog to three times its natural size, to avoid presenting something badly pixelated. The effect, visible on the small screenshot added here, is that a low-resolution image opens in a noticeably smaller window than every other kind of image.

That is the behavior I have been asking for in the discussion around this: a reader who asks to enlarge an image should get the same enlargement whatever the image is made of. So the cap is gone, and every image now fills the same space regardless of resolution. A small source image is soft at that size, and that is the honest presentation — the reader sees what is actually there, at the size they asked for. Two variables existed only to support the cap and go with it, which is most of that diff.

PR #3171, opened 2026-08-21
HTML: let the DoenetML coordinator manage a page's activities

dqnykamp:
Proposed on pretext-dev: https://groups.google.com/g/pretext-dev/c/kkF9ZJst3NI

## The problem

A page carrying many DoenetML activities starts all of them the moment it loads, and then holds all of them in memory for as long as the reader stays. On modest hardware the activities can take minutes to appear, or fail outright with "The document viewer could not be started."

## What this adds

DoenetML publishes a coordinator script alongside the bundle the activities already load. Given one script tag on the page that *holds* the activities, it takes charge of their iframes:

* it starts them a couple at a time as the reader nears them, so an activity the reader never scrolls to never starts at all;
* it parks the ones the reader has left well behind — flushing their state first and handing it back if the reader returns, so no work is lost. The iframe element stays in the layout while parked, so nothing shifts;
* with shared core workers, the activities draw on one pool of background workers instead of each taking a dedicated one, which is the bulk of what an activity costs.

The activity pages need no cooperation: the coordinator marks each activity's URL and the DoenetML bundle recognizes the mark. So the whole of it is the script tag, plus a `data-doenet-coordinate` attribute on the activity iframes so that only DoenetML activities are swept up — the script's own default selector would take every `*-if.html` iframe, which for PreTeXt means every other platform's interactives too.

The script goes in the head shared by content pages, and deliberately not in the one iframe pages get — it belongs on the page that holds the activities, not inside one. Standalone pages for a single interactive do not get it either, having nothing to coordinate.

## Measurements

The section of Active Calculus that prompted this (ten activities, https://mattboelkins.github.io/active-calculus-single-mbx/doenet/sec-1-3-derivative-pt.html), in headless Chrome, memory summed (PSS) across every browser process. Both arms pin DoenetML 0.7.25, so the coordinator is the only variable:

| scenario | without coordinator | with coordinator |
| --- | --- | --- |
| open the page and leave it | peak 2871 MB, settled 1446 MB | peak 986 MB, settled 552 MB |
| scroll through the whole page | peak 2881 MB, settled 1426 MB | peak 1481 MB, settled 775 MB |

In the second row every one of the ten activities is brought into view and read, so everything boots at least once; even then the peak is halved, and memory comes back down when the reader stops, which without the coordinator it never does. The first activity also appears sooner, since the page no longer starts all ten at once.

## Publisher entries

New entries on `html/interactives/doenetml`, documented in the publisher's guide:

* `@coordinator` — `"yes"` (default) or `"no"`;
* `@shared-core-workers` — `"yes"` (default) or `"no"`;
* `@max-live-viewers` and `@max-concurrent-boots` — positive integers for a publisher who wants to tune how much is allowed at once (defaults 3 and 2).

The schema source, the `.rnc`, and the `.rng` are all updated; the numeric entries are declared as positive integers, so the schema rejects what the guide forbids.

## Versions

The DoenetML version in use is a property of the source (`docinfo/doenetml/@version`), not of the publication file. A book that names no version or a partial one such as `0.7` gets the newest release, which is what these options are written for. The coordinator script first shipped in DoenetML `0.7.21`, so a book pinned earlier simply will not find it and behaves as before; a book pinned into `0.7.21`–`0.7.24` should set `@coordinator` to `no` — the build now warns when it can see that combination (a pin in that range together with a Runestone or SCORM host) — because in that range a restored activity is rebuilt once per party answering its request for saved state, and on some hosts the slower, staler answer could overwrite the reader's newer work. Both points are spelled out in the guide text this PR adds; `0.7.25`, current as of this PR, is safe.

## The SCORM commit

The second commit hardens `js/ptx_scorm_events.js` independently of any publisher choice. A `SPLICE.getState` request that arrives before the SCORM restore data is ready is queued and answered later, and the answer posts to the window that asked. With the coordinator managing the page, that window may be gone by then: parking an activity points its iframe at `about:blank`, which discards the browsing context the request came from. Measured in Chrome, the `MessageEvent.source` captured at arrival then reads back as `null` — while the iframe's own `contentWindow` stays valid — so posting to it throws, and since the throw happens inside the loop that drains the queue, every activity still queued behind it silently lost its saved state too. A missing window is now skipped and the post is guarded, so one departed frame costs only its own response.

🤖 Generated with [Claude Code](https://claude.com/claude-code) using Claude Fable 5, after earlier drafting by Claude Opus 5

### DIFF
--- a/doc/guide/publisher/online-html.xml
+++ b/doc/guide/publisher/online-html.xml
@@ -109,6 +109,23 @@
             <p>A window that allows for entering, and executing, computer programs can be made available for every page.  The reader can click on a pencil icon to activate this window.  Some languages can run in a web browser as part of any <init>HTML</init> output, while some other languages require infrastructure on a Runestone server to execute, and so are only available when you specify that hosting option.  Note that the publisher will select a single language for use with the entire document. See <xref ref="online-activecode"/> for details on specifying this option.</p>
         </subsection>
 
+        <subsection xml:id="online-html-doenetml">
+            <title>DoenetML Activities</title>
+            <idx><h>DoenetML</h><h>coordinator</h></idx>
+
+            <p>A DoenetML activity is a substantial piece of software.  A page carrying a dozen of them once started every one the moment the page opened, and held every one in memory for as long as the reader stayed<mdash/>enough, on a modest machine, that the activities could be slow to appear, or fail to appear at all.  So a page carrying DoenetML activities loads a small <term>coordinator</term> script, served from the same place as the activities themselves, which takes charge of them: it starts them a couple at a time as the reader nears them, so an activity the reader never scrolls to is never started, and it puts away the ones the reader has left well behind.  A reader's work is saved before an activity is put away and restored if the reader returns to it, and the space the activity occupies on the page is held while it is away, so nothing shifts.  This is on by default and needs nothing from an author, though a publisher can turn it off (<xref ref="online-doenetml-options"/>).</p>
+
+            <p>The coordinator also lets the activities on a page share a pool of the background workers they compute in.  A worker is the largest part of what an activity costs in memory, so the pool is most of the coordinator's saving; it too can be turned off, giving each activity a worker of its own.  Turning the coordinator off turns the pool off with it, since the coordinator is what owns the pool.</p>
+
+            <p>Two limits tune how much the coordinator allows at once: the number of activities kept running, and the number allowed to be starting up at the same time.  A book whose activities are small might raise them; but raising them asks more of the reader's machine, which is what the coordinator exists to avoid, and the defaults suit a reader on modest hardware.</p>
+
+            <p>The version of DoenetML in use, including the version of the coordinator, comes from the <tag>doenetml</tag> element of <tag>docinfo</tag> rather than from the publication file, since it is a property of the source: see <xref ref="topic-docinfo"/>.  A book that names no version, or names a partial one such as <c>0.7</c>, gets the newest release, which is what this machinery is written for.  A book that pins an exact version should pin <c>0.7.25</c> or later.</p>
+
+            <p>Two earlier boundaries matter to a book that does pin one.  The coordinator first shipped with DoenetML <c>0.7.21</c>, so a book pinned before that will not find the script at all and behaves as if the coordinator were turned off.  From <c>0.7.21</c> through <c>0.7.24</c> the script is there, but it does not yet share a page well with a book that saves the reader's work itself, as a Runestone or <init>SCORM</init> deployment does.  Two parties then answer an activity's request for its saved work<mdash/>the coordinator, from what it took down when it put the activity away, and the book's own saving<mdash/>and before <c>0.7.25</c> each answer rebuilt the activity in turn, so every restore built the document twice over.  That is the very cost the coordinator is there to avoid.  And since the answer that arrived last was the one kept, a book whose saving answers out of storage it has to go and fetch could have the fresher copy the coordinator was holding overwritten by an older one, leaving the reader to find work they had just done replaced.  A book pinned into that range should turn the coordinator off; a build that can see the combination will say so.</p>
+
+            <p>See <xref ref="online-doenetml-options"/> for the publication file entries controlling all of the above.</p>
+        </subsection>
+
         <subsection xml:id="online-base-url">
             <title>Base URL</title>
             <idx>base URL</idx>
--- a/doc/guide/publisher/publication-file.xml
+++ b/doc/guide/publisher/publication-file.xml
@@ -1099,6 +1099,51 @@
             </p>
         </subsection>
 
+        <subsection xml:id="online-doenetml-options">
+            <title>DoenetML Options</title>
+            <idx><h>publisher options</h><h>DoenetML</h></idx>
+            <idx><h>DoenetML</h><h>coordinator</h></idx>
+
+            <p>
+                A page carrying DoenetML activities loads a small <term>coordinator</term> script that manages them, so that a page carrying many activities need not run them all at once.  The
+                <cd>
+                    <cline>/publication/html/interactives/doenetml</cline>
+                </cd>
+                element can have an attribute <attr>coordinator</attr> with values:<ul>
+                    <li><c>yes</c>: the default, let the coordinator manage the DoenetML activities on a page.</li>
+                    <li><c>no</c>: start every activity on a page at once, and keep every one of them running.</li>
+                </ul>
+            </p>
+
+            <p>
+                The same element can have an attribute <attr>shared-core-workers</attr> with values:<ul>
+                    <li><c>yes</c>: the default, let the activities on a page share a pool of the background workers they compute in.</li>
+                    <li><c>no</c>: give each activity a worker of its own.</li>
+                </ul>
+            </p>
+
+            <p>
+                Two attributes tune how much the coordinator allows at once: <attr>max-live-viewers</attr>, the number of activities kept running (default <c>3</c>), and <attr>max-concurrent-boots</attr>, the number allowed to be starting up at the same time (default <c>2</c>).  Both take a positive integer; any other value is reported during the build and the default used.
+            </p>
+
+            <p>
+                An example, for a book whose activities are small enough that more of them can be alive at once:
+                <cd>
+                    <cline>&lt;publication&gt;</cline>
+                    <cline>    &lt;html&gt;</cline>
+                    <cline>        &lt;interactives&gt;</cline>
+                    <cline>            &lt;doenetml max-live-viewers="6" max-concurrent-boots="3"/&gt;</cline>
+                    <cline>        &lt;/interactives&gt;</cline>
+                    <cline>    &lt;/html&gt;</cline>
+                    <cline>&lt;/publication&gt;</cline>
+                </cd>
+            </p>
+
+            <p>
+                See <xref ref="online-html-doenetml"/> for more information, including a caution for a book that pins its DoenetML version.
+            </p>
+        </subsection>
+
         <subsection xml:id="online-read-aloud-options">
             <title>Read Aloud Options</title>
             <idx><h>read aloud</h><h>publisher options</h></idx>
--- a/js/ptx_scorm_events.js
+++ b/js/ptx_scorm_events.js
@@ -484,16 +484,34 @@
     // Subject must be "SPLICE.getState.response" with the message_id echoed back.
     // State must be a JS object (not a JSON string) because Doenet checks state.cid directly.
     function respondToGetState(sourceWindow, messageId) {
+        // A request that arrives before the restore data is ready is queued and
+        // answered later, by which time its frame may be gone.  The DoenetML
+        // coordinator parks an activity the reader has scrolled well past by
+        // pointing its iframe at about:blank, and that discards the browsing
+        // context the request came from: the MessageEvent.source captured when
+        // the request arrived reads back as null (verified in Chrome; the
+        // iframe's own contentWindow stays valid, it is the captured source
+        // that goes), so posting to it throws.  Unguarded, that throw escapes
+        // the loop draining the queue and every activity still behind it
+        // loses its saved state.
+        if (!sourceWindow) {
+            console.log('[PTX-SCORM] SPLICE.getState: requesting frame is gone; skipping response');
+            return;
+        }
         var divId = resolveIframeId(sourceWindow);
         var stateObj = divId ? loadDoenetState(divId) : null;  // JS object; Doenet checks state.cid
         console.log('[PTX-SCORM] SPLICE.getState from "' + (divId || '?') + '" — ' +
                     (stateObj ? 'sending saved state (cid: ' + (stateObj.cid || '?') + ')' : 'no saved state (first visit)'));
         if (divId) _doenetSentStateAt[divId] = Date.now();
-        sourceWindow.postMessage({
-            subject:    'SPLICE.getState.response',
-            message_id: messageId,
-            state:      stateObj || null
-        }, '*');
+        try {
+            sourceWindow.postMessage({
+                subject:    'SPLICE.getState.response',
+                message_id: messageId,
+                state:      stateObj || null
+            }, '*');
+        } catch (e) {
+            console.log('[PTX-SCORM] SPLICE.getState: could not reach the requesting frame: ' + e);
+        }
     }
 
     /**
--- a/schema/publication-schema.rnc
+++ b/schema/publication-schema.rnc
@@ -337,13 +337,25 @@
                     calcplot3d |
                     circuitjs |
                     d3 |
-                    doenetml |
                     iframe |
                     javascript |
                     jsxgraph |
                     sage
                 ) {
                     attribute resize-behavior { "fixed-height" | "responsive" }?
+                }
+                | Doenetml
+
+            # DoenetML activities are additionally managed on the page by a
+            # "coordinator" script, which starts them a few at a time as the
+            # reader nears them and puts away the ones left well behind.
+            Doenetml =
+                element doenetml {
+                    attribute resize-behavior { "fixed-height" | "responsive" }?,
+                    attribute coordinator { "yes" | "no" }?,
+                    attribute shared-core-workers { "yes" | "no" }?,
+                    attribute max-live-viewers { xsd:positiveInteger }?,
+                    attribute max-concurrent-boots { xsd:positiveInteger }?
                 }
 
             Knowls =
--- a/schema/publication-schema.rng
+++ b/schema/publication-schema.rng
@@ -960,25 +960,70 @@
     </element>
   </define>
   <define name="InteractiveTypes">
-    <element>
-      <choice>
-        <name>geogebra</name>
-        <name>desmos</name>
-        <name>calcplot3d</name>
-        <name>circuitjs</name>
-        <name>d3</name>
-        <name>doenetml</name>
-        <name>iframe</name>
-        <name>javascript</name>
-        <name>jsxgraph</name>
-        <name>sage</name>
-      </choice>
+    <choice>
+      <element>
+        <choice>
+          <name>geogebra</name>
+          <name>desmos</name>
+          <name>calcplot3d</name>
+          <name>circuitjs</name>
+          <name>d3</name>
+          <name>iframe</name>
+          <name>javascript</name>
+          <name>jsxgraph</name>
+          <name>sage</name>
+        </choice>
+        <optional>
+          <attribute name="resize-behavior">
+            <choice>
+              <value>fixed-height</value>
+              <value>responsive</value>
+            </choice>
+          </attribute>
+        </optional>
+      </element>
+      <ref name="Doenetml"/>
+    </choice>
+  </define>
+  <!--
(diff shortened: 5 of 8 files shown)


rbeezer:
I like the attribution to AI - adding the model would be even better.  The link to the session might be interesting, too, but now it is a dead link.

rbeezer:
Thanks for this — the XSL is carefully written, and the comments explain the parts that would be hard to reconstruct later. Notes below; the first is the substantial one.

### Publisher options are documented in two places, cross-linked both ways

This is a firm convention in the Guide, and the PR currently does half of it.

A new publication-file option gets **both**:

1. **A terse entry in the reference chapter**, `doc/guide/publisher/publication-file.xml` — the `<cd><cline>/publication/…/@attribute</cline></cd>` line, the permitted values and defaults, an `<idx>`, and an `<xref>` out to the fuller treatment. That chapter is organized by output and concern so a publisher can scan it. It is not where explanations live.
2. **A conversational treatment** — for HTML options, `doc/guide/publisher/online-html.xml`. This is where the discussion goes: why the feature exists, what it does for a reader, when a publisher would change it. It ends with an `<xref>` back to the reference entry.

The pair of xrefs is what makes the two halves navigable, and both directions are required.

The new subsection's own neighbours are the models. `online-read-aloud-options` closes with "See `<xref ref="online-html-read-aloud"/>` for more information", and `online-html-read-aloud` links back to it. Annotation, the embedded calculator, and the favicon all follow the same shape.

As it stands `online-doenetml-options` has no conversational counterpart and no xref to one — its only xref goes to `topic-docinfo`, which covers `docinfo/doenetml/@version` rather than these switches. The discussion has gone into the reference chapter instead, and at 666 words it is now the longest of that chapter's 71 subsections. The median is 74; the next longest is 594.

Concretely:

- **Keep in `publication-file.xml`:** the `/publication/html/interactives/doenetml` entry, the four attributes with their values and defaults, the `<idx>` entries, the example, and a closing "See `<xref ref="online-html-doenetml"/>`".
- **Move to a new `online-html-doenetml` subsection in `online-html.xml`:** the opening paragraph on why the coordinator exists, the note that `shared-core-workers` has no effect when `coordinator` is `no`, the guidance on raising the two limits, and the version-boundary paragraph. Close it with `<xref ref="online-doenetml-options"/>`.

That keeps the reference scannable, and gives the version discussion room to be as long as it needs to be.

### A hazard the build could catch

The version paragraph documents something sharp: pinned between `0.7.21` and `0.7.24`, in a book that saves the reader's work itself, a reader can find work they just did replaced by an older copy. The remedy offered is that the publisher should set `coordinator` to `no`.

Both facts needed to detect that combination are available during the build — `$docinfo/doenetml/@version`, and `$b-host-runestone` / `$b-host-scorm`. Would a `PTX:WARNING` be worth adding? Losing a reader's work seems like a lot to rest on whether the publisher read a particular paragraph.

### Smaller things

`data-doenet-coordinate="true"` is placed on every DoenetML `interactive`, but the coordinator script is added to `file-wrap-simple-head-cache`, and standalone interactive pages are built from `file-wrap-basic-head-cache`. So `interactive-doenetml-example.html` marks its activity and then loads nothing to manage it. It is inert, since the attribute is only the coordinator's selector, but the two conditions no longer agree.

Could `js/dist/ptx_scorm_events.js` come out of the PR? Regenerating the built bundles is a maintainer step at merge time, and committed dist files in a PR tend to conflict irreconcilably across PRs. The source change is the part to review, and it reads correctly. This is a newish policy and easy to miss.

Two small commit points: the convention here is a subject line plus a `Co-Authored-By` trailer and nothing else, so the `Claude-Session:` lines would come out; and the two subjects disagree on the name, "DoenetML" in one and "Doenet" in the other.

### What I checked

The derived schema files match a fresh regeneration exactly. The Guide validates with nothing reported in the new material. `PTX:FALLBACK` is the right severity here, and the positive-integer validation follows the existing `braille-page-width` idiom. The generated markup is right: one script tag carrying the four data attributes, the selector confined to marked iframes, and the shared-core-workers value stringified the way the script reads it.

One honest limit: I could not confirm runtime behaviour. DoenetML activities do not render in my test environment on master either — the applet mounts and stops, with no console errors — so while I can watch the coordinator load and take control of an iframe, I could not see deferred starting or state restore actually happen.

dqnykamp:
Thank you for the careful review — every point was right, and all of them are now in the branch (force-pushed).

**Docs split.** `online-doenetml-options` in the reference chapter is now a terse entry: the element path, the four attributes with values and defaults, the index entries, and the example, closing with an xref out. The discussion — why the coordinator exists, the worker pool, the guidance on the two limits, and the version boundaries — moved to a new `online-html-doenetml` subsection of `online-html.xml`, which closes with the xref back. Thanks for spelling out the convention and pointing at the neighbouring models; that made the split unambiguous.

**The build now catches the hazard.** Yes — a warning was worth adding, and it cost little: when the document pins a version from `0.7.21` through `0.7.24` and the build is for a Runestone or SCORM host (with the coordinator on), a `PTX:WARNING` names the pinned version, says a reader's saved work can be replaced by an older copy, and offers both remedies. A partial pin such as `0.7` floats to the newest release and stays quiet, as do exact pins outside the range. Verified through the stylesheets directly: pinned `0.7.22` with a SCORM build warns; the same pin without a host, `0.7.25`, no pin, and `0.7` do not.

**Standalone pages.** The `doenetml-coordinator` call moved from `file-wrap-simple-head-cache` into `file-wrap-basic-head-cache`, so the marker and the script agree everywhere: a standalone interactive page now loads the coordinator for the one activity it holds, which under the defaults simply boots at once.

**Dist file.** `js/dist/ptx_scorm_events.js` is out of the PR; only the source change remains. Thanks for the pointer to the policy.

**Commits.** Both messages are now a subject line plus the `Co-Authored-By` trailer and nothing else — the session link is gone — and the second subject now says DoenetML to match the first. On attribution: the model has been named in the commits all along, in the `Co-Authored-By: Claude Fable 5` trailers, which the PR page keeps out of sight; the PR description's closing line now names it too.

On the honest limit: activities not rendering in your test environment even on master is something we would like to fix regardless of this PR — if you can share what the environment is (OS, browser, how the built book is served), I'm happy to dig into it as a separate issue on the Doenet side.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rbeezer:
Merged, with code as-is.  But I failed to notice that we need more granularity in the commits, so I broke this into smaller bits like we do.  I will have my assistant post something about testing complications, I'm not sure I'm following all that.

dqnykamp:
> I will have my assistant post something about testing complications, I'm not sure I'm following all that.

I'm not sure I am following either, but if our assistants can help each other out...

rbeezer:
Taking you up on the offer to look at the rendering problem — here is exactly what I was trying to do, and where it stops. This is a separate matter from the PR, which I am satisfied with.

### What I was trying to test

The coordinator's observable behaviour, rather than the markup that turns it on: that activities start a couple at a time as the reader nears them, that ones left well behind are put away, that the space they held is kept, and that a reader's work survives the round trip. None of that can be watched without first getting a single activity to render, and that is the step that never completes here.

I can confirm the coordinator itself is wired up and doing something: on a page with several activities, an iframe I have not scrolled to is parked at `about:blank`, which is the coordinator holding it back. So I could verify the PR's generated markup and the coordinator taking control, but not the behaviour it exists for.

### What happens instead

The viewer mounts, removes its "Waiting on the page to load…" placeholder, and leaves the container as

```html
<div data-theme="light" lang="en" dir="ltr"><div class="before-keyboard"></div></div>
```

and then nothing further. The document never appears.

The sharpest clue: on a bare test page with the virtual keyboard enabled, **the keyboard renders** — the full math keypad paints — while the document area stays at those 85 characters. So React mounts and paints sibling UI; it is specifically the document content that never arrives.

### What I ruled out

- **The document.** A trivial activity, a single `<p>` and nothing else, behaves identically.
- **PreTeXt's markup, and this PR.** Same result on the merge-base, and same on a hand-written minimal page that loads `@doenet/standalone@0.7.25` directly and calls `renderDoenetViewerToContainer` on a container I built myself.
- **Loading and parsing.** The real activity pulls 43 chunks, all `200`, including `graph`, `point`, `vector`, `line`, `polygon`, `sideBySide` and `JSXGraphRenderer`. The trivial one pulls 10, and none of those. The chunk set is content-dependent, so the source is being parsed and its component set resolved before the render stalls.
- **The worker.** A blob-URL classic worker is constructed, with no exception at construction and no `error` event on it afterwards.
- **WebAssembly.** Neither `WebAssembly.instantiate` nor `instantiateStreaming` is ever called.
- **Anything surfacing an error.** No console output at all — not one message — no `window` error event, and no `unhandledrejection`.
- **Cross-origin isolation.** `SharedArrayBuffer` was unavailable under a plain static server, which looked promising, so I re-served the identical build with `Cross-Origin-Opener-Policy: same-origin` and `Cross-Origin-Embedder-Policy: credentialless`. `crossOriginIsolated` became `true` and `SharedArrayBuffer` became available; the outcome was unchanged.

### Diagnosis

Everything up to the render succeeds — fetch, parse, component resolution, worker construction, React mount — and then the document render is a silent no-op, with no error raised anywhere I can observe it.

### Environment

- Chrome `143.0.0.0`, Linux x86_64, 32 hardware threads, `navigator.deviceMemory` 8
- book built with the `pretext` script, served over a plain local static HTTP server, never `file://`
- reproduced three ways: the generated single-activity `*-if.html`, the same activity inside a full book page, and a hand-written minimal page with no PreTeXt involvement
- `@latest` resolves to `@doenet/standalone@0.7.25`

### What would help

Whether that empty shell is a failure mode you recognise, and whether there is a minimal snippet known to render, so we can tell an embedding problem from a package one. I can run arbitrary probes inside the page — message traffic to and from the worker, internal state, anything you want instrumented — so if there is a specific thing to look at, name it and I will report back.

dqnykamp:
Thanks for the detailed writeup — it was enough to work from, and two of the observations turn out to be pointing the wrong way. Also: thank you for merging, and for breaking it into the granularity you use.

### That empty shell is the normal pre-render state, not a terminal one

`<div data-theme="light" lang="en" dir="ltr"><div class="before-keyboard"></div></div>` is exactly what the viewer looks like from mount until the document is ready to paint. The viewer renders nothing at all while the core is coming up — it does not render a container it then fills — and the virtual keyboard is a *sibling* of the document, not a child, so "keyboard paints, document area stays at 85 characters" is the expected shape of any stall before first paint rather than a clue that the document specifically is broken. Your 85 characters are the wrapper plus the keyboard's anchor.

So the question is not "why did the render no-op" but "which phase of the boot never finished".

### WebAssembly runs inside the worker, where a page-side probe cannot see it

The core is WASM, but it is instantiated inside the core worker — a separate realm. Patching `WebAssembly.instantiate`/`instantiateStreaming` on the page's `window` will never observe it, on a healthy boot or a broken one. (The worker is booted from a blob URL with the WASM handed to it inline as a `data:` URL, so there is no WASM fetch on the network panel either — consistent with your 43 chunks all being JS.) That observation doesn't rule core start in or out.

### The boot has three phases, and there is a hook that names them

`stateLoad` (the saved-state load that precedes a boot) → `handshake` (the round trip to the core worker, watchdogged) → `generate` (the un-watchdogged evaluation that produces the document). A host can observe all three in a production bundle by setting `__doenetTestCoreInitHook` on `window.doenetGlobalConfig` **before** the bundle loads. Which phase is last tells you where it stops:

- **no phase at all** — the viewer never started booting (it is gated on being on-page, or on a host telling it to render)
- **`stateLoad`, then nothing** — the saved-state load never settles; it waits on IndexedDB, so a wedged IndexedDB hangs here silently
- **`handshake`, then nothing** — the core worker isn't answering; after 15s you should also see a `DocViewer: core worker handshake attempt 1 of 3 failed` warning, and after three attempts a failure pane replaces the blank
- **`generate`, then nothing** — the worker is alive and the document is being evaluated; on a large document this can genuinely take a long time

Your "no console output at all" is itself informative: if the handshake had been reached and hung, the watchdog would have said so at 15 seconds. That points at the two phases *before* it.

### A probe page

Self-contained, tested; drop it beside your book and open it. It reports the phase trace with timings, independently checks IndexedDB and a blob-worker-plus-WASM round trip, and prints the container at intervals past the watchdog.

```html
<!doctype html>
<html>
<head>
<meta charset="utf-8">
<title>DoenetML boot probe</title>
<script>
// Must run BEFORE the bundle loads: the bundle adopts this object.
window.doenetGlobalConfig = window.doenetGlobalConfig || {};
var t0 = Date.now();
function log(msg) {
  var line = "[" + ((Date.now() - t0) / 1000).toFixed(1) + "s] " + msg;
  console.log(line);
  var out = document.getElementById("out");
  if (out) { out.textContent += line + "\n"; }
}
window.doenetGlobalConfig.__doenetTestCoreInitHook = function (phase, attempt) {
  log("PHASE reached: " + phase + " (attempt " + attempt + ")");
};
window.addEventListener("error", function (e) { log("window error: " + e.message); });
window.addEventListener("unhandledrejection", function (e) { log("unhandled rejection: " + e.reason); });
</script>
</head>
<body>
<h3>Probe</h3>
<div id="c" class="doenetml-applet" data-doenet-add-virtual-keyboard="false"><script type="text/doenetml">&lt;p&gt;hello there&lt;/p&gt;</script></div>
<hr>
<pre id="out" style="font:12px monospace"></pre>
<script type="module">
log("page script running");

// 1. Can IndexedDB be opened at all?  The saved-state load waits on it.
(function () {
  var settled = false;
  var req;
  try { req = indexedDB.open("doenet-probe-db", 1); }
  catch (e) { log("IndexedDB threw synchronously: " + e); return; }
  req.onsuccess = function () { if (!settled) { settled = true; log("IndexedDB: opened OK"); } };
  req.onerror = function () { if (!settled) { settled = true; log("IndexedDB: error " + req.error); } };
  req.onblocked = function () { log("IndexedDB: blocked"); };
  setTimeout(function () { if (!settled) { settled = true; log("IndexedDB: NEVER SETTLED after 10s <-- suspicious"); } }, 10000);
})();

// 2. Does a blob-URL worker running WebAssembly answer at all?
(function () {
  var src = "self.onmessage=async()=>{try{" +
    "const bytes=new Uint8Array([0,97,115,109,1,0,0,0]);" +
    "await WebAssembly.instantiate(bytes);" +
    "self.postMessage('worker-ok-wasm-ok');" +
    "}catch(e){self.postMessage('worker-ok-wasm-FAILED: '+e);}};";
  var w = new Worker(URL.createObjectURL(new Blob([src], { type: "application/javascript" })));
  var answered = false;
  w.onmessage = function (e) { answered = true; log("blob worker: " + e.data); w.terminate(); };
  w.onerror = function (e) { answered = true; log("blob worker ERROR: " + e.message); };
  w.postMessage("go");
  setTimeout(function () { if (!answered) log("blob worker: NO ANSWER after 10s <-- suspicious"); }, 10000);
})();

// 3. The viewer itself.
const { renderDoenetViewerToContainer } = await import("https://cdn.jsdelivr.net/npm/@doenet/standalone@0.7.25/doenet-standalone.js");
log("bundle imported");
renderDoenetViewerToContainer(document.getElementById("c"));
log("renderDoenetViewerToContainer called");

// 4. What is in the container as time passes?  Watch past the 15s watchdog.
[2, 5, 10, 20, 30, 60].forEach(function (s) {
  setTimeout(function () {
    var html = document.getElementById("c").innerHTML;
    var rendered = html.indexOf("hello there") >= 0;
    log("t+" + s + "s container: " + html.length + " chars, document " +
        (rendered ? "RENDERED" : "NOT rendered") + " :: " + html.slice(0, 100));
  }, s * 1000);
});
</script>
</body>
</html>
```

Here (Chrome 150, Linux x86_64, headless, that same `0.7.25` straight from jsdelivr, plain static server, no cross-origin isolation) it reads:

```
[0.0s] blob worker: worker-ok-wasm-ok
[0.0s] IndexedDB: opened OK
[0.5s] PHASE reached: stateLoad (attempt 0)
[1.8s] PHASE reached: handshake (attempt 0)
[1.8s] PHASE reached: generate (attempt 0)
[2.5s] t+2s container: 326 chars, document RENDERED
```

So a minimal embedding of the published bundle does render, which makes this environmental rather than a package fault — and the trace from your machine should say which of the four it is.

### One specific thing worth ruling out first

If you are driving Chrome through automation, check whether the harness advances timers artificially — Chrome's `--virtual-time-budget` and the like. Doing that reproduced **your exact DOM string, character for character** on my otherwise-healthy machine: virtual time races the 15-second handshake watchdog while the worker is still doing real work, the handshake is abandoned as timed out, and the viewer sits at the blank shell. Real-time run of the same page, same bundle: renders in 2.5 seconds. That is the cheapest thing to eliminate, and it would explain a silent stall in an environment where nothing is actually wrong.

If the trace says `stateLoad`, the next probe is IndexedDB under the exact profile you test with; if it says `handshake`, I would want the worker's own console (it will not appear in the page's), and I can give you a build that echoes the handshake traffic.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

oscarlevin:
BTW, I'm releasing v2.50.0 of the CLI that has this fix in case @bnmnetp wants to use it to build a new version of the Active Calculus books.

rbeezer:
That probe settled it, and the answer is entirely on our side. Thank you — both for the correction and for building something I could just run.

### The trace

```
[0.2s] PHASE reached: stateLoad (attempt 0)
[0.5s] PHASE reached: handshake (attempt 0)
[0.5s] PHASE reached: generate (attempt 0)
```

IndexedDB opened, the blob worker answered with WebAssembly instantiated, no watchdog warning, no retries. All three phases complete in half a second — and the container still read 85 characters at t+2s, 5s, 10s, 20s, 30s and 60s.

### The cause

The tab is hidden. `document.visibilityState` is `"hidden"` and `document.hasFocus()` is false, because the browser automation drives a background tab. Chrome gives a hidden tab no animation frames, and the first paint is committed from a `requestAnimationFrame` callback, so the document was evaluated and waiting the whole time with nothing to paint it.

The proof is the number you measured: on the one occasion a frame did occur here, the document painted immediately, at exactly **326 characters**. `requestAnimationFrame` callbacks queued from the automation otherwise never run at all, and `Page.captureScreenshot` times out against the same tab — not a frozen renderer, just an unpainted one.

### Two things I got wrong, now clear

You were right that the empty shell is the normal pre-render state. The other one is the same artifact: layout still runs in a hidden tab even though paint does not, so the virtual keyboard's text was readable through `innerText` while the document's DOM had not been committed. "Keyboard paints, document does not" was never a clue about the document — it was two different things being measured, one of which does not need a frame.

Your suggestion about automation advancing timers was the right shape, too. The mechanism turned out to be missing frames rather than a raced watchdog, but it was the same instinct: look at the harness before the package.

So there is nothing to fix in DoenetML, and nothing to chase on IndexedDB or the core worker. I have withdrawn the conclusion from my earlier note.

### Kept

The probe is preserved on our side, along with the phase meanings, because the `__doenetTestCoreInitHook` trace is the fastest way to tell a real stall from an environmental one, and it will be the first thing we reach for next time. The lasting lesson for us is narrower and ours: never call a component broken from an automated tab without checking `document.visibilityState` first.

Thanks again — that was a genuinely useful exchange, and the coordinator work is the better for having been looked at this closely.

PR #3172, opened 2026-08-21
WeBWorK: use server's MathJax config

Alex-Jordan:
Head up: I don't think this clashes with #3147, but all the same, it would be better to process #3147 before this one.

In the past, when we supported old versions of WeBWorK, we needed to hard code the MathJax config here in `pretext-webwork.js`. Now that we only support 2.19 and 2.20, we do not need to do that, and we should use the webwork2 server's MathJax (v3) config. So that is the main change here. Most of the red in the diff here is replaced by one line of green at the new line 345.

Nearby, I changed some lines to interpolate a js variable, to make the code more readable.

Tested directly with 2.20. For 2.19, I located some live 2.19 servers and just checked that the path which is used here for a MathJax config file is valid on those servers. So it should work exactly the same.

[Sometime coming up soon, we will add support for 2.21 (with MJ v4). The change here will make that project cleaner. But I want to be clear that this PR alone is not bringing support for 2.21.]

### DIFF
--- a/js/src/pretext-webwork.js
+++ b/js/src/pretext-webwork.js
@@ -342,61 +342,12 @@ async function handleWW(ww_id, action) {
         }
 
         let iframeContents = '<!DOCTYPE html><head>' +
-            '<script src="' + ww_domain + '/webwork2_files/node_modules/jquery/dist/jquery.min.js"></script>' +
-            `<script>
-                window.MathJax = {
-                    tex: { packages: { '[+]': ['noerrors'] } },
-                    loader: { load: ['input/asciimath', '[tex]/noerrors'] },
-                    startup: {
-                        ready() {
-                            const AM = MathJax.InputJax.AsciiMath.AM;
-
-                            // Modify existing AsciiMath triggers.
-                            AM.symbols[AM.names.indexOf('**')] = {
-                                input: '**',
-                                tag: 'msup',
-                                output: '^',
-                                tex: null,
-                                ttype: AM.TOKEN.INFIX
-                            };
-
-                            const i = AM.names.indexOf('infty');
-                            AM.names[i] = 'infinity';
-                            AM.symbols[i] = { input: 'infinity', tag: 'mo', output: '\u221E', tex: 'infty', ttype: AM.TOKEN.CONST };
-
-                            return MathJax.startup.defaultReady();
-                        }
-                    },
-                    options: {
-                        processHtmlClass: "process-math",
-                        renderActions: {
-                            findScript: [
-                                10,
-                                (doc) => {
-                                    for (const node of document.querySelectorAll('script[type^="math/tex"]')) {
-                                        const math = new doc.options.MathItem(
-                                            node.textContent,
-                                            doc.inputJax[0],
-                                            !!node.type.match(/; *mode=display/)
-                                        );
-                                        const text = document.createTextNode('');
-                                        node.parentNode.replaceChild(text, node);
-                                        math.start = { node: text, delim: '', n: 0 };
-                                        math.end = { node: text, delim: '', n: 0 };
-                                        doc.math.push(math);
-                                    }
-                                },
-                                ''
-                            ]
-                        },
-                        ignoreHtmlClass: 'tex2jax_ignore'
-                    }
-                };
-            </script>` +
-            '<script src="' + ww_domain + '/webwork2_files/node_modules/mathjax/es5/tex-chtml.js" id="MathJax-script" defer></script>' +
+            `<script src="${ww_domain}/webwork2_files/node_modules/jquery/dist/jquery.min.js"></script>` +
+            `<script src="${ww_domain}/webwork2_files/js/MathJaxConfig/mathjax-config.js"></script>` +
+            `<script src="${ww_domain}/webwork2_files/node_modules/mathjax/es5/tex-chtml.js" id="MathJax-script" defer></script>` +
             `<script src="${courseUrlBase}_static/pretext/js/dist/knowl.js" defer></script>` +
-            '<link rel="stylesheet" href="' + ww_domain + '/webwork2_files/node_modules/bootstrap/dist/css/bootstrap.min.css"/>' +
-            '<script src="' + ww_domain + '/webwork2_files/node_modules/bootstrap/dist/js/bootstrap.bundle.min.js" defer></script>';
+            `<link rel="stylesheet" href="${ww_domain}/webwork2_files/node_modules/bootstrap/dist/css/bootstrap.min.css"/>` +
+            `<script src="${ww_domain}/webwork2_files/node_modules/bootstrap/dist/js/bootstrap.bundle.min.js" defer></script>`;
 
         // Determine javascript and css dependencies
         const extra_css_files = [];


Alex-Jordan:
This was rebased after #3147 was merged. I tested, and things are still working as expected.

rbeezer:
We have had enough problems with MathJax running in different contexts (knowls, Runestone assignment builder) that it makes me nervous to cede more control to whatever configuration WW is producing -- now, or later.  Yes, I know this is inside an `iframe`.

More coming from Claude, just now.

rbeezer:
Verified the 2.19 claim from source rather than by probing paths: `htdocs/js/MathJaxConfig/mathjax-config.js` is present on the `WeBWorK-2.19` and `WeBWorK-2.20` tags, and the two files are identical apart from a MathJax menu-rerender workaround added in 2.20. So this should indeed behave the same on both. Tested against the public PreTeXt WeBWorK server: rendered math is character-for-character identical to master across three problems, including expansion of a macro defined in `docinfo/macros`, with no MathJax errors and nothing left unprocessed.

One change I'd ask for before this goes in.

**The iframe stops honouring `process-math`.** The inline config being removed set `processHtmlClass: "process-math"`; no version of the server config sets it — not 2.19, 2.20, or 2.21. After this change, `process-math` means nothing inside the WeBWorK iframe.

Today that is harmless, and I confirmed it: the iframe contains no `tex2jax_ignore` subtree, and the one `process-math` element in it — the `PTX_MACROS` span — is a child of `<main>`, so it is typeset either way.

But it leaves the iframe running a different math-processing contract from every other page PreTeXt produces, and that is the part I think is worth not accepting. On an ordinary page, `<body class="ignore-math">` means nothing is typeset unless PreTeXt marks it `process-math` — opt out globally, opt in per element. `process-math` is PreTeXt's own convention, emitted from several places in `pretext-html.xsl`: the inline-math wrapper, the display-math wrapper, the units `\pi` wrapper, JSXGraph containers, and the hidden macros div. The iframe is not foreign territory — PreTeXt writes its `<main>`, injects the macros span, and loads `knowl.js` into it, which exists precisely to pull more PreTeXt HTML in at runtime. Any of that markup arriving inside a `tex2jax_ignore` subtree would now silently fail to typeset, and the invariant that keeps it safe is not written down anywhere, not enforced, and not tested.

It also becomes a standing dependency on a third party: whether PreTeXt's own class is honoured is now decided by whatever webwork2 ships next.

Keeping the server config wholesale and re-asserting just this one PreTeXt-specific option costs one line, between the config script and `tex-chtml.js`:

```js
'<script>if (window.MathJax) (window.MathJax.options ||= {}).processHtmlClass = "process-math";</script>' +
```

I tested that: the server's AsciiMath additions all survive, `processHtmlClass` is back to `process-math`, and rendering is unchanged.

Two smaller notes.

Worth saying out loud in the description, because it undersells the change: the server config is not just the old config relocated. It adds AsciiMath triggers for consistency with MathObjects — `DNE`, `NONE`, `Inf`/`Infinity`/`INF`, `Re`, `Im`, `log10`, `U`, `><`. I confirmed these are live in the iframe now. Answers typed into a PreTeXt-hosted problem will accept the same forms a native WeBWorK page does, which is a real improvement for students.

The readability pass missed one line two below the converted block, still concatenated while its neighbours are template literals:

```js
'<script src="' + ww_domain + '/webwork2_files/node_modules/iframe-resizer/js/iframeResizer.contentWindow.min.js"></script>' +
```

Alex-Jordan:
Imo Claude is just wrong here, missing some context needed to understand everything. This passage in particular:

> The iframe is not foreign territory — PreTeXt writes its `<main>`, injects the macros span, and loads knowl.js into it, which exists precisely to pull more PreTeXt HTML in at runtime.

as well as its commentary on `knowl.js` after that.

I think there is a Monday morning drop-in session, and I'll go there. Unless anyone thinks I really should try to explain in writing here.

Alex-Jordan:
Hmm, drop-in is Tuesday this week, not the Monday that I thought. And tomorrow (Tuesday) I have conflicts. So I will try to discuss here. Apologies in advance I can't be brief.

`<background>`

We put a static representation on a web page, with an Activate button. That button creates an iframe where a live exercise will render. There are two (and only two) reasons why we do not just directly let the iframe have a `src` attribute such that everything inside it would be coming from the webwork2 server.

1. The original reason: so that we could style the live exercise more like PreTeXt. There used to be knowls inside a rendered PG exercise that were for revealing hints and solutions. Rather than use the `knowl.js` that WeBWorK had, we hotwired those things to work with the `knowl.js` that PreTeXt has. So this was especially about making the hint and solution behavior feel PreTeXt-y.
2. A bit later, we wanted Runestone to be able to intercept student answers and the correctness assessment that the webwork2 server undertook, so that Runestone could record scores and previous attempted answers. Again, a reason not to just toss in some `src` from the webwork2 server.

So what we do is build a `srcdoc` and intervene as needed to accomplish 1 and 2. Along the way, there were some hiccups with gathering answers to record in Runestone. Namely with various types of multiple choice exercises, and a lot of that was because of a lack of structure for those things in PG at the time. There was also an issue with the Runestone Academy assignments page, where certain URLs being relative paths that work fine for a problem in an HTML book don't work when the problem is rendered in a RS Assignments page.

Worth noting: WeBWorK/PG stopped using a javascript-based knowl mechanism for solutions and answers a few versions back. Solutions and hints are just handled with native HTML `details/summary` now. The place near the code in this PR that is loading PTX's `knowl.js` can be removed, pending testing that I haven't overlooked something about that. It's just a different thing than this PR, which is about the MathJax config, and tosses in a little bit of code tidying in the immediate vicinity for readability.

`</background>`

OK, so Claude is talking about `knowl.js` being here, and how that means there might be a knowl in a rendered WW problem. And if you open that knowl, you would get PTX content from elsewhere in the book, and it should have the PTX MathJax config, or at least recognize the `process-math` class. But that's all never going to happen. There is never a WeBWorK exercise that cross-references something else in a PTX book. That was a design understanding from the start, since PG problems can go have a life somewhere else apart from the PTX book. The presence of `knowl.js` here was only ever to use it to open the PG problem's own solution and hint knowls.

So this sentence from Claude starting out as "The iframe is not foreign territory ..." is a misunderstanding. The body content of the iframe really is foreign territory coming from the webwork2 server. It won't ever "pull more PreTeXt HTML" in the way Claude is describing.

And then to step back, it is right to let the math content of the exercise be governed by the webwork2 server's MathJax config. That content was born where that particular config has been curated. Back in the day, the webwork2 server did not serve a standalone MathJax config file. Rather, webwork2 wrote the config into each page, inline. So we were forced to make this MathJax config here that the PR wants to eliminate. But now that we are supporting only 2.19+, the need is gone. Meanwhile, these exercises were never using the "real" MathJax config that PTX uses. They have been using the config being deleted here. I can't explain why `process-math` shows up here in the config being deleted. My guess is that at the time, I just mistakenly copied things from the PTX MathJax config along with the config webwork2 was using at the time, just for good measure. But these recent thoughts have me understanding there is no situation where `process-math` is going to appear within this iframe.

As far as Runestone's interests goes, that's all about intercepting the answer hash to record scores and answer attempts, not what the math looks like. All this time in RS, even in the Assignments page, the math has rendered based on the stale MathJax config being deleted here. Not a good MathJax config that PTX usually uses, or a good one that RS usually uses.

The post has gotten long enough, so I'll see if there's response from anyone (be they human or Claude) about this. I'll just note that for 2.21, the drift on the WW config is going to be significant. This is where WW/PG upgraded to MJ4. And now with 2.22 under devlopment, more as WeBWorK/PG moves away from using script tags to delimit math, and simplifies to just using `\(`. Letting `pretext-webwork.js` just naturally use the correct config file will save a lot of trouble.

sean-fitzpatrick:
> There is never a WeBWorK exercise that cross-references something else in a PTX book. 
> That was a design understanding from the start, since PG problems can go have a life somewhere else apart from the PTX book.

I think there is possibility for author error here... For example, I don't think the schema forbids an xref inside a webwork. 

That said: I can think of some places in APEX where there are xref knowls, but the most common situation is in the introduction to an exercise group. There are few (if any) references back to the main text.
E.g. there is one set of questions where the reader is asked to carry out a particular task (compute arc length, say), and later, another set that asks for a different task (let's say surface area). The introduction will say something like "Note: these are the same functions as in Exercises 15 - 20."

But here's the thing: in HTML, "Exercises 15 - 20" is a knowl, and it appears in the Exercise group introduction, but that's just a regular paragraph in the HTML. It doesn't appear in the static version of the corresponding exercises.
When you activate one of the questions, the exercise group introduction gets copied into the statement of the question, but "Exercises 15 - 20" *is no longer a knowl*. It's just plain text.

This makes sense to me. Maybe the reference to Exercises 15 - 20 doesn't make sense if this problem is used out of context on a WeBWorK server, but we never intended our exercises to be used out of context. Either way, I think it makes sense that the xref ceases to be a knowl on WeBWorK.

Alex-Jordan:
OK, following the -dev thread, and I'll bow out and close this.

Alex-Jordan:
@sean-fitzpatrick You're completely correct here. When I wrote "never cross-references" I meant never in a dynamic way, with a knowl. Claude's concern here is that you could open a knowl leading to who knows what.

sean-fitzpatrick:
Right -- in other words, not a valid concern.

PR #3177, opened 2026-08-25
Printout improvements

oscarlevin:
This fixes three separate issues with printouts (worksheets and handouts).

1. There apparently is a bug with how we were swapping out stylesheets in the newest version of chrome.  That's fixed.  Thanks Claude Opus 5 :)
2. Footnotes had never worked in printouts: they would be opened mid paragraph as a knowl.  Now they migrate to the bottom of their page.
3. A paragraph that had a long list (ol/ul/dl) would overflow a page.  Since we now do a better job fixing page overflows in general, it was a simple fix to add this for the list case.

Mostly javascript, with a few css tweaks for the footnotes (so both js and css needs a rebuild).  Examples of both footnotes and multipage lists added to the sample article.

### DIFF
--- a/css/targets/print-worksheet/print-worksheet.scss
+++ b/css/targets/print-worksheet/print-worksheet.scss
@@ -259,6 +259,148 @@ article.printout {
   }
 }
 
+// A list too tall to fit on one page is cut into one single-item list per row,
+// so that pagination has somewhere to break inside it (see
+// flattenOversizedListsIn()).  The pieces have to go on reading as the one list
+// they were, which means the space between two of them must be the space
+// between two items -- `li` carries that (`_lists.scss`) -- and not that plus a
+// list's own outer margins on top.  So the pieces give up their outer margins;
+// the first piece is left behind inside whatever held the list, where it keeps
+// the margins the original had.
+.onepage > .split-list {
+  // The gap between two items lives on the piece, not on the item inside it.
+  //
+  // `li` normally carries that spacing (see _lists.scss), but a margin on the
+  // item collapses straight out through the piece's border-free top edge: the
+  // space is there on screen, while the piece's offsetHeight does not include
+  // it.  Pagination measures rows with offsetHeight plus margins, so every
+  // piece then measured about 7px short -- nothing on its own, but once per
+  // item it put a whole page's budget out by well over an inch, which
+  // over-filled pages and left the footnote block below them hanging off the
+  // bottom of the sheet.
+  //
+  // Carried as a top margin rather than a bottom one so that it collapses with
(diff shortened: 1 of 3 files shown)


oscarlevin:
Just force pushed after a rebase on master.

rbeezer:
Built this and drove the printouts; all three fixes do what they say.

| fix | what I saw |
|---|---|
| stylesheet swap | pagination completes, `print-worksheet.css` in force, the theme link gone |
| footnotes | `worksheet-dot-products` paginates to two pages, each carrying its own `.footnotes` block with exactly its page's note, sitting at the foot |
| long lists | the new handout splits nine items then three across two pages, with nothing overflowing the page box |

Both `css/dist` and `js/dist` rebuilt from the branch to test, since the change is invisible without them.

The structure holds up too. The `isHeaderFooterEl` to `isPageFurnitureEl` rename is fully converted with no stragglers, and the new `.footnotes` class does not collide with anything the XSL emits, so treating it as page furniture cannot swallow authored content. Charging each row its own footnote height, rather than charging the page, is the right shape — it keeps the cost moving with the row and leaves nothing circular for the page-break search to resolve. Removing the theme link outright is safe because leaving a preview is a full navigation, so nothing needs it back.

One thing worth a follow-up, on the comment rather than the code.

The stylesheet-swap comment states that Chromium does not fire `load` on a link whose href is rewritten in place, and that waiting on one therefore hangs forever. I could not reproduce that in Chrome 143. I tried four ways and `load` fired every time:

- a plain href rewrite on the existing link
- the original ordering, href set first and the listener attached after, with the target stylesheet already warm in cache
- a rewrite pointed at a 404, where both `load` and `error` fired
- setting the href to the value it already had

The fix itself should stay regardless — a fresh element with both `load` and `error` handled is more robust than rewriting in place, and adding the new link before removing the old one is what keeps the preview from flashing unstyled. It is the explanation I would want corrected. That comment is long and confident, and as written it will send the next person debugging a printout hang after a mechanism that does not appear to exist. If you still have the conditions that produced the hang, those are what belong there — a real repro is worth more than the generalisation, and it is possible something specific to the printout flow is involved that my minimal cases do not capture.

One small correction to the new handout, separate from the above.

Its list items say "It should be on the first page of the handout" twice and "It should be on the second page of the handout" once, but the split does not fall there: items one through nine land on the first page and the last three on the second, so the third item is on page one. Worth fixing because those sentences read as the expected result. Anyone checking this printout later, after a change to the page-break search, would take the third item appearing on page two as correct when it would in fact be the regression.

rbeezer:
Mergrd as-is, will go look at 3178 now.

PR #3178, opened 2026-08-25
Printout code migration

oscarlevin:
The javascript to control printouts is getting quite large.  This moves the code into its own file is `js/src/`, along with the required wiring in jsbuilder.

Note that I made this commit on top of those from #3177.  If that can be merged first, I can then do a rebase and this should just work.  Or we could merge all the commits from this current PR and close the other one completely.  Just needed to do this sometime.


rbeezer:
Checked this as a move rather than as a change, and it holds up.

Every one of the 2,340 lines removed from `pretext_add_on.js` reappears in `pretext-printouts.js`, and nothing reappears that was not removed except a thirteen-line file header — which is exactly the difference between the two file counts. Nothing was dropped in either direction.

The thing I most wanted to check was references across the new boundary, since esbuild gives each bundled file its own module scope: a call from one file to a function left behind in the other would build cleanly and then fail at runtime. There are none. Fifty-six functions moved and three stayed, no function or top-level variable is named across the split, and the single global in play, `window.i18next`, is both set and read on the same side.

The wiring is right too. The import sits between `pretext_add_on.js` and `knowl.js`, which preserves the load-listener ordering that the entry point's own header warns about, and touching only the descriptive comment in `jsbuilder.mjs` is correct given `pretext-core` is already a bundle target and esbuild follows the import.

Behaviour is unchanged where it counts. Building the bundle from this branch and driving the printouts gives the same results as before the split: the footnote worksheet paginates to two pages carrying the same notes at the same offsets, the long-list handout splits nine items then three with nothing overflowing its page box, and the console stays clean.

One small thing, and it is not really this PR's doing. The list of what `pretext-core` bundles appears in two places, and only one is updated here. The other is in `xsl/pretext-html.xsl`, where the comment above the script tag still reads "pretext.js + pretext_add_on.js + knowl.js". That list was already out of date before this change — it omits `pretext-dialog.js`, `pretext-dropdown.js`, `readability-options.js` and `pretext-embed.js` — so this adds a fifth omission to an existing four rather than creating the problem. Worth either refreshing it or letting it point at `js/src/pretext-core.js` as the one place that actually lists the contents.

rbeezer:
Picked up the reoranization commit cleanly, and merged that.  JS files regenerated - some little bit of movment, just three functions it seems.  All done here.  Thanks for the maintenance.

PR #3180, opened 2026-08-26
HTML: render empty table headings as td

ascholerChemeketa:
Fix for immediate issue reported here:
https://groups.google.com/g/pretext-a11y/c/LlGCpMDTmCk/m/ROhzFu3hAgAJ

Pre screenshot with WAVE:
image

Post:
image

Heads up to @oscarlevin who reported issue

### DIFF
--- a/examples/sample-article/sample-article.xml
+++ b/examples/sample-article/sample-article.xml
@@ -6673,6 +6673,29 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
                 </tabular>
             </table>
 
+            <p>Generating empty header cells as <tag>th</tag> elements is an accessibility issue. They should render as <tag>td</tag> elements instead.  This next table tests that behavior.</p>
+
+            <table>
+                <title>Empty Header Stress Test</title>
+                <tabular row-headers="yes">
+                    <row header="yes">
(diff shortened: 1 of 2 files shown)


rbeezer:
`normalize-space` does not do exactly what you may think it does.

`normalize-space(.)` applied to `<cell><fillin characters="5"/></cell>`  will produce a `td` here.

Suggestion:

Define a `$cell-nonempty` (or similar) by `normalize-space(.) or *` and then use it three times.  Better results and easier to read.  `nbsp` could fool it, so you could adjust to `*[not(self::nbsp)]`,  If so, maybe explain why the exception for `nbsp` in the code.

More elaborate constructions can get fooled, like `<cell><p><nbsp/></p></cell>` but folks should not be putting `p` into headings.

ascholerChemeketa:
Fixed. Thanks - I had not considered contents inserted via means other than literal text.

Confirmed that the fillin is considered meaningful content by WAVE and screen reader.

Also added a SA example with those test cases.

rbeezer:
Merged, with two silent changes.

*  Added a code comment about why `nbsp` was in the test.

*  Removed the `!= ''` from the `normaize-space()` part of the test, since it is redundant.

As always, thanks for all the accessibility work!

PR #3186, opened 2026-08-28
Bug fix: ensure correct parbreaks around image and tabular in sidebyside

oscarlevin:
This fixes two related bugs with the latex conversion.

1. If a worksheet contains a sidebyside to lay out exercises, those exercises can contain a mix of paragraphs and images.  As in a stack, a paragraph followed by an image in a sidebyside needs special care to ensure there is a paragraph break between the paragraph and the image.  The test for this only checked whether the image was in a stack.
2. The same sort of check needs to be applied to tabulars after paragraphs.  Claude uncovered that this is not happening because the tabular-inclusion code lived inside a `\resizebox{...}{`.  See the pdf of the sample article corresponding to this [Paragraph](https://pretextbook.org/examples/sample-article/html/section-side-by-side.html#stacking-side-by-side-8-1-3) 
https://pretextbook.org/examples/sample-article/html/section-side-by-side.html#stacking-side-by-side-8-1-3

Claude claimed that the same thing was happening with the text after an image, specifically the text after the QR codes for interactives.  This may be, but those QR codes are wide enough that the next word went to the following line anyway.  But in theory it does make sense to guard against the "after" case too.

### DIFF
--- a/xsl/pretext-latex-common.xsl
+++ b/xsl/pretext-latex-common.xsl
@@ -7345,14 +7345,42 @@ along with PreTeXt.  If not, see <http://www.gnu.org/licenses/>.
     </xsl:choose>
 </xsl:template>
 
-<!-- Second: images already constrained by side-by-side panels -->
-<xsl:template match="image[ancestor::sidebyside]">
-    <!-- get a newline if inside a "stack" -->
-    <xsl:if test="parent::stack and preceding-sibling::*">
+<!-- An "image" or a "tabular" in a side-by-side panel is set as a  -->
+<!-- bare box, so if a paragraph is already in progress the box is  -->
+<!-- appended to it, hanging off that paragraph final line, and the -->
+<!-- leading "\noindent" is a no-op.  So break the paragraph, and   -->
+<!-- start a fresh one for whatever follows.  Two panels need no    -->
+<!-- break: one that *is* the object, and a "figure" or "table",    -->
+<!-- whose "caption" is a sibling but is deferred to the bottom of  -->
+<!-- its box.  The panels that do are the containers holding        -->
+<!-- several things: a "stack", a list item, and the "statement"    -->
+<!-- of an "exercise" (or of its "task") in a worksheet or handout. -->
+<xsl:template match="image|tabular" mode="panel-paragraph-break">
+    <xsl:param name="side"/>
+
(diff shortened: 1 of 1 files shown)


oscarlevin:
Before/after PDFs compared with Adobe's compare tool show only expected diffs (besides a few random "image replaced" with seem to just be artifacts of tikz not being 100% raster-perfect.

rbeezer (commented):
This looks right to me, and the reasoning in the new comment holds up under the cases I could think to try: a `figure` or `table` panel whose only sibling is a `caption`, two images stacked (exactly one break between them rather than two), an image alone in a list item, and the object that *is* the panel. Each gets the breaks it should and no others.

I checked the thing most likely to bite — the removal of the `parent::stack` test from `tabular-inclusion`. That mode is reachable only through `tabular[not(ancestor::sidebyside)]` and `tabular[ancestor::sidebyside]`, which partition, and a `stack` exists only inside a `sidebyside`. So the removed test could never have fired on the path that no longer carries it, and behaviour is preserved there.

Building the sample article and the workbook against the merge base:

- 85 `\par` added to the sample article, three to the workbook, and **none removed**.
- The PDF compiles with no errors, 301 pages before and after.
- Typesetting badness goes down rather than up: Overfull 392 → 377, Underfull 475 → 469.
- Five pages differ in text layout.

Page 226 of the sample article shows the problem plainly. Before, exercise 1's sentence stops at "…a system of linear equations for the", the circuit diagram is set into the middle of that paragraph, and the orphaned "amperages." lands below the diagram at the left margin; exercise 2 does the same beneath its figure. After, each sentence finishes and the diagram follows it.

Worth knowing that most of the 85 breaks change nothing on the page — panel content is usually wide enough that following text was already forced onto its own line. Those five pages are where it shows.

One small thing, on the commit subject rather than the code: the convention here quotes element names, so `"image"`, `"tabular"` and `"sidebyside"`.

rbeezer:
Merged with a rebase on master, and so then mild commit message editing.

Thanks for this.  We keep chasing this one, or similar.  For example see `\leavevmode` in the code.  I think we whittled that down to two places finally.

This is reaching the limits of the kind of complication I like to tolerate (not **your** fault!) .  The necessity is just too complicated for my tastes.  Anyway, I have a mind to send Fable off for something of a more universal solution - maybe more automatic `\par` in various places.

PR #3188, opened 2026-08-30
CSS: hide image-expansion button on printouts

oscarlevin:
The recently merged work on image expansions left a dangling image expansion button on all images in printouts (always, not just on hover).  

This PR hides that, like we do for the image description icon.

Will require a rebuild of the distributed CSS.


rbeezer:
That looks pretty bad!  Merged, and rebuilding website samples right now.

PR #3189, opened 2026-08-31
Accessibility fix for data files in HTML

ascholerChemeketa:
Image datafiles currently have no path to emit an @alt for accessibility.

This renders an alt using the shortdescription with fallback to the filename is that is not provided.


rbeezer:
I need to look at #3149, so I've been thinking this might conflict.  Are you OK revisting this after I get to that one?

ascholerChemeketa:
You mean #3139?

Yeah, this is pretty minor.

PR #3191, opened 2026-08-31
Bug fix: don't delete latex-images for cdn hosted resources builds

oscarlevin:
First there was the "portable" builds that were meant to create a single HTML file whenever possible.  We inlined the svgs created by latex-images, so in the pretext script, deleted all the generated latex-images.

Then we added cdn hosted js/css as a separate publication option to reduce the bundle size of things like SCORM and not require PreTeXt Plus to copy the large number of js files all the time.  But this does not inline svgs, so we still need the latex-images generated folder.

In setting up the cdn hosted resources, I was sloppy and tied both cases to a single boolean.  

This PR fixes that mistake.  There are already publisher variables being reported, so it was a minimal fix.

I am aware that there is some cleanup that can be done (switching the pub var being reported to the boolean, adding additional generated svgs to the list of things that shouldn't be copied over in true portable builds), but I wanted to get this fixed as a first step.

### DIFF
--- a/pretext/lib/pretext.py
+++ b/pretext/lib/pretext.py
@@ -4235,8 +4235,14 @@ def html(xml, pub_file, stringparams, xmlid_root, file_format, extra_xsl, out_fi
     # names for scratch directories
     tmp_dir = common.get_temporary_directory()
 
+    # Two variants of limiting the files produced for HTML: CDN resources are
+    # used for portable HTML as well as resources/@host="cdn", either of which
+    # set b-cdn-resources to "true" in the publisher file.  Some generated
+    # assets are inlined for the portable HTML (but not for CDN resources),
+    #  which is indicated by portable-html="true"
     pub_vars = common.get_publisher_variable_report(xml, pub_file, stringparams)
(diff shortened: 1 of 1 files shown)


rbeezer:
Verified with builds of the sample article. At this PR's merge-base, a `resources/@host="cdn"` build deletes `generated/latex-image` — 90 SVG files the built pages reference, so images break; that reproduces the report. With this change the same build keeps all 90 files, and a build with `platform/@portable="yes"` still omits the directory, since its images are inlined — the portable behavior is unchanged. The reported `portable-html` value and the computed `b-cdn-resources` both check out against the publisher-variable report, confirming the analysis in the description.

One correction needed in the new comment block: it ends with `portable-html="true"`, but the reported value — and what the code just below tests — is `"yes"`. Since this comment exists precisely to untangle the two value conventions (`"true"` for the computed boolean, `"yes"` for the plain variable), quoting the wrong one would send the next reader into exactly the trap it documents. Rob will fix this.

rbeezer:
Rebased, edited commit message, silently fixed "true -> yes".  Thanks for the fix.

PR #3194, opened 2026-09-01
Offline mathematics is serialized as XML

rbeezer:
**Nothing in PreTeXt's output changes.** All four modes we run — SVG, MathML, braille, speech — come out byte-identical. This is about a latent defect in a script we share with PreFigure, and about the two copies of that script staying the same text.

`script/mjsre/mj-sre-page.js` also offers `--svgenhanced`, which adds Speech Rule Engine enrichment to the SVG. PreTeXt never invokes it; PreFigure does, for the labels in its diagrams. Enrichment writes the author's LaTeX into `data-semantic-attributes`, and nothing filters that attribute — so a macro as ordinary as `\newcommand{\lt}{<}` puts a raw `<` inside an attribute value, and the HTML serialization the script used produces something no XML parser will read.

Davide Cervone diagnosed this on davidaustinm/prefigure#87 and pointed at the remedy: `adaptor.serializeXML()` in place of `adaptor.outerHTML()`, which escapes the character and keeps the attributes.

The reason we could not simply take that one line is that `serializeXML` declares the XHTML namespace on the mock page's `html` root, and `xsl/support/package-math.xsl` matched `body`, `div/mjx-data`, `mjx-braille` and `mjx-speech` with no prefix. Every one of those would then miss — measured, five of fourteen offline-MathJax runs went from real content to zero bytes, every braille and every speech one, while SVG and MathML survived on their own namespaces.

So the serializer and its reader move together here, in one commit: the output line, and the stylesheet gaining `xmlns:xhtml` plus prefixes on eight name tests across six templates. `math:math` and `svg:svg` are untouched. Literal result elements stay namespace-less, so the packaged representations keep their existing shape.

### Verification

- **Fourteen offline-MathJax runs byte-identical** against master: sample article, EPUB sampler, braille test, fonts, slideshow and humanities, across SVG, MathML, braille and speech as each document supports.
- **Seven consumer builds** — EPUB in all four flavours, braille emboss and electronic, and the reveal.js slideshow — every one exits cleanly, and every differing file differs only in a build timestamp. That includes `package.opf`, whose two differing characters sit inside `dcterms:modified`.
- **The defect is fixed**: `--svgenhanced` on a page defining and using `\newcommand{\lt}{<}` now parses as XML, with the attribute written `latex:\newcommand{\lt}{&lt;}`.

### On the shared script

PreFigure carries its own copy, identical to ours apart from the output block. Their reader needs the matching change — an `xhtml` prefix on three XPath lookups in `label_tools.py` — and that is theirs to make when it suits them. The two copies differ by one line until then, and match again afterwards. Neither project has to wait on the other, and nothing breaks in either while they are out of step; the hazard would be copying the file wholesale between the repositories before the receiving reader is ready.

### DIFF
--- a/script/mjsre/mj-sre-page.js
+++ b/script/mjsre/mj-sre-page.js
@@ -386,15 +386,14 @@ if (!(argv.svg || argv.svgenhanced)) {
   //
   await html.renderPromise();
   //
-  //  Output the resulting document.  This is deliberately the HTML
-  //  serialization: the XML serialization would add the XHTML
-  //  namespace to the bare "html" root of the mock page, and every
-  //  namespace-less element match in the packaging stylesheet
-  //  (xsl/support/package-math.xsl) would then miss.  The attribute
-  //  escaping it would add is not needed: no output attribute can
-  //  hold a raw "<" now that the "data-latex" attributes are
-  //  filtered away.
+  //  Output the resulting document as XML.  Speech enrichment writes
+  //  the author's LaTeX into "data-semantic-attributes", which nothing
+  //  filters, so a macro such as \newcommand{\lt}{<} would put a raw
+  //  "<" into an attribute value; the XML serialization escapes it,
+  //  where the HTML serialization does not.  The XHTML namespace this
+  //  puts on the mock page's "html" root is expected by the packaging
+  //  stylesheet (xsl/support/package-math.xsl), which matches on it.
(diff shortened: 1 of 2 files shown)


rbeezer:
Motivated by https://github.com/davidaustinm/prefigure/pull/87

PR #3201, opened 2026-09-02
CSS: make permalink indent adjustments style aware

ascholerChemeketa:
Inspired by all the fun we are having over on https://github.com/PreTeXtBook/pretext/pull/3168

Current permalink positioning has a similar problem. We have ad hoc rules to change the offset of the permalinks based on what containers do in terms of padding and left border in some styles. But the current rules apply to elements regardless of what style is applied.

This reengineers the positioning to be based on things like "permalink is immediate child of a heading box" instead of "permalink is an immediate child of an example-like"

Results are slightly cleaner at a pixel level (handles different offset amounts for different kinds of boxes) and basically the same at a macro level.

Heads up for @oscarlevin

### DIFF
--- a/css/components/chunks/helpers/_C-box-mixin.scss
+++ b/css/components/chunks/helpers/_C-box-mixin.scss
@@ -61,6 +61,17 @@ $heading-color: var(--block-head-color) !default;
       border-bottom-left-radius: $border-radius;
     }
 
+    // Note - once inherit() is widely supported, refactor permalinks
+    & > .autopermalink {
+      // own permalink is offset by border width
+      left: calc(-1 * ($border-width + 2em));
+    }
+
+    & > * > .autopermalink {
+      // child element permalink is offset by border width and padding on side with border
+      left: calc(-1 * ($padding-left + $border-width + 2em));
+    }
+
     & > .heading:first-child {
       background-color: $heading-background;
       display: inline-block;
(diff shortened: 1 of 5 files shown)


rbeezer:
Permalink offsets inside blocks were set by a hard-coded list in `_permalinks.scss` — `assemblage-like`, `example-like`, `project-like`, `remark-like` and a few nested combinations — all given the same `left: -2.9em` regardless of what the active style actually does with borders and padding. Where a theme's box geometry differed from the one that constant was chosen for, the permalink landed on the block's left border.

This derives the offset instead from the box the permalink hangs off, inside the four mixins that draw those boxes: a block's own permalink is offset by its border width, and a child element's by border width plus the padding on the bordered side.

Measured on the sample article across six themes, comparing the compiled themes on identical HTML and checking every visible permalink against its own parent's box:

| page | on the border, before → after |
|---|---|
| `block-samples` | denver 9 → 0 |
| `section-interesting-corollary` | default-modern 1 → 0, denver 1 → 0 |
| `section-worksheet-solutions-testing` | default-modern 7 → 0 |

Eighteen permalinks that sat on a border are clear of it, none is introduced, and no permalink intrudes past the padding edge into the text column, before or after. Visible-permalink counts are identical throughout: this moves permalinks, it never reveals or hides one.

One thing left for later: the same eleven-line block, including the `inherit()` note, now appears in four mixin files, differing only in the name of the padding variable. A shared mixin taking the padding as an argument would leave one copy.

rbeezer:
Merged as-is after a rebase and a tag in the commit message.  Thanks for getting this all in order.

PR #3204, opened 2026-09-03
Script: ensure entities resolve even with newest lxml

oscarlevin:
The issue reported on [-support](https://groups.google.com/g/pretext-support/c/QaM3kiiHns4) appears to be cause by `uvx` downloading the latest security patch of lxml, which breaks PreTeXt's entities tricks.  This PR should fix that.  Without this, the next time the CLI or anyone using the pretext script upgrades their lxml, their pretext will also break.

### DIFF
--- a/pretext/lib/common.py
+++ b/pretext/lib/common.py
@@ -280,7 +280,14 @@ def xsltproc(xsl, xml, result, output_dir=None, stringparams={}):
     control = None
     if output_dir:
         control = ET.XSLTAccessControl(write_file=True)
-    xsl_tree = ET.parse(xsl)
+    # Our stylesheets use an internal DTD subset to pull in "../entities.ent"
+    # (see the "entity tricks" comment atop each stylesheet) via an external
+    # parameter entity. lxml 6.1.3 stopped resolving external parameter
+    # entities by default (LP#2165901), so this must be requested explicitly.
+    # The stylesheet is always our own bundled/trusted XSL, never
+    # user-supplied XML content, so re-enabling resolution here is safe.
+    xsl_parser = ET.XMLParser(resolve_entities=True)
+    xsl_tree = ET.parse(xsl, parser=xsl_parser)
     xslt = ET.XSLT(xsl_tree, access_control=control)
 
     # do the transformation, with parameterization


rbeezer:
Thanks for the quick work on this one, @oscarlevin!  Merged as-is.

rbeezer:
This one is right, and it is worth landing quickly — anyone whose lxml updates loses every build until it is in. The code needs no changes. Everything below is about the comment.

## Verified

Reproduced the breakage and the fix by running the sample article through the script with lxml 6.1.3 ahead of the usual dependencies. On master the build dies before producing anything:

```
File ".../xsl/utilities/report-publisher-variables.xsl", line 25
lxml.etree.XMLSyntaxError: Entity 'entities' not defined, line 25, column 15
```

On this branch the same build completes, 175 HTML files.

I bisected the releases rather than take the version on trust:

| lxml | master | this branch |
|---|---|---|
| 4.9.4 | works | works |
| 5.4.0 | works | works |
| 6.0.4 | works | works |
| 6.1.1 | works | works |
| 6.1.2 | works | works |
| 6.1.3 | **fails** | works |

So 6.1.3 in the comment is exactly right. `load_dtd=True` on its own does not fix it; `resolve_entities=True` is the lever that matters.

Three more things worth recording:

- **No regression on the current lxml.** Sample-article builds from master and from this branch, both on 6.1.1, differ only in the two build-timestamp lines.
- **Backward compatible.** `resolve_entities=True` behaves correctly on 4.9.4 and 5.4.0 as well, so leaving `lxml` unpinned in `pretext/requirements.txt` stays fine and no lower bound is needed.
- **Coverage is complete.** `ET.XSLT(` occurs exactly once in the codebase, at the line this commit touches, so this single parse serves all 38 stylesheets that use the entity trick — 28 in `xsl/`, 6 in `xsl/latex/`, 2 in `xsl/utilities/`, 1 in `xsl/tests/`, and `schema/pretext-validation-plus.xsl`. Stylesheets reached by `xsl:import` are compiled by libxslt rather than by this parser, so they were worth checking separately: `pretext-html.xsl` imports both `pretext-common.xsl` and `pretext-assembly.xsl`, each of which uses the trick, and the full 6.1.3 build proves they resolve.

## Two corrections to the comment

**The stylesheet can be user-supplied.** `-X` / `--XSL` takes a stylesheet path from whoever runs the build. It becomes `extra_xsl`, then `extraction_xslt`, and reaches this same `xsltproc()` in the HTML, revealjs, and LaTeX paths. The Guide's transition appendix even shows authors the entity trick for use in their own stylesheets, so a custom stylesheet both arrives here and legitimately needs the setting.

The conclusion holds, but the reason written down does not. What is true, and is a sound justification: the stylesheet is always chosen by the person running the build — bundled with PreTeXt, or named with `-X` — and never taken from the document being processed, while `no_network` stays at its default `True`, so an entity cannot reach off the machine. Worth getting right, since this comment is the standing record of why a security-relevant setting is safe.

**What 6.1.3 changed is the other way around.** The comment has lxml "stopped resolving external parameter entities by default". Reading the cited bug, `resolve_entities='internal'` was always meant to exclude external entities, and a gap in the parameter-entity callback let them through anyway, permitting local file reads. 6.1.3 closed that gap.

Which makes this a nicer change than the comment claims: PreTeXt had been relying on the gap, and `resolve_entities=True` is not a workaround but the intended way to ask for the behaviour on purpose. Saying that makes the safety argument above self-evident, and it still reads correctly years from now, when which release changed it no longer matters.

## Smaller

- `LP#2165901` is opaque, and an abbreviation. Spell out Launchpad or give the URL — there is no other bug citation in this file, so the choice is open.
- The comment names `"../entities.ent"`. Across the stylesheets that spelling appears 9 times; `"entities.ent"` appears 25, `"./entities.ent"` 3, and `"../xsl/entities.ent"` once. Naming `entities.ent` without a prefix covers all of them.
- Optional, and in keeping with the file: state the second flag as well.

  ```python
  xsl_parser = ET.XMLParser(resolve_entities=True, no_network=True)
  ```

  It is already the default, but two lines above this the file says "this is the default, but we are explicit here" about access control, and `pretext.py` spells out all three flags where it parses SVG. The build still completes under 6.1.3 with it stated.

Commit subject, topic and granularity are all as they should be, and the whitespace is clean.

rbeezer:
I might take up some of the things in "Smaller", but you should not wait on that now.

rbeezer:
> I might take up some of the things in "Smaller"

See #3206

PR #3213, opened 2026-09-10
Texstyles: the preamble defines the text-mode symbols

rbeezer:
The texstyle preamble driver (`texstyle/ptx-preamble` in `xsl/latex/pretext-latex-texstyle.xsl`) now calls the `text-symbols` named template, between `math-packages` and `pdfpages-package`, where the regular and classic preamble drivers call it.  That template loads `pifont` and defines the end-of-environment marks (`\ptxsquare`, `\ptxtriangle`, `\ptxdiamond`) and the unordered-list markers (`\ptxlistdisc`, `\ptxlistcircle`, `\ptxlistsquare`).  The list templates in `pretext-latex-common.xsl` use the markers in every LaTeX conversion, so a texstyle document with an unordered list has been failing to compile since the markers were introduced (2026-03-03): `Undefined control sequence. \labelitemi ->\ptxlistdisc`.

**Testing.**  Sample article through the AMS texstyle (`<journal name="ams"/>`).  LaTeX before and after differs only by the ten-line `text-symbols` block in the preamble; before, the file uses `\ptxlistdisc` 49 times, `\ptxlistcircle` 8 and `\ptxlistsquare` 3 with no definition of any of them.  PDF before: the compile stops at page 16 (Subsection 4.5, the first unordered list) with the undefined control sequence.  PDF after: no undefined control sequence; the compile runs to page 186, where it stops on `Environment openconjecture undefined` — a separate gap in the texstyle route, not touched here.

*Claude Fable 5.1, acting as a coding assistant for Rob Beezer*


rbeezer:
@oscarlevin - I don't think this is controversial, but if so, let me know - easily reverted.

PR #3215, opened 2026-09-12
Widths of images inside a worksheet/sidebyside/exercise

oscarlevin:
Previously the width of an image that is a descendant of a sidebyside always get the full width of their panel.  But sidebyside can hold exercises when inside a worksheet (or similar), and images inside these should be allowed to shrink.  This implements this, as well as the case where the image is inside a li and a p in a sidebyside.


rbeezer:
Gave the following to my assistant.  Response (next) may be enough for your assistant to work out easily, and force-push?

Please carefully review #3215.  Some points to consider:

1.  Commit messages mention #worksheet.  Fix seems much broader, as I do not see #worksheet really mentioned in any change.

2.  sidebyside/image versus sidebyside/p/ul/li/image do seem to me to be separate situations - the first is a panel, the latter is some component of the narrative text.  And so width is a different story.

3.  Not obvious to me that the new entity is the right filter.  Panels are children of #sidebyside, or of sidebyside/stack.  Isn't that it?

4.  In pretext-common on master at line 2976 only *some* conditions get edited.  Shouldn't they all be edited?

rbeezer:
Built the sample article at the merge-base and at `3a515cb0e` (HTML, LaTeX,
PDF, assembled form validated): HTML differs only in the worksheets page,
LaTeX only at the five worksheet images nested inside exercise panels plus
the new example, no new messages, no new schema errors.  Measured in a
browser: the new example's image box is 50% of its panel, centered; a probe
`p/ul/li/image` at 50% is 50% of the list item; figure-panel images are
unchanged at 100%.  The change does what it says.  Requests:

1. **Scope of the subjects and description.**  Nothing here is
   worksheet-specific: any `image` below a panel of any `sidebyside` is
   affected (list items inside `p`, `ol`/`ul`/`dl`, `list` and `stack`
   panels; `exercise` panels in worksheets and handouts, hints and solutions
   included).  Please word the commit subjects and the description for the
   rule, with the worksheet exercise as the example, and quote element names
   per the repository convention (`"sidebyside"`, `"exercise"`).
2. **The entity.**  A panel is a child of a `sidebyside` or of a `stack` in
   one; the third clause, `parent::figure/parent::sidebyside`, is not a
   panel test but "the image that is a figure panel's content".  It is
   needed (it keeps every `sidebyside/figure/image` at full panel width), but
   the name promises a panel test, and it sits next to the `SBSPANEL`
   element list.  Suggest: make the entity the panel test,
   `parent::sidebyside or parent::stack/parent::sidebyside`, and write the
   figure case at the image matches,
   `image[&SBS-PANEL-FILTER; or parent::figure[&SBS-PANEL-FILTER;]]`, so the
   rule reads "the image, or the figure holding it, is a panel".
3. **Edit all six conditions the same way**, in both `get-width-percentage`
   templates: `audio`, `video`, `interactive`, `jsxgraph`, `slate` still test
   `ancestor::sidebyside`.  Those placements are deprecated or removed, so
   nothing live changes, but the two complementary patterns should state one
   rule.
4. **Docinfo defaults.**  With this change `docinfo/defaults/images`
   `@width` and `@margins` apply to an image inside an exercise panel that
   has none of its own (the four circuit-worksheet images in the sample
   article would become 60% if its commented-out default were active).  That
   follows from the rule, but please say so in the description.
5. **"Relative to the panel".**  The reference width is the text block the
   image sits in: in a list item it is the item's width, not the panel's.
   The comments in `entities.ent` and `pretext-html.xsl`, and the sample
   prose, should say "the available width where it sits" or similar; and
   "now relative to" in the entity comment narrates the change — state the
   rule.
6. **Sample article.**  Use `<tag>sidebyside</tag>` in the prose as the rest
   of the article does; drop the `xml:id` on the new image (a `@source`
   image needs none, and the value abbreviates).
7. **Same rule in the other conversions.**  `pretext-markdown.xsl` keeps
   its own copy of the panel lookup, keyed on `ancestor::sidebyside` in both
   its alignment test and its width test, so a nested image there still takes
   a panel's layout width (a probe image at 50% inside an exercise panel comes
   out `width="45%"`); please use the entity in both places, after which the
   -common layout supplies the author's width.  `pretext-revealjs.xsl` keys
   its paused-image wrapper on `ancestor::sidebyside`; same entity.
   `pretext-fo.xsl` sizes images through -common and inherits the change, but
   its image comment ("makes an image fill its panel when inside a
   sidebyside") should be reworded.  The remaining conversions import HTML or
   LaTeX and inherit.

*Claude Fable 5.1, acting as a review assistant for Rob Beezer*

oscarlevin:
A panel of a `"sidebyside"` is a child of the `"sidebyside"`, or a child of a `"stack"` in it.  Only a panel, or the content of a `"figure"` panel, takes its width from the layout of the `"sidebyside"`.  Previously *any* `"image"` below a `"sidebyside"` filled the width available to it, however deeply it was nested.  Now an object nested more deeply keeps its own `@width` and `@margins`, relative to the available width where it sits.  That covers an `"image"` in an `"exercise"` panel of a worksheet or handout (hints and solutions included), or in a list item within a `"p"`, `"ol"`, `"ul"`, `"dl"`, `"list"` or `"stack"` panel.

Implementation:

- `&SBS-PANEL-FILTER;` in `entities.ent` is the panel test, `parent::sidebyside or parent::stack/parent::sidebyside`.  The figure case is written at each use, as `X[&SBS-PANEL-FILTER; or parent::figure[&SBS-PANEL-FILTER;]]`.
- -common: all six conditions (`image`, `audio`, `video`, `jsxgraph`, `interactive`, `slate`) in both complementary `get-width-percentage` templates use the same rule.  So do the docinfo image defaults in `layout-parameters`.  The non-image placements are deprecated or removed, so nothing live changes for them.
- HTML, LaTeX, markdown (alignment and width), and revealjs (paused-image wrapper) use the same test.  The XSL-FO image comment is reworded; FO inherits the change from -common.

**Docinfo defaults.**  A consequence of the rule: `docinfo/defaults/images` `@width` and `@margins` now apply to an image nested inside a panel (e.g. in an `"exercise"` panel) that has neither attribute of its own.  They still do not apply to panels or to the content of `"figure"` panels.

Checked against the previous head of this branch (sample article, plus the sample slideshow for revealjs).  LaTeX and HTML differ only in the reworded sample prose, and revealjs is identical.  Markdown changes only at the worksheet images nested in `"exercise"` panels: they now take their own width (the new example at 50%, the others at the 100% default) instead of the panel's 45%.  No new messages in any conversion.

The sample article gains an `"image"` at 50% inside an `"exercise"` panel of a worksheet `"sidebyside"`.

Co-authored by Claude Opus 5

oscarlevin:
Oops, let me clean up Claude's lengthy commit messages.  Force pushed the clean version.

rbeezer:
Thanks for this Oscar - looking good.  Re-review from Claude, I will make a few changes on merge, so no need to do anymore on this one.

rbeezer:
Re-reviewed `5dbc4ddfa` the same way as before (sample article at the
merge-base and at the head in HTML, LaTeX, PDF, Markdown and XSL-FO, the
sample slideshow for reveal.js, assembled forms validated, a probe with
nested cases and the docinfo default switched on, widths measured in a
browser).  All seven requests are addressed: HTML and LaTeX match the
previous head except for the reworded prose; Markdown now gives a nested
image its own width (the probe's 50% and the 60% default, `align="left"`
for a flush margin); reveal.js output is identical; XSL-FO inherits the
rule (`fo:external-graphic/@width` 50%/60% for the probe images, figure
panels unchanged).  No new messages, no new schema errors.  Nothing further
is needed from you; @rbeezer will take care of the rest at merge:

1. **One more entity, as a follow-on commit by @rbeezer.**
   `&SBS-PANEL-FILTER; or parent::figure[&SBS-PANEL-FILTER;]` now appears
   23 times across five stylesheets, and the panel test is never used
   alone.  That repetition is the literal reading of my earlier request, so
   the fault is mine.  A second entity defined in terms of the panel test
   (say "sized by the layout of the sidebyside: a panel, or the content of
   a figure panel") will replace the compound at every site.
2. **Two comment nits, melded in by @rbeezer.**  The reflowed XSL-FO
   comment stops a line short ("components.  N.B. an SVG file" is left as
   a stub before the untouched lines), and the pre-existing line above the
   layout-width template in `pretext-common.xsl`, "occurs in a figure, not
   contained in a sidebyside", describes the opposite of that template.
3. **The XSL commit may be split by conversion** at merge, retaining your
   authorship.

*Claude Fable 5.1, acting as a review assistant for Rob Beezer*

rbeezer:
Merged with changes as described.  Thanks for catching this one, @oscarlevin!

PR #3219, opened 2026-09-17
Improvements for journal support including CSL specification

oscarlevin:
This PR contains the next step for improving support for creating latex for journals from pretext source: using the CSL style preferred by a journal, and grabbing the CSL file required on demand.  

There is also a small commit to update the latex support files for the springer nature journal file.

This was all done using Claude Opus 5, which also created a helpful roadmap in the doc folder (this is on its own commit in case you want to drop it entirely).

Note that I did not check whether the pretext/pretext script needs additional infrastructure to support this; if it does, I'll follow up.


rbeezer:
Just a short note, no need to change anything yet, there will be more later.

I think maybe a development road map should just become an issue on GitHub.  It could have tasks, and we could just let our assistants manage it.

Not sure I want a lot of AI messing-about in the repo, and I'm not sure diffs on the git history would be the best way to see history/progress.

My work converting PDFs of research articles keeps turning up subtle bugs (excellent!) and I'm just having those sessions make issues.  Then I can point a development assistant to the issue for the real work.  I think a roadmap could be utilized the same way.

Tangential: where/how we get the CSL style files has been a big question in my mind, so I will take some time to scrutinize this one carefully.

